PR #1, opened 2016-12-07
Add babel as a default loader

itkrt2y:


dhh:
Great! I think we have to exclude node_modules from that loader for performance reasons, though, right?

guilleiguaran:
+1, this should exclude node_modules folder.

PR #2, opened 2016-12-07
Change directory structure

itkrt2y:
I think that `app/packs/javascripts` is closer to the Rails way of naming things, rather than `app/javascript/packs/`. Also, `app/packs/javascripts` is more future-proof, as it will let us add other assets, for example stylesheets under `app/packs/stylesheets/`.


dhh:
I'll explain this properly in the README. But the intention is that app/javascript is where all your app-like JS lives. Within that, we just use packs/ as a convention to create entry targets for webpack. So whatever you put in packs/ is what you want to have compiled via webpack as an entry. Every bit of "real code" should just live in app/javascript, like app/javascript/calendar/index.js or whatever.

itkrt2y:
Thanks, I understand.

Only one more thing, I think it would be better to use plural name like `app/javascripts/` than `app/javascript/`.
What do you think about it? @dhh

dhh:
I went with `app/javascript` specifically to highlight the difference to `app/assets/javascripts`. To me, JavaScripts is great when you consider these things just that: A collection of independent scripts. Kinda what we call sprinkles.

But for this, we're building systems, not individual scripts. Kinda like if we did app/ruby rather than app/rubies.

Hope that makes sense.

itkrt2y:
OK, now I totally understand. Thanks a lot!

PR #5, opened 2016-12-07
Require `lodash.keyby` instead of `lodash`

serihiro:
I tried this gem with `rails5.0.1.rc1` , and I got an error when execute `bin/webpack-watcher`:

```
$ » bin/webpack-watcher
todo/config/webpack/shared.js:8
  entry: _.keyBy(glob.sync('../app/javascript/packs/*.js'), function(entry) { return path.basename(entry, '.js') }),
           ^

TypeError: _.keyBy is not a function
    at Object.<anonymous> (todo/config/webpack/shared.js:8:12)
    at Module._compile (module.js:541:32)
    at Object.Module._extensions..js (module.js:550:10)
    at Module.load (module.js:458:32)
    at tryModuleLoad (module.js:417:12)
    at Function.Module._load (module.js:409:3)
    at Module.require (module.js:468:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (todo/config/webpack/development.js:7:31)
    at Module._compile (module.js:541:32)
```

The lodash that I installed version is `4.17.2` (latest version for now) and lodash `4.17.2` does not include function `keyBy`.  `keyBy` is exported as an other  [npm package](https://www.npmjs.com/package/lodash.keyby).

So I think we should require `lodash.keyby` in `shared.js` to use `keyBy`.

### DIFF
--- a/lib/install/config/shared.js
+++ b/lib/install/config/shared.js
@@ -1,11 +1,11 @@
 // Note: You must restart bin/webpack-watcher for changes to take effect
 
-var path = require('path')
-var glob = require('glob')
-var _    = require('lodash')
+var path  = require('path')
+var glob  = require('glob')
+var keyBy = require('lodash.keyby')
 
 module.exports = {
-  entry: _.keyBy(glob.sync(path.join('..', 'app', 'javascript', 'packs', '*.js')),
+  entry: keyBy(glob.sync(path.join('..', 'app', 'javascript', 'packs', '*.js')),
     function(entry) { return path.basename(entry, '.js') }),
 
   output: { filename: '[name].js', path: path.join('..', 'public', 'packs') },


dhh:
Hmm, that's pretty odd. I have 4.17.2 here as well and I don't get that error. I can confirm it through node alone like so:

```
vendor david$ node
> var _ = require('lodash')
Expression assignment to _ now disabled.
undefined
> _.keyBy
[Function]
```

dhh:
I think there's a fair consideration as to whether we need a lodash dependency just to do keyBy. Maybe we can just write that natively.

guilleiguaran:
@dhh I'm using [webpack-merge](https://www.npmjs.com/package/webpack-merge) that was created specifically for those cases, I'll send a PR for this in a moment so we can check what approach seems better.

guilleiguaran:
nvm, I put my comment in the wrong place, this was for a commit not for this PR.

serihiro:
@dhh 
I noticed my mistake.
I used `lodash` installed by `npm`. Then it worked well as you said.
I did not read `install/template.rb` well. So sorry...

guilleiguaran:
Thanks, we aren't using lodash.keyBy anymore so I'm closing this

PR #6, opened 2016-12-07
fix node_modules path

Nitrino:
By default, node packages are installed in the node_modules folder in root project


dhh:
That's not how it's setup through this approach. We're using vendor/node_modules. The rest of the configuration is made to match that.

PR #7, opened 2016-12-07
change lodash.keyBy to pure javascript

Nitrino:


dhh:
Sweet! We can kill lodash from the install/template.rb definition then too.

Nitrino:
lodash is still used for _.merge. @guilleiguaran solves this.

dhh:
Oh right.

dhh:
Let's nix the ;'s btw.

Nitrino:
Do you plan to use ES6(babel) by default? 
In ES6, instead of `lodash.merge` can use the native method `Object.assign`

dhh:
Yeah, would actually be nice to have that setup for the config files too.
Please to investigate!

nathanmarks:
`lodash.merge` !== `Object.assign`

`_.merge` does a deep merge, `Object.assign` is a shallow merge.

I'd recommend sticking with a deep merge for webpack configuration.

nathanmarks:
Also, you don't need babel to use `Object.assign`, that's dependent on the node runtime being used when it comes to this config.

Nitrino:
@nathanmarks Oh sorry, you're right, `Object.assign` only does a shallow merge. 
It would be better to use [webpack-merge](https://www.npmjs.com/package/webpack-merge)

koenpunt:
Or use https://www.npmjs.com/package/deepmerge for merging.

PR #9, opened 2016-12-09
Add yarn binstub

itkrt2y:
I added the yarn binstub because we cannot run `rails webpacker:install` on apps predating Rails 5.1.

### DIFF
--- a/lib/install/bin/yarn.tt
+++ b/lib/install/bin/yarn.tt
@@ -1,0 +1,10 @@
+<%= shebang %>
+VENDOR_PATH = File.expand_path('../vendor', __dir__)
+Dir.chdir(VENDOR_PATH) do
+  begin
+    exec "yarnpkg #{ARGV.join(" ")}"
+  rescue Errno::ENOENT
+    puts "Yarn executable was not detected in the system."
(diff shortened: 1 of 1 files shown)


Nitrino:
I ran into the same problem and wrote binstab uses system yarn. 
Maybe it will be better, because the yarn has to be installed in the system?
```
<%= shebang %>

require 'mkmf'

yarn = find_executable 'yarn'

if yarn
  exec(['yarn', ARGV, '--modules-folder vendor/node_modules'].join(' ') )
else
  puts("Yarn executable was not detected in the system.")
  puts("Download Yarn at https://yarnpkg.com/en/docs/install")
end
```

itkrt2y:
I intended to make the same binstub as Rails 5.1. I don't know which is better, making this gem's own one or making the same as Rails 5.1.

What do you think? @Nitrino cc: @dhh 

https://github.com/rails/rails/pull/27245/files#diff-835c0a70de9cf02b472c84bf0a72323a

guilleiguaran:
I think is a better idea use same binstub as Rails 5.1

dhh:
Hmm, I'd rather that someone made a yarn plugin for pre-5.1 apps. Webpacker should be able to just assume that Yarn is available.

PR #10, opened 2016-12-09
Switch from lodash.merge to webpack-merge for per environment webpack config

guilleiguaran:


guilleiguaran:
Fixing conflicts with master now...

PR #14, opened 2016-12-10
Use APP_PATH when searching for webpack bin

garbles:


Nitrino:
`node_modules` installed in the vendor directory. 
It should look like this:
`WEBPACK_BIN = "#{VENDOR_PATH}/node_modules/webpack/bin/webpack.js"`
or in this case, quite right:
`WEBPACK_BIN    = "./node_modules/webpack/bin/webpack.js"`

guilleiguaran:
@Nitrino is right, `node_modules` is not in `APP_PATH` in Rails 5.1 apps

garbles:
Ah yup. `bin/yarn` did not exist in the project. Thanks!

Nitrino:
@garbles see this PR about yarn binstub #9

PR #18, opened 2016-12-11
Use old fashioned tag code

maclover7:
The new `tag.script` code is not available on `5-0-stable`; it is onl
available on Rails master, which is tracking 5.1.

Discovered this while trying to write a test for this helper method. :)

### DIFF
--- a/lib/webpacker/helper.rb
+++ b/lib/webpacker/helper.rb
@@ -2,6 +2,6 @@ require 'webpacker/source'
 
 module Webpacker::Helper
   def javascript_pack_tag(name, **options)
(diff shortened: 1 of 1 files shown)


guilleiguaran:
Not sure if we expect to support Rails 5.0 in this plugin, @dhh?

dhh:
I was intending to target 5.1+. Because we will need other concessions from Rails, like asset pipeline recording digests.

maclover7:
Yep, that makes sense. Without this patch, you'd only be able to use webpacker on apps running Rails master though 😬

dhh:
Until we have need for a 5.1+ feature, let's just keep 5.0 compatibility.

PR #19, opened 2016-12-11
Add rails-erb-loader for ERB support

guilleiguaran:
This allow to get support for ERB inside of JS files, e.g you can put Ruby code inside of a .js.erb file to be able to access to asset pipeline files:

```js
// app/javascripts/packs/application.js.erb
var imgPath = '<%= ActionController::Base.helpers.asset_path "rails.png" %>';
console.log(imgPath);
```

is compiled to:

```js
var a="/assets/rails-03bb170d74e34a9d3ab9157113c6ab68d2eb4a61a1a7ec5f77e45bc55d9e168e.png";
console.log(a);
```

### DIFF
--- a/lib/install/config/shared.js
+++ b/lib/install/config/shared.js
@@ -2,11 +2,12 @@
 
 var path = require('path')
 var glob = require('glob')
+var extname = require('path-complete-extname')
 
 module.exports = {
-  entry: glob.sync(path.join('..', 'app', 'javascript', 'packs', '*.js')).reduce(
+  entry: glob.sync(path.join('..', 'app', 'javascript', 'packs', '*.js*')).reduce(
     function(map, entry) {
-      basename = path.basename(entry, '.js')
+      basename = path.basename(entry, extname(entry))
       map[basename] = entry;
       return map;
     }, {}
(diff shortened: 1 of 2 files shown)


guilleiguaran:
btw, I'm waiting for https://github.com/usabilityhub/rails-erb-loader/pull/21 before of getting this merged to change the option name from `rails` to `runner`.

lib/install/config/shared.js
```diff
@@ -16,15 +17,25 @@ module.exports = {
...
         }
       }
+    ],
+
+    preLoaders: [
+      {
+        test: /\.erb$/,
+        loader: 'rails-erb-loader',
+        query: {
+          rails: '../bin/rails'
```
guilleiguaran: After of getting https://github.com/usabilityhub/rails-erb-loader/pull/21 this needs to be changed to `runner: '../bin/rails runner'`

guilleiguaran:
This is ready for merge.

@dhh do you think ERB support should be a default in webpacker or that we should add this to docs instead?

dhh:
Yeah, I think this is great. Gives us an immediate solution to getting digested stamps from the asset pipeline, even though I'd prefer a pure JS solution as well.

PR #25, opened 2016-12-14
added lib/install/vue/hello_vue.js, rails webpacker:install:vue

ytbryan:
First draft. Open for discussion if necessary. 

- `rails webpacker:install:vue` will ship with Vue 2.0 and its arsenal of tools (vuex, vue-router, vue-validator, vue-loader + vue-resource) 
- `vue/hello_vue.js` renders "Hello Vue" at the bottom of the page. `hello_vue.js` did not use Vue's JSX yet. Vue's JSX may require [babel-plugin-transform-vue-jsx](https://github.com/vuejs/babel-plugin-transform-vue-jsx) as a plugin.
```
Vue.component('jsx-example', {
  render (h) { // <-- h must be in scope
    return <div id="foo">bar</div>
  }
})
``` 
- Used the React's presets because I can't find a working Vue's presets. Known presets on Evan You's npm does not work when I do `webpacker:compile`. 

Suggestion, do we wish to change `app/javascript/packs/*.js` to `app/packs/javascript/*js`? Because we may require a `app/packs/templates/*.jsx or *.vue` next time? And IMO a javascript directory at the same level as assets, may be confusing. 

Please let me know your thoughts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ make any changes for a basic setup out the box. But this is where you do go if y
 more advanced.
 
 The configuration for what Webpack is supposed to compile by default rests on the convention that
-every file in app/javascript/packs/* should be turned into their own output files (or entry points, 
+every file in app/javascript/packs/* should be turned into their own output files (or entry points,
 as Webpack calls it).
 
 Let's say you're building a calendar. Your structure could look like this:
@@ -65,6 +65,10 @@ To use Webpacker with React, just create a new app with `rails new myapp --webpa
 will be added via yarn and changes to the configuration files made. Now you can create JSX files and
 have them properly compiled automatically.
 
+## Ready for Vue
+
+To use Webpacker with React, just create a new app with `rails new myapp --webpack=vue` (or run `rails webpacker:install:vue` on a Rails 5.1 app already setup with webpack), and all the relevant dependencies
+will be added via yarn and changes to the configuration files made. 
 
 ## Work left to do
 
--- a/README.md
+++ b/README.md
@@ -61,14 +61,12 @@ pipeline does it.
 
 ## Ready for React
 
-To use Webpacker with React, just create a new app with `rails new myapp --webpack=react` (or run `rails webpacker:install:react` on a Rails 5.1 app already setup with webpack), and all the relevant dependencies
-will be added via yarn and changes to the configuration files made. Now you can create JSX files and
+To use Webpacker with React, just create a new app with `rails new myapp --webpack=react` (or run `rails webpacker:install:react` on a Rails 5.1 app already setup with webpack), and all the relevant dependencies will be added via yarn and changes to the configuration files made. Now you can create JSX files and
 have them properly compiled automatically.
 
 ## Ready for Vue
 
-To use Webpacker with React, just create a new app with `rails new myapp --webpack=vue` (or run `rails webpacker:install:vue` on a Rails 5.1 app already setup with webpack), and all the relevant dependencies
-will be added via yarn and changes to the configuration files made. 
+To use Webpacker with Vue, just create a new app with `rails new myapp --webpack=vue` (or run `rails webpacker:install:vue` on a Rails 5.1 app already setup with webpack), and all the relevant dependencies will be added via yarn and changes to the configuration files made.
 
 ## Work left to do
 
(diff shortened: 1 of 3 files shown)


chrisvfritz:
I think installing vuex, vue-resource, vue-router, and vue-validator out of the box may be a bit overkill (also, vue-resource has been removed from official support). There are some other tweaks I'd make as well. I'm a maintainer of the official Webpack templates for Vue and a member of the core team. @ytbryan Would it be OK if put together an "officially blessed" PR for Vue integration to replace this one (likely within the next week)?

README.md
```diff
@@ -65,6 +65,10 @@ To use Webpacker with React, just create a new app with `rails new myapp --webpa
 will be added via yarn and changes to the configuration files made. Now you can create JSX files and
 have them properly compiled automatically.
 
+## Ready for Vue
+
+To use Webpacker with React, just create a new app with `rails new myapp --webpack=vue` (or run `rails webpacker:install:vue` on a Rails 5.1 app already setup with webpack), and all the relevant dependencies
```
toshimaru: should be "Webpacker with Vue"??

> Webpacker with React

ytbryan:
@chrisvfritz of course. I am all for a better Webpacker. 👍🏻

rlafranchi:
@chrisvfritz You've got my blessing. I will keep an eye out for your pr to test it out.

rafaelfranca:
@dhh why this gem is not using the generators API to generate files?

dhh:
@rafaelfranca In part because these generators or not intended to be configurable. I found it less overhead to just use the Rails template feature for the initial configuration. And for the framework-specific configurations, it's mostly changing existing files, which the generator API doesn't really help much with either.

rafaelfranca:
:+1:. But, templates and generators share the same infrastructure so generators can do the same things are templates but are more flexible. Now that we have different frameworks, having different rake tasks seems weird. Even more weird to use rake when Rails 5.1 is aiming to stop using rake in favor of our commands infrastructure.

I think it is fine for now, but we should revaluate in the future when we have the commands infrastructure in place.

dhh:
Rafael, happy to see it changed to generators if you want to take a stab at that. Or someone else does.

toshimaru:
+1 for using generators API to make template flexible.

dhh:
Thanks for working on this, guys. Going to close this PR pending the new one you got cooking.

cguillemette:
Just wondering where I could find that new PR, can you please share the link? Thanks!

chrisvfritz:
The Vue team is discussing what we'd like to do, as it's proven a bit more complicated than I originally hoped. This project intentionally only allows JS files to be bundled, which removes most of the power of Webpack and precludes Vue component files, which are by far the most popular feature made possible through Webpack. Given that, I can imagine it being quite confusing for users when the only supported feature is the much less popular JSX plugin. What we're leaning towards at this point is a vue_jsx integration that makes this more clear.

rlafranchi:
@chrisvfritz If it's any help, I've got a quick prototype for an app working with .vue extensions.  It's somewhat based off of the existing vuejs webpack-simple template.  Excuse the Gemfile, since it's pointing to a local directory for webpacker.  I modified the webpack config file (config/webpack/shared.js) post install to get it working.
[rlafranchi/rails-webpack-vue](https://github.com/rlafranchi/rails-webpack-vue)

I'm not sure if that's what you were aiming for, but let me know if there is anything I can do to help.

anlek:
Hey guys, I'm trying to get this working but I'm a bit lost on what to do?
I need to get a existing vuejs project pulled into an existing rails app but currently am having a hard time with Babel!
Any directions to give this PR a try, would be great. Thanks!

chrisvfritz:
@rlafranchi @anlek The decision isn't yet final, but we're actually leaning towards not providing an official Vue integration. Only supporting JSX just wouldn't be very useful to most users, so at this point, I'd probably recommend Vue users to not use Webpacker.

__However__, using Vue with Rails is still very possible and I have a repo with [several integration strategies for Vue](https://github.com/chrisvfritz/rails-javascript-integrations). I hope that helps!

rlafranchi:
Thanks @chrisvfritz, using Vue in an existing application has worked out well for me just by loading it in on a controller by controller basis where it will come in handy.  I agree that it doesn't make a whole lot of sense to create a webpack integration that doesn't support all the features.  If I were to start an SPA, I would lean towards initializing rails in --api mode and manage all frontend code using the Vue webpack template.

I'm sorry we couldn't get the official blessing from the Vue team, but agree with the decision and appreciate the time put into it.

samandmoore:
Does this mean that webpacker is not open to supporting CSS and other
assets down the line?

dhh:
Not to me. But we have to start somewhere and starting with JavaScript is the highest value bit for the largest group. So we start there. Besides, Webpacker itself is just a starting point. There's not much actual software here, just default configurations of existing packages, such that getting going is easier for more people. Those starting points can be extended and tweaked.

PR #37, opened 2016-12-20
Terminate statements with semicolons

Austio:
Hello, thank you for the time taken to create a webpack solution inside of rails.

I was running this locally and saw that the a few of the statements didn't have semi-colons at the end of them.  Not a huge deal as odds of weirdness are pretty small, but more of a best practice to have these.  

I look forward to helping on this project, thanks for your time!


dhh:
Excited for Webpack and Rails as well. We are intentionally going with a no semicolon style. Feel free to suggest other changes!

Issue #36, opened 2016-12-20
Turn on tree shaking by default


dhh:
As detailed in https://moduscreate.com/webpack-2-tree-shaking-configuration/, but using es2017.

guilleiguaran:
Done in 97ecd82b5b5

PR #39, opened 2016-12-20
es2015 modules false for webpack 2 tree shaking

Nitrino:
#36 Now that should be enough to work webpack 2 tree shaking


guilleiguaran:
Sorry, I didn't noticed this before and did it similarly in 97ecd82b5b5

PR #44, opened 2016-12-28
Use javascript_include_tag inside javascript_pack_tag

guilleiguaran:
With this is possible to use`config.action_controller.asset_host` to set host


kaspth:
👍 Probably needs to mention that it supports that config option too 😄

PR #47, opened 2016-12-31
Add middleware to wait for bundles to complete compiling

drewhamlett:
I know this isn't perfect but just wanted to throw out the idea.  I'm also not 100% on the error side of this.

To get middleware working in development

```config.middleware.insert_before ActionDispatch::Flash, "Webpacker::Middleware"```

Thanks for any help getting this in.

### DIFF
--- a/lib/install/config/development.js
+++ b/lib/install/config/development.js
@@ -23,6 +23,6 @@ module.exports = merge(config, {
       debug: true
     }),
 
-    new BundleTracker({ filename: './tmp/webpack-stats.json' })
+    new BundleTracker({ path: __dirname, filename: '../../tmp/webpack-stats.json' })
   ]
 })
--- a/lib/install/config/development.js
+++ b/lib/install/config/development.js
@@ -1,8 +1,9 @@
 // Note: You must restart bin/webpack-watcher for changes to take effect
 
-var path    = require('path')
-var webpack = require('webpack')
-var merge   = require('webpack-merge')
+var path          = require('path')
+var webpack       = require('webpack')
+var merge         = require('webpack-merge')
+var BundleTracker = require('webpack-bundle-tracker')
 
 var config = require('./shared.js')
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by guilleiguaran** [comment]
I like the idea, @kirs had something similar to track and raise Webpack errors on Rails so he might have a opinion on this.

**Comment by kirs** [comment]
We should totally have this check.

I had no idea there's `webpack-bundle-tracker` as a package and I used my own dirty JS code that syncs the bundle status.

One suggestion I'd like to address: this doesn't have to be a middleware. If the request is hitting API part of app, as a backend developer I don't really need to bundle the assets. I think it should better live in `javascript_pack_tag` and the check should only be invoked when someone is using webpack bundle on that page.

**Comment by drewhamlett** [comment]
@kirs I didn't know about webpack-bundle-tracker either until I saw the django-webpack repo.  It's from the same author.  This code is roughly based on that.

Good point on the middleware.  I guess the tag could pause execution of page.  That would be nice.

File: lib/install/config/development.js [hunk 1]
```diff
@@ -20,6 +21,8 @@ module.exports = merge(config, {
   plugins: [
     new webpack.LoaderOptionsPlugin({
       debug: true
-    })
+    }),
+
+    new BundleTracker({ path: __dirname, filename: '../../tmp/webpack-stats.json' })
```
abouthiroppy: You should use `path.join()` for windows :)
`filename: path.join('..', '..', 'tmp', 'webpack-stats.json')`

merqlove:
You can also use websocket's for such task :) Also seems that this must be dev. env only feature.

p0wl:
I'm not quite sure if I understand the purpose of this pr. Can't you use the webpack-dev-server? It blocks requests until the compilation is finished.

kirs:
@p0wl that's a great idea. Would you be interested in contributing to webpacker and adding webpack-dev-server?

p0wl:
@kirs webpack-dev-server is already part of webpacker and can be used with the `bin/webpacker-dev-server` binstub =)

ytbryan:
`./bin/webpack-dev-server` instead of `./bin/webpacker-dev-server`

Issue #52, opened 2017-01-08
Intentions regarding webpack-dev-server?


nathanstitt:
Hello,

I've also been experimenting with Ruby / Webpack integration.  My focus has mainly been on getting the developer experience similar to the existing asset-pipeline, but with the assets served via webpack-dev-server, so that source maps work and to utilize hot-reloading.

To that end I've created a few small wrappers for [yarn](https://github.com/nathanstitt/knitter) and [webpack](https://github.com/nathanstitt/ruby_webpack).  Both are extremely embryonic at the moment, and I'm quite happy to abandon them in favor of other solution(s) that are perhaps more established.

Do you envision `webpacker` controlling webpack-dev-server in any way?  My intended solution is to start it in a background process and provide asset tag helpers that would point to the appropriate address/port and paths.   I'd also like the solution to be Rails agnostic so it would work stand-alone, much like Sprockets does, since I also do some work with Sinatra.   If you see those goals as compatible, I'd love to drop the work I've done and contribute to `webpacker`.

Thoughts?

renchap:
I am also thinking about how to be able to use `webpack-dev-server` with Webpacker, specifically to enable hot releoading with my `webpacker-react` gem.
It would be great if we can use `webpack-dev-server` the same way `./bin/webpack-watcher` is currently used, and this would not need much : a config file and a change to `javascript_pack_tag` to load the assets from it.

I planned to do this from my gem, but it would be much better to have this support directly into Webpacker. If this is not desirable,  an alternate solution would be to have a `webpacker-dev-server` gem that would add this support and that you can depend on.

dhh:
I'm open to exploring the use of the dev-server, but I'm not convinced it's worth the extra complexity out the box. I want to make sure that everything is working just as well and easy as with the simple watcher. I heard several stories about hot loading not working too well in all scenarios.

But feel free to propose a PR and let's look at it!

renchap:
I agree this needs to be optional and only enabled if the user has a need for it (or if he uses a webpacker plugin that requires it).

I am thinking about a new `./bin/webpack-dev-server` stub as long as a config option for the user to put in `config/environments/development.rb` to set the dev server host / port (with a default). Probably worth displaying usage for this option when you launch the dev server through the stub if its not set (I dont know if there is a proper way for this?). This way it is all optional.

Would that be a good starting point for you?

dhh:
Yeah, let's give it a try and see how it feels.

renchap:
[...]

PR #55, opened 2017-01-10
Add support for webpack-dev-server

renchap:
As discussed in #52, here is a proposal.
Documentation needs to be done, but I wanted some feedback first.

This is the simplest implementation possible I think. An alternative to install `webpack-dev-server` when installing Webpacker would be to add a check in `bin/webpack-dev-server` and ask the user to add it to `vendor/package.json` and run Yarn.
Another solution would be to not install the binstub, and create a new `webpack::install::dev_server` generator.

Note: this does not enable Hot Module Reloading or anything special, it only serves the packs using `webpack-dev-server` running on port 8080 by default. Adding HMR requires changes to your Webpack config (and changes to your application code / React integration / ... to be really useful) and I would prefer the user or a plugin to do it (I plan to add this to my React integration gem).

Any thoughts ?

### DIFF
--- a/lib/install/bin/webpack-dev-server.tt
+++ b/lib/install/bin/webpack-dev-server.tt
@@ -1,0 +1,23 @@
+<%= shebang %>
+
+RAILS_ENV   = ENV['RAILS_ENV'] || 'development'
+WEBPACK_ENV = ENV['WEBPACK_ENV'] || RAILS_ENV
+
+APP_PATH    = File.expand_path('../', __dir__)
+VENDOR_PATH = File.expand_path('../vendor', __dir__)
+
+SET_NODE_PATH  = "NODE_PATH=#{VENDOR_PATH}/node_modules"
+WEBPACKER_BIN    = "./node_modules/.bin/webpack-dev-server"
+WEBPACK_CONFIG = "#{APP_PATH}/config/webpack/#{WEBPACK_ENV}.js"
+
+# Warn the user if the configuration is not set
+RAILS_ENV_CONFIG = File.join("config", "environments", "#{RAILS_ENV}.rb")
+
+# Look into the environment file for a non-commented variable declaration
+unless File.foreach(File.join(APP_PATH, RAILS_ENV_CONFIG)).detect { |line| line.match(/^\s*[^#]*config\.x\.webpacker\[\:dev_server_host\].*=/) }
+  puts "Warning: if you want to use webpack-dev-server, you need to tell Webpacker to serve asset packs from it. Please set config.x.webpacker[:dev_server_host] in #{RAILS_ENV_CONFIG}.\n\n"
+end
+
+Dir.chdir(VENDOR_PATH) do
+  exec "#{SET_NODE_PATH} #{WEBPACKER_BIN} --config #{WEBPACK_CONFIG} --content-base public/packs #{ARGV.join(" ")}"
+end
--- a/lib/webpacker/source.rb
+++ b/lib/webpacker/source.rb
@@ -8,7 +8,9 @@ class Webpacker::Source
   end
 
   def path
-    if digesting?
+    if config[:dev_server_host].present?
+      "#{config[:dev_server_host]}/#{filename}"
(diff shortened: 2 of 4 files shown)


lib/install/bin/webpack-dev-server.tt
```diff
@@ -0,0 +1,15 @@
...
+
+APP_PATH    = File.expand_path('../', __dir__)
+VENDOR_PATH = File.expand_path('../vendor', __dir__)
+
+SET_NODE_PATH  = "NODE_PATH=#{VENDOR_PATH}/node_modules"
+WEBPACKER_BIN    = "./node_modules/.bin/webpack-dev-server"
+WEBPACK_CONFIG = "#{APP_PATH}/config/webpack/#{WEBPACK_ENV}.js"
+
+Dir.chdir(VENDOR_PATH) do
+	exec "#{SET_NODE_PATH} #{WEBPACKER_BIN} --config #{WEBPACK_CONFIG} --content-base public/packs #{ARGV.join(" ")}"
```
guilleiguaran: extra indentation here

lib/install/config/shared.js
```diff
@@ -13,7 +13,7 @@ module.exports = {
     }, {}
   ),
 
-  output: { filename: '[name].js', path: path.join('..', 'public', 'packs') },
+  output: { filename: '[name].js', path: path.resolve('..', 'public', 'packs') },
```
guilleiguaran: I think this change is already done in master

lib/webpacker/source.rb
```diff
@@ -29,4 +31,8 @@ def digested_filename
     def filename
       "#{name}.js"
     end
+
+    def dev_server?
+      Rails.configuration.x.webpacker[:dev_server] && Rails.configuration.x.webpacker[:dev_server] != false
+    end
```
dhh: Instead of this, I'd like to add an accessor method like this:

```ruby
def config
  Rails.configuration.x.webpacker
end
```

Which we can then use like:

```ruby
def path
  if config[:dev_host].present?
    "#{config[:dev_host]}/#{filename}"
  elsif digesting?

  ...
end
```

That'll follow the pattern from `asset_host` we already use.

renchap:
Rebased to master and fixed your comments.

Do you think there should be a check in the binstub to point to the configuration if `Rails.configuration.x.webpacker[:dev_server]` is not set? What would be the most correct way to do this?

lib/webpacker/source.rb
```diff
@@ -8,7 +8,9 @@ def initialize(name)
   end
 
   def path
-    if digesting?
+    if config[:dev_server].present?
```
dhh: Let's use `dev_host` to match `asset_host`

dhh:
I think that would be a kind thing to do, but it would require booting all of Rails to provide that check. Not sure that's worth the trade-off.

lib/webpacker/source.rb
renchap: `dev_server_host` to make it clear this is about `webpack-dev-server` ?

dhh:
That works!

renchap:
Changed the config option to `dev_server_host` and added some documentation to README.

For the configuration check, I was thinking about reading `config/environment/#{RAILS_ENV}.rb` for something like `/[^#].*config\.x\.webpacker\[\:dev_server_host\].*=/`. Though I dont know if you want this kind of regexp check and this will not work if this is set elsewhere. An alternative would be to always print an information message before launching `webpack-dev-server`.

I dont have any preference here, I will let you decide what is the best.

dhh:
Great. I think the regexp check is fine for now 👍

renchap:
Here you go!

As a side note, `environment` in a generator inserts at the top of the environment file, just after `Rails.application.configure do`. Is there a reason for not appending at the end of the block? It would probably make the environment file look cleaner with base Rails options at first, then the ones from gems.

guilleiguaran:
@renchap thank you very much for this!!!

I agree we need to check the `environment` method in generators on Rails side.

PR #62, opened 2017-01-16
Use config digests path within rake

merqlove:
Subj

### DIFF
--- a/lib/tasks/webpacker.rake
+++ b/lib/tasks/webpacker.rake
@@ -1,17 +1,17 @@
-PACKS_PATH        = Rails.root.join('public/packs')
-PACK_DIGESTS_PATH = PACKS_PATH.join('digests.json')
-
 WEBPACKER_APP_TEMPLATE_PATH = File.expand_path('../install/template.rb', File.dirname(__FILE__))
 
 namespace :webpacker do
   desc "Compile javascript packs using webpack for production with digests"
-  task :compile do
+  task :compile => :environment do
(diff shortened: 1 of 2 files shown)


merqlove:
Looks like default value is needed for `:digests_path`

dhh:
One is provided in https://github.com/rails/webpacker/blob/f4cc31d445e54300d48e381b0381d87d31204588/lib/webpacker/railtie.rb#L15, but not initialized as such. So yeah, need to initialize that before using it here.

merqlove:
Ok, also i think fix up chunk names fetcher. 
In the case when we do production build with SourceMaps (i.e. for Rollbar, ...), here is `Array` per pack:
`["chunk name.js", "chunk_name.js.map"]`, instead of `"chunk name.js"`. 
I have this fix local already.

merqlove:
I have pushed few optimizations, maybe best to have it as single PR?

https://github.com/merqlove/webpacker/tree/develop

Features:
- Shared options per webpack config.
- Defaults for `:digests_path`.
- `:packs_dist_path`: Destination folder, useful way to extract production builds and/or change location of dev build.
- Fixed per chunk `array` problem, when we building SourceMaps in prod.

Cleanup

merqlove:
Link to compare: https://github.com/rails/webpacker/compare/master...merqlove:develop?expand=1

merqlove:
So, seems that https://github.com/rails/webpacker/pull/67 is better. I will close current for now,

PR #66, opened 2017-01-21
Can link to sprockets assets using rails-erb-loader

stevejackson:
Using rails-erb-loader, it's possible to use the sprockets asset helpers. 

It's not ideal to have to specify `helpers = ActionController::Base.helpers` at the beginning of each file, but it's the best simple solution I've thought of so far.

rails-erb-loader also seems to have problems with spring, so I've just disabled it.

I tested this in both development & production modes, and it works with and without sprockets adding digest hashes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,9 +93,7 @@ to your javascript file, then you can use Sprockets' asset helpers:
 var railsImagePath = "<%= helpers.image_path('rails.png') %>";
 ```
 
-This is enabled by the `rails-erb-loader` loader rule in `config/webpack/shared.js`. If you need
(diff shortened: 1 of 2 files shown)


kaspth:
I think we could get around the `helpers` assignment by going for something like: https://github.com/rails/rails/blob/92703a9ea5d8b96f30e0b706b801c9185ef14f0e/actionview/test/template/erb/helper.rb#L2-L17

In other words make `rails-erb-loader` adopt a lightweight `ViewContext`. Not sure if we can borrow one from Sprockets or if we have to write our own.

cc @guilleiguaran 😊

PR #67, opened 2017-01-21
Shared configs. Default variables. Fixed task. Cleanup. Composite patch.

merqlove:
So here it is. Something like composite patch.
Maybe some comments here.

- Shared options per webpack config. Useful to share same data between both config's, For example here i provide path which can (and seriously *must*) be used in build cleaner's.
- Defaults for `:digests_path`.
- `:packs_dist_path`: Destination folder, useful way to extract production builds and/or change location of dev build.
- Fixed per chunk array problem, it is when we builds SourceMap for prod.

And few cleanup's.

P.S. I can close my previous one.

### DIFF
--- a/lib/install/config/shared.js
+++ b/lib/install/config/shared.js
@@ -1,10 +1,16 @@
 // Note: You must restart bin/webpack-watcher for changes to take effect
 
 var path = require('path')
+var process = require('process')
 var glob = require('glob')
 var extname = require('path-complete-extname')
+var distPath = process.env.WEBPACK_DIST_PATH
 
-module.exports = {
+if(distPath === undefined) {
+  distPath = 'packs'
+}
+
+config = {
   entry: glob.sync(path.join('..', 'app', 'javascript', 'packs', '*.js*')).reduce(
     function(map, entry) {
       var basename = path.basename(entry, extname(entry))
@@ -13,7 +19,7 @@ module.exports = {
     }, {}
   ),
 
-  output: { filename: '[name].js', path: path.resolve('..', 'public', 'packs') },
+  output: { filename: '[name].js', path: path.resolve('..', 'public', distPath) },
 
   module: {
     rules: [
@@ -53,4 +59,9 @@ module.exports = {
   resolveLoader: {
     modules: [ path.resolve('../vendor/node_modules') ]
   }
+}
+
+module.exports = {
+  distPath: distPath,
+  config: config
 }
--- a/lib/tasks/webpacker.rake
+++ b/lib/tasks/webpacker.rake
@@ -1,17 +1,17 @@
-PACKS_PATH        = Rails.root.join('public/packs')
-PACK_DIGESTS_PATH = PACKS_PATH.join('digests.json')
-
 WEBPACKER_APP_TEMPLATE_PATH = File.expand_path('../install/template.rb', __dir__)
 
 namespace :webpacker do
   desc "Compile javascript packs using webpack for production with digests"
-  task :compile do
+  task :compile => :environment do
     webpack_digests_json = JSON.parse(`WEBPACK_ENV=production ./bin/webpack --json`)['assetsByChunkName'].to_json
 
-    FileUtils.mkdir_p(PACKS_PATH)
-    File.open(PACK_DIGESTS_PATH, 'w+') { |file| file.write webpack_digests_json }
+    digests_path = Rails.application.config.x.webpacker[:digests_path]
+    packs_path = File.dirname(digests_path)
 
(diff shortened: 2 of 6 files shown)


merqlove:
P.P.S. Also i can add few lines to the README, if needs.

lib/install/config/shared.js
```diff
@@ -1,10 +1,16 @@
 // Note: You must restart bin/webpack-watcher for changes to take effect
 
 var path = require('path')
+var process = require('process')
 var glob = require('glob')
 var extname = require('path-complete-extname')
+var distPath = process.env.WEBPACK_DIST_PATH
```
cgarvis:
```js
var DIST_PATH = process.env.WEBPACK_DIST_PATH || 'packs';
```

cgarvis: Is this different from `packs_dist_path` in `lib/webpacker/railtie.rb`?

lib/install/config/shared.js
```diff
@@ -53,3 +59,10 @@ module.exports = {
     modules: [ path.resolve('../vendor/node_modules') ]
   }
 }
+
+module.exports = () => { 
+  return {
+    distPath: distPath,
+    config: config
+  }
+}
```
cgarvis: I don't see any reason to return a function here.  It's not like these config values will/should change.

lib/install/config/shared.js
merqlove: Same. used to provide packs path from rails config into webpack configs.
For example i'm use it like this in production:
```js
new CleanWebpackPlugin([sharedConfig().dist], {
      root: path.resolve('..', 'public'),
      exclude: ['.keep']
  })
```

lib/install/config/shared.js
merqlove: Nice thought, thanks. Better if it be an object.

.gitignore
```diff
@@ -1 +1,2 @@
 /.bundle
+.idea
```
guilleiguaran: this should be in your global gitignore not in the project.

lib/install/config/shared.js
guilleiguaran: 👍

merqlove: Fixed

merqlove:
Fixed

guilleiguaran:
Sorry for the delay reviewing this, can you rebase it?

merqlove:
Hi, yay

merqlove:
Seems done. But for the future, looks like we need some specs, easy to mistake)

I think it can be demo webpack project, and some Task/Generator specs.

merqlove:
Hello guys, well @guilleiguaran @cgarvis,  anyone have time to look here?)

lib/tasks/webpacker.rake
```diff
@@ -3,16 +3,22 @@ WEBPACKER_APP_TEMPLATE_PATH = File.expand_path('../install/template.rb', __dir__
 namespace :webpacker do
   desc "Compile javascript packs using webpack for production with digests"
   task :compile => :environment do
-    webpack_digests_json = JSON.parse(`WEBPACK_ENV=production ./bin/webpack --json`)['assetsByChunkName'].to_json
+    dist_path = Rails.application.config.x.webpacker[:packs_dist_path]
+    webpack_digests = JSON.parse(`WEBPACK_DIST_PATH=#{dist_path} WEBPACK_ENV=production ./bin/webpack --json`)['assetsByChunkName']
+
+    js_webpack_digests_json = webpack_digests.each_with_object({}) do |(chunk, file), h|
+      h[chunk] = file.is_a?(Array) ? file.first : file
+    end.to_json
```
guilleiguaran: can you explain the changes done in the three lines above?

merqlove: This fix for case when you have to export source-map with your JS. In this case each chunk will be array [source.js, source.js.map]. So we have to cleanup it here.

merqlove: Any other questions?)

guilleiguaran: the file order is guaranteed to be always in that way?  (first the js and then the source map)

merqlove: Yes, because source always first.

merqlove: Other way is to filter by .js & pick first.

merqlove: Added this filter with some refresh.

merqlove: So now we have no dependency on file extension. Ability to use css, images, fonts and so on.

merqlove:
LGTM ?)

PR #68, opened 2017-01-23
Expose `RAILS_ENV` as `process.env.NODE_ENV`

afgomez:
As seen in [optimizing performance][1], React makes use of the `process.env.NODE_ENV` to conditionally exclude code when creating a production build.

This pull request uses exposes the value of `RAILS_ENV` inside the packs, so webpack can take advantage of it and make the production build smaller.

With a fresh Rails 5 app, this is the size before the change:
```
$ RAILS_ENV=production bin/webpack
Hash: 205a8b63e1f09ece9742
Version: webpack 2.2.0
Time: 1900ms
                              Asset     Size  Chunks                    Chunk Names
hello_react-205a8b63e1f09ece9742.js   728 kB       0  [emitted]  [big]  hello_react
application-205a8b63e1f09ece9742.js  3.01 kB       1  [emitted]         application
...
```

And this is after:
```
$ RAILS_ENV=production bin/webpack
Hash: b2c19f7c4e64877b870a
Version: webpack 2.2.0
Time: 1559ms
                              Asset     Size  Chunks                    Chunk Names
hello_react-b2c19f7c4e64877b870a.js   669 kB       0  [emitted]  [big]  hello_react
application-b2c19f7c4e64877b870a.js  3.01 kB       1  [emitted]         application
...
```

[1]: https://github.com/facebook/react/blob/30c7c4fbe6f35af93ddec0c7ae661d592821b296/docs/docs/optimizing-performance.md

### DIFF
--- a/lib/tasks/webpacker.rake
+++ b/lib/tasks/webpacker.rake
@@ -5,10 +5,10 @@ namespace :webpacker do
   desc "Compile javascript packs using webpack for production with digests"
   task :compile => :environment do
     dist_path = Rails.application.config.x.webpacker[:packs_dist_path]
-    result    = `WEBPACK_DIST_PATH=#{dist_path} WEBPACK_ENV=production ./bin/webpack --json`
-    
+    result    = `WEBPACK_DIST_PATH=#{dist_path} NODE_ENV=production ./bin/webpack --json`
+
     exit! $?.exitstatus unless $?.success?
-    
+
     webpack_digests = JSON.parse(result)['assetsByChunkName'].each_with_object({}) do |(chunk, file), h|
       h[chunk] = file.is_a?(Array) ? file.find {|f| REGEX_MAP !~ f } : file
     end.to_json
--- a/lib/install/bin/webpack-dev-server.tt
+++ b/lib/install/bin/webpack-dev-server.tt
@@ -2,14 +2,13 @@
 
 RAILS_ENV   = ENV['RAILS_ENV'] || 'development'
 NODE_ENV    = ENV['NODE_ENV'] || RAILS_ENV
-WEBPACK_ENV = ENV['WEBPACK_ENV'] || RAILS_ENV
 
 APP_PATH    = File.expand_path('../', __dir__)
 VENDOR_PATH = File.expand_path('../vendor', __dir__)
 
 SET_NODE_PATH  = "NODE_PATH=#{VENDOR_PATH}/node_modules"
 WEBPACKER_BIN    = "./node_modules/.bin/webpack-dev-server"
-WEBPACK_CONFIG = "#{APP_PATH}/config/webpack/#{WEBPACK_ENV}.js"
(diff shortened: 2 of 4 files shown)


cgarvis:
It would be great to get all environment variables.  You could do something like this:

```js
new webpack.EnvironmentPlugin(Object.keys(process.env))
```
See [docs](https://github.com/webpack/docs/wiki/list-of-plugins#environmentplugin)

Once you have this, I would set the `NODE_ENV` via the ruby task.  What do you think?

afgomez:
@cgarvis I can see its usefulness to expose things like API keys. Also, cases where you might want `NODE_ENV` to be different from `RAILS_ENV`, like having a `NODE_ENV=production` build in a `RAILS_ENV=staging` server.

afgomez:
I wonder if we now need the `WEBPACK_ENV` variable at all or we can just use `NODE_ENV` to get the same results

dhh:
This looks good to me.

p0wl:
I'm strongly against exposing ALL env variables, this may do much more harm than good (exposing secrets without even knowing).

The reason why you want to set `process.env` to `"PRODUCTION"` is explained in the very first comment: It is a known pattern in JS applications to rely on this env variable. So this is a common pattern, whereas exposing all env variables to the frontend is definitely not.

Please remove commit https://github.com/rails/webpacker/pull/68/commits/c01f062746cd53e62096a7b0d74a13e87166f52d

afgomez:
@p0wl it should not be a huge problem unless you explicitly try to access the variables. Webpack doesn't create a `process.env` object in the frontend with all your enviroment variables. It just tries to statically replace all uses of `process.env.MY_VAR` for whatever its value is in the final package.

...and as I write that paragraph I can see the potential for a security issue. Somebody could craft a malicious package that tries to access `process.env.DB_PASSWORD` for example.

I think we can explicitly expose `RAILS_ENV` and `NODE_ENV` for its usefulness, and allow the user to expose whatever else they want (facebook or error tracker API keys, for example).

@cgarvis you proposed the first change. What do you think?

p0wl:
why did you merge this? there is still an error (https://github.com/rails/webpacker/pull/68#discussion_r102157882). Now master is broken...

lib/tasks/webpacker.rake
```diff
@@ -5,7 +5,7 @@ namespace :webpacker do
   desc "Compile javascript packs using webpack for production with digests"
   task :compile => :environment do
     dist_dir = Rails.application.config.x.webpacker[:packs_dist_dir]
-    result   = `WEBPACK_DIST_DIR=#{dist_dir} WEBPACK_ENV=production ./bin/webpack --json`
+    result   = `WEBPACK_DIST_DIR=#{dist_dir} NODE_ENV_ENV=production ./bin/webpack --json`
```
p0wl: one `_ENV` too much.

Issue #41, opened 2016-12-22
third party tools blowing up trying to find package.json and node_modules


denniscollective:
Howdy! Thanks for the tool! I'm not trying to reopen a bike-shed, but don't know how to deal with the issues that this is raising. I'm down for any solution that works :)

I'm trying to get eslint set up and it's blowing up, trying to find the package.json.

I got it set up another way, and then when it was linting, it was raising a bunch of "cant_resolve module" issues, which I'm assuming is because node_modules is buried in vendor, but I could be jumping to a conclusion.

My webpack knowledge is very limited (which is why I'm using this tool), is there some magic env variable I should be using to point the tooling at package.json in a certain way? Should I just be running this from within /vendor ?

If you can point me in the right direction if you already have a workaround, I could help document it better, otherwise, heads up, this is happening.

I'm super psyched on how all of this is evolving.

```
sparkly$ vendor/node_modules/.bin/eslint --init
? How would you like to configure ESLint? Answer questions about your style
? Are you using ECMAScript 6 features? Yes
? Are you using ES6 modules? Yes
? Where will your code run? Browser
? Do you use CommonJS? Yes
? Do you use JSX? Yes
? Do you use React? Yes
? What style of indentation do you use? Spaces
? What quotes do you use for strings? Single
? What line endings do you use? Unix
? Do you require semicolons? No
? What format do you want your config file to be in? JavaScript
Could not find a package.json file. Run 'npm init' to create one.
Error: Could not find a package.json file. Run 'npm init' to create one.
    at check (/Users/dennis/workspace/sparkly/vendor/node_modules/eslint/lib/util/npm-util.js:76:15)
    at Object.checkDevDeps (/Users/dennis/workspace/sparkly/vendor/node_modules/eslint/lib/util/npm-util.js:124:12)
    at installModules (/Users/dennis/workspace/sparkly/vendor/node_modules/eslint/lib/config/config-initializer.js:79:35)
    at processAnswers (/Users/dennis/workspace/sparkly/vendor/node_modules/eslint/lib/config/config-initializer.js:245:5)
    at null.completed (/Users/dennis/workspace/sparkly/vendor/node_modules/eslint/lib/config/config-initializer.js:468:30)
    at PromptUI.onCompletion (/Users/dennis/workspace/sparkly/vendor/node_modules/inquirer/lib/ui/prompt.js:57:10)
[...]

PR #84, opened 2017-02-06
Add option to use rails root for node

dleavitt:
I was reading through rails/rails#27245, and thought the best and most "Railsey" solution would be to use `vendor` by default but make other options easy as well.

This PR adds a new rake task, `webpacker:install:root`, which uses `Rails.root` as the root directory for yarn, installing `node_modules`, `package.json` and `yarn.lock` there.

`webpacker:install` continues to function as it always has, installing to `vendor`. 

This change is backwards compatible. It just adds some conditionals to the install template; none of the runtime or asset compilation code is affected. All of that as well as all of the instructions in the README still just work, whether you install to vendor or to root.

I think this is a good solution in that it leaves `vendor` as the default but puts the choice in the hands of early adopters, allowing you guys to gather feedback on both approaches.

**Why it's important to have this choice**

I know this has already been discussed, but I want to reiterate why it's important to have this as an option. The Javascript ecosystem has matured, and having `package.json` and `node_modules` in the project root (or at least above all JS code) is one of its few universals. Because of this:
- Using `vendor` breaks most existing tooling. It can be worked around in some cases, but not in others. E.g. I don't think there's any way at the moment to get Atom's `eslint` package to handle this without major nastiness. Bundling one-off solutions for every popular use case (eg #41) is not a great solution.
- Using `vendor` breaks `package.json` scripting, JS land's `bundle exec` / `rake` / binstubs. Instead JS commands have to be wrapped in bin scripts or rake tasks that call out to Javascript. This is gross to implement and especially painful on JRuby. Wouldn't it be nice just to run `yarn start` to boot up the frontend?
- If you're using webpack, your frontend is likely just as serious as your backend. Using `vendor` marginalizes the frontend.
- Using `vendor` raises the barrier to getting Javascript developers excited about Rails. Their tooling doesn't work, their conventions don't apply, they need to edit obscure Ruby files to do basic things they could otherwise do straightforwardly in JS. Rails has gotten a reputation as being behind the curve on modern frontend development. This won't help.

**Other notes / topics for bikeshedding**
- There aren't tests for this repo yet, so I made a little fixture app here to demonstrate that this works: https://github.com/dleavitt/webpacker-testcase
- Is `webpacker:install:root` the cleanest way to do this? Should it be a flag or something instead?
- Not sure what the best practice is for passing info to the `.tt` files. Currently just using an ENV var.
- Not sure if my variable names are the best.
- `webpacker:install:angular` won’t quite work out of the box after `webpacker:install:root` - a couple paths will need to be adjusted in the generated tsconfig file.
- This will probably necessitate corresponding changes to Rails’s yarn generator.

In summary:

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,13 @@ As the rubygems version isn't promised to be kept up to date until the release o
 gem 'webpacker', github: 'rails/webpacker'
 ```
 
+
+## Installing outside of vendor
+
+By default, `webpacker:install` uses `vendor` as the root path for yarn. This can cause problems for tooling, node modules and Javascript developers, all of whom expect Javascript source code to be in a subdirectory of the yarn root path.
+
+To install to the application root, run `bin/rails webpacker:install:root`. Everything else should work normally.
+
 ## Binstubs
 
 Webpacker ships with three binstubs: `./bin/webpack`, `./bin/webpack-watcher` and `./bin/webpack-dev-server`.
--- a/lib/install/bin/webpack-dev-server.tt
+++ b/lib/install/bin/webpack-dev-server.tt
@@ -1,12 +1,14 @@
 <%= shebang %>
 
+<% yarn_root = ENV['WEBPACKER_TARGET'] == 'root' ? 'APP_PATH' : 'VENDOR_PATH' %>
+
 RAILS_ENV   = ENV['RAILS_ENV'] || 'development'
 WEBPACK_ENV = ENV['WEBPACK_ENV'] || RAILS_ENV
 
 APP_PATH    = File.expand_path('../', __dir__)
-VENDOR_PATH = File.expand_path('../vendor', __dir__)
+<%= "VENDOR_PATH = File.expand_path('../vendor', __dir__)" unless ENV['WEBPACKER_TARGET'] == 'root' %>
 
-SET_NODE_PATH  = "NODE_PATH=#{VENDOR_PATH}/node_modules"
+SET_NODE_PATH  = "NODE_PATH=#{<%= yarn_root %>}/node_modules"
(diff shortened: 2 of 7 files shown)


pstjean:
I want to chime in to agree on the subject of `vendor` breaking existing tooling. Even very mature and well maintained projects treat projects without `package.json` and `node_modules` in the root as second class citizens.

I spent the day yesterday configuring Jest testing for a vanilla Rails and Webpacker setup, after seeing a week old, unanswered [question on Stack Overflow](http://stackoverflow.com/questions/42164582/react-testing-setup-with-rails-webpacker) inquiring how to get JavaScript testing (of any description) working alongside Webpacker. I figured it couldn't be that hard. I was completely wrong.

I wrestled with documentation in both the Jest and Babel projects that was self-contradictory and outdated. We can self-righteously condemn the maintainers of these projects for inadequate documentation, but the simple fact of the matter is that only a very tiny minority of their users ever use their product in a project that doesn't follow the convention of placing `package.json` and `node_modules` in the project root. They don't prioritize these users, and they can't be expected to.

If this is the situation with two extremely well maintained and highly regarded projects, it's easy to see why a lot of smaller Node projects simply omit support for non-standard configurations like the one implemented by Webpacker. We'll be left in a situation where we're forking JavaScript libraries to support the peculiar Rails way of handling JavaScript dependencies, and at that point we're right back at JavaScript gems.

dhh:
[Going back to root for everything](https://github.com/rails/rails/pull/28093). Please do update this PR just assume the root 👍

dleavitt:
Will do! Thanks!

dleavitt:
@guilleiguaran @dhh closing this, updated pr is #101

PR #85, opened 2017-02-07
Fix angular/typescript compilation issue

mikerob215:
When compiling typescript with webpack using the --json option, the output will have a string saying the name of the loader before the expected object. This causes JSON.parse to fail. Fix removes the string before the object.

### DIFF
--- a/lib/tasks/webpacker.rake
+++ b/lib/tasks/webpacker.rake
@@ -6,7 +6,12 @@ WEBPACKER_APP_TEMPLATE_PATH = File.expand_path('../install/template.rb', __dir__
 namespace :webpacker do
   desc "Compile javascript packs using webpack for production with digests"
   task :compile do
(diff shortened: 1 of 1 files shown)


spion:
Another option is to add

```
options: {
          silent: true
}
```
to the `ts-loader` rule.

guilleiguaran:
Can we do what @spion suggest instead?

If this start to get more complex we will need to use a Manifest plugin like https://www.npmjs.com/package/webpack-manifest-plugin

p0wl:
I'd also argument for using the [Manifest Plugin](https://github.com/danethurber/webpack-manifest-plugin) or the [Asset Plugin](https://github.com/kossnocorp/assets-webpack-plugin) (both are basically doing the same.

mikerob215:
Issue was resolved

PR #92, opened 2017-02-10
Fix dependent railties & activesupport version

tricknotes:
Since https://github.com/rails/webpacker/pull/12, webpacker only supports Rails 5.x.
Also activesupport too.


webpacker.gemspec
```diff
@@ -9,7 +9,7 @@ Gem::Specification.new do |s|
 
   s.required_ruby_version = '>= 1.9.3'
 
-  s.add_dependency 'activesupport', '>= 3.0.0', '< 5.1'
+  s.add_dependency 'activesupport', '~> 5'
```
rafaelfranca: `~> 5` is meaningless. Can you change to `>= 5.0`? Same on railties.

tricknotes: Ah, right!
Now I updated this PR.

PR #95, opened 2017-02-13
Add .babelrc when installing react

danbroooks:
React install was missing `.babelrc` so was throwing errors when I was running the build (unexpected `<` when getting to the JSX). This PR adds copying a `.babelrc` to the project root as part of the react install.

### DIFF
--- a/lib/install/react/.babelrc
+++ b/lib/install/react/.babelrc
@@ -1,0 +1,3 @@
+{
+  "presets": ["es2015", "react"]
+}
--- a/lib/tasks/webpacker.rake
+++ b/lib/tasks/webpacker.rake
@@ -46,6 +46,10 @@ namespace :webpacker do
 
       File.write config_path, config
 
+      puts "Copying .babelrc to project directory"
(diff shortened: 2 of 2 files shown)


notrab:
I'd personally avoid another dotfile. I know how much @dhh loves them but could we not go down the approach of specifying this from within the [`package.json`](https://babeljs.io/docs/usage/babelrc/#use-via-package-json)?

sealocal:
I probably put my comment in the wrong place. Any thoughts? 

https://github.com/rails/webpacker/commit/c99f9ede27a6dddaae5cede8ab25dc9fb0c9eccb#commitcomment-21021014

sealocal:
Never mind. The output does not direct me to make changes if starting with a new Rails app.

```console
$ rails webpacker:install:react
Replacing loader presets to include react in [my_app]/config/webpack/shared.js
Replacing loader test to include react in [my_app]/config/webpack/shared.js
Copying .babelrc to project directory
```

Issue #81, opened 2017-02-04
`bin/rake webpacker:compile` fails with angular example


taboularasa:
rails (5.0.1)
webpacker (0.2)

Steps to reproduce:

```shell
$ rails new foo
$ cd foo
$ bin/rake webpacker:install
$ bin/rake webpacker:install:angular
$ bin/rake webpacker:compile
```
compile fails with the following output:

```shell
** Invoke webpacker:compile (first_time)
** Execute webpacker:compile
rake aborted!
JSON::ParserError: 743: unexpected token at 'ts-loader: Using typescript@2.1.5 and /Users/taboularasa/work/foo/tsconfig.json
...
```

I'm willing an motivated to get this working but could really use some input from anyone more familiar

joerodrig3:
I spent a little time looking at this and it looks like this is being caused by the require statement in the `hello_angular.js` file that's being generated at [line 7](https://github.com/rails/webpacker/commit/331deaeb5a5a9ac25d3be886e72d4a89b35b7abd#diff-cfabf609842d53ffbfb34e4508592450R7). I'm not very familiar with Angular, but that require statement doesn't look like it fits into that file. I was able to get the app to compile by moving the require statement to the application.js and changing it to an import statement.
ie `import HelloAngular from "./hello_angular"`.

taboularasa:
@joerodrig3 thanks for taking some time to look at this! Thats a good observation you made about `hello_angular.js:7` requiring itself, I don't understand what the intention is there. I was able to reproduce your suggested fix by removing the require to `application.js` that does allow `bin/rake webpacker:compile` to succeed but that's also just a noop, the app is never actually booted. `application.js` is then requiring an empty file. The other curious detail is the app does successfully run locally without any modification via `bin/webpack-watcher`

I think I'll look at some other Angular 2 / webpack examples and see how it's being done there.

mikerob215:
Ran into this issue in a work project with our own webpack integration. The issue is that when you use ts-loader, and compile using webpack --json, the output will contain that string that caused the error in the op and then the object that the webpack:compile rake task is looking for. I've had this also happen with awesome-typescript-loader, I'm not sure what other loaders could cause this. I've submitted a pr that will cause the rake to task to ignore any string that precedes the JSON it is looking for.

gauravtiwari:
Does this work now @taboularasa? There has been some updates lately, which I think fixes this problem. Please close this issue if it does :)

taboularasa:
[...]

PR #101, opened 2017-02-21
Use the project root (not vendor) as the yarn root

dleavitt:
This replaces #84, hardcoding the rails root (rather than vendor) as the node/yarn root.

There aren't yet tests for this gem, but I made a fixture app here that can be used to test that everything is working: https://github.com/dleavitt/webpacker-testcase

Both angular and react presets should still work, though the the angular preset won't compile properly because of #81.

This PR makes it possible to remove the javascript binstubs entirely if desired - all yarn/webpack stuff should now be runnable directly or via package.json scripts.


p0wl:
thanks for this. this makes deploying to heroku much easier, because the heroku buildpack nodejs requires a package.json in the app root directory. We worked around that, but with your pr we can remove these ugly hacks \o/

PR #110, opened 2017-02-24
es6 variables and style changes

5minpause:
Since you use ES2017 the variable declarations should be made using `let` and `const` instead of `var`. Also in the webpack configuration files and in the react-files there were semicolons missing from the end of the lines. I do not know whether this was intentional.
Since it is marked as a warning when using the AirBNB ESlint configuration (which is widely used), I added these.

Let me know what you think.

### DIFF
--- a/lib/install/config/development.js
+++ b/lib/install/config/development.js
@@ -1,9 +1,9 @@
 // Note: You must restart bin/webpack-watcher for changes to take effect
 
-var webpack = require('webpack')
-var merge   = require('webpack-merge')
+const webpack = require('webpack')
+const merge   = require('webpack-merge')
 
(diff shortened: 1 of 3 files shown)


dhh:
Omitting semicolons is intentional, but we should indeed use `const`. If you want to pare down the PR for just that change, I can merge 👍. Thanks!

5minpause:
Great, I removed the commit that added the semicolons. 
Would you mind explaining the rationale behind your intentions regarding semicolons?

dhh:
Awesome, thanks! Omitting semicolons for aesthetics. None of the ES6 code I personally write has semicolons, unless I explicitly need to disambiguate something.

PR #114, opened 2017-02-25
Minify production builds with uglifyjs

schpet:
for smaller production builds and to support tree shaking (#36)

### DIFF
--- a/lib/install/config/production.js
+++ b/lib/install/config/production.js
@@ -11,6 +11,7 @@ module.exports = merge(sharedConfig.config, {
   plugins: [
     new webpack.LoaderOptionsPlugin({
       minimize: true
-    })
+    }),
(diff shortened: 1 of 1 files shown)


guilleiguaran:
wait, one second...

The files were being compressed already before of merging this?

schpet:
@guilleiguaran i don't think the files were being compressed by default for some reason. you either have to run webpack with `-p` or`--optimize-minimize` flag, or set this up in the webpack.config. i can try and confirm this behavior with a demo if that would be helpful to you.

schpet:
#132 has  this same change, and also gzipping which makes sense

guilleiguaran:
yup, I didn't noticed #132 before, now it needs to be rebased before of merge, will try to do it now

PR #121, opened 2017-02-26
Removing version number to fix a compile issue with hello_vue.js

ytbryan:
Previously, [a version number (2.1.10)](https://github.com/rails/webpacker/blob/master/lib/tasks/webpacker.rake#L137
) was specified as Vue pointed to 2.2.0.beta1 (which breaks webpack compilation).  

[This PR](https://github.com/rails/webpacker/pull/121/files) removes that version number. 

Users are encouraged to use `./bin/yarn` to control the package version in the event of dependencies breaking default pack again.


guilleiguaran:
already merged a similar change in #122

PR #123, opened 2017-02-26
Allow Rails 4.x support

nickpith:
Technically, nothing in webpacker is dependent on Rails 5.x.
Although there is only official support for Rails 5.1, this
minor change allows the usage of webpacker for Rails 4.x. This
is helpful for enterprise users of Rails that are still working
on supporting Rails 5.x but really want to use the new conventions
for supporting Webpack before fully upgrading to Rails 5.1.

### DIFF
--- a/lib/tasks/webpacker.rake
+++ b/lib/tasks/webpacker.rake
@@ -26,7 +26,11 @@ namespace :webpacker do
 
   desc "Install webpacker in this application"
   task :install do
-    exec "./bin/rails app:template LOCATION=#{WEBPACKER_APP_TEMPLATE_PATH}"
+    if Rails::VERSION::MAJOR >= 5
+      exec "./bin/rails app:template LOCATION=#{WEBPACKER_APP_TEMPLATE_PATH}"
+    else 
+      exec "./bin/rake rails:template LOCATION=#{WEBPACKER_APP_TEMPLATE_PATH}"
+    end
   end
 
   namespace :install do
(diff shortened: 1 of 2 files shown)


qrush:
Big +1. Test suite should perhaps use the Appraisals gem to ensure it works on both.

rafaelfranca:
I really don't want to support Rails versions that are not supported anymore in new gems. I know the gem works with rails 4.0 but 4.0, 4.1 and soon 4.2 are not supported anymore by the Rails team.

nickpith:
@rafaelfranca That makes sense. Sounds like those that would like that support will just fork the repo to allow it.

rafaelfranca:
I think it is fine to support 4.2. Can you change the PR?

nickpith:
Yeah, I can change the PR later tonight.

nickpith:
Updated the restriction to 4.2

rafaelfranca:
Thanks!

aharpervc:
Should the readme be updated to reflect that this now works with 4.2+ instead of 5.1+?

nickpith:
I don't believe it should be updated because that is not the official stance

aharpervc:
The approval of this pr suggests otherwise

guilleiguaran:
Let's say that we will do our best effort to support 4.2 but it is not guaranteed and can be dropped if in future versions we found compatibility problems.

nickpith:
OK. I can put up a new PR with that wording

Issue #117, opened 2017-02-25
Webpack executable not found on production


gauravtiwari:
Related to #116. Basically, when using the --production flag, or NODE_ENV=production environment variable, Yarn will not install any package listed in devDependencies. https://yarnpkg.com/lang/en/docs/cli/install/#toc-yarn-install-production

In our config files we are using packages that also needs to be available on production to compile the assets. Therefore, the dependencies which are used in production should be moved to dependencies block instead of devDependencies. 

I can make a PR once #111 is merged

dleavitt:
Background: right now we're putting build deps under `devDependencies` even though they're required in production.

Pretty sure that setting `NPM_CONFIG_PRODUCTION=false` on production fixes this problem. This should probably be hardcoded into the rake task, in fact.

I think the "correct" solution would be to move everything under `dependencies` rather than `devDependencies`. Right now they're being misused in a way that's convenient (and common) but that doesn't actually reflect their semantics.

justin808:
@dleavitt The only things that should go under devDependencies are things like eslint, flow, the webpack dev server. Any build tools need to be under "dependencies.

dleavitt:
Indeed, that's what I was (trying to) say above. Sure makes package.json prettier when you do it the wrong way though.

dhh:
Yes let's fix the root issue and move away from errant use of devDependencies.

elliotjyoung:
What about Webpack plugins (non babel)? Such as loaders?

PR #126, opened 2017-02-26
Add more deployment help to README

dleavitt:
Heroku as well as other deployment strategies don't currently work out of the box, so I have further bloated the README with some instructions on how to make them work. See #117 for an example of the issue and what might be a better ultimate solution.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,11 @@ will invoke the production configuration, which includes digesting. The `javascr
 method will automatically insert the correct digest when run in production mode. Just like the asset
 pipeline does it.
 
+### Troubleshooting
+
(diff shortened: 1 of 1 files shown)


p0wl:
I'm against changing `NPM_CONFIG_PRODUCTION`, all dependencies should be moved from `devDependencies` to `dependencies`, see https://github.com/rails/webpacker/issues/117

:+1: for the note on the heroku buildsets. I also thought about adding a hint that webpacker hooks into the `assets:precompile` automatically ([here](https://github.com/rails/webpacker/blob/master/lib/tasks/webpacker.rake#L143-L148)), because the current readme sounds like you have to invoke `webpacker:compile` manually.

dleavitt:
@p0wl I think I agree on `NPM_CONFIG_PRODUCTION` - I just think it should be documented until it's been fixed (it's a frequent point of confusion.)

dleavitt:
Not need because #137

Issue #109, opened 2017-02-23
Add option to have css/images managed through webpack


dleavitt:
I think it would be useful to have an additional generator that supports _all_ FE assets being managed through Webpack. I'd love to gauge interest and get feedback before I start on a PR.

Background:

As far as I know there, are two basic options for managing images/css/etc in a webpack + rails app.

**Option 1:** Manage Javascript through Webpack and everything else through Sprockets.
This is only option currently supported out of the box.

- **Advantages:** Easy to set up, legacy app friendly, easy to reference images, etc. from erb.
- **Disadvantages:** Still clunky (as noted in README). Still no good system for managing vendor CSS dependencies. Doesn't make CSS a first class citizen of the front end app, which means no loading css from node_modules and no support for modular CSS approaches (used by Vue, Material Design, etc.)

**Option 2:** Manage _all_ assets through Webpack, remove Sprockets.

This involves adding some loaders to webpacker (probably file-loader and sass/css/style loaders) as well as a manifest generator that Rails can use to get properly fingerprinted paths in erb and such.
  - **Advantages:**
    - Allows easy use of Vue, CSS modules, and lots of other things that frontend developers dig.
    - More broadly, this is the way most frontend apps are structured nowadays. Tutorials and example apps will work out-of-the-box with Rails.
    - Removes Sprockets entirely for apps that don't need it, no more need for two separate asset pipelines. No need to mess with bower or https://rails-assets.org.
   - Fixes a bunch of things in one stroke, eg #91, #79, etc.
  - **Disadvantages:**
    -  Ultimately need to modify or duplicate all existing asset helpers to work with the webpack-generated manifest.
    - Additional webpack complexity around manifest generation and CSS extraction and other things we want production-ready defaults for. Potentially more moving pieces on the webpack side.

I'd be interested in creating a generator to support option 2 (either as an alternative to `webpacker:install` or as an additional generator that can be run after it.) Option 1 would be left as the default.

I think this feature could also provide the first step on a gradual path to moving Sprockets out of 
core. I think the technical risk of relying on Webpack (or another successor library) is less than than the risk of relying on Sprockets. Sprockets is a very complex library (and I suspect a maintenance pain point.) It was a major step forward for its time but is showing its age; most of its functionality is better provided by Webpack, which will soon be part of Rails.

dhh:
Happy to see a PR explore option #2 👍

dleavitt:
Great, I'll get started!

dhh:
[...]

PR #127, opened 2017-02-27
Fix distDir layout for advanced webpack features

dleavitt:
Right now this gem is built on the assumption that `app/javascript/packs` will always match `public/packs`. This holds when the JS entry points are the only files emitted and avoids users having to mess with `publicPath` in the webpack config.

But this isn't going to be the case for most webpack-based apps of any complexity. Eg, file-loader breaks this assumption. I believe it gets broken under a number of other scenarios as well: commons chunks, ExtractTextPlugin, some kinds of code splitting, etc. 

These are not scenarios we need to support out-of-the-box, but per #109 it makes sense to set up defaults that make them possible.

The easiest solution I could think of was to make the structure of the "packs" dir match the structure of `app/javascript`.

Let's say your `app/javascript` dir looked like this:

```
app/javascript
+-- images
|   +-- image.png
+-- packs
|   +-- application.js
|   +-- admin.js
+-- nonentry.js
```

If you ran `bin/webpack`, your `public/packs` dir would look like this:

```
public/packs
+-- images
|   +-- image.png
+-- application.js
+-- admin.js
```

Doh! Images is now a sibling of your pack files! The structure doesn't match the structure of `app/javascript`, causing the need for additional configuration and pain. Instead let's make it look like this:

```
public/dist
+-- images
|   +-- image.png
+-- packs
|   +-- application.js
|   +-- admin.js
```

Nice, it matches! I also renamed the top-level dir from `packs` to `dist` to make clear that it's _not_ the equivalent of the `packs` dir in `app/javascript`, and because `assets` was already taken.

Thoughts? Any lower-impact alternatives?

cc @gauravtiwari

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ If you'd rather not have to run the two processes separately by hand, you can us
 Alternatively, you can run `./bin/webpack-dev-server`. This will launch a
 [Webpack Dev Server](https://webpack.github.io/docs/webpack-dev-server.html) listening on http://localhost:8080/
 serving your pack files. It will recompile your files as you make changes. You also need to set
-`config.x.webpacker[:dev_server_host]` in your `config/environments/development.rb` to tell Webpacker to load
+`config.x.webpacker[:dev_server_host]` in your `config/environments/development.rb` and uncomment the `output.publicPath` line in your `config/webpack/development.js` to tell Webpacker to load
 your packs from the Webpack Dev Server. This setup allows you to leverage advanced Webpack features, such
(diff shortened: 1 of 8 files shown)


gauravtiwari:
@dleavitt Good idea, however shall we wait until we have the other changes we discussed in #109 in place in a PR so, we can bring it all together and discuss it through.

dleavitt:
Yeah, I left this open because I was hoping to get some feedback on this from somebody with knowledge of webpack internals. Haven't gotten around to poking people yet though.

PR #132, opened 2017-02-27
Minify and compress production assets

AndrewHYi:


schpet:
is the `compression-webpack-plugin` available? i assume it would need to be added to the package.json

PR #135, opened 2017-02-28
Make jsx files actually compile with webpack

the-architect:
bin/webpack-dev-server would fail to compile the hello_react.jsx file with the default config, so I added this and it fixed the issue. 
Don't know if I overlooked something else though :)


lib/install/config/shared.js
```diff
@@ -28,7 +28,7 @@ const config = {
     rules: [
       { test: /\.coffee(\.erb)?$/, loader: 'coffee-loader' },
       {
-        test: /\.js(\.erb)?$/,
+        test: [/\.js(\.erb)?$/, /\.jsx$/],
```
dleavitt: should be `/\.jsx?(\.erb)?$/` I think

the-architect: This works too. I'll update. Thank you!

guilleiguaran:
We are doing this in React installer: https://github.com/rails/webpacker/blob/master/lib/tasks/webpacker.rake#L54

the-architect:
Well that's weird, because I ran this 👍  Obviously I did something wrong when installing it.

guilleiguaran:
Can you test again and see if the file is updated or if you get a `Couldn't automatically update loader test` message?

the-architect:
```Bash 
Couldn't automatically update loader test in .../application/config/webpack/shared.js. Please set test: /.jsx?(.erb)?$/.
```
Yes indeed I got the message now.

the-architect:
Found that the check was missing some backslash escapes.

the-architect:
squashed

lib/install/config/shared.js
```diff
@@ -28,7 +28,7 @@ const config = {
     rules: [
       { test: /\.coffee(\.erb)?$/, loader: 'coffee-loader' },
       {
-        test: /\.js(\.erb)?$/,
+        test: /\.jsx?(\.erb)?$/,
```
guilleiguaran: I think we shouldn't have `jsx` here by default or people will think that this is going to work even when the `react` preset isn't set for Babel and it is not by default

lib/tasks/webpacker.rake
```diff
@@ -49,11 +49,11 @@ namespace :webpacker do
         puts "Couldn't automatically update loader presets in #{config_path}. Please set presets: [ 'react', [ 'latest', { 'es2015': { 'modules': false } } ] ]."
       end
 
-      if config.include?("test: /\\.js(.erb)?$/")
+      if config.include?("test: /\\.js(\\.erb)?$/")
         puts "Replacing loader test to include react in #{config_path}"
-        config.gsub!("test: /\\.js(.erb)?$/", "test: /\\.jsx?(.erb)?$/")
+        config.gsub!("test: /\\.js(\\.erb)?$/", "test: /\\.jsx?(\\.erb)?$/")
```
guilleiguaran: 👍

justin808:
My guess is that you probably don't want to include anything regarding the use of optional front end frameworks, etc. in this gem as it's going to be a never ending slippery slope of things getting out of date.

lib/install/config/shared.js
the-architect: Agreed, it was just the first file I looked into without completely understanding the process.
The rake task will now do it and this change is not necessary anymore. I will adjust the PR.

the-architect:
removed the change in shared.js Thank you guys for the feedback!

Issue #103, opened 2017-02-22
Should "rails new --webpack=..." stop when bin/yarn fails?


marcusmolchany:
Locally I was running node version `5.0.0` which caused an error during the `yarn install` step of `rails new`.

loader-runner (a sub-dependency of webpack) failed because of its [`engines` requirement in package.json](https://github.com/webpack/loader-runner/blob/c875814f169348ef712b78e22c6404a7f585df3f/package.json#L30-L32). This prevented `./bin/webpack-watcher` from running because it was not installed in the `node_modules` folder.

My question is should the `rails new --wepback=...` command fail if there is an issue during the `yarn install`? After `rails new` ran and completed, with errors further up in the logs, it took some time to trace back why `./bin/webpacker-watcher` was not running.

---

Below is the full output of my rails new commands (omitted large output for brevity using `[...]`). The current working directory of the command is the root directory of cloning `rails/rails` from github. The error occurs in the logs after `run  ./bin/yarn add --dev webpack webpack-merge webpack-dev-server path-complete-extname babel-loader babel-core babel-preset-latest coffee-loader coffee-script rails-erb-loader glob from "."`

```sh
$ ./rails/railties/exe/rails new my5oneapp --edge --webpack=react
      create  
      create  README.md
      create  Rakefile
      [...]
      create  vendor/.keep
      create  package.json
      remove  config/initializers/cors.rb
         run  bundle install
Fetching https://github.com/rails/rails.git
[...]
Fetching dependency metadata from https://rubygems.org/.
Resolving dependencies...
Using rake 12.0.0
Using concurrent-ruby 1.0.4
[...]
Using webpacker 0.2 from https://github.com/rails/webpacker.git (at master@35586c7)
Using rails 5.1.0.alpha from https://github.com/rails/rails.git (at master@ecca24e)
Using sass-rails 6.0.0.beta1 from https://github.com/rails/sass-rails.git (at master@dfbcc6a)
Bundle complete! 17 Gemfile dependencies, 70 gems now installed.
Use `bundle show [gemname]` to see where a bundled gem is installed.
       rails  webpacker:install
      create  app/javascript
      create  app/javascript/packs/application.js
       exist  bin
      create  bin/webpack-dev-server
      create  bin/webpack-watcher
      create  bin/webpack
    conflict  bin/yarn
Overwrite /Users/mmolchany/development/my5oneapp/bin/yarn? (enter "h" for help) [Ynaqdh] 
       force  bin/yarn
[...]

PR #149, opened 2017-03-02
fail hard when webpacker:install ./bin/yarn fails

sealocal:
This is the most simple solution I could compose for #103, similar to [this method](https://github.com/rails/webpacker/blob/9ef79dd9d1a5ab0a2e9ada35b471a81bbe1ab132/lib/tasks/webpacker/compile.rake#L9).

Does it make sense to have two separate failure checks, or should there be only one check after both `./bin/yarn add [...]` commands have run?  Any other thoughts?


gauravtiwari:
@sealocal Hmm, wouldn't it be simple instead to just check for yarn executable and fail if not present? The rest of the commands wouldn't fail anyway unless there is some network issue or if module doesn't exist, and yarn will take care of that. Also, I guess it needs to be added here in the install task, so, the task doesn't run up to that point. 

In the last comment David suggested something similar, think this needs to be added to app_base generator as well in Rails so, it works for both (with --webpack option and independent gem install).

What do you think?

sealocal:
I think maybe there's a little confusion around what's failing.  I belive there are two problems:

`rails new my_app --webpack=[...]` fails when

  1. yarn is not installed
  2. when webpack and its dependencies fail to install via `./bin/yarn add [...]`

If I understand, `rails new my_app --webpack[...]` simply delegates to the webpacker gem by installing the gem and calling `rails webpacker:install` (delegation [here](https://github.com/rails/rails/blob/d50380f4e3f9e8d9398c8fdcbe5f52f784d14f5f/railties/lib/rails/generators/app_base.rb#L429), described [here](https://github.com/rails/rails/blame/88857f33956512270bbc410fd006f07075ee09d9/railties/CHANGELOG.md#L75)).


The first commit in this PR fixes problem 2.

This is my repro of the problem:

```console
$ ruby --version
ruby 2.3.2p217
$ node --version
v5.0.0
$ yarn --version
0.19.1
$ rails --version
Rails 5.1.0.beta1
$ rails new my_app --webpack=react
...
run bundle install
...
Fetching https://github.com/rails/sass-rails.git
Fetching https://github.com/rails/webpacker.git
...
Using webpacker 1.0 from https://github.com/rails/webpacker.git (at master@a85a3ab)
Using rails 5.1.0.beta1
Using sass-rails 6.0.0.beta1 from https://github.com/rails/sass-rails.git (at master@dfbcc6a)
...
         run  ./bin/yarn add webpack webpack-merge path-complete-extname babel-loader babel-core babel-preset-env coffee-loader coffee-script compression-webpack-plugin rails-erb-loader glob from "."
yarn add v0.19.1
info No lockfile found.
[2/4] 🚚  Fetching packages...
error loader-runner@2.3.0: The engine "node" is incompatible with this module. Expected version ">=4.3.0 <5.0.0 || >=5.10".
error Found incompatible module
info Visit https://yarnpkg.com/en/docs/cli/add for documentation about this command.
         run  ./bin/yarn add --dev webpack-dev-server from "."
yarn add v0.19.1
```

So, what's happening is that yarn *is* installed, but the output of this command actually fails:

*template.rb*
```
run  ./bin/yarn add webpack webpack-merge path-complete-extname babel-loader babel-core babel-preset-env coffee-loader coffee-script compression-webpack-plugin rails-erb-loader glob
```

It fails because:

```
error loader-runner@2.3.0: The engine "node" is incompatible with this module. Expected version ">=4.3.0 <5.0.0 || >=5.10".
error Found incompatible module
```

David's suggestion was that `rails new --webpack` should fail hard when yarn is not installed, but the above failure is due to a version of node being used that is not supported by the `loader-runner` module, which `webpack` depends on.


So the rake task installs webpack and its related dependencies via yarn, but with errors reported, while Rails continues the rest of the installation task. The first commit (1a5099e) fixes this problem by exiting `rails webpacker:install` so that the user understands something is not right.

Here's an example output, failing :

```console
$ rails webpacker:install
         run  ./bin/yarn --version from "."
0.19.1
       exist  app/javascript
   identical  app/javascript/packs/application.js
       exist  bin
   identical  bin/webpack-dev-server
   identical  bin/webpack-watcher
   identical  bin/webpack
   identical  bin/yarn
       exist  config/webpack
   identical  config/webpack/development.js
   identical  config/webpack/production.js
   identical  config/webpack/shared.js
      append  .gitignore
         run  ./bin/yarn add webpack webpack-merge path-complete-extname babel-loader babel-core babel-preset-env coffee-loader coffee-script compression-webpack-plugin rails-erb-loader glob from "."
yarn add v0.19.1
info No lockfile found.
[2/4] 🚚  Fetching packages...
error loader-runner@2.3.0: The engine "node" is incompatible with this module. Expected version ">=4.3.0 <5.0.0 || >=5.10".
error Found incompatible module
info Visit https://yarnpkg.com/en/docs/cli/add for documentation about this command.
Failed to install webpack!
```


As for the first problem, where yarn is not installed, I think the best place for now is to add it to the beginning of `template.rb`, before `./bin/yarn` is called invoked.  The last commit has some comments to explain what I found.

Any other thoughts? Concerns?

lib/tasks/webpacker/install.rake
```diff
@@ -3,6 +3,13 @@ WEBPACKER_APP_TEMPLATE_PATH = File.expand_path('../../install/template.rb', __di
...
+    # Calling exec before invoking app:tempalte will not proceed
+    # exec 'yarn --version'
+    # Checking yarn binstub - will not work because yarn binstub will not exist yet
+    # `./bin/yarn --version`
+    # Check yarn global - could work, but we are already depning on the yarn binstub,
+    # so why not utilize it as the central point truth for existence of yarn global?
+    # `yarn --version`
     if Rails::VERSION::MAJOR >= 5
       exec "./bin/rails app:template LOCATION=#{WEBPACKER_APP_TEMPLATE_PATH}"
     else
```
sealocal: @gauravtiwari Actually, I take that back.  The yarn binstub is created with `rails new` when using Rails 5.1.0 -- with or without the `webpack` option.  I think `./bin/yarn --version` might actually work here to check that yarn is installed prior to running invoking the `rails app:template`.  I can keep testing to see if that way works out.

lib/install/bin/yarn.tt
```diff
@@ -6,5 +6,6 @@ Dir.chdir(VENDOR_PATH) do
   rescue Errno::ENOENT
     puts "Yarn executable was not detected in the system."
     puts "Download Yarn at https://yarnpkg.com/en/docs/install"
+    exit false
   end
```
sealocal: This change appears to cause a conflict with the yarn binstub that's created by `rails new`.  If the check in `install.rake` works out, then I think this change could easily be discarded.

lib/tasks/webpacker/install.rake
gauravtiwari: @sealocal Sure! Sorry, I got busy with another PR. I will check this out at some point :)

PR #152, opened 2017-03-02
Make sure Digest::lookup always pick the correct js file name.

### LINKED ISSUE #151: `Webpacker::Digests.lookup` issue
The content of `digests.json` file that my project generated in production environment is:

```json
{"conversation":["conversation-4ab9768dd0430c0910c6.js","conversation-4ab9768dd0430c0910c6.js.map"],"hello_react":"hello_react-4ab9768dd0430c0910c6.js","packs-bundle":"packs-bundle-4ab9768dd0430c0910c6.js"}
```
So the [lookup](https://github.com/rails/webpacker/blob/master/lib/webpacker/digests.rb#L29) method will return an array string instead of the correct digested filename if I want to use `conversation` script.

yiliangt5:
Fix #151

### DIFF
--- a/lib/webpacker/digests.rb
+++ b/lib/webpacker/digests.rb
@@ -27,7 +27,12 @@ class Webpacker::Digests
   end
 
   def lookup(name)
-    @digests[name.to_s]
+    lookedup = @digests[name.to_s]
+    if lookedup.respond_to?(:select)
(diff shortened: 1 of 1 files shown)


gauravtiwari:
Looks like this won't be needed after #153 as it flattens the manifest file.

PR #157, opened 2017-03-04
Cleanup Travis configuration

vipulnsward:


ytbryan:
@gauravtiwari was it your intention to have docker in the travis configuration? cc @guilleiguaran

gauravtiwari:
@ytbryan yeah, read in the docs that the builds are generally faster with docker (and apparently it does), but no personal reasons. Perhaps, I am missing something?

aried3r:
> Perhaps, missing something?

I think you read [this blog post from 2014](https://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/) which basically says that test runs are automatically run inside a docker container when using `sudo: false` in `.travis.yml`.

What the removed line in this PR would have done is install docker to [run and build docker images](https://docs.travis-ci.com/user/docker/).

So to run within docker (see first link), you also have to switch from `sudo: required` to `sudo: false`.

PR #164, opened 2017-03-08
Set the value of NODE_ENV for compilation from RAILS_ENV. 

lilfaf:
Supports custom rails environments (ex. staging). Much appreciated during
Capistrano deployments.

### DIFF
--- a/lib/tasks/webpacker/compile.rake
+++ b/lib/tasks/webpacker/compile.rake
@@ -1,10 +1,11 @@
 REGEX_MAP = /\A.*\.map\z/
+NODE_ENV = ENV['RAILS_ENV'] || 'production'
 
 namespace :webpacker do
   desc "Compile javascript packs using webpack for production with digests"
   task :compile => :environment do
     dist_dir = Rails.application.config.x.webpacker[:packs_dist_dir]
-    result   = `WEBPACK_DIST_DIR=#{dist_dir} NODE_ENV=production ./bin/webpack --json`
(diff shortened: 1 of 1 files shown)


dleavitt:
`NODE_ENV` has special meaning to React, Vue and other frontend frameworks, where setting it to `production` disables dev mode.

So more often than not `NODE_ENV` should be `production` even in environments where `RAILS_ENV` is set to something else (like `staging`.)

ytbryan:
@dleavitt actually some tools like vue-devtools are disabled in production mode and only works in dev mode. Is there example of those "special meaning"? 


I am not sure if it is best to set `NODE_ENV` to `ENV['RAILS_ENV']` or product by default.
Maybe there should be a `ENV['NODE_ENV']`?

gauravtiwari:
@dleavitt Agree.  @ytbryan with special meaning he means - 

1. https://facebook.github.io/react/docs/optimizing-performance.html#use-the-production-build
2. https://vuejs.org/v2/guide/deployment.html#With-Build-Tools

The compile task is made for deployments (production or staging), much like `assets:precompile` task and regardless of env it should always be production because both staging and production are expected same in behaviour. As for development, the setup is different - we got watcher and dev-server and they will build assets for us when run.

gauravtiwari:
@lilfaf Is this still a issue for you? You could always set the `NODE_ENV=production` unless in development. The compile task will set this anyway.

gauravtiwari:
Feel free to close if this no longer is a problem 👍

gauravtiwari:
This can be closed as per #197

lilfaf:
@gauravtiwari oh sorry I missed your comment. Indeed this issue could be closed as its not a problem anymore. 👍

PR #166, opened 2017-03-09
Use relative node_modules path for webpack resolve.modules config

nrako:
Yarn (and NPM3+) try to flatten the node_modules directory, but when different versions of the same package are required Yarn (and NPM as well) will fallback to a nested structure.

If `node_modules` is set as an absolute path, webpack will only search in that directory, ignoring a potential nested node packages.

> A relative path will be scanned similarly to how Node scans for node_modules, by looking through the current directory as well as it's ancestors (i.e. ./node_modules, ../node_modules, and on).

`resolve.modules` defaults to:
```javascript
modules: ["node_modules"]
```

Ref: https://webpack.js.org/configuration/resolve/#resolve-modules

### DIFF
--- a/lib/install/config/shared.js
+++ b/lib/install/config/shared.js
@@ -57,7 +57,7 @@ const config = {
     extensions: ['.js', '.coffee'],
     modules: [
       path.resolve('app/javascript'),
-      path.resolve('node_modules')
+      'node_modules'
     ]
   },
 


gauravtiwari:
We can close this one @nrako #153 is merged

nrako:
@gauravtiwari I don't think so, how is #153 closing this? 

Sure #153 moved quite a lot of stuff around but I don't believe it fixed what that PR intended to.

This line:
https://github.com/gauravtiwari/webpacker/blob/d7278d28a3e8f66ec7e5df6205e37550a68f0eb6/lib/install/config/webpack/shared.js#L44

Should be `path.node_modules,` without the `resolve()`, otherwise webpack receive an absolute path and won't lookup modules relatively the way it should.

nrako:
I would also just use `'node_modules',` instead of [`path.node_modules`](https://github.com/rails/webpacker/pull/153/files#diff-7db49308619fd22a85547012e1c34deeR5) IMHO this is useless verbosity.

gauravtiwari:
Ohh shoot, confused it with `resolveLoader`. Perhaps, you could open this one up again?

nrako:
@gauravtiwari - sure. What do you think of also removing node_modules for the `paths.yml` file?

Issue #158, opened 2017-03-04
Getting lots of  WARNING in EnvironmentPlugin with rvm variables


djforth:
So I am setting up webpacker in rails 5 project and it mostly seems ok, but I am getting the following warnings a bunch of warnings like this:
```
 WARNING in EnvironmentPlugin - rvm_use_flag environment variable is undefined.
22:02:25 hot.1  | 
22:02:25 hot.1  | You can pass an object with default values to suppress this warning.
22:02:25 hot.1  | See https://webpack.js.org/plugins/environment-plugin for example.
22:02:25 hot.1  | 
```

They all seem to be related to rvm, I've rectified it by adding the following to the webpack config:

```
plugins: [
   new webpack.EnvironmentPlugin(Object.keys(process.env).filter((key)=>!key.match(/rvm/)))
],
```
It looks to me that because I use rvm it is applying a bunch of variables to the RAILS_ENV.
It seems that in bin/webpack and webpack-dev-server seem to be passing all of the environment  variables from rails to node

```
ENV['NODE_ENV'] ||= RAILS_ENV
NODE_ENV    = ENV['NODE_ENV']
```

Just wondering if it's worth being more selective on how this is done to avoid issues like this?

Issue #167, opened 2017-03-10
Issue with multiple enviroments


djforth:
If you have a multiple environments for you app the maybe a scenario when you want to set a different NODE_ENV to RAILS_ENV.  For example if you have a staging environment in React app, so you want staging webpack set up (for pushing assets to s3 for example), but in order to remove all of Reacts development code you need to set the NODE_ENV to production.  So you  might want to call 

```bash
RAILS_ENV=staging NODE_ENV=production bin/webpack
```
I'd expect this to trigger the webpack/staging.js but it will always call webpack/production.js

I looks like in bin/webpack your setting:

```ruby
WEBPACK_CONFIG = "#{ESCAPED_APP_PATH}/config/webpack/#{NODE_ENV}.js"
```

If you tweak this to:

```ruby
WEBPACK_CONFIG = "#{ESCAPED_APP_PATH}/config/webpack/#{RAILS_ENV}.js"
```

Then everything works.  Is this something that is worth tweaking in the default set up?

PR #168, opened 2017-03-10
Fix webpack for multiple enviroments and rvm issue with EnvironmentPl…

djforth:
Fixes for https://github.com/rails/webpacker/issues/167 & https://github.com/rails/webpacker/issues/158

### DIFF
--- a/lib/install/config/shared.js
+++ b/lib/install/config/shared.js
@@ -50,7 +50,7 @@ const config = {
   },
 
   plugins: [
-    new webpack.EnvironmentPlugin(Object.keys(process.env).filter((key)=>!key.match(/rvm/))
+    new webpack.EnvironmentPlugin(Object.keys(process.env).filter(key => !key.match(/rvm/)))
   ],
 
   resolve: {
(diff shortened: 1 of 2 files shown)


lib/install/config/shared.js
```diff
@@ -50,7 +50,7 @@ const config = {
   },
 
   plugins: [
-    new webpack.EnvironmentPlugin(Object.keys(process.env))
+    new webpack.EnvironmentPlugin(Object.keys(process.env).filter((key)=>!key.match(/rvm/))
```
ytbryan: A missing closing parenthesis

djforth: Sorry my bad, my sublimelinter didn't pick it up. Just pushing fix

gauravtiwari:
@djforth This is not just related to rvm, but with other env keys that are either `null or undefined`. I have wrapped up this in another PR - https://github.com/rails/webpacker/pull/153/files#diff-67f77b829953e34e910db8d7c0c285c3R66 

That will fix the problem once merged 🎉

djforth:
Ah I see ok excellent :)

noma4i:
This PR doesn't fix https://github.com/rails/webpacker/issues/158

gauravtiwari:
@noma4i but this will #153, once merged.

PR #170, opened 2017-03-11
Add a javascript_pack_url helper

renchap:
This is useful for example if you want to create a `<link rel="prefetch">` tag for some of your packs

### DIFF
--- a/lib/webpacker/helper.rb
+++ b/lib/webpacker/helper.rb
@@ -2,7 +2,7 @@ require 'webpacker/source'
 
 module Webpacker::Helper
   # Creates a script tag that references the named pack file, as compiled by Webpack per the entries list
-  # in config/webpack/shared.js. By default, this list is auto-generated to match everything in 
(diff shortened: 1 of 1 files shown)


gauravtiwari:
@renchap Think the `stylesheet_pack_tag` helper added after #153 can take care of this. Perhaps, we can set a alias method name to reflect it's purpose?

renchap:
I feel it is a good idea to have a way to get an URL for a webpacker asset.
Once your PR is merge, it makes sense to have a single `webpacker_asset_url` (or another more coherent name).

gauravtiwari:
@renchap I have wrapped a `asset_pack_path` helper here (uses asset_path helper under the hood) - https://github.com/rails/webpacker/pull/153/files#diff-24bd59c9028b8c9d1ea0fe27cf0314b7R55 

What do you think?

renchap:
👍 looks good!

gauravtiwari:
we can close this one @renchap #153 is merged

PR #177, opened 2017-03-20
Fix Typescript compilation

WoH:
Without this, .ts files have to be referenced directly.


WoH:
Should be fixed in the angular.rake task

PR #178, opened 2017-03-20
Fix adding .ts to extensions

WoH:
Format is :
> const extensions = ['.js', '.coffee']

not:

> const extensions: ['.js', '.coffee']

This fixes adding the typescript extension in the shared webpack config while runnning webpacker:install:angular


gauravtiwari:
Thanks for making this PR. This is  already fixed in #153.

WoH:
Ok, thanks for responding, I only briefly glanced over the other PR's and they didn't seem to be related, I should've checked yours more in depth. :+1:

Issue #194, opened 2017-03-27
RAILS_ENV=test ./bin/rake webpacker:compile fails


mfazekas:
with latest branch

```sh
rails _5.1.0.rc1_ new webpacker-test
cd webpacker-test
echo 'gem "webpacker", github: "rails/webpacker"' >> Gemfile
bundle install
./bin/rails webpacker:install
./bin/rails webpacker:install:react
RAILS_ENV=test ./bin/rake wepack:compile

RAILS_ENV=test ./bin/rake webpacker:compile
Running via Spring preloader in process 95126
Webpacker is installed 🎉 🍰
Using /private/tmp/foo/webpacker-test/config/webpack/paths.yml file for setting up webpack paths
Compiling webpacker assets 🎉
(node:95127) DeprecationWarning: loaderUtils.parseQuery() received a non-string value which can be problematic, see https://github.com/webpack/loader-utils/issues/56
parseQuery() will be replaced with getOptions() in the next major version of loader-utils.
Compiled digests for all packs in /private/tmp/foo/webpacker-test/public/packs: 
rake aborted!
Errno::ENOENT: No such file or directory @ rb_sysopen - /private/tmp/foo/webpacker-test/public/packs/manifest-test.json
/Users/boga/.rvm/gems/ruby-2.4.0/bundler/gems/webpacker-afe9f69fd028/lib/tasks/webpacker/compile.rake:16:in `read'
/Users/boga/.rvm/gems/ruby-2.4.0/bundler/gems/webpacker-afe9f69fd028/lib/tasks/webpacker/compile.rake:16:in `block (2 levels) in <top (required)>'
/Users/boga/.rvm/gems/ruby-2.4.0/gems/activesupport-5.1.0.rc1/lib/active_support/dependencies.rb:286:in `load'
/Users/boga/.rvm/gems/ruby-2.4.0/gems/activesupport-5.1.0.rc1/lib/active_support/dependencies.rb:286:in `block in load'
/Users/boga/.rvm/gems/ruby-2.4.0/gems/activesupport-5.1.0.rc1/lib/active_support/dependencies.rb:258:in `load_dependency'
/Users/boga/.rvm/gems/ruby-2.4.0/gems/activesupport-5.1.0.rc1/lib/active_support/dependencies.rb:286:in `load'
-e:1:in `<main>'
Tasks: TOP => webpacker:compile
(See full trace by running task with --trace)
```

ytbryan:
[...]

PR #195, opened 2017-03-27
Use correct env for compiling webpack

mfazekas:
Fixes: #194

Even if Rails.env is non production we want to keep NODE_ENV=production don't we? See comments on https://github.com/rails/webpacker/pull/164
So we want to do most of our checks based on RAILS_ENV even in js

### DIFF
--- a/lib/install/bin/webpack-dev-server.tt
+++ b/lib/install/bin/webpack-dev-server.tt
@@ -14,7 +14,7 @@ APP_PATH    = File.expand_path("../", __dir__)
 CONFIG_PATH = File.join(APP_PATH, "config/webpack/paths.yml")
 
 begin
-  paths = YAML.load(File.read(CONFIG_PATH))[NODE_ENV]
+  paths = YAML.load(File.read(CONFIG_PATH))[RAILS_ENV]
 
   NODE_MODULES_PATH   = File.join(APP_PATH.shellescape, paths["node_modules"])
   WEBPACK_CONFIG_PATH = File.join(APP_PATH.shellescape, paths["config"])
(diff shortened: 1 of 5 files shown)


renchap:
`NODE_ENV` and `Rails.env` dont really map together. I am not sure how to handle this, except by having a mapping between the 2. In any case, I think in the precompile task `NODE_ENV` should default to production if not set, but should respect the env variable if present.

mfazekas:
@renchap i've updated my PR, so instead of adjusting `NODE_ENV` it now check `RAILS_ENV` of `NODE_ENV` in javascript too.

gauravtiwari:
@mfazekas `NODE_ENV` should be good enough as we are signing RAILS_ENV to NODE_ENV anyway. Any reason to use RAILS_ENV?

gauravtiwari:
Semantically it seems NODE_ENV would be idea within webpack config instead of RAILS_ENV, plus most of frontend libraries and frameworks use that env during compilation.

mfazekas:
@gauravtiwari as discussed on #164 NODE_ENV has it's own semantic. So like on RAILS_ENV=staging we still want NODE_ENV=production.
I assume RAILS_ENV=development we want NODE_ENV=development. Not sure about RAILS_ENV=test i assume we'd want NODE_ENV=production?

mfazekas:
In any case i'm not sure if there can be one to one mapping between `NODE_ENV`  and `RAILS_ENV` so we should be using either NODE_ENV or RAILS_ENV when selecting which type of config to use.

The merged pr uses `NODE_ENV` on most places expect `Rails.env` here: https://github.com/rails/webpacker/pull/191/files#diff-fbcf8ee6233af5f453a42e3c8e3fd490R40

gauravtiwari:
@mfazekas Yes, I think that seems like a good approach. What you think about this? https://github.com/rails/webpacker/pull/197

mfazekas:
see #197

PR #198, opened 2017-03-27
Allow multiple entry points with same basename

fsateler:
This allows having eg, application.css and application.js

### DIFF
--- a/lib/install/config/webpack/shared.js
+++ b/lib/install/config/webpack/shared.js
@@ -18,7 +18,10 @@ module.exports = {
   entry: packPaths.reduce(
     (map, entry) => {
       const localMap = map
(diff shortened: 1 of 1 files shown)


gauravtiwari:
@fsateler Thanks 👍  and sorry this one has been hanging in for a while. Basically, the idea is to keep only Webpack entry files (JS) inside packs directory. Obviously, the main module can live anywhere inside the `app/javascript` or where ever the JS source is as long as it can be referenced and bootstrapped inside packs. The module should be responsible for all styles and assets instead of packs/entries. That way it will create the `.css` bundle for you after compilation (if any styles), which will be same name as entry/pack name. Does that make sense?

PR #201, opened 2017-03-29
Allow pack files to be placed in subdirectories

kuroda:
With this commit, you can organize JavaScript files with namespaces.

For example:

```text
app/javascript/packs/admin/hello_vue.js
app/javascript/packs/admin/hello.vue
app/javascript/packs/hello_vue.js
app/javascript/packs/hello.vue
```

### DIFF
--- a/lib/install/config/webpack/shared.js
+++ b/lib/install/config/webpack/shared.js
@@ -3,7 +3,7 @@
 /* eslint import/no-dynamic-require: 0 */
 
 const webpack = require('webpack')
-const { basename, join, resolve } = require('path')
+const { basename, dirname, join, relative, resolve } = require('path')
 const { sync } = require('glob')
 const { readdirSync } = require('fs')
 const ExtractTextPlugin = require('extract-text-webpack-plugin')
@@ -11,14 +11,15 @@ const ManifestPlugin = require('webpack-manifest-plugin')
 const extname = require('path-complete-extname')
 const { env, paths, publicPath, loadersDir } = require('./configuration.js')
 
-const extensionGlob = `*{${paths.extensions.join(',')}}*`
+const extensionGlob = `**/*{${paths.extensions.join(',')}}*`
 const packPaths = sync(join(paths.source, paths.entry, extensionGlob))
 
 module.exports = {
   entry: packPaths.reduce(
     (map, entry) => {
       const localMap = map
-      localMap[basename(entry, extname(entry))] = resolve(entry)
+      const namespace = relative(join(paths.source, paths.entry), dirname(entry))
+      localMap[join(namespace, basename(entry, extname(entry)))] = resolve(entry)
       return localMap
     }, {}
   ),


ronnyhaase:
I think **not** considering sub-folders is intended, because `app/javascript/packs` only contains your entry points and you should probably move the admin folder into `app/javascript`, and either import it in `hello_vue.js` or add a `hello_vue_admin.js` doing so.

kuroda:
@ronnyhaase 

Thanks for your comment.

I suppose that you assume a kind of _single page application_ with a very few entry points. Under this assumption, organizing entry points using subdirectories seems an overdo.

I personally, however, want to have a lot of entry points.

I often create a small Vue component in order to fullfil a small functionality, which used to be done with jQuery. In such case, there are times to need a separate `vue` file for each _action_.

Suppose that we have `admin/users` and `portal/users` resources, for example. I want to have a `vue` file for the `admin/users#new` action, and another for the `portal/users#new` action. In my opinion, it is very natural to put the former under the `app/javascript/packs/admin` directory, and the latter under the `app/javascript/packs/portal` directory.

This use case might not have been under consideration by the core team, but I think it is very important especially for Vue developers.

ronnyhaase:
I see your point, that you might have a requirement for subfolders in a huge app, where you have multiple areas like "admin" and "portal", each containing multiple pages and therefor entry points.

But I'm still not sure if Webpacker should go for that case by default, but this decision has to be made by the core contributors.

p0wl:
whats the downside of this pr? the `app/javascript/packs` folder is the root of entries of webpack and I don't see a reason why you would want subfolders in this directory if not for organizing entries.

So I think this pr enforces developers to keep the packs folder clean from anything but entry points, which I think is a good thing.

gauravtiwari:
@kuroda How about you just name space them within file name? For ex - `admin-new` or `portal-new` . I am not sure, but subdirectories seems redundant, when all we are keeping here are packs/entries name, which is essentially going to be one per module or bundle. 

Please correct me if I am assuming wrong?

This is what we have in readme, 

```js
// app/javascript/packs/calendar.js
require('calendar')

app/javascript/calendar/index.js // gets loaded by require('calendar')
app/javascript/calendar/components/grid.jsx
app/javascript/calendar/styles/grid.sass
app/javascript/calendar/models/month.js
```

If you have multiple namespaces, just name space the pack/entry name - `admin-calendar.js`, `app-calendar.js`. 

I guess this change will create some edge cases

dhh:
I like the idea of using actual directories for namespaces. Follows how we
do namespaces in the rest of a Rails app as well. Did you find any problems
with this change, Gaurav?

ronnyhaase:
I think there are arguments for both sides, but I agree with @dhh it fits with Rails namespaces. And thinking of bigger apps I would prefer directories instead of @gauravtiwari suggestion as well.

As an alternative option, how about only allowing one level of sub-directories by replacing the ** with a single * ? Just an idea, not sure if I like it.

dhh:
I don't like trying to second guess what's reasonable here. We're designing
for adults and they can decide for themselves. We should of course nudge
them in a good direction, but think we're already doing that.

kuroda:
@gauravtiwari

Vue.js allows JS components to pick their template from an HTML element.

If we have this code as `app/javascript/packs/hello_world`:

```javascript
import Vue from 'vue/dist/vue.esm'

document.addEventListener('DOMContentLoaded', () => {
  new Vue({
    el: "#app",
    data: { name: "world" },
    methods: {
      changeName: function() {
        this.name = "Alice"
      }
    }
  })
})
```

then, you can use this component like this: 

```html
<div id="app">
  <p>Hello, {{ name }}</p>
  <button v-on:click="changeName">Click me</button>
</div>
<%= javascript_pack_tag "hello_world" %>
```

With this approach, we can introduce Vue.js to an existing Rails app without massive code change.

I assume that this app has a lot of (hundreds of, maybe) pages which are organized using namespaces.
And we want to create a _separate_ Vue.js component for _each_ page.

In this circumstance, I don't want to see hundreds of JS files scattered in the `app/javascript/packs` directory.

gauravtiwari:
Thank you, that totally makes sense and apologies for under-sight on my part. @dhh No, everything works (just tested it). I was just worried that this might break the paths generated in the manifest :)

Thanks @kuroda 👍  🍰

PR #209, opened 2017-03-30
Adds .jpg extension to assets matcher

lbrito1:
Previously only `.jpeg` was matched.


renchap:
I think it should be added here too : https://github.com/rails/webpacker/blob/master/lib/install/config/webpack/paths.yml

gauravtiwari:
Thanks. Looks good for merge :)

ytbryan:
this should be merged.

PR #214, opened 2017-03-31
Update README.md

tannermares:
### Why this was needed
- `github` option is deprecated and `git` with `https` preferred. 
- `webpacker` command is under `rake` not `rails`

### How I fixed it
- used preffered `git` option with `https` url 
- changed invalid `rails` command with `rake`

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,15 +24,15 @@ Webpacker is currently compatible with Rails 4.2+, but there's no guarantee it w
 in the future.
 
 You can either make use of Webpacker during setup of a new application with `--webpack`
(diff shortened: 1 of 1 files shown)


tannermares:
Just realized this might be a Rails 4.2 vs Rails 5 thing. Since README says 4.2+ might need to mention both?

ytbryan:
Yes, Webpacker was intended for Rails 5.1.x initially. https://github.com/rails/webpacker/pull/123 changes it. I think it's fine to keep those instruction.

tannermares:
@ytbryan sounds good!

PR #220, opened 2017-04-02
Default to ERB for rails-erb-loader

atwoodjw:
Looks like `rails-erb-loader` has a default ERB template engine of `erubis`, not `erb`. This results in a build error on a freshly generated Rails 5.1.0.rc1 app.

https://github.com/usabilityhub/rails-erb-loader#options

```
ERROR in ./app/javascript/packs/sessions.js.erb
Module build failed: Error: Command failed: DISABLE_SPRING=1 bin/rails runner /Developer/testapp/node_modules/rails-erb-loader/erb_transformer.rb __RAILS_ERB_LOADER_DELIMETER__ erubis
/Users/testuser/.rvm/gems/ruby-2.3.1@testapp/gems/activesupport-5.1.0.rc1/lib/active_support/dependencies.rb:292:in `require': cannot load such file -- erubis (LoadError)
```

### DIFF
--- a/lib/install/config/loaders/core/erb.js
+++ b/lib/install/config/loaders/core/erb.js
@@ -4,6 +4,7 @@ module.exports = {
   exclude: /node_modules/,
   loader: 'rails-erb-loader',
   options: {
-    runner: 'DISABLE_SPRING=1 bin/rails runner'
+    runner: 'DISABLE_SPRING=1 bin/rails runner',
+    engine: 'erb'
   }
 }


renchap:
In Rails 5.1 the ERB engine switched to `erubi` (https://github.com/jeremyevans/erubi). Maybe it should be added as an engine in `rails-erb-loader`?

rhys-vdw:
> Maybe it should be added as an engine in rails-erb-loader?

Alternatively, does Rails provide a generic interface to render an ERB template that is consistent between different versions? That way webpacker could default to a `engine: "rails"` option that would do the right thing for all versions.

rhys-vdw:
You can now update to `rails-erb-loader` version 5 to have erubi support and a default engine of `erb`.

atwoodjw:
Great. That makes this PR unnecessary. Closing.

rhys-vdw:
It would probably be appropriate to provide versions for npm dependencies of this project unless you want them to break with time. If you didn't have to change anything for this to work then that means you're vulnerable to unaddressed breaking changes.

Perhaps https://greenkeeper.io/ combined with a local `package.json`?

PR #230, opened 2017-04-04
Correct "Linking to static assets" paragraph

ronnyhaase:
The example is misleading, because doing it like in the the current state would cause Webpack to compile hello-react.sass two times, since it's imported by `hello_react.jsx` and considered as entry point when being in `app/javascript/packs`


gauravtiwari:
thanks @ronnyhaase. We can merge this one 👍

PR #234, opened 2017-04-05
Only load non-hidden .js files as loader

nielsslot:
Some editors, vim for example, create temporary hidden swap files when editing a file. This PR makes the webpack configuration use only non-hidden JavaScript files from the loaders directory. Which means it ignores swap files, instead of giving an error because they aren't valid JavaScript.

### DIFF
--- a/lib/install/config/webpack/shared.js
+++ b/lib/install/config/webpack/shared.js
@@ -5,7 +5,6 @@
 const webpack = require('webpack')
 const { basename, dirname, join, relative, resolve } = require('path')
 const { sync } = require('glob')
-const { readdirSync } = require('fs')
 const ExtractTextPlugin = require('extract-text-webpack-plugin')
 const ManifestPlugin = require('webpack-manifest-plugin')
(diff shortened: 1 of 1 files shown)


gauravtiwari:
@nielsslot Just checked this. There is one issue with the code, it's going to load full path twice due to sync and join. So, we need to remove that extra join inside the block. 

```js
  module: {
    rules: sync(join(loadersDir, '*.js')).map(file => require(file))
  },
```

If you could fix that, I will merge 👍

gauravtiwari:
Perhaps, rename file to loader - 

```js
  module: {
    rules: sync(join(loadersDir, '*.js')).map(loader => require(loader))
  },
```

nielsslot:
@gauravtiwari, thanks for the feedback. I checked the Rails project in which I originally made the change and see that I forgot the copy the removal of the second join to here.

I just updated my commit with the changes you requested.

gauravtiwari:
Awesome 🎉  Thank you :)

PR #239, opened 2017-04-06
Fixed a bug where node 6.10 would be converted to 6.1

jamespearson:
The previous way converted a s "6.10" to 6.1 before comparison with 6.4. This meant that Node 6.10 (the LTS running on Heroku) was converted to 6.1, and was less than the required 6.4

This change now compares on major version first, then an integer of the major and minor.


aganov:
This is already covered by #232

PR #240, opened 2017-04-06
Vue does not require an eventListener

ytbryan:
It seems that Vue does not require an eventListener for DOMContentLoaded. 

Can somebody verify this PR? It seems to work without eventListener on my end.

### DIFF
--- a/lib/install/examples/vue/hello_vue.js
+++ b/lib/install/examples/vue/hello_vue.js
@@ -7,13 +7,11 @@
 import Vue from 'vue/dist/vue.esm'
 import App from './app.vue'
 
-document.addEventListener('DOMContentLoaded', () => {
-  document.body.appendChild(document.createElement('hello'))
-  const app = new Vue({
-    el: 'hello',
-    template: '<App/>',
-    components: { App }
-  })
-
-  console.log(app)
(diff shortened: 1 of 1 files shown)


ankitrg:
This works fine! I am not sure if this causes memory leak, as my `destroyed` hook is never called and `mounted` hook is called every time I navigate to the page.

gauravtiwari:
@ytbryan Did you checked this with turbolinks? I noticed this behaviour earlier and it seemed vue does the mounting automatically, but I am not sure if this causes problems if you are using turbolinks and don't use turbolinks event listener.

ankitrg:
If you will import all js files at once with `packs/application.js` and then try to mount it, this will cause problems with turbolinks. Try navigating to other page and coming back, the js doesn't work then.

gauravtiwari:
@ytbryan I get this if I use a pre-defined DOM node,  which makes sense so, lets just keep it. 

```js
const app = new Vue({
  el: '#vue-app',
  template: '<App/>',
  components: { App }
})

console.log(app)
```

gauravtiwari:
@ankitrg Thanks for confirming, makes sense 👍

PR #249, opened 2017-04-10
webpacker can also manage non-JS assets [ci skip]

claudiob:
Not sure if the whole paragraph in the README should go away, but after https://github.com/rails/webpacker/pull/153 it's not true that webpacker is _only_meant for JS files.


gauravtiwari:
@claudiob Yeah, you are right. We can rewrite this whole paragraph and may be get rid of linking with asset pipeline info?

dhh:
[Reformulated the purpose here](https://github.com/rails/webpacker/commit/96ca7f760fa70c07a7ef1fccdf7ca40999b50ecd) 👍

PR #250, opened 2017-04-10
Enable Tree Shaking (for react)

p0wl:
In order to enable [tree shaking](https://webpack.js.org/guides/tree-shaking/) out of the box, webpack needs to do the compiling of "module" statements (`import` and `export` statements).

Therefore, module compilation should be disabled for `babel-preset-env` in `.babelrc`.

Note: this only applies to `react`, as the other examples do not have a `.babelrc` right now


gauravtiwari:
Great, thanks @p0wl 👍

Issue #216, opened 2017-04-01
Question: Usage of `rails-erb-loader`


rhys-vdw:
Hey maintainers, just had a question about your usage of rails-erb-loader. (I'm not a consumer of this project, but I am the maintainer of rails-erb-loader).

Why [disable spring](https://github.com/rails/webpacker/blob/a858f8e2cd2c736d09f936902a59d33b32ba73fb/lib/install/config/loaders/core/erb.js#L7) in your `runner` config option? The loader spins up a new instance of rails in a child process for every `.erb` file fed into it. I would expect this to make rails-erb-loader grind to an intolerably slow speed.

matthewd:
@gauravtiwari

ytbryan:
It's first added here https://github.com/rails/webpacker/pull/66 cc @stevejackson

rhys-vdw:
Thanks @ytbryan for the info.

> rails-erb-loader also seems to have problems with spring, so I've just disabled it.
- from https://github.com/rails/webpacker/pull/66#issue-202299186

@stevejackson I wonder if you can remember the problems you were hitting? If so could you open an issue in [https://github.com/usabilityhub/rails-erb-loader](the rails-erb-loader repo) with a details.

ytbryan:
@rhys-vdw please take a look at this https://github.com/rails/webpacker/pull/220

stevejackson:
@rhys-vdw

It was me who wrote that, sorry for the slow reply. It was likely a mistake on my part - I was testing out webpacker in RAILS_ENV=production, which doesn't have spring installed - therefore it couldn't find the spring executable.

We could probably remove this `DISABLE_SPRING=1`.

rhys-vdw:
> We could probably remove this DISABLE_SPRING=1.

Great. Well then I would suggest removing it then.

stevejackson:
Will do - I'll take care of this soon.

gauravtiwari:
Feel free to make a PR for this 😄

gauravtiwari:
Fixed by #256

PR #256, opened 2017-04-12
Re-enable Spring for rails-erb-loader

stevejackson:
I'm the committer who originally disabled spring on this line. It was likely a mistake - see further discussion in the issue https://github.com/rails/webpacker/issues/216.

With `DISABLE_SPRING=1` removed, I've ran `bin/webpack` with both RAILS_ENV=development & production - both work.


gauravtiwari:
Thanks @stevejackson 👍

PR #271, opened 2017-04-19
Use a hash in js filenames for all environments

### LINKED ISSUE #267: Prevent browser caching of javascript packs, using a hash in the filename
Chrome keeps running old versions of application.js, unless I disable the cache in developer tools. This doesn't happen with sprockets because it includes a hash of the file's contents in the filename, preventing chrome from loading the file from cache unless the hash is the same.

Can we add this feature to webpacker? How would you recommend it be implemented?

mikeyhew:
Fixes #267

### DIFF
--- a/lib/install/config/webpack/production.js
+++ b/lib/install/config/webpack/production.js
@@ -7,8 +7,6 @@ const CompressionPlugin = require('compression-webpack-plugin')
 const sharedConfig = require('./shared.js')
 
 module.exports = merge(sharedConfig, {
-  output: { filename: '[name]-[chunkhash].js' },
-
   plugins: [
     new webpack.optimize.UglifyJsPlugin(),
     new CompressionPlugin({
--- a/lib/install/config/webpack/shared.js
+++ b/lib/install/config/webpack/shared.js
@@ -24,7 +24,7 @@ module.exports = {
   ),
 
   output: {
-    filename: '[name].js',
+    filename: '[name]-[chunkhash].js',
     path: resolve(paths.output, paths.entry),
     publicPath
   },


gauravtiwari:
@mikeyhew Apparently, we can't use `[chunkhash]` in conjunction with `webpack-dev-server` so, an alternative would be to use `[hash]` in shared.js and keep production.js as it was earlier. Someone from webpack team commented on similar issue so, I am not sure if we should have this since after #264 there will be only `webpack-dev-server` and no watcher - 

> You should not use [chunkhash] or [hash] for development. This will cause many other issues, like a memory leak, because the dev server does not know when to clean up the old files. 

https://github.com/webpack/webpack-dev-server/issues/377#issuecomment-241258405
https://github.com/webpack/webpack/issues/1363#issue-101426739

p0wl:
more than that, it will not work because of a conceptual problem:

bin/webpack-dev-server (and bin/webpack) updates the manifest.json when the compilation is finished. If you refresh before the compilation is finished, rails uses the old manifest.json, which points to the old file names (old hashes). I guess rails handles this by waiting to render the script tag until the compilation of the requested file is done (?).

I think the right way to go would be to set the headers for every file in public/packs to something like `Cache-Control: no-cache, no-store, must-revalidate` in development.

mikeyhew:
@gauravtiwari That's... interesting. I changed my app's config today to put a hash in the filename, just like in this PR, and it works really well. Although I just checked my `public/packs` folder and there's a hundred different files in there 😆 

That comment farther down by @SpaceK33z is pretty naive though. I have some remote assets that aren't stored on my local machine, and the browser saves a lot of bandwidth by caching them. It especially helps when you're tethering your phone and want to minimize data charges.

>No, you should have every caching feature off in dev. There's no real benefit; you're working on localhost anyway, so downloading is only limited by your ~imagination~ disk.

mikeyhew:
@p0wl 

>I think the right way to go would be to set the headers for every file in public/packs to something like Cache-Control: no-cache, no-store, must-revalidate in development.

If that works, it would be a fine solution.

gauravtiwari:
@mikeyhew Btw, you have been using watcher or dev server in development?

dhh:
Sprockets doesn't include a hash in development either. Relies on cache controls. We should do the same.

rafaelfranca:
> Sprockets doesn't include a hash in development either. Relies on cache controls. We should do the same.

It actually does since Rails 4.2,, but it relies on cache controls.

mikeyhew:
@gauravtiwari I'm using watcher. I tried using the dev server, to see if it would stop things from being cached, but I think it still had the same problem.

Issue #160, opened 2017-03-05
DevDependencies vs Dependencies


artemgurzhii:
Current behaviour
```json
"dependencies": {
    "babel-core": "^6.23.1",
    "babel-preset-env": "^1.2.0",
    "coffee-loader": "^0.7.3",
    "coffee-script": "^1.12.4",
    "compression-webpack-plugin": "^0.3.2",
    "glob": "^7.1.1",
    "path-complete-extname": "^0.1.0",
    "rails-erb-loader": "^3.2.0",
    "webpack": "^2.2.1",
    "webpack-merge": "^3.0.0"
  }
```

expected
```json
"devDependencies": {
    "babel-core": "^6.23.1",
    "babel-preset-env": "^1.2.0",
    "coffee-loader": "^0.7.3",
    "coffee-script": "^1.12.4",
    "compression-webpack-plugin": "^0.3.2",
    "glob": "^7.1.1",
    "path-complete-extname": "^0.1.0",
    "rails-erb-loader": "^3.2.0",
    "webpack": "^2.2.1",
    "webpack-merge": "^3.0.0"
  }
```

Installed via yarn packages should be added to the `devDependencies` as they used only in development.

artemgurzhii:
Should I create pull request for this?

gauravtiwari:
Hey @artemgurzhii, please checkout this issue - https://github.com/rails/webpacker/issues/117, it's discussed here. Feel free to close the issue, if that answers your question :)

artemgurzhii:
What about this?
```ruby
if ENV['NODE_ENV'] == 'production'
  run './bin/yarn add webpack webpack-merge path-complete-extname babel-loader babel-core babel-preset-env coffee-loader coffee-script compression-webpack-plugin rails-erb-loader glob'
  run './bin/yarn add --dev webpack-dev-server'
else
  run './bin/yarn add --dev webpack webpack-merge path-complete-extname babel-loader babel-core babel-preset-env coffee-loader coffee-script compression-webpack-plugin rails-erb-loader glob webpack-dev-server'
```

gauravtiwari:
@artemgurzhii This installer is run only once, when you add webpacker using `webpacker:install`. Obviously, you can do that, but it's not something that will work,

1. You will have to install webpacker in production in all your deploys, manually (for container-based)
2. Your package.json will go out-of-sync, say if you are using Git. You commit something and then production installs something differently, resulting in conflicts. 
[...]

PR #273, opened 2017-04-20
Install JS packages as dev-dependency

toshimaru:
I think the packages installed by `webpacker:install` (such as webpack, babel) don't have to be installed as `dependencies` in `package.json`. 

This PR makes the packages `devDependencies`.

### DIFF
--- a/lib/install/template.rb
+++ b/lib/install/template.rb
@@ -19,7 +19,7 @@ EOS
 end
 
 puts "Installing all JavaScript dependencies"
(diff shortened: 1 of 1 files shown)


ytbryan:
Thanks for this. is this related to https://github.com/rails/webpacker/issues/160? (see explanation)

toshimaru:
thanks for the information. I read #117 and I understood those packages are needed in production, for example, on Heroku.

I close this change.

Issue #281, opened 2017-04-24
assets:clobber erases entire public folder


Rishka:
When running `rails assets:clobber` in a default rails 5.1.0.rc2 app with webpacker from master, it removes the entire public folder. 

### Expected behavior 
`public/assets` and `public/packs` should be removed

### Actual behavior
`public` folder is removed, including `robots.txt`, `404.html` etc

gauravtiwari:
Fixed by #282

PR #282, opened 2017-04-24
Fix clobber behavior

Rishka:
Ref #281

This erases the correct folder when running `rails assets:clobber`


gauravtiwari:
Thanks Adam 👍

Issue #278, opened 2017-04-22
webpacker:install leads to a bin/yarn conflict


rubys:
### Steps to reproduce

```
rails new wpinstall
cd wpinstall/
echo "gem 'webpacker', github: 'rails/webpacker'" >> Gemfile
bundle install
rails webpacker:install
```

### Expected behavior
On a clean repository, this should complete without conflict.

### Actual behavior
```
conflict  bin/yarn
Overwrite /home/rubys/git/awdwr/edition4/work-51/depot/bin/yarn? (enter "h" for help) [Ynaqdh] Y
       force  bin/yarn
```

### System configuration
**Rails version**: 5.1.0.rc2

**Ruby version**: 2.4.0

ytbryan:
Yup @y-yagi fixed it. This can be closed. Thanks for reporting, rubys

matthewd:
webpacker should be more explicit about only creating that file if it doesn't exist, so it's only supplying it for pre-5.1 Rails apps. Otherwise we're going to have ongoing cross-version compatibility issues trying to keep them in sync.

gauravtiwari:
Fixed by #285

PR #285, opened 2017-04-25
explicit yarn

ytbryan:
Ref https://github.com/rails/webpacker/issues/278

Basically, copy the three binstubs
And copy yarn only if it does not exist. Tell the user when we are copying it. 

cc @matthewd

### DIFF
--- a/lib/install/template.rb
+++ b/lib/install/template.rb
@@ -6,7 +6,7 @@ puts "Copying binstubs"
 template "#{__dir__}/bin/webpack-dev-server", "bin/webpack-dev-server"
 template "#{__dir__}/bin/webpack-watcher", "bin/webpack-watcher"
 template "#{__dir__}/bin/webpack", "bin/webpack"
-if !File.exists?("bin/yarn")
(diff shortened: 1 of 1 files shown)


lib/install/template.rb
```diff
@@ -3,7 +3,13 @@
 directory "#{__dir__}/javascript", "app/javascript"
 
 puts "Copying binstubs"
-directory "#{__dir__}/bin", "bin"
+template "#{__dir__}/bin/webpack-dev-server", "bin/webpack-dev-server"
+template "#{__dir__}/bin/webpack-watcher", "bin/webpack-watcher"
+template "#{__dir__}/bin/webpack", "bin/webpack"
+if !File.exists?("bin/yarn")
```
y-yagi: Since `File.exists?` deprecated in Ruby 2.1, it is better to use `File.exist?`.
Ref: https://github.com/ruby/ruby/blob/v2_1_2/file.c#L1413

ytbryan: thank you

Issue #280, opened 2017-04-24
Uglifyjs es6 syntax support


Davidzhu001:
```
 DEBUG [33cc528e] Running [ -d $HOME/.rbenv/versions/2.4.0 ] as deploy@1.1.1.1
 DEBUG [33cc528e] Command: [ -d $HOME/.rbenv/versions/2.4.0 ]
 DEBUG [33cc528e] Finished in 1.138 seconds with exit status 0 (successful).
 DEBUG [ec0b7f33] Running /usr/bin/env which passenger as deploy@1.1.1.1
 DEBUG [ec0b7f33] Command: /usr/bin/env which passenger
 DEBUG [ec0b7f33] 	/usr/bin/passenger
 DEBUG [ec0b7f33] Finished in 0.397 seconds with exit status 0 (successful).
  INFO [3265e493] Running /usr/bin/env mkdir -p /tmp as deploy@1.1.1.1
 DEBUG [3265e493] Command: ( export RBENV_ROOT="$HOME/.rbenv" RBENV_VERSION="2.4.0" ; /usr/bin/env mkdir -p /tmp )
  INFO [3265e493] Finished in 1.541 seconds with exit status 0 (successful).
 DEBUG Uploading /tmp/git-ssh-video_casts-production-weichengzhu.sh 0.0%
  INFO Uploading /tmp/git-ssh-video_casts-production-weichengzhu.sh 100.0%
  INFO [51b0992a] Running /usr/bin/env chmod 700 /tmp/git-ssh-video_casts-production-weichengzhu.sh as deploy@1.1.1.1
 DEBUG [51b0992a] Command: ( export RBENV_ROOT="$HOME/.rbenv" RBENV_VERSION="2.4.0" ; /usr/bin/env chmod 700 /tmp/git-ssh-video_casts-production-weichengzhu.sh )
  INFO [51b0992a] Finished in 0.129 seconds with exit status 0 (successful).
  INFO [77ed106c] Running /usr/bin/env git ls-remote git@github.com:username/video_casts.git HEAD as deploy@1.1.1.1
 DEBUG [77ed106c] Command: ( export RBENV_ROOT="$HOME/.rbenv" RBENV_VERSION="2.4.0" GIT_ASKPASS="/bin/echo" GIT_SSH="/tmp/git-ssh-video_casts-production-weichengzhu.sh" ; /usr/bin/env git ls-remote git@github.com:username/video_casts.git HEAD )
[...]

PR #290, opened 2017-04-26
Use export default instead of module.exports in Vue component example

jaredcwhite:
I encountered a problem importing new modules within the `<script>` tag of the example Vue component, and it turned out that `module.exports =` was the culprit. It looks like vue-loader prefers the `export default` syntax, as depicted here: https://vue-loader.vuejs.org/en/start/spec.html

Once I made that change in my example file, I was able to add `import` statements to the top of the `<script>` tag without a problem.

### DIFF
--- a/lib/install/examples/vue/app.vue
+++ b/lib/install/examples/vue/app.vue
@@ -5,7 +5,7 @@
 </template>
 
 <script>
-module.exports = {
+export default {
   data: function () {
     return {
       message: "Hello Vue!"


Davidzhu001:
what are the differences, if you don't mind to share/

jaredcwhite:
I believe it's due to a difference between ES6 module syntax and CommonJS syntax. There's a pretty good rundown of the issue here: http://stackoverflow.com/a/40295288/551775

Bottom line: if you're using ES6 `import` statements to mange your modules (as vue-loader/webpack does), you should use the equivalent ES6 `export default` statements as well.

Davidzhu001:
@jaredcwhite It makes a lot sense, thanks so much

gauravtiwari:
@jaredcwhite Doesn't es6 syntax have issues with webpack uglify plugin? https://github.com/rails/webpacker/issues/280

Davidzhu001:
@gauravtiwari it does, I have to remove webpack uglify plugin from my production.js for a temp solution.
Or we can use webpack uglify plugin the harmony one to instead, it seems not that stable yet.

jaredcwhite:
Wouldn't uglify operate at the post-compilation stage — aka after all ES6 code has been translated down to browser-friendly ES5? Seems weird that wouldn't work. (I'm talking out of my depth now, I'm pretty new to Rails/Webpack integration.)

gauravtiwari:
Thanks @jaredcwhite. It seems like babel config issue. @Davidzhu001 Could you take a look at this PR please #291

gauravtiwari:
Thanks @jaredcwhite

jaredcwhite:
👍

PR #293, opened 2017-04-26
get rid of babel loader warning

Davidzhu001:
`webpack 2.x | babel-loader >= 7.x (recommended) (^6.2.10 will also work, but with deprecation warnings)`
https://github.com/babel/babel-loader


gauravtiwari:
Thanks 👍

PR #294, opened 2017-04-27
Add troubleshooting react component(s) not loading to README.

Najtmare:
In case someone (like me) spent 20 mins wondering why his components weren't being loaded.


gauravtiwari:
@Najtmare See the example comments. We have this already inline 😄  https://github.com/rails/webpacker/blob/master/lib/install/examples/react/hello_react.jsx

gauravtiwari:
Closing this now @Najtmare in favour of better react installer that takes care of this step.

gauravtiwari:
Thanks for contributing though 👍

PR #299, opened 2017-04-28
Add more tests

kimquy:


test/webpacker_test.rb
```diff
@@ -7,7 +7,7 @@
 
 module TestApp
   class Application < ::Rails::Application
-    config.root = File.dirname(__FILE__)
+    config.root = File.join(File.dirname(__FILE__), "test_app")
```
gauravtiwari: Perhaps we can create a `test_helper.rb` and extract these common paths configuration to that? 

```rb
require "minitest/autorun"
class Minitest::Test
  @@app_path = File.join(File.dirname(__FILE__), "test_app")
  # Other common setup here ....
end
```

test/configuration_test.rb
```diff
@@ -0,0 +1,42 @@
+require "webpacker_test"
+
+class ConfigurationTest < Minitest::Test
+  def test_config_path
+    config_path = File.join(File.dirname(__FILE__), "test_app/config", "webpack").to_s
```
gauravtiwari: Perhaps use setup or before block to extract these paths out so that test only have assertions? What do you think?

gauravtiwari:
Amazing work @kimquy with setting up tests 💯  🍰

gauravtiwari:
//cc @rafaelfranca

PR #302, opened 2017-04-28
Don't set ASSET_HOST environment if no asset host configured

mattmacleod:
This change avoids setting the ASSET_HOST environment variable in the compilation rake task if no asset host is set in Rails' configuration.

Previously—introduced in #224—if no asset host was configured, this environment variable would be set to an empty string, which caused the default `ifHasCDN` check in `configuration.js` to always return `true` in production.


gauravtiwari:
thanks @mattmacleod 👍

Issue #300, opened 2017-04-28
Example app should use runtime-only version of vue?


RavenXce:
The hello world example app is currently referencing the full version of vue which includes the compiler. This isn't really needed and isn't the default way of running Vue applications (`import Vue from 'vue'` defaults to the runtime version).

Should the example use this style instead? : https://github.com/vuejs-templates/webpack/issues/215#issuecomment-287652462

gauravtiwari:
@RavenXce Feel free to make a PR please 👍  Looks good to me

irrg:
Possibly related, possibly worth a new issue (possibly worth telling me I'm Doing It Wrong)… I tried to modify the new example to use vue-router. Could never get past `Unknown custom element: <router-view>` etc. 

Considering vue-router (and vuex) are staples…I'd love to see an example that shows what I'm doing wrong.

RavenXce:
That doesn't seem related to this issue. Did you add the component during initialization?

```
import VueRouter from ...

MyComponent new Vue({
   ...,
   components: { VueRouter, ... },
   ...,
});
```

Reference: https://vuejs.org/v2/guide/components.html#Local-Registration

irrg:
I finally got a router to work within the context of webpacker (example: https://github.com/irrg/rails-webpacker-vue-example) but I haven't been able to port my Vue into it yet. I'll figure it out and if needed ask help elsewhere.

That said, the moment I tried to add hot reloading to my linked example, vue-router broke again. So some stuff definitely feels fickle, but it's likely outside of the scope of the webpacker project (and probably firmly in webpack's camp).

PR #304, opened 2017-04-29
Remove template strings from vue example

RavenXce:
This gives a better convention to start users of with -
all templates are pre-compiled through webpack/vue-loader,
and the runtime-only version of vue is imported instead.

Resolves https://github.com/rails/webpacker/issues/300.

### DIFF
--- a/lib/install/examples/vue/hello_vue.js
+++ b/lib/install/examples/vue/hello_vue.js
@@ -4,16 +4,12 @@
 // like app/views/layouts/application.html.erb.
 // All it does is render <div>Hello Vue</div> at the bottom of the page.
 
-import Vue from 'vue/dist/vue.esm'
+import Vue from 'vue'
 import App from './app.vue'
 
 document.addEventListener('DOMContentLoaded', () => {
   document.body.appendChild(document.createElement('hello'))
-  const app = new Vue({
-    el: 'hello',
-    template: '<App/>',
-    components: { App }
-  })
+  const app = new Vue(App).$mount('hello')
 
   console.log(app)
 })


gauravtiwari:
Thanks 👍🍰

zigomir:
Ahh, nice! I was complaining about this issue here: https://github.com/vuejs-templates/webpack-simple/pull/100

yeonhoyoon:
@RavenXce @gauravtiwari
What are some ways to pass data into components without importing the full compiler-included version of vue?

With this commit's change, the runtime-only version of vue is imported.
Now when you try to pass data to components like following as per the [README instruction](https://github.com/rails/webpacker#vue-1), 

```js
const app = new Vue({
    data: {name: 'test'},
    el: '#vue-app',
    template: '<App/>',
    components: { App }
  })
```

an error shows in the console as follows.

`You are using the runtime-only build of Vue where the template compiler is not available. Either pre-compile the templates into render functions, or use the compiler-included build`

RavenXce:
@yeonhoyoon seems like the README is wrong. You should simply do:

```
new Vue(App).$mount('#hello-vue');
```

Note that the problem is with template string compilation, not passing in of data, so the following *should* be valid too:

```
new App({ el: '#hello-vue', data: { name: 'test' } })
```

This assumes you have imported `App` from a `*.vue` single-file component.

yeonhoyoon:
@RavenXce 

I can confirm that the following works fine,
```
new Vue(App).$mount('#vue-app');
```

But the following snippet with data passed in throws an error.
```js
new App({ el: '#vue-app', data: { name: 'test' } })
```

```
Uncaught TypeError: __WEBPACK_IMPORTED_MODULE_2__app_vue___default.a is not a constructor
    at HTMLDocument.eval (eval at <anonymous> (application.js:1110), <anonymous>:15:3)
```

RavenXce:
@yeonhoyoon sorry, that only works if `App` is registered as a component constructor using `Vue.extend`, of course.

You could just try the following instead?
```
new Vue({
   ...App, // Requires ES7
   data
}).$mount('#vue-app');
```

Or:
```
new Vue({
   data,
   render: h => h(App)
}).mount('#vue-app');
```

yeonhoyoon:
@RavenXce 
Your snippet works nicely. Thanks for taking time to look into this!

```js
new Vue({
   {
     foo: 'bar'
   },
   render: h => h(App)
}).$mount('#vue-app');
```

PR #310, opened 2017-05-01
Extract CSS and have smaller build size.

zigomir:
[Extracting CSS into a Single File vue-loader](http://vue-loader.vuejs.org/en/configurations/extract-css.html).

This gets us ~7,6K smaller build, because we don 't need JS to inject styles into head during runtime.

Comparison:

Before:
css-injected-into-head

After:
extracted-css-file


gauravtiwari:
Great thanks 👍

Issue #313, opened 2017-05-02
Add Elm support


bkuhlmann:
I'd like to add support for Elm out-of-the box so that this option is available to those that want to run [Elm](http://elm-lang.org) on top of Rails: `webpacker:install:elm`.

I don't mind doing the work to make this happen if there are no objections? If so, are there any guidelines I need to be aware of before creating a PR (I imagine it'll be several commits consisting of documentation, examples, implementation + specs, generators, etc.)?

dhh:
Go ahead, Brooke. You can follow the patterns from angular/react/vue.
There's not that much to it. Once you have a PR that includes everything,
someone will help review it. Thanks for helping out!

bkuhlmann:
Awesome, will do. I'll work to have a PR ready soon. Thanks!

bkuhlmann:
Support is on `master`, closing this issue. Thanks @gauravtiwari and @guilleiguaran for the support.

PR #324, opened 2017-05-04
Added Elm support.

bkuhlmann:
## Overview

Provides support for easily adding and configuration Elm to run within your Rails app.

## Details

- [Feature](https://github.com/rails/webpacker/issues/313).
- See commits for details.

## Notes

Not sure what the best way to test the new logic is. Almost need a type of integration spec that verifies file changes, bill of materials, etc. The majority of this work builds upon the framework already put in place that is leveraged by the other installers.

## Screenshot

demo_app

## Screencast

[asciicast](https://asciinema.org/a/118946)

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,6 +215,13 @@ To use Webpacker with Angular, just create a new app with `rails new myapp --web
 
 To use Webpacker with Vue, just create a new app with `rails new myapp --webpack=vue` (or run `rails webpacker:install:vue` on a Rails app already setup with webpacker). Vue and its supported libraries will be added via yarn and changes to the configuration files made. An example component is also added to your project in `app/javascript` so that you can experiment Vue right away.
 
+## Ready for Elm
+
+To use Webpacker with [Elm](http://elm-lang.org), create a new app with `rails new myapp
+--webpack=elm` (or run `rails webpacker:install:elm` on a Rails app already setup with webpacker).
+The Elm library and core packages will be added via Yarn and Elm itself. An example `Main.elm` app
+is also added to your project in `app/javascript` so that you can experiment with Elm right away.
+
 ## Troubleshooting
 
 *  If you get this error `ENOENT: no such file or directory - node-sass` on Heroku
--- a/lib/install/config/loaders/installers/elm.js
+++ b/lib/install/config/loaders/installers/elm.js
@@ -1,0 +1,5 @@
+module.exports = {
+  test: /\.elm$/,
+  exclude: [/elm-stuff/, /node_modules/],
+  loader: 'elm-hot-loader!elm-webpack-loader?verbose=true&warn=true&debug=true'
+}
--- a/lib/install/elm.rb
+++ b/lib/install/elm.rb
@@ -1,0 +1,22 @@
+require "webpacker/configuration"
+
+puts "Copying elm loader to #{Webpacker::Configuration.config_path}/loaders"
+copy_file "#{__dir__}/config/loaders/installers/elm.js",
+          "#{Webpacker::Configuration.config_path}/loaders/elm.js"
+
+puts "Copying elm example entry file to #{Webpacker::Configuration.entry_path}"
+copy_file "#{__dir__}/examples/elm/Main.elm", "#{Webpacker::Configuration.entry_path}/Main.elm"
+
+puts "Copying elm app file to #{Webpacker::Configuration.entry_path}"
+copy_file "#{__dir__}/examples/elm/hello_elm.js",
+          "#{Webpacker::Configuration.entry_path}/hello_elm.js"
+
+puts "Updating .gitignore to include elm-stuff folder"
+insert_into_file ".gitignore", "/elm-stuff\n", before: "/node_modules\n"
+
+puts "Installing all elm dependencies"
+run "./bin/yarn add elm"
+run "./bin/yarn add --dev elm-hot-loader elm-webpack-loader"
+run "yarn run elm package install -- --yes"
+
+puts "Webpacker now supports elm 🎉"
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /.bundle
+/pkg
 node_modules
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+- Added [Elm](http://elm-lang.org) support. You can now add Elm support via the following methods:
+  - New app: `rails new <app> --webpack=elm`
+  - Within an existing app: `rails webpacker:install:elm`
+
 ## [1.2] - 2017-04-27
 Some of the changes made requires you to run below commands to install new changes.
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    webpacker (1.1)
+    webpacker (1.2)
       activesupport (>= 4.2)
       multi_json (~> 1.2)
       railties (>= 4.2)
@@ -134,4 +134,4 @@ DEPENDENCIES
   webpacker!
 
 BUNDLED WITH
-   1.14.3
+   1.14.6
--- a/lib/install/config/webpack/paths.yml
+++ b/lib/install/config/webpack/paths.yml
@@ -9,6 +9,7 @@ default: &default
   source: app/javascript
   extensions:
     - .coffee
+    - .elm
     - .js
     - .jsx
     - .ts
--- a/lib/install/examples/elm/Main.elm
+++ b/lib/install/examples/elm/Main.elm
@@ -1,0 +1,54 @@
+module Main exposing (..)
+
+import Html exposing (Html, h1, text)
+import Html.Attributes exposing (style)
+
+-- MODEL
+
+type alias Model =
+  {
+  }
+
+-- INIT
+
+init : (Model, Cmd Message)
+init =
+  (Model, Cmd.none)
+
+-- VIEW
+
+view : Model -> Html Message
+view model =
+  -- The inline style is being used for example purposes in order to keep this example simple and
+  -- avoid loading additional resources. Use a proper stylesheet when building your own app.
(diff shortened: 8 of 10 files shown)


bkuhlmann:
:information_source: Wasn't aware of the yarn linter that is causing build failure due to my use of double quotes. Can't fix tonight but will resolve tomorrow.

bkuhlmann:
:information_source: Fixed the CI, ES Lint, build issues (via interactive rebase). Sorry about that. Ready for review.

gauravtiwari:
@bkuhlmann Great, thanks 👍  I guess other loaders that comes by default with webpacker works as usual right - style, file and css?

bkuhlmann:
@gauravtiwari Yeah, you bet. Actually, I've been seeing some weirdness with the default CSS/SCSS loaders where loading styles via the `stylesheet_pack_tag` will cause the embedded Elm app to not render properly. I was planning to work on that more this weekend. I'm not sure if it is Elm-related, a misconfiguration with Webpack, or something else entirely. I don't think that would prevent this this work from being merged as there is an [Elm CSS](https://github.com/rtfeldman/elm-css) package for styling your app natively. So there are workarounds but, yes, it is important to verify that there are no issues with using native CSS stylesheets as well.

I can close this PR to double check those issues and re-open later if you like?

gauravtiwari:
@bkuhlmann Okay cool, just wanted to confirm because the styles are added a bit differently in an elm app compared to other JS frameworks.  Like you mentioned about the elm CSS package, it seems that CSS and styles loaders might be redundant here and that would mean we need to nix some of those base dependencies if someone is using elm integration. Yepp, sounds like we should explore a bit more and see how things are working together - but either way I don't see why this shouldn't be merged given the integration is working.  What do you think?

bkuhlmann:
@gauravtiwari I agree, it's OK to merge this as it'll allow others to easily have a working environment for their Elm code. As I dig into this more, I can issue a new PR in the future with any additional fixes/enhancements that makes running Elm atop Rails that much smoother (especially concerning stylesheets). This is the first step, at least.

Feel free to merge, I'll have more for you later. :)

gauravtiwari:
One last bit, could you please update changelog?

README.md
```diff
@@ -201,6 +201,42 @@ var railsImagePath = "<%= helpers.image_path('rails.png') %>";
...
+    Rails.application.routes.draw do
+      resource :demo, controller: :demo, only: :show
+    end
+
+### Controllers
+
+    class DemoController < ApplicationController
+      def show
+      end
+    end
```
gauravtiwari: Lets remove routes and controllers part to make things simple to follow up and just follow what other integration suggests (we are not using views anyway).

README.md
```diff
@@ -201,6 +201,42 @@ var railsImagePath = "<%= helpers.image_path('rails.png') %>";
...
+Add `<%= javascript_pack_tag "hello_elm" %>` to `app/views/layouts/application.html.erb`.
+
+### Launch
+
+In two terminal windows, launch each of the following:
+
+    rails s
+    bin/webpack-dev-server
+
+Now you can open `http://localhost:3000/demo` in your default browser and see Elm running on Rails!
```
gauravtiwari: This can be just root url without route :)

lib/install/elm.rb
```diff
@@ -0,0 +1,21 @@
+require "webpacker/configuration"
+
+puts "Copying elm loader to #{Webpacker::Configuration.config_path}/loaders"
+copy_file "#{__dir__}/config/loaders/installers/elm.js", "config/webpack/loaders/elm.js"
```
gauravtiwari: Lets use `Webpacker::Configuration.config_path/loaders/elm.js`

README.md
bkuhlmann: [Fixed](https://github.com/bkuhlmann/webpacker/commit/9b9a11a1045b).

README.md
bkuhlmann: [Fixed](https://github.com/bkuhlmann/webpacker/commit/9b9a11a1045b). Actually, I added this doc because I've never seen this work without actually adding a `demo` controller to as I always get the "Yay, you're on Rails!" page. This doesn't happen for Elm, but for Vue, React, etc. as well.

lib/install/elm.rb
bkuhlmann: [Fixed](https://github.com/bkuhlmann/webpacker/commit/baf36a94bd41). NOTE: This is an issue with vue, react, and angular. Can issue a follow-up PR for those, though.

bkuhlmann:
❗️ Fixes applied. Ready for review. Let me know if approved as I'd like to rebase fixes so that Git can automatically clean them up in order to keep a clean Git history.

README.md
gauravtiwari: @bkuhlmann Yeah, that's good idea, but that more rails specific. Perhaps, we move this section to a common place so, it's documented for all in another PR 👍

README.md
```diff
@@ -201,6 +201,15 @@ var railsImagePath = "<%= helpers.image_path('rails.png') %>";
...
 This is enabled by the `rails-erb-loader` loader rule in `config/webpack/shared.js`.
 
+## Ready for Elm
+
+To use Webpacker with [Elm](http://elm-lang.org), create a new app with `rails new myapp
+--webpack=elm` (or run `rails webpacker:install:elm` on a Rails app already setup with webpacker).
+The Elm library and core packages will be added via Yarn and Elm itself. An example `Main.elm` app
+is also added to your project in `app/javascript` so that you can experiment with Elm right away.
+
+Now you can open `http://localhost:3000/demo` in your default browser and see Elm running on Rails!
```
gauravtiwari: Lets remove this line too since we don't have demo controller.

gauravtiwari:
Looks good @bkuhlmann - one small fix and then ready to merge 👍

README.md
bkuhlmann: That's true it is Rails-specific. Yeah, moving to another section of the README is a good idea since it applies to all. I'd like to experiment with more because if there is a way to do this automatically without manual steps, that'd be ideal.

README.md
bkuhlmann: Ah, yes, missed that. Fixed (via rebase).

bkuhlmann:
:information_source: Rebased and changes applied. Ready for review.

lib/install/elm.rb
```diff
@@ -0,0 +1,22 @@
...
+copy_file "#{__dir__}/examples/elm/hello_elm.js",
+          "#{Webpacker::Configuration.entry_path}/hello_elm.js"
+
+puts "Updating .gitignore to include elm-stuff folder"
+insert_into_file ".gitignore", "/elm-stuff\n", before: "/node_modules\n"
+
+puts "Installing all elm dependencies"
+run "./bin/yarn add elm"
+run "./bin/yarn add --dev elm-hot-loader elm-webpack-loader"
+run "elm package install --yes"
```
gauravtiwari: @bkuhlmann Sorry one last question - would this work globally - `elm package install` ?

README.md
```diff
@@ -201,6 +201,13 @@ var railsImagePath = "<%= helpers.image_path('rails.png') %>";
 
 This is enabled by the `rails-erb-loader` loader rule in `config/webpack/shared.js`.
 
+## Ready for Elm
```
guilleiguaran: [nitpick] I think we maintain a order in this README (most recent integrations are added to the bottom), so I think this info should be added below `Ready for Vue` section

gauravtiwari: 👍

lib/install/elm.rb
guilleiguaran: Same question, I think this will work only if Elm is installed globally in the system

bkuhlmann: Good catch, thanks. Too used to working with my current configuration and wasn't thinking about those that might not have installed Elm via `brew install elm`.

[Fixed](https://github.com/bkuhlmann/webpacker/blob/elm/lib/install/elm.rb#L20) (via rebase). Notice that I've prefixed the command with `yarn run` since the Elm installer has Yarn install Elm prior to this command as a package dependency. The use of `--` before `--yes` is to ensure that option gets passed to Elm and not Yarn so that Elm won't prompt anyone to confirm the Elm packages being installed (in order to keep this process non-interactive).

README.md
bkuhlmann: Yeah, that makes sense.

In terms of being able to scan the document quickly, how about alpha-sorting the options instead? That would improve general organization and easy of scanning the document for available options.

gauravtiwari: Lets do this for now and we can revisit this later. I have some plans for README 👍

gauravtiwari: Please add elm one at the bottom

bkuhlmann: [Fixed](https://github.com/bkuhlmann/webpacker/commit/8a9e9e177d2b7e22c4c6324cc80324dc1a630ca6) (rebased). I've moved it to the bottom just before *Troubleshooting* section.

gauravtiwari:
Thanks @bkuhlmann  👍  🍰

lib/install/config/loaders/installers/elm.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  test: /\.elm$/,
+  exclude: [/elm-stuff/, /node_modules/],
+  loader: 'elm-hot-loader!elm-webpack-loader?verbose=true&warn=true&debug=true'
```
eeue56: Hi, I'm the author of the webpack loader. It's _highly_ recommended that you set the [cwd](https://github.com/elm-community/elm-webpack-loader#cwd-default-null-recommended) option

guilleiguaran: @eeue56 do you want to send a PR adding it? 😊 🙏

bkuhlmann: I can pick this up if @eeue56 doesn't beat me to it.

guilleiguaran: please do :)

PR #330, opened 2017-05-06
Updated Elm Webpack loader to use current working directory.

bkuhlmann:
## Overview

Ensures the [Elm Webpack Loader](https://github.com/elm-community/elm-webpack-loader#cwd-default-null-recommended) sets the current working directory to the Rails app root so that directives in the `elm-package.json` are fully honored.

This allows an Elm developer to add multiple Elm package dependencies as well as layout their app via multiple sub-directories/modules and Webpack will automatically pick up and compile them.

## Details

- Addresses a concern that @eeue56 pointed out in his [PR](https://github.com/rails/webpacker/pull/324#discussion_r115103572).
- See commits for details.

### DIFF
--- a/lib/install/config/loaders/installers/elm.js
+++ b/lib/install/config/loaders/installers/elm.js
@@ -1,5 +1,9 @@
+const path = require('path')
+
+const elmSource = path.resolve(process.cwd())
+
 module.exports = {
   test: /\.elm$/,
   exclude: [/elm-stuff/, /node_modules/],
-  loader: 'elm-hot-loader!elm-webpack-loader?verbose=true&warn=true&debug=true'
+  loader: `elm-hot-loader!elm-webpack-loader?verbose=true&warn=true&debug=true&cwd=${elmSource}`
 }
--- a/lib/install/elm.rb
+++ b/lib/install/elm.rb
@@ -19,4 +19,8 @@ run "./bin/yarn add elm"
 run "./bin/yarn add --dev elm-hot-loader elm-webpack-loader"
 run "yarn run elm package install -- --yes"
 
+puts "Updating elm source location"
+source_path = File.join Webpacker::Configuration.source, Webpacker::Configuration.paths.fetch(:entry, "packs")
+gsub_file "elm-package.json", /\"\.\"\n/, %("#{source_path}"\n)
+
 puts "Webpacker now supports elm 🎉"
--- a/lib/install/examples/elm/hello_elm.js
+++ b/lib/install/examples/elm/hello_elm.js
@@ -1,11 +1,11 @@
 // Run this example by adding <%= javascript_pack_tag "hello_elm" %> to the head of your layout
 // file, like app/views/layouts/application.html.erb. It will render "Hello Elm!" within the page.
 
-import App from './Main'
+import Elm from './Main'
 
 document.addEventListener('DOMContentLoaded', () => {
   const target = document.createElement('div')
 
   document.body.appendChild(target)
-  App.Main.embed(target)
+  Elm.Main.embed(target)
 })


lib/install/config/loaders/installers/elm.js
```diff
@@ -1,5 +1,9 @@
+const path = require('path')
+
+const elmSource = path.join(__dirname, '..', '..', '..')
```
gauravtiwari: Is this elm app source directory? We can just use `path.resolve(paths.source)` from `paths.yml`

lib/install/elm.rb
```diff
@@ -19,4 +19,7 @@
 run "./bin/yarn add --dev elm-hot-loader elm-webpack-loader"
 run "yarn run elm package install -- --yes"
 
+puts "Updating elm source location"
+gsub_file "elm-package.json", /\"\.\"\n/, %("app/javascript/packs"\n)
```
gauravtiwari: We should use `Webpacker::Configuration.entry_path` instead of hardcoding the directory.

lib/install/config/loaders/installers/elm.js
bkuhlmann: I'm afraid that won't work because `elmSource` must point to the root of the generated Rails app where the `elm-package.json` is located. This allows Webpack to load the configuration as defined in the `elm-package.json` along with any Elm package dependencies.

lib/install/elm.rb
bkuhlmann: Yeah, that would be nice but I'm afraid that method yields the absolute path. We need a relative path here because if you happen to move the generated Rails app to a different folder, your Elm app will fail to compile. Example:

```
# elm-package.json with Webpacker::Configuration.entry_path:
{
    "source-directories": [
        "/Users/bkuhlmann/Downloads/example/app/javascript/packs"
    ]
}

# elm-package.json without Webpacker::Configuration.entry_path (using code in this PR):
{
    "source-directories": [
        "app/javascript/packs"
    ]
}

```

lib/install/config/loaders/installers/elm.js
gauravtiwari: @bkuhlmann Right I see, I don't know much about elm that's why asked. In that case lets use `path.resolve('elm-package.json')`

lib/install/elm.rb
gauravtiwari: @bkuhlmann I see then in that case, lets just pull in the raw paths from paths.yml - `File.join(Webpacker::Configuration.source, Webpacker::Configuration.paths.fetch(:entry, 'packs'))`

gauravtiwari: You will get relative paths like so - `"app/assets/javascript/entries"` or `"app/assets/packs"`

lib/install/config/loaders/installers/elm.js
bkuhlmann: [Fixed](https://github.com/bkuhlmann/webpacker/blob/elm-webpack-loader/lib/install/config/loaders/installers/elm.js#L3) (rebase). No worries, happy to explain. This is good feedback and your solution above is much better.

lib/install/elm.rb
bkuhlmann: [Fixed](https://github.com/bkuhlmann/webpacker/blob/elm-webpack-loader/lib/install/elm.rb#L23-L24) (rebase). Ah, thanks, that's much better.

lib/install/elm.rb
```diff
@@ -19,4 +19,9 @@
 run "./bin/yarn add --dev elm-hot-loader elm-webpack-loader"
 run "yarn run elm package install -- --yes"
 
+puts "Updating elm source location"
+source_path = File.join Webpacker::Configuration.source,
```
gauravtiwari: Lets put this in one line :)

gauravtiwari:
@bkuhlmann All looks good if you could just put that in one line I will merge 👍

lib/install/elm.rb
bkuhlmann: [Fixed](https://github.com/bkuhlmann/webpacker/blob/elm-webpack-loader/lib/install/elm.rb#L23). Yeah, my personal limit is a 100 line column.

bkuhlmann:
@gauravtiwari ok, all set. Thanks for the review!

gauravtiwari:
Awesome work 👍

PR #332, opened 2017-05-06
Updated gem version to use semantic versioning.

bkuhlmann:
## Overview

- It's a good practice to use semantic versioning as a standard for
  communicating to others what has changed in this gem when updating
  dependencies in downsteam projects.
- [Documentation](http://semver.org).

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    webpacker (1.2)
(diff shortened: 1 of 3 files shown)


matthewd:
> It's a good practice to use semantic versioning

Disagree.

"Here's a PR that just changes the project versioning policy" probably isn't the best way to open that conversation, either way.

PR #339, opened 2017-05-07
Update README.md with related projects

justin808:
We can add some other related projects, like https://rubygems.org/gems/webpacker-react. However, that one might not be ready for production yet, with less than 2700 downloads as of now.  React on Rails has almost 330,000 downloads. While Webpacker Lite is new, it will have heavy traffic due to React on Rails and it offers a viable alternative to Webpacker for those that do not want to auto-configure webpack.

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,3 +249,7 @@ completed the compilation successfully before loading a view.
 
 ## License
 Webpacker is released under the [MIT License](https://opensource.org/licenses/MIT).
+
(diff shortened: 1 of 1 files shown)


dhh:
I appreciate this sentiment, but I think it's probably better captured in more fully rounded blog posts explaining why someone might want to go with one direction or another. I don't think just these links really serve to explain the differences well enough.

Issue #315, opened 2017-05-02
Feature Request - Remove PreCSS - add cssnext


djforth:
PreCSS seems like it's not well maintained having not been touched in 2 years, when I tried to use in a project with just postcss is caused a lot of issues.  Also not sure what the benefit is of having scss-esque in css when scss is already included?

http://cssnext.io (postcss-cssnext) might be a better option to include? It well maintained and is pushing itself as a sort of babel for CSS, supporting upcoming features like the new  custom properties - http://cssnext.io/features/#custom-properties-var.  It also has autoprefixer included with it.

gauravtiwari:
@djforth Sounds good 👍  Wanna make a PR for this ? 😄

djforth:
Will do :)

PR #344, opened 2017-05-08
Add cssnext and remove PreCSS

djforth:
As mentioned in #315.

I've added postcss-cssnext and remove PreCSS.  I also removed autoprefixer as it included in cssnext and will cause issues if loaded twice


guilleiguaran:
Can you remove the changes in Gemfile.lock and test/test_app/log/development.log from this commit?

djforth:
Looks like the log wasn't in the previous commits so have just removed it.  and reset the gemfile.lock back.

gauravtiwari:
I think we also need to update css next section in README - https://github.com/rails/webpacker#css-next

renchap:
I am not sure why I am getting this error, but after updating to the latest `master` I got a PostCSS error saying that the `cssnext` module was not found.

I tried using `postcss-cssnext` (and not `cssnext`) in my `.postcssrc.yml` config file and it works fine. Not sure why it worked for you and not for me.

Issue #340, opened 2017-05-07
extractCSS in vue-loader does not work for sass and scss styles


morr:
Bug of #310 PR

Video http://take.ms/a7Pe2 (30sec)

This code generates css files
```html
<style scoped>
p {
  font-size: 2em;
  text-align: center;
  color: red;
}
</style>
```

This code does not generate css files
```html
<style scoped lang="scss">
p {
  font-size: 2em;
  text-align: center;
  color: red;
}
</style>
```
Example project (just generated new rails app and added webpacker to it) https://github.com/morr/rails-webpacker-vue-loader

gauravtiwari:
@morr Want to make a PR for this? I think this needs adding explicit loaders per - https://github.com/vuejs/vue-loader/blob/master/docs/en/configurations/extract-css.md

inopinatus:
Not sure this is fixed, or perhaps there is a regression? My test repo for SCSS at https://github.com/inopinatus/vue-scss-webpacker does not emit CSS files.

inopinatus:
Update: thanks to Chris @ Gorails.com who diagnosed this for me.
This PR set `const extractStyles = false` in `loader/vue.js`, which breaks `stylesheet_pack_tag`.

gauravtiwari:
@inopinatus Ahh yes, probably this option needs to be configurable based on env.

sealocal:
If `extractStyles` is (or will be) configurable in `vue.js`, should it also be configurable in `react.js`?

gauravtiwari:
@sealocal Yes that's the plan actually 😉

inopinatus:
I am getting an intermittent "Uncaught TypeError: inherits is not a function" with subcomponents when `extractStyles = true`.

gauravtiwari:
@inopinatus We will disable the `stylesheet_pack_tag` as well in development so it will return `nil`

PR #346, opened 2017-05-08
vue-loader with support of extract styles (#340)

morr:
this PR fixes two things

1. #340 issue
2. Styles parsing in `.vue` should work exactly the same way it works for regular `.css`/`.sass` files. This means that in `loaders/vue` we should use the same loaders configuration that we use in `loaders/sass`. But I'm not sure where to put shared loaders configuration. In this PR I just copy-pasted them.

I also added ability to disable styles extraction with `extractStyles` variable because in some cases it's better to inline styles rather than extract them. 

Any suggestions? How about such file?
`loaders/core/common(or shared or configuration).js`
```js
const cssLoader = [
  { loader: 'css-loader', options: { minimize: env === 'production' } },
  'postcss-loader'
]
module.exports = {
  css: cssLoader,
  sass: cssLoader.concat(['sass-loader?indentedSyntax'])
  scss: cssLoader.concat(['sass-loader'])
}
```

### DIFF
--- a/lib/install/config/loaders/installers/vue.js
+++ b/lib/install/config/loaders/installers/vue.js
@@ -1,13 +1,41 @@
+const ExtractTextPlugin = require('extract-text-webpack-plugin')
+const { env } = require('../configuration.js')
+
+// Change it to false if you prefer Vue styles to be inlined by javascript in runtime
+const extractStyles = false
+
+const cssLoader = [
+  { loader: 'css-loader', options: { minimize: env.NODE_ENV === 'production' } },
+  { loader: 'postcss-loader', options: { sourceMap: true } },
+  'resolve-url-loader'
+]
+const sassLoader = cssLoader.concat([
+  { loader: 'sass-loader', options: { sourceMap: true, indentedSyntax: true } }
+])
+const scssLoader = cssLoader.concat([
+  { loader: 'sass-loader', options: { sourceMap: true } }
+])
+
+function vueStyleLoader(loader) {
+  if (extractStyles) {
+    return ExtractTextPlugin.extract({
+      fallback: 'vue-style-loader',
+      use: loader
+    })
+  }
+  return ['vue-style-loader'].concat(loader)
+}
+
 module.exports = {
   test: /.vue$/,
   loader: 'vue-loader',
   options: {
-    extractCSS: true,
     loaders: {
       js: 'babel-loader',
       file: 'file-loader',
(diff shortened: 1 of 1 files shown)


lib/install/config/loaders/installers/vue.js
```diff
@@ -1,13 +1,36 @@
...
+const scssLoader = cssLoader.concat(['sass-loader'])
+
+function vueLoader(loader) {
+  if (extractStyles) {
+    return ExtractTextPlugin.extract({
+      fallback: 'vue-style-loader',
+      use: loader
+    })
+  }
+  return ['vue-style-loader'].concat(sassLoader)
```
RavenXce: Shouldn't this be  `loader` not `sassLoader`?

lib/install/config/loaders/installers/vue.js
```diff
@@ -1,13 +1,36 @@
...
+const extractStyles = true
+
+const cssLoader = [
+  { loader: 'css-loader', options: { minimize: env === 'production' } },
+  'postcss-loader'
+]
+const sassLoader = cssLoader.concat(['sass-loader?indentedSyntax'])
+const scssLoader = cssLoader.concat(['sass-loader'])
+
+function vueLoader(loader) {
```
RavenXce: Seems like a rather confusing name as this is prepending the `vue-style-loader`, not `vue-loader`.

lib/install/config/loaders/installers/vue.js
morr: oh my... yes this should be `loader`

gauravtiwari:
@morr Sorry for late review - please could you add `postcss` too in the loaders list.

morr:
@gauravtiwari it was already added to loaders list. You mean this, right?
monosnap 2017-05-26 14-26-44

I rebased pull request on current `rails/webpacker#master` branch and added to `vue.js` latest loaders changes from `loaders/core/sass.js`

gauravtiwari:
@morr Ahh yes! Looks good to me - I will merge 👍  🍰 

Thank you :)

lib/install/config/loaders/installers/vue.js
```diff
@@ -1,13 +1,41 @@
+const ExtractTextPlugin = require('extract-text-webpack-plugin')
+const { env } = require('../configuration.js')
+
+// Change it to false if you prefer Vue styles to be inlined by javascript in runtime
+const extractStyles = false
```
inopinatus: Does this need `const extractStyles = true` to avoid a regression on #310?

PR #350, opened 2017-05-08
Adjust Elm compilation flags for production

skarger:
Two independent changes:

1. The more important change is to not compile elm with debug in the `production` environment. Otherwise a web page with a compiled Elm app will display the debugger window:

elm-debugger-small

2. Less important/more uncertain change: do not user `elm-hot-loader` in production.
The docs say that Hot Module Replacement is usable in production but experimental and possibly buggy. See the [Caveats section](https://webpack.github.io/docs/hot-module-replacement-with-webpack.html#caveats) at https://webpack.github.io/docs/hot-module-replacement-with-webpack.html.
I am not a webpack expert and would be happy to change this back if someone provides a rationale, but my take is that HMR shouldn't be the default for production.

For both I switched only on the `production` environment, but there is a case that it should have the same behavior for `staging`.

### DIFF
--- a/lib/install/config/loaders/installers/elm.js
+++ b/lib/install/config/loaders/installers/elm.js
@@ -4,11 +4,11 @@ const { env } = require('../configuration.js')
 const elmSource = path.resolve(process.cwd())
 
 const loaderOptions = () => {
-  if (env.NODE_ENV === 'production') {
-    return `elm-webpack-loader?cwd=${elmSource}`
+  if (['development', 'test'].includes(env.NODE_ENV)) {
+    return `elm-hot-loader!elm-webpack-loader?cwd=${elmSource}&verbose=true&warn=true&debug=true`
   }
 
-  return `elm-hot-loader!elm-webpack-loader?cwd=${elmSource}&verbose=true&warn=true&debug=true`
+  return `elm-webpack-loader?cwd=${elmSource}`
 }
 
 module.exports = {
--- a/lib/install/config/loaders/installers/elm.js
+++ b/lib/install/config/loaders/installers/elm.js
@@ -2,11 +2,17 @@ const path = require('path')
 const { env } = require('../configuration.js')
 
 const elmSource = path.resolve(process.cwd())
-const elmLoader = env.NODE_ENV === 'production' ?  'elm-webpack-loader' : 'elm-hot-loader!elm-webpack-loader'
-const debug = env.NODE_ENV === 'production' ? 'false' : 'true'
+
+const loaderOptions = () => {
+    if (env.NODE_ENV === "production") {
+        return `elm-webpack-loader?cwd=${elmSource}`
+    }
+
+    return `elm-hot-loader!elm-webpack-loader?cwd=${elmSource}&verbose=true&warn=true&debug=true`
+}
 
 module.exports = {
   test: /\.elm$/,
   exclude: [/elm-stuff/, /node_modules/],
-  loader: `${elmLoader}?verbose=true&warn=true&debug=${debug}&cwd=${elmSource}`
+  loader: loaderOptions()
 }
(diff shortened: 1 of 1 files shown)


gauravtiwari:
@skarger This sounds like correct behaviour. If you could fix the spacing issue on line 5, I will merge 👍 

```
/home/travis/build/rails/webpacker/lib/install/config/loaders/installers/elm.js
  5:52  error  Multiple spaces found before ''elm-webpack-loader''  no-multi-spaces
```

lib/install/config/loaders/installers/elm.js
```diff
@@ -1,9 +1,12 @@
...
 
 const elmSource = path.resolve(process.cwd())
+const elmLoader = env.NODE_ENV === 'production' ?  'elm-webpack-loader' : 'elm-hot-loader!elm-webpack-loader'
+const debug = env.NODE_ENV === 'production' ? 'false' : 'true'
 
 module.exports = {
   test: /\.elm$/,
   exclude: [/elm-stuff/, /node_modules/],
-  loader: `elm-hot-loader!elm-webpack-loader?verbose=true&warn=true&debug=true&cwd=${elmSource}`
+  loader: `${elmLoader}?verbose=true&warn=true&debug=${debug}&cwd=${elmSource}`
```
gauravtiwari: Lets do it like this instead - 

```js
const loaderOptions = () => {
  if (env.NODE_ENV === "production") {
    return `elm-webpack-loader?&cwd=${elmSource}`
  }

  return `elm-hot-loader!elm-webpack-loader?verbose=true&warn=true&debug=true&cwd=${elmSource}`
}

{ 
  loader: loaderOptions()
}
```

skarger:
Hi @gauravtiwari, I made your suggested change.

gauravtiwari:
Thanks @skarger Looking good 👍 

@bkuhlmann Do you want take another look please?

lib/install/config/loaders/installers/elm.js
```diff
@@ -4,11 +4,11 @@ const { env } = require('../configuration.js')
 const elmSource = path.resolve(process.cwd())
 
 const loaderOptions = () => {
-    if (env.NODE_ENV === "production") {
-        return `elm-webpack-loader?cwd=${elmSource}`
-    }
+  if (env.NODE_ENV === 'production') {
```
bkuhlmann: :art: Thanks for doing this! Would you want to check for *development* and *test* environments instead? This way it'd be more of a *whitelist*. Those environments in the whitelist would get elm-hot-loader, debug, etc. support while the rest would get a more production-like setup. I mention this because I've worked in environment where you might have the following production-like environments:

- review
- stage
- production

By limiting the additional options to only development and test environments this might allow others to add additional environments, if need be, without having to modify the code further. Might make for a nice experience.

skarger: @bkuhlmann sure, made that change.

gauravtiwari: @bkuhlmann It's just an ENV flag so, it could be toggled in any environment. `NODE_ENV=production ./bin/webpack` so, don't think we need to be that explicit

skarger: The current revisions is the mirror image of that - if you wanted debug-compilation in your staging environment for example, you could say `NODE_ENV=development ./bin/webpack`.

To fully support various scenarios people might have, I think we'd need to expose an env variable allowing people to customize the loader/compiler flags. I didn't want to go down that road yet, in hopes that this PR will provide reasonable defaults without additional user config options.

gauravtiwari: @skarger If you could kindly revert the last commit and I will merge. Basically, the webpack config used within webpacker follows this convention so, lets stick to that 👍

skarger: @gauravtiwari OK makes sense, done.

gauravtiwari:
thanks 👍

lib/install/config/loaders/installers/elm.js
bkuhlmann: @gauravtiwari, @skarger Ah, yes, sorry about that. Read that incorrectly. Makes sense.

PR #352, opened 2017-05-09
Update the list of file formats to be compresses.

mareksuscak:
*.woff and *.woff2 file formats are already compressed
and further compressions produces diminishing returns.

I used the following list from Fastly:
https://www.fastly.com/blog/new-gzip-settings-and-deciding-what-compress/

cc @gauravtiwari


gauravtiwari:
Thanks 👍  🍰

PR #355, opened 2017-05-09
binstubs: do not exec via the shell, but rather execute the programs directly

fsateler:
Windows cmd does not support the ENVVAR=VALUE syntax prepended to a command, so the binstubs did not work
there. To avoid that, use the exec form that does not go via the shell, and instead executes the program directly.

Fixes: #245

--

It appears the testsuite does not test these binstubs, and I don't know how to generate the resulting file to execute....


gauravtiwari:
Looks good to me 👍

Issue #351, opened 2017-05-09
Elm doesn't work "out of the box"


joshuaclayton:
Hi!

I recently tried out the new Elm support introduced by @bkuhlmann (thank you so much for this!) and found it didn't work directly out of the box. I created a [GitHub repo](https://github.com/joshuaclayton/rails_elm_webpacker_heroku_example_app) demonstrating the steps I had to take, but one of the more confusing ones was [adding the `.elm` extension to the JS import](https://github.com/joshuaclayton/rails_elm_webpacker_heroku_example_app/commit/a45ee7d6303655a064dd5c6f6162cc821b7aa6c7). Another alternative would be to [add `.elm` to the list of extensions supported by Webpack](https://github.com/joshuaclayton/rails_elm_webpacker_heroku_example_app/commit/cf20c6f301602e042be16caaa601f761bd80d231).

Is there a "right solution" in this case? I'm leaning towards the latter, but wanted to get some feedback.

In either case, does it make sense to adjust `hello_elm.js`, adding the extension, or update `elm.rb`'s generated output to remind developers to add the `.elm` extension to the list? I'm happy to help here!

gauravtiwari:
@joshuaclayton Hey, I think .elm is already in the extensions list. Did you re-run `webpacker:install` to replace paths.yml?
 - https://github.com/rails/webpacker/blob/master/lib/install/config/webpack/paths.yml

joshuaclayton:
@gauravtiwari aha, it is! Re-running it added it in correctly. Does it make sense to let users know to do that, or automate it somehow?

bkuhlmann:
@joshuaclayton Thanks, glad you are making use of it. :)

@gauravtiwari is correct, the `.elm` extension is added to the `paths.yml` so you'll get that.

[...]

PR #356, opened 2017-05-09
Fixed Elm install issue with older Webpacker gem versions.

bkuhlmann:
## Overview

Ensures the `elm.rb` script remains the single source of all operations related to adding Elm support to Rails/Webpack (in case older versions of this gem is used during app generation/install).

## Details

- Fixes an [issue](https://github.com/rails/webpacker/issues/351) that @joshuaclayton discovered.
- @gauravtiwari This might be a worthwhile practice to adopt for the other JavaScript installers.

### DIFF
--- a/lib/install/config/webpack/paths.yml
+++ b/lib/install/config/webpack/paths.yml
@@ -9,7 +9,6 @@ default: &default
   source: app/javascript
   extensions:
     - .coffee
-    - .elm
     - .js
     - .jsx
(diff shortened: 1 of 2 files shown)


gauravtiwari:
@bkuhlmann Great 👍  Lets follow this for any new installers (since the old ones are already there)

joshuaclayton:
❤️  👏

PR #360, opened 2017-05-10
Lazy compile in test env

javan:
* Adds new `Webpacker.compile` API (that invokes the `webpacker:compile` rake task)
* Lazily compiles in test env to avoid `Webpacker::FileLoader::*` exceptions
* Eliminates the need enhance Rails test tasks

Differences from https://github.com/rails/webpacker/pull/341
* Doesn't require any changes to an application's test classes
* Doesn't attempt to recompile when files change

/cc @gauravtiwari @dhh

### DIFF
--- a/lib/tasks/webpacker/compile.rake
+++ b/lib/tasks/webpacker/compile.rake
@@ -18,17 +18,6 @@ namespace :webpacker do
     puts "Compiled digests for all packs in #{Webpacker::Configuration.packs_path}: "
     puts JSON.parse(File.read(Webpacker::Configuration.manifest_path))
   end
-
-  desc "Compile javascript packs using webpack for test with digests"
-  task compile_before_test: ["webpacker:compile"] do
-    Webpacker::Manifest.load(Webpacker::Manifest.file_path)
-  end
-end
-
-# Compile packs prior to system and controller tests running
-if Rake::Task.task_defined?("test:system")
-  Rake::Task["test:system"].enhance(["webpacker:compile_before_test"])
-  Rake::Task["test:controllers"].enhance(["webpacker:compile_before_test"])
 end
 
 # Compile packs after we've compiled all other assets during precompilation
--- a/lib/tasks/webpacker/compile.rake
+++ b/lib/tasks/webpacker/compile.rake
@@ -8,7 +8,7 @@ namespace :webpacker do
     puts "Compiling webpacker assets 🎉"
     asset_host = Rails.application.config.action_controller.asset_host
     asset_env = asset_host ? "ASSET_HOST=#{asset_host}" : ""
-    result = `#{asset_env} NODE_ENV=#{Webpacker::Env.current} ./bin/webpack --json`
+    result = `#{asset_env} NODE_ENV=#{Webpacker.env} ./bin/webpack --json`
 
     unless $?.success?
       puts JSON.parse(result)["errors"]
--- a/lib/webpacker.rb
+++ b/lib/webpacker.rb
@@ -1,8 +1,3 @@
-require "webpacker/env"
-require "webpacker/configuration"
-require "webpacker/manifest"
-require "webpacker/compiler"
-
 module Webpacker
   extend self
 
@@ -18,4 +13,8 @@ module Webpacker
   end
 end
 
+require "webpacker/env"
+require "webpacker/configuration"
+require "webpacker/manifest"
+require "webpacker/compiler"
 require "webpacker/railtie" if defined?(Rails)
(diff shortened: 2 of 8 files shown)


gauravtiwari:
@javan This looks really good 👍  After we discussed the other day I realised that re-compilation isn't necessary if we restart the `dev-server/watcher` programmatically whenever a pack is added/deleted (at the moment it doesn't).  That would be good enough because it will ensure that manifest gets updated, which will ensure that tests will run just fine. I think we will still need #292 to make sure manifest change gets picked up in test environment though.

jackjennings:
Would love to get this in a release—any plans for cutting a new version of the gem with this change soon?

dhh:
Yeah. Will cut a release tomorrow 👍

gauravtiwari:
Shall we wait until Monday? I was thinking to finish up README PR for this release too.

dhh:
Sure, that's fine. Do wrap it up. Monday it is then.

pmk1c:
Is it correct, that I still have to run webpack-dev-server during tests with this change? Seems to be the case. The assets get compiled during test, but the asset url Webpacker generates is "http://localhost:8080/packs/application.js"

zacharyw:
Is it best practice to check these compiled test assets into version control, or no?

javan:
@zacharyw, no. I'd .gitignore them.

Jeremy-Walton:
@javan I updated my webpacker gem like @pmk1c and now it will generate the `public/packs-test` when I run my tests. However, once I have that folder, any changes I make to my packs are not being recompiled when I run the tests and I have to delete my `public/packs-test` folder to get it to regenerate it. Is there something I am doing wrong here? Shouldn't it recompile for the tests when there have been changes made?

zacharyw:
@Jeremy-Walton I'm seeing the same thing. As a workaround I'm back to manually compiling assets before spec runs:

```
config.before(:suite) do
    `bin/webpack`
  end
```

richseviora:
@Jeremy-Walton Also observing the same.
@zacharyw thanks for the workaround!

gauravtiwari:
The idea was to keep this simple but if you need to recompile during tests just call `compile` on webpacker module: 

```rb
before(:suite) do
  Webpacker.compile
end
```

screen shot 2017-06-03 at 22 56 09

gauravtiwari:
The above will ensure that packs are recompiled and manifest is reloaded - ref: https://github.com/rails/webpacker/blob/549988b795ca1b1dcec97ecef392e2dceb347bf7/lib/webpacker.rb#L10

richseviora:
It sounds like we save unnecessary compilation calls at the expense of introducing an implicit requirement that the user invokes `WebPacker.compile` (manually or before suite execution) whenever their JS code changes.

Is there a better way we could communicate this? I totally appreciate that we don't want to recompile the entire pack for *every* test run, but I think a lot of users are going to run into similar concerns.

justin808:
@richseviora See #464. Also, I wrote the code for [React on Rails](https://github.com/shakacode/react_on_rails) that does this. We use the file mtimes.

gauravtiwari:
Yepp, actually there was a PR made (#341 - https://github.com/rails/webpacker/pull/341/files#diff-23ac5b11daf84548b507c40511c4aa2eR21) to add very same feature but in the end went with more simpler and implicit solution.

Perhaps we should add that too so both options work

// cc @javan

raldred:
For me the cached version would be favourable. The speed of compilation whilst not an issue on CI is pretty frustrating developing locally. Particularly on a large react app.

stevehanson:
Compiling webpack in the `before(:suite)` every time I ran my tests was slowing me down quite a bit, so I ended up pulling @gauravtiwari's solution from his now-closed PR #341 into my codebase (slightly modified), and it's working great for me. 

Using his approach, Webpack now only compiles if my source has changed. Thanks @gauravtiwari for that and for all of your diligent work on this gem. Would love to see your solution included into the Webpacker gem, as I think the functionality would be broadly useful.

For anyone looking for a quick fix, [here is a gist of the code](https://gist.github.com/stevehanson/d51aa2e70dc0c9dac07c0fb6317644b4) I ended up using, or check out PR #341.

raldred:
Thanks for sharing @stevehanson I wonder if #341 could be reopened and included as well. Both gives everyone the option depending on their use.

gauravtiwari:
@stevehanson Thanks for sharing. Looks great 👍 

@raldred @stevehanson Seems like there is real requirement for this so I will reopen #341 and we can review it together. It would be nice to have both options.

kernow:
I've just run into this issue too, a solution where webpack is recompiled only when something has changed before the test suite is run is whats needed IMHO. 👍 for something like #341

PR #361, opened 2017-05-10
execute the same ruby interpreter as is currently running for all installer shell commands

AshlessCinder:
an extension of 01749eecaa669f461b1361a7dc1f7dc1516426ae
to the other shell commands

related to https://github.com/rails/webpacker/issues/245


gauravtiwari:
Thanks @AshlessCinder 👍

PR #364, opened 2017-05-10
Ensure child processes are killed

leifcr:
When running webpacker dev server under foreman or guard-process, the node child process can sometimes continue to run. This ensure the child process is killed.

To test, install foreman ```gem install foreman```, then add this to a Procfile:

```
web: bundle exec rails server
webpacker: ./bin/webpack-dev-server
```
Run ```foreman start```

Start a new terminal and run ```foreman stop```, and ```ps aux | grep webpacker``` will sometimes show the child process continue in the background.

Hitting ctrl+c when running interactive with foreman terminates the command properly, while start/stop does not. Seems like SIGINT is captured by the child process, while SIGTERM is not.

### DIFF
--- a/lib/install/bin/webpack-dev-server.tt
+++ b/lib/install/bin/webpack-dev-server.tt
@@ -4,6 +4,15 @@ $stdout.sync = true
 require "shellwords"
 require "yaml"
 
+def descendant_processes(base=nil)
+  return [] if base.nil?
+  descendants = Hash.new{|ht,k| ht[k]=[k]}
+  Hash[*`ps -eo pid,ppid`.scan(/\d+/).map{|x|x.to_i}].each{|pid,ppid|
+    descendants[ppid] << descendants[pid]
+  }
+  descendants[base].flatten - [base]
+end
+
 ENV["RAILS_ENV"] ||= "development"
 RAILS_ENV = ENV["RAILS_ENV"]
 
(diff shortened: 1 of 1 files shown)


lib/install/bin/webpack-dev-server.tt
```diff
@@ -30,6 +39,26 @@ end
...
+Signal.trap("INT") do
+  exit 1 if $webpacker_dev_server_pid.nil?
+  # Kill the subprocesses
+  descendant_processes($webpacker_dev_server_pid).each do |d|
+    Process.kill(9, d)
+  end
+end
+
 Dir.chdir(APP_PATH) do
-  exec newenv, *cmdline
```
matthewd: As it's currently using `exec`, the immediate "child" process takes over here. That would seem to make any handling of further processes a matter for that one (the real `webpack-dev-server`). Does that process not manifest the same problem if run directly?

gauravtiwari:
@leifcr Actually never had this issue with foreman. Please could you add `--watch-stdin` or ` --stdin` arg to binstub and see if the issue still persist?

```rb
cmdline = [DEV_SERVER_BIN, "--", "--progress", "--color", "--watch-stdin", "--config", DEV_SERVER_CONFIG] + ARGV
```

leifcr:
I've tested more and found that the initial test app I created used Rails 5.1 0 + webpack 1.2, which has a bit different webpack-dev-server bin from PR #367.

PR #367 solves the problem by using ```yarn run webpack-dev-server```, which also tracks child processes, and properly kills them.

Closing this, as the solution in #367 is merged.

PR #366, opened 2017-05-11
Configs are .js or .yml?

arthurnn:


gauravtiwari:
@arthurnn Thanks for the PR 👍  But seems like we need to explain a bit more here than changing extension - apparently both file types live there and are essential config files.

gauravtiwari:
Lets address this in #372

gauravtiwari:
Please feel free to submit any notes in #372

Issue #306, opened 2017-04-29
Assets Precompile fails


ankitsinghaniyaz:
I think I am missing something, I updated my gems and since then this is failing. 

I also ran 
```
rails webpacker:install
rails webpacker:install:vue
```

Here is my trace:

```
aks:formester$ RAILS_ENV=production rake assets:precompile --trace
** Invoke assets:precompile (first_time)
** Invoke assets:environment (first_time)
** Execute assets:environment
** Invoke environment (first_time)
** Execute environment
** Invoke yarn:install (first_time)
** Execute yarn:install
yarn install v0.21.3
[1/4] Resolving packages...
success Already up-to-date.
Done in 0.55s.
** Execute assets:precompile
rake aborted!
ExecJS::RuntimeError: SyntaxError: Unexpected token: punc ())
/home/aks/playground/formester/app/assets/config/manifest.js:2
JS_Parse_Error.get ((execjs):3538:621)
(execjs):4060:47
(execjs):1:102
Object.<anonymous> ((execjs):1:120)
Module._compile (module.js:570:32)
Object.Module._extensions..js (module.js:579:10)
Module.load (module.js:487:32)
tryModuleLoad (module.js:446:12)
Function.Module._load (module.js:438:3)
Module.runMain (module.js:604:10)
/home/aks/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/execjs-2.7.0/lib/execjs/external_runtime.rb:39:in `exec'
/home/aks/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/execjs-2.7.0/lib/execjs/external_runtime.rb:21:in `eval'
/home/aks/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/execjs-2.7.0/lib/execjs/external_runtime.rb:46:in `call'
/home/aks/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/uglifier-3.2.0/lib/uglifier.rb:195:in `run_uglifyjs'
/home/aks/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/uglifier-3.2.0/lib/uglifier.rb:167:in `compile_with_map'
/home/aks/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/sprockets-4.0.0.beta4/lib/sprockets/uglifier_compressor.rb:52:in `call'
[...]

PR #369, opened 2017-05-11
Fix production UglifyJS constant error

Davidzhu001:
issue at

#291  #306


gauravtiwari:
@Davidzhu001 Thanks, but please use this solution locally. Don't think it's stable enough to merge. 👍  Perhaps, best fix is to document the uglify usage somewhere so it's clear to people.

Davidzhu001:
@gauravtiwari Ok, thanks

PR #374, opened 2017-05-12
Use https in readme's gem instructions.

ktusznio:
Using the git protocol, yields the following warning during a `bundle install`:

> The git source `git://github.com/rails/webpacker.git` uses the `git` protocol, which transmits data without encryption. Disable this warning with `bundle config git.allow_insecure true`, or switch to the `https` protocol to keep your data secure.

Use https instead.


gauravtiwari:
@ktusznio Thanks, started doing some work here #372. Please submit any comments there 👍

PR #385, opened 2017-05-14
Added Table of Contents auto-generation to README.

bkuhlmann:
## Overview

Per the conversation in this [issue](https://github.com/rails/webpacker/pull/372), this provides auto-generation, auto-updating of the Table of Contents for this project.

## Details

- [Documentation](https://github.com/bkuhlmann/tocer)
- @gauravtiwari created this as a PR as I figured it would be easier to review that way.

### DIFF
--- a/.tocerrc
+++ b/.tocerrc
@@ -1,0 +1,1 @@
+:label: "## Table of Contents"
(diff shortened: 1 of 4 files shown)


bkuhlmann:
:information_source: I apologize for the noise this has created but I'm afraid we can't use Tocer for this project as Webpacker needs to support older versions of Ruby and the Tocer gem makes use of the Ruby *refinements* language feature that is incompatible with older versions of Ruby. 😢

PR #387, opened 2017-05-14
feat: add resolve-url-loader to sass loader

### LINKED ISSUE #384: css-loader cannot resolve resources (.eot, .woff, etc.)
In `hello_react.jsx`, I import `hello_react.css`:
``` typescript
import "./hello_react.css";
```
In `hello_react.css`, I import the CSS contained in a node_module:
``` css
@import "~@blueprintjs/core/dist/blueprint";
```

When webpacker builds, I get the following error:

> ERROR in ./\~/css-loader?{"minimize":false}!./\~/postcss-loader/lib!./~/sass-loader/lib/loader.js!./app/javascript/packs/hello_react.css

> Module not found: Error: Can't resolve '../resources/icons/icons-16.eot' in '/Users/btc/Documents/src/respectable/r3/react/app/javascript/packs'

`icons-16.eot` resides within `node_modules/@blueprintjs/core/resources/icons`. It is not in `app/javascript/packs`, but it seems like `app/javascript/packs` is the only place where the tool is configured to look. Is this a bug? 

- [ ] My first instinct is that this may be a problem with the configuration of the `css-loader` in `config/webpacker/loaders`. 

- [ ] My second instinct is that this may be a problem with the configuration in `config/webpack/shared.js`.

- [ ] My third instinct is that locating these font files requires https://github.com/bholloway/resolve-url-loader. However, my attempt to integrate the loader hasn't fixed the problem. It results in this error:

> resolve-url-loader cannot operate: CSS error

> no such file or directory, open app/javascript/packs/blueprint.css.map'

It's looking in `app/javascript/packs` again! That file is right next to `blueprint.css` in the directory of the node_modules dependency.

```
react (master) λ. ls -l node_modules/@blueprintjs/core/dist/                                                                                                                             1
total 5424
drwxr-xr-x   8 btc  staff      272 May 12 18:17 accessibility
-rw-r--r--   1 btc  staff   275808 May 12 14:02 blueprint.css
-rw-r--r--   1 btc  staff  1706972 May 12 14:02 blueprint.css.map
```

- [ ] My fourth instinct is that webpack's context is incorrect.

I understand that this may not actually be a problem with webpacker itself, and instead might be a problem with a dependency or god forbid a user error. However, it feels like this initial question belongs here. If I am mistaken, I apologize.

btc:
Fixes #384. Reference - https://github.com/webpack-contrib/sass-loader#problems-with-url

### DIFF
--- a/lib/install/config/loaders/core/sass.js
+++ b/lib/install/config/loaders/core/sass.js
@@ -7,8 +7,9 @@ module.exports = {
     fallback: 'style-loader',
     use: [
       { loader: 'css-loader', options: { minimize: env.NODE_ENV === 'production' } },
-      'postcss-loader',
-      'sass-loader'
+      { loader: 'postcss-loader', options: { sourceMap: true } },
+      'resolve-url-loader',
+      { loader: 'sass-loader', options: { sourceMap: true } }
     ]
   })
 }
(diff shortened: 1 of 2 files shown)


cleverboy32:
i cant fix this question

PR #388, opened 2017-05-14
Silence task output in test environment.

tristandunn:
Once lazy compiling in #360 is released, these tasks will generate output in tests. This prevents it in the test environment.


gauravtiwari:
@tristandunn Probably #539 might help with the output noise. What do you think?

gauravtiwari:
And sorry it took a while to review this one 😞

javan:
I'd prefer to see the current setup inverted. Move webpack command invocations to the lib so `Webacker.compile` doesn't need to call a rake task.

dhh:
This is fixed via #627 👍

tristandunn:
@dhh Thanks. Works like a charm.

Issue #383, opened 2017-05-13
Missing pathMake from elm-webpack-loader


gauravtiwari:
@bkuhlmann Just tried elm support and it seems that loader can't access `elm-make` file when installed locally with `yarn`.

```
hvw00000gn/T/117413-10953-10rs2tp.wpu82xogvi.js
Could not find Elm compiler "elm-make". Is it installed?
```

After adding `pathMake` to loader config, it works. 

```js
const path = require('path')
const { env } = require('../configuration.js')

const elmSource = path.resolve(process.cwd())

const loaderOptions = () => {
  if (env.NODE_ENV === 'production') {
    return `elm-webpack-loader?cwd=${elmSource}&pathToMake=${elmSource}/node_modules/.bin/elm-make`
  }

  return `elm-hot-loader!elm-webpack-loader?cwd=${elmSource}&verbose=true&warn=true&debug=true&pathToMake=${elmSource}/node_modules/.bin/elm-make`
}

module.exports = {
  test: /\.elm$/,
  exclude: [/elm-stuff/, /node_modules/],
  loader: loaderOptions()
}

```

bkuhlmann:
@gauravtiwari I'm not able to reproduce locally. Here are my steps:

```
brew uninstall elm # <-- I have to do this step first since I like to have Elm installed globally.
cd ~/scratch
git clone https://github.com/rails/webpacker.git
cd webpacker
bundle exec rake install
cd ~/scratch
rails new demo --webpack=elm
cd demo
rails s
bin/webpack-dev-server # <-- Run in a second terminal
```

Also, if I run the following within the `demo` app directory:

`yarn elm make`

I get the following output:

```
yarn elm-make v0.23.4
$ "~/scratch/demo/node_modules/.bin/elm-make"
Success! Compiled 0 modules.
✨  Done in 0.25s.
```

It is weird to see that only 0 modules are compiled, though.

bkuhlmann:
Opps, forgot to add the source path when running `elm make`. Here is what you'd want to run:

`yarn elm make app/javascript/packs/**/*.elm`

...and now the resulting output looks better:

```
yarn elm v0.23.4
$ "~/scratch/demo/node_modules/.bin/elm" make app/javascript/packs/Main.elm
Success! Compiled 1 module.
Successfully generated index.html
✨  Done in 0.42s.
```

Now we see that one module id compiled (i.e. `Main.elm`).

gauravtiwari:
[...]

PR #393, opened 2017-05-16
Fixed missing path to Elm Make.

bkuhlmann:
## Overview

Resolves an issue with running the `bin/webpack` binstub where the path to Elm Make could not be resolved due to being installed via Yarn. This would work if Elm was globally installed, though. In this case we don't want to make any assumptions on a globally installed version of Elm and only use what is provided by Yarn.

## Details

- Resolves this [Issue](https://github.com/rails/webpacker/issues/383).


gauravtiwari:
Thank you 🍰🙂

PR #397, opened 2017-05-17
Fix asset host support & improve output path

### LINKED ISSUE #320: javascript_pack_tag seems to add path twice with asset host
Since upgrading to 1.2 `javascript_pack_tag` no longer seems to render the correct url.

I am using an asset host, specified in `production.rb`. 

In previous versions, the`manifest.json` file copiled for production looked something like this:

```json
{
  "admin.js" : "/packs/admin-s089sdf098.js"
}
```

But now it explicitly includes the asset host path, rendering something like this:

```json
{
  "admin.js": "jiu8a9d8f9.cloudfront.net/packs/admin-s089sdf098.js"
}
```

The effect of this is that the `javascript_pack_tag` helper renders the path twice as follows:

`https://jiu8a9d8f9.cloudfront.net/javascripts/jiu8a9d8f9.cloudfront.net/packs/admin-s089sdf098.js`

Which doesn't work. 

Note that "javascripts" has made it into the link too, which I think is incorrect. 

I have run webpacker:install as part of the upgrade process, but perhaps I have missed something else.

Any ideas would be welcome!

javan:
* Fixes duplicated asset hosts (Fixes #320, Closes #335)
* Fixes that asset hosts would be ignored when defined as `Proc` or when set on `ActionController::Base.asset_host` directly
* Fixes that `webpacker:compile` and `bin/webpack` would generate dev server URLs in development mode. This should only happen when using `bin/webpack-dev-server`.
* **Changes** `paths.output` so that it's:
  1. Relative to `public/`
  2. Distinct from `paths.entry`
  3. Not broken if you change it. Previously webpacker would compile into `paths.output/paths.entry`, but wouldn't include `paths.output` in generated URLs. So, if you changed `paths.output`, you'd end up with busted URLs:
    
    Before:

    ```ruby
    >> Webpacker::Configuration.paths[:entry]
    => "packs"

    >> Webpacker::Configuration.paths[:output]
    => "public/assets"

    >> Rails.application.config.action_controller.asset_host
    => "https://cdn.example.com"

    >> Webpacker::Manifest.lookup("application.js")
    => "https://cdn.example.com/packs/application.js"

    # That ^ URL would not work because the compiled file isn't there
    >> Rails.root.join("public/packs/application.js").exist?
    => false

    # Because the compiled file is here
    >> Rails.root.join("public/assets/packs/application.js").exist?
    => true
    ```

    After:
    ```ruby
    >> Webpacker::Configuration.paths[:entry]
    => "packs"

    >> Webpacker::Configuration.paths[:output]
    => "assets"

    >> Webpacker::Manifest.lookup("application.js")
    => "https://cdn.example.com/assets/application.js"

    >> Rails.root.join("public/assets/application.js").exist?
    => true
    ```

### DIFF
--- a/lib/install/config/webpack/configuration.js
+++ b/lib/install/config/webpack/configuration.js
@@ -10,17 +10,29 @@ const loadersDir = join(__dirname, 'loaders')
 const paths = safeLoad(readFileSync(join(configPath, 'paths.yml'), 'utf8'))[env.NODE_ENV]
 const devServer = safeLoad(readFileSync(join(configPath, 'development.server.yml'), 'utf8'))[env.NODE_ENV]
 
-// Compute public path based on environment and ASSET_HOST in production
-const ifHasCDN = env.ASSET_HOST !== undefined && env.NODE_ENV === 'production'
-const devServerUrl = `http://${devServer.host}:${devServer.port}/${paths.entry}/`
-const publicUrl = ifHasCDN ? `${env.ASSET_HOST}/${paths.entry}/` : `/${paths.entry}/`
-const publicPath = env.NODE_ENV !== 'production' && devServer.enabled ? devServerUrl : publicUrl
+function removeOuterSlashes(string) {
+  return string.replace(/^\/*/, '').replace(/\/*$/, '')
+}
+
+function formatPublicPath(host = '', path = '') {
+  host = removeOuterSlashes(host)
+  path = removeOuterSlashes(path)
+  if (host && !/^http/i.test(host)) {
+    host = `//${host}`
+  }
+  return `${host}/${path}/`
+}
+
+const output = {
+  path: resolve('public', paths.output),
+  publicPath: formatPublicPath(env.ASSET_HOST, paths.output)
+}
 
 module.exports = {
   devServer,
   env,
   paths,
   loadersDir,
-  publicUrl,
-  publicPath
+  output,
+  formatPublicPath
 }
--- a/lib/install/config/webpack/configuration.js
+++ b/lib/install/config/webpack/configuration.js
@@ -33,6 +33,5 @@ module.exports = {
   env,
   paths,
   loadersDir,
-  output,
-  formatPublicPath
+  output
 }
--- a/lib/install/config/webpack/configuration.js
+++ b/lib/install/config/webpack/configuration.js
@@ -15,12 +15,12 @@ function removeOuterSlashes(string) {
 }
 
 function formatPublicPath(host = '', path = '') {
-  host = removeOuterSlashes(host)
-  path = removeOuterSlashes(path)
-  if (host && !/^http/i.test(host)) {
-    host = `//${host}`
+  let formattedHost = removeOuterSlashes(host)
+  if (formattedHost && !/^http/i.test(formattedHost)) {
+    formattedHost = `//${formattedHost}`
   }
-  return `${host}/${path}/`
+  const formattedPath = removeOuterSlashes(path)
+  return `${formattedHost}/${formattedPath}/`
 }
 
 const output = {
--- a/lib/install/config/webpack/development.server.js
+++ b/lib/install/config/webpack/development.server.js
@@ -2,17 +2,33 @@
 
 const { resolve } = require('path')
 const merge = require('webpack-merge')
+const ManifestPlugin = require('webpack-manifest-plugin')
 const devConfig = require('./development.js')
-const { devServer, publicPath, paths } = require('./configuration.js')
+const { devServer, paths, output, formatPublicPath } = require('./configuration.js')
+
+const { host, port } = devServer
+const contentBase = output.path
+const publicPath = formatPublicPath(`http://${host}:${port}`, paths.output)
+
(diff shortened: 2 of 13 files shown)


lib/install/config/webpack/configuration.js
```diff
@@ -10,17 +10,29 @@ const loadersDir = join(__dirname, 'loaders')
...
+  if (formattedHost && !/^http/i.test(formattedHost)) {
+    formattedHost = `//${formattedHost}`
+  }
+  const formattedPath = removeOuterSlashes(path)
+  return `${formattedHost}/${formattedPath}/`
+}
+
+const output = {
+  path: resolve('public', paths.output),
+  publicPath: formatPublicPath(env.ASSET_HOST, paths.output)
```
gauravtiwari: How about we set `ASSET_HOST` to dev server host in development when it's enabled? That way we don't need to replace manifest plugin.

```js
if (env.NODE_ENV === "development") {
  env.ASSET_HOST = `//${host}:${port}` // relative to origin
}
```

OR

```js
if (env.NODE_ENV === "development") {
  env.ASSET_HOST = `${devServer.https ? 'https': 'http'}://${devServer.host}:${devServer.port}`
}
```

We would be able to do this in development config without `if`, but guess we can't  since we have much shared logic here.

lib/install/config/webpack/development.server.js
```diff
@@ -1,18 +1,33 @@
...
 const devConfig = require('./development.js')
-const { devServer, publicPath, paths } = require('./configuration.js')
+const { devServer, paths, output, formatPublicPath } = require('./configuration.js')
+
+const { host, port } = devServer
+const contentBase = output.path
+const publicPath = formatPublicPath(`http://${host}:${port}`, paths.output)
+
+// Remove ManifestPlugin so we can replace it with a new one
+devConfig.plugins = devConfig.plugins.filter(plugin => plugin.constructor.name !== 'ManifestPlugin')
```
gauravtiwari: If we consider using `ASSET_HOST` env in development then I guess we don't need any of the changes here :)

lib/install/config/webpack/paths.yml
```diff
@@ -1,12 +1,12 @@
...
-  config: config/webpack
-  entry: packs
-  output: public
-  manifest: manifest.json
-  node_modules: node_modules
-  source: app/javascript
+default: &default            # ~ = Rails.root
+  source: app/javascript     # ~/:source
+  entry: packs               # ~/:source/:entry
+  output: packs              # ~/public/:output
```
gauravtiwari: Awesome 👍   been thinking to do this

lib/webpacker/configuration.rb
```diff
@@ -4,42 +4,54 @@
...
+    def config_path
+      Rails.root.join(fetch(:config))
+    end
+
+    def file_path(root: Rails.root)
+      root.join("config/webpack/paths.yml")
+    end
+
+    def default_file_path
+      file_path(root: Pathname.new(__dir__).join("../install"))
```
gauravtiwari: 👍

lib/install/config/webpack/configuration.js
gauravtiwari: How about we set ASSET_HOST to dev server host in development when it's enabled? That way we don't need to replace manifest plugin.

```js
if (env.NODE_ENV === "development") {
  env.ASSET_HOST = `http://${host}:${port}`
}
```

OR

```js
if (env.NODE_ENV === "development") {
  env.ASSET_HOST = `${devServer.https ? 'https': 'http'}://${devServer.host}:${devServer.port}`
}
```
We would be able to do this in development config without `if`, but guess we can't  since we have much of the shared logic here.

lib/install/config/webpack/development.server.js
gauravtiwari: All this will go away if we use `ASSET_HOST` in development too :)

lib/install/config/webpack/paths.yml
gauravtiwari: 👍  🍰

lib/webpacker/configuration.rb
```diff
@@ -4,42 +4,54 @@
...
+      paths.fetch(key, default_paths[key])
+    end
+
+    def paths
+      load if Webpacker.env.development?
+      raise Webpacker::FileLoader::FileLoaderError.new("Webpacker::Configuration.load must be called first") unless instance
+      instance.data
+    end
+
+    def default_paths
```
gauravtiwari: 🍰  🎉

lib/install/config/webpack/configuration.js
javan: I described the reason for *not* doing it this way in the PR:

> Fixes that `webpacker:compile` and `bin/webpack` would generate dev server URLs in development mode. This should only happen when using `bin/webpack-dev-server`.

That said, I cleaned things up in 2a94cdf1c93e0509af0cce971a788e5eea6ca8e3 by constructing and setting `ASSET_HOST` in `bin/webpack-dev-server`.

mltsy:
You guys are awesome! Thanks for working on this feature :clap:

PR #402, opened 2017-05-20
Fix install task message of webpacker.rake

hatappi:
I found mistake in `webpacker.rake`.
That's `angular2`.
We will call `Angular` instead of `angular2, angular4, angular...`.

It was announced [NG-BE 2016 Keynote by Igor Minar](https://youtu.be/aJIMoLgqU_o?t=1389).

This PR is correcting the messages of other tasks by referring [here](https://github.com/rails/webpacker/blob/74cde3c0f028516f11a793c17235ea71af21bca6/lib/tasks/installers.rake#L2-L5
).


gauravtiwari:
Thanks 👍

PR #417, opened 2017-05-24
Revert the extension path to not expand to sub-directories to stop OOM

robdimarco:
After upgrading to 2.0, I was consistently running out of RAM when trying to run webpack.

After some digging, I came across a comment on the webpack project that helped shed some light.

https://github.com/webpack/webpack/issues/1914#issuecomment-303837359

It seems that in PR 201 (https://github.com/rails/webpacker/pull/201), the `extensionGlob`
was changed from `*{${paths.extensions.join(',')}}*` to `**/*{${paths.extensions.join(',')}}*`

This change removes the `**/` in the extension and fixes my OOM issue.

I am not sure about the ramifications for allowing pack files in subdirectories
(the original intent of the PR that introduced the change). Maybe we need a comment or a
flag when generating the configuration that will set it one way or the other?

### DIFF
--- a/lib/install/config/webpack/shared.js
+++ b/lib/install/config/webpack/shared.js
@@ -11,7 +11,7 @@ const ManifestPlugin = require('webpack-manifest-plugin')
 const extname = require('path-complete-extname')
 const { env, settings, output, loadersDir } = require('./configuration.js')
 
-const extensionGlob = `**/*{${settings.extensions.join(',')}}*`
+const extensionGlob = `*{${settings.extensions.join(',')}}*`
 const entryPath = join(settings.source_path, settings.source_entry_path)
 const packPaths = sync(join(entryPath, extensionGlob))
 


gauravtiwari:
@robdimarco What files you have under `packs` or whatever folder you are using for entries/packs?

robdimarco:
I have 5 files directly in `app/javascript/packs` and 131 files all told if you include files in subdirectories. This includes Jest tests (in __tests__ subdirectories).

guilleiguaran:
@joerodrig this looks similar to the issue you found the last week.

guilleiguaran:
I guess the official reply for this is that this is totally expected since we expect to have in `app/javascripts/packs` only the 'packs' that are equivalent to webpack entry points, anything else should be in `app/javascripts` but outside of `packs`, that's just the convention that we have adopted for this.

robdimarco:
OK, that works for me. Thanks.

gauravtiwari:
Perhaps we make this more clear in README 👍

marz0:
@robdimarco I'm curious to know if you've run into any issues with this change? We're looking to make this change for our app as well and have noticed that it reduces our deployment time substantially, but would like to make sure it doesn't have any downstream effects

robdimarco:
@marz0 worked great for us. No side-effects.

PR #431, opened 2017-05-26
Use localhost instead of 0.0.0.0 for webpack-dev-server

### LINKED ISSUE #424: "Invalid Host Header" in JS console since upgrading to Webpacker 2.0
Since upgrading to webpacker 2.0, I'm getting a bunch of "invalid host header" errors in the JS console. Other sources online have said to pass `disableHostCheck true` when starting webpack-dev-server. I've modified my webpacker.yml to look like this:

```
development:
  <<: *default

  dev_server:
    host: 0.0.0.0
    port: 8080
    disableHostCheck: true
    https: false
```

And then development.js, I've modified the devServer object to include this line: 
`disableHostCheck: settings.dev_server.disableHostCheck,`

Is that appropriate? Is anyone else getting these errors?

catskull:
Resolves #424

### DIFF
--- a/lib/install/config/webpacker.yml
+++ b/lib/install/config/webpacker.yml
@@ -25,7 +25,7 @@ development:
   <<: *default
 
   dev_server:
-    host: 0.0.0.0
+    host: localhost
     port: 8080
     https: false
 


renchap:
A bit of explanations here:
`0.0.0.0` is a special IP address. It should only be used in an HTTP server setting: listening on `0.0.0.0` means listening on all IP addresses available on the system: your LAN IP address, localhost (127.0.0.1), …

It should never be used to connect to an HTTP server. You need to use a proper IP address, like localhost/127.0.0.1, your server's public IP, …

From what I understand, `dev_server.host` is used to configure both the IP `webpack-dev-server` listens to (so `0.0.0.0 is correct here`) and the host used to create assets URLs (and `0.0.0.0` must not be used here).

This setting should be split in two, maybe `host` and `listen_address`.

For local development, we can have:

```ruby
host = 'localhost'
listen_address = 'localhost' # or 0.0.0.0
```

For remote development (for example Cloud9, the use case which prompted this change), we can use:

```ruby
host = '<Cloud9 environment public IP>' # The one you use in your browser to
                                        # connect to your Rails server
listen_address = '0.0.0.0' # Webpack needs to listen on all addresses
```

gauravtiwari:
Superseded by #482

PR #434, opened 2017-05-27
Fix publicPath in file-loader config

renchap:
70eace2b8b7035954fb921c7066a24d75e775a86 changed the export to `output.publicPath`, but this has not been updated.


gauravtiwari:
Thanks @renchap 🍰

PR #439, opened 2017-05-27
Do not write files in test env

### LINKED ISSUE #435: Running unit tests with Karma - Error: EACCES: permission denied, mkdir '/_karma_webpack_'
Posting here to see if others are having this issue and if it's one with the configuration Webpacker creates.

I had this working with many previous versions of the wonderful JavaScript toolchain, but now with the latest of everything, I get

```
Error: EACCES: permission denied, mkdir '/_karma_webpack_'
```

The source for karma-webpack seems to [be totally wrong](https://github.com/webpack-contrib/karma-webpack/blob/master/src/karma-webpack.js#L50-L51), but it's been like that for 10+ months, so I think there's a subtly I'm not aware of.

I'm using the latest Webpacker-generated configuration files, with *zero changes*.

My devDependencies are:

```json
  "devDependencies": {
    "jasmine": "^2.6.0",
    "jasmine-core": "^2.6.2",
    "karma": "^1.7.0",
    "karma-jasmine": "^1.1.0",
    "karma-phantomjs-launcher": "^1.0.4",
    "karma-webpack": "^2.0.3",
    "testdouble": "^2.1.2"
  }
```

My `karma.conf.js` looks like:

```javascript
module.exports = function(config) {
  config.set({
    frameworks: ['jasmine'],
    files: [
      '**/*.spec.js'
    ],
    preprocessors: {
      '**/*.spec.js': [ 'webpack' ]
    },
    webpack: require('../../config/webpack/test.js'),
    browsers: ['PhantomJS'] })
}
```

If I `sudo mkdir /_karma_webpack_ && sudo chmod 777 /_karma_webpack` everything works.  This is not a great solution, but i have no idea how to change it or why it's now broken.

Anyone have any ideas?

davetron5000:
# Problem

When running unit tests with Karma, karma forcibly and inflexibly overrides the `output.path` of Webpack to be `/_karma_webpack_`, which in almost every system is not writable.  Since Karma has been like this for almost a year, it seems that it's assuming no files are getting written to the filesystem during a test run.  While odd, this appears to be the case and isn't causing anyone else problems per se.

For Rails devs, this means running Karma-based unit tests requires a writable file in the root directory of their machine, and any CI environment, which is not ideal.

# Solution

Configure the `ManifestPlugin` to not write files in the "test" environment.

## Notes

I'm not sure if this is the right way to solve this problem.  Since it's being done to afford Karma-based unit testing, this might not be the right thing for Webpacker.  But, it would be nice if this behavior could be controlled in some other way.  It seems that the overall theory of Webpacker is that the developer should not be messing with `config/webpack/{development,shared,production,test}.js` and instead be changing `config/webpacker.yml`  so I'm open to any ideas on how to make this configurable in there if that's the right thing.

Fixes #435 (also more details)

### DIFF
--- a/lib/install/config/webpack/shared.js
+++ b/lib/install/config/webpack/shared.js
@@ -40,7 +40,7 @@ module.exports = {
     new ExtractTextPlugin(env.NODE_ENV === 'production' ? '[name]-[hash].css' : '[name].css'),
     new ManifestPlugin({
       publicPath: output.publicPath,
-      writeToFileEmit: true
+      writeToFileEmit: env.NODE_ENV !== 'test'
     })
   ],
 


gauravtiwari:
@davetron5000 Hey, thanks for this but I think this will cause problems for other testing frameworks that rely on the same feature. I suspect that capybara tests would fail because that would require files to be there.

gauravtiwari:
You might wanna override the plugin in `test.js` to not write files in test environment.

davetron5000:
@gauravtiwari yup understood.  Can you point me to how that might work?  It's not clear to me how to change just the config of that plugin without having to replicate the entire plugin list in `test.js`

gauravtiwari:
@davetron5000 For now, I think simplest you could do is this - 

```js
const merge = require('webpack-merge')
const sharedConfig = require('./shared.js')

sharedConfig.plugins = sharedConfig.plugins.filter(plugin => plugin.constructor.name !== 'ManifestPlugin')
module.exports = merge(sharedConfig, {})

```

gauravtiwari:
Btw the `writeFile` option only works with dev server so if you are using normal webpack binary then it will output manifest anyway regardless of what you have set.

davetron5000:
I was not using the dev server in any part of the testing toolchain.  I was using karma-webpack, which runs webpack's compiler internally.  Without the change posted, karma-webpack instructions the ManifestPlugin to write files to `/` (for reasons I don't know o_O)

gauravtiwari:
@davetron5000 Lets explore this in #435

PR #449, opened 2017-05-31
Remove duplicated resolveLoader config declaration

frozenfung:
resolveLoader default setting is like below:

```
{
    modules: ["node_modules"],
    extensions: [".js", ".json"],
    mainFields: ["loader", "main"]
}
```

Is there any reason we duplicately declare it in `shared.js` ?

reference:
https://webpack.js.org/configuration/resolve/#resolveloader

### DIFF
--- a/lib/install/config/webpack/shared.js
+++ b/lib/install/config/webpack/shared.js
@@ -50,9 +50,5 @@ module.exports = {
       resolve(settings.source_path),
       'node_modules'
     ]
-  },
-
-  resolveLoader: {
-    modules: ['node_modules']
   }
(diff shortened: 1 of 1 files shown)


gauravtiwari:
@frozenfung Both are different in the sense that resolve - works for app modules whereas resolveLoader is for internal webpack related loaders and plugins. 

> This set of options is identical to the resolve property set above, but is used only to resolve webpack's loader packages. Default:

frozenfung:
@gauravtiwari You are totally right. I need a cup of covfefe to wake me up.

Issue #379, opened 2017-05-13
webpacker is installed... ./bin/webpack: not found


christophermlne:
I'm not sure how to interpret this error that I get when trying to precompile assets on the server.
1) webpacker is installed
2) but webpack is not found
3) A JSON text must at least contain two octets! (thanks but which JSON file?)

```
app@vagrant:/var/www/app/application$ bundle exec rake assets:precompile
Webpacker is installed 🎉 🍰
Using /var/www/app/application/config/webpack/paths.yml file for setting up webpack paths
Compiling webpacker assets 🎉
sh: 1: ./bin/webpack: not found
rake aborted!
JSON::ParserError: A JSON text must at least contain two octets!
/home/app/.rbenv/versions/2.3.1/bin/bundle:23:in `load'
/home/app/.rbenv/versions/2.3.1/bin/bundle:23:in `<main>'
Tasks: TOP => webpacker:compile
(See full trace by running task with --trace)
```

btc:
Perhaps you need `yarn global add webpack`

christophermlne:
@btc i just tried your suggestion but unfortunately it didn't work. i get the same error.

this line https://github.com/rails/webpacker/blob/master/lib/tasks/webpacker/compile.rake#L11
is allowed to execute after `webpacker:verify_install` succeeds. i'm wondering if `webpacker:verify_install` ought to have failed here, since a working webpack installation doesn't seem to be present? (even though i can run 'webpack' in the terminal and it works)

ytbryan:
Are you using rails 5? I think it's because you are lacking the binstub `webpack` in your bin folder.

christophermlne:
@ytbryan you are right. i have a bin folder on my local machine but not on the remote server.

my webpacker version is 1.2

christophermlne:
i ran `bundle exec bin/rails app:update:bin` and so now i have a bin folder.

i still get the same error though.

christophermlne:
the issue was caused by the bin folder not being present, and updating the .gitignore file and redeploying fixed it.

raldred:
What the solution for getting the webpack binstubs? in a rails 5.0.2 app? same as OP can't upgrade to 5.1 yet. 
no ./bin in gitignore 
`bundle exec bin/rails app:update:bin` runs no ./bin/webpack
`rails webpacker:install:react` errors `Webpack binstubs not found.`

ytbryan:
[...]

PR #450, opened 2017-05-31
Have you run `rails webpacker:install`?

ytbryan:
Do you think we need such a simple reminder in the error message? 

To prevent this https://github.com/rails/webpacker/issues/379#issuecomment-305165973


lib/tasks/webpacker/check_webpack_binstubs.rake
```diff
@@ -3,6 +3,7 @@ namespace :webpacker do
   task :check_webpack_binstubs do
     unless File.exist?("bin/webpack") && File.exist?("bin/webpack-dev-server")
       puts "Webpack binstubs not found.\n"\
+           "Have you run `rails webpacker:install`?\n"\
```
gauravtiwari: We don't need those back-ticks? 😄

ytbryan: 😅 removed!

gauravtiwari:
Thanks 👍

PR #453, opened 2017-06-01
Updated README.md to add related projects

justin808:
This is similar to how https://github.com/reactjs/react-rails lists related projects. I believe that it's useful to show the community alternatives.

Why did [ShakaCode](http://www.shakacode.com) fork [rails/webpacker](https://github.com/rails/webpacker)?

3 reasons:

1. React on Rails needed only the helpers in Webpacker to obtain the correct file path or relative URL to Webpack created assets, given different Rails environments, fingerprinting of assets, hot-reloading, etc.
2. We preferred a simpler configuration to get the core functionality needed. You configure just one thing: The directory within `/public` where Webpack will create the manifest and output file. Then you configure your Webpack config to generate a simple manifest that maps the base output names to the possibly fingerprinted versions. Note, unlike Webpacker, Webpacker Lite wants your manifest to **NOT** contain any host information.
3. We needed the ability to quickly make changes needed by [react_on_rails](https://github.com/shakacode/react_on_rails)

For more details on how this project differs from Webpacker and why we forked, please see [Webpacker Lite: Why Fork Webpacker?](https://blog.shakacode.com/webpacker-lite-why-fork-webpacker-f0a7707fac92).

The [Webpacker Lite README](https://github.com/shakacode/webpacker_lite) has more details on some subtle differences. For example, Webpacker puts detailed server path information in the manifest.json. Webpacker Lite wants just minimal information in the manifest.json. There's other little differences like that. I wanted Webpacker Lite to make the minimal requirements for the hand-coded Webpack config. Since Webpacker handles the config, it can put logic in either on the Rails or Webpack side.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1113,6 +1113,10 @@ completed the compilation successfully before loading a view.
 - HMR - [#188](https://github.com/rails/webpacker/issues/188)
 - Support rails engines - [#348](https://github.com/rails/webpacker/issues/348)
 
-
 ## License
 Webpacker is released under the [MIT License](https://opensource.org/licenses/MIT).
+
+## Related Projects
+- [react\_on\_rails Gem](https://github.com/shakacode/react_on_rails): Integration of Rails with React, Redux, React-Router, including Server Rendering utilizing [Webpacker Lite](https://github.com/shakacode/webpacker_lite).
+- [webpacker\_lite Gem](https://github.com/shakacode/webpacker_lite): A slimmed down version of Webpacker that provides only the helpers with customizations for React on Rails. See article [Webpacker Lite: Why Fork Webpacker?](https://blog.shakacode.com/webpacker-lite-why-fork-webpacker-f0a7707fac92).
+- [react-rails Gem](https://github.com/reactjs/react-rails): Integration of Rails with React utilizing Webpacker.


kaspth:
Thanks, but I interpret https://github.com/rails/webpacker/pull/339#issuecomment-299710867 as meaning this has already been turned down once 😊

justin808:
@dhh Any chance on reconsidering a section to indicate related projects?

justin808:
I had a good talk with @gauravtiwari! We're working together on a few ideas, so let's leave this one closed for now. Potentially something good is cooking!

gauravtiwari:
Just to add some context here - @justin808 and I discussed this PR and why alternatives aren't related. Obviously, as an open-source community we can help each other out where possible, but alternatives listed aren't really alternatives - which I think is very clear.

We also discussed a great deal about `webpacker_lite` since it only works with `react_on_rails` and hence can't be called `webpacker_lite` - very confusing to end users. We discussed to either change `webpacker_lite` to more `react_on_rails` specific name, so it's more clear that this package is a companion for `react_on_rails` *OR* merge it into `react_on_rails` - much easier to maintain.

@justin808 liked the idea of changing `webpacker_lite` to `webpacker_helpers` in the light that perhaps `webpacker` too can depend on `webpacker_helpers` going forward. However later I realized that `webpacker` as a Gem is tiny in itself so this isn't a viable option. This will make maintenance hard and turn off contributors.

Therefore I would suggest to either:

1. Merge webpacker_lite into react_on_rails
2. Consider using `webpacker` instead without webpack config and provide the one ships with `react_on_rails`. We can iron out any missing pieces like HMR etc. through PR's

If react_on_rails chooses 2nd option, then we can link it in README under react section  👍

justin808:
Thanks, @gauravtiwari! 

I think option 2 is great.

I will open an issue ticket to get confirmation that we agree on any changes before I do the work of a PR.

Issue #428, opened 2017-05-26
Rake verify+check tasks - output to stdout / stderr ? 


asaletnik:
I was recently trying to deploy my Rails 4.2.8 app to OpsWorks (using Chef).
It was failing silently without giving me any clue in the log, ending with:

> yarn install v0.24.5
> [1/4] Resolving packages...
> [2/4] Fetching packages...
> info "fsevents@1.1.1" is an optional dependency and failed compatibility check. Excluding it from installation.
> [3/4] Linking dependencies...
> [4/4] Building fresh packages...
> Done in 24.75s.
> STDERR:
> warning No license field
> warning fsevents@1.1.1: The platform "linux" is incompatible with this module.

The last lines of stdout were coming from `webpacker:yarn_install` which runs in Rails 4.2, then it should follow to `webpacker:compile` first verifying the install, but there was nothing more on that in the log file.

I have run a `assets:precompile --trace`, and noticed that last running task is `webpacker:check_webpack_binstubs` - which was failing because I had the `bin/` directory in a `.gitignore` file (my bad :disappointed:).

Now to the point - wouldn't it be better to pass the errors to `stderr` or at least `stdout` instead of just using `puts`?

gauravtiwari:
@asaletnik Good catch. I think we should update these statements to use `stdout` or `stderr` instead of `puts`. If you have some time please feel free to make a PR or I will make one at some point.

gauravtiwari:
Big thanks to @ytbryan. This is fixed in #455

PR #455, opened 2017-06-01
Use stderr or stdout instead of puts 

ytbryan:
As mentioned here https://github.com/rails/webpacker/issues/428

Do I also need to change those "puts" in `lib/install/*.rb`?


gauravtiwari:
Good stuff 👍  No, I think this should be fine since install is only run on local system.

PR #456, opened 2017-06-01
Added typed asset_path_pack versions for consistency

AlexMouton:
...with the link tag brethren.
Separated pack asset name from link helpers.


gauravtiwari:
@AlexMouton What's your use case here?

PR #458, opened 2017-06-01
Ignore packs generated for tests in git

michel-zimmer:
Should we also add a warning to [[2.0] - 2017-05-24 / Breaking Change](https://github.com/rails/webpacker/blob/master/CHANGELOG.md#breaking-change)?
And maybe later specify up to which version this applies when this fix lands in a gem release?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,6 +74,12 @@
   rm config/webpack/development.server.js
   ```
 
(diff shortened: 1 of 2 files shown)


gauravtiwari:
@michel-zimmer Yepp, adding it to changelog sounds great for now. If you could do that under breaking change, I will merge 👍

PR #473, opened 2017-06-05
`Error: listen EADDRINUSE 0.0.0.0:8080` in readme

ytbryan:
Add `Error: listen EADDRINUSE 0.0.0.0:8080` to readme and reformat the troubleshooting section.

Currently, troubleshooting is at the bottom of the readme. I brought it up to benefit the newcomers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,6 @@ in which case you may not even need the asset pipeline. This is mostly relevant 
 - [Prerequisites](#prerequisites)
 - [Features](#features)
 - [Installation](#installation)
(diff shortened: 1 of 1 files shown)


gauravtiwari:
@ytbryan Hey, not sure how much this would help since people are only going to refer to this when they have an issue. Therefore, I think it would makes sense to keep it at the bottom - we have TOC anyway for easy navigation. 

If you please move it down where it was we can merge it 👍  thank you :)

ytbryan:
@gauravtiwari okay. done.

Issue #476, opened 2017-06-06
Compiling twice for system test when there is no cache


S-YOU:
I believe this is not the issue with webpacker but please help me figure how to debug this issue.

```
% rake test:system
/usr/local/lib/ruby/gems/2.4.0/gems/webpacker-2.0/lib/tasks/installers.rake:1: warning: already initialized constant INSTALLERS
/usr/local/lib/ruby/gems/2.4.0/gems/webpacker-2.0/lib/tasks/installers.rake:1: warning: previous definition of INSTALLERS was here
/usr/local/lib/ruby/gems/2.4.0/gems/webpacker-2.0/lib/tasks/webpacker/install.rake:1: warning: already initialized constant WEBPACKER_APP_TEMPLATE_PATH
/usr/local/lib/ruby/gems/2.4.0/gems/webpacker-2.0/lib/tasks/webpacker/install.rake:1: warning: previous definition of WEBPACKER_APP_TEMPLATE_PATH was here
Webpacker is installed � �
Using ..../config/webpacker.yml file for setting up webpack paths
Webpacker is installed � �
Using ..../config/webpacker.yml file for setting up webpack paths
[Webpacker] Compiling assets �
[Webpacker] Compiling assets �
...
```
Webpacker 2.0, rails 5.1.1, ruby 2.4.1, osx

I actually asked on stackoverflow, but no one answers. (https://stackoverflow.com/q/44351894)

ytbryan:
This might not work, but have you tried to restart your mac? And try it again?

S-YOU:
It is somehow unfortunately reproducible for me on el capitan, and sierra. both are upgrading from webpacker 1.x to webpacker 2.0 with manual install, webpacker:install and webpacker:install:vue.

RAILS_ENV=test rake webpacker:clobber
RAILS_ENV=test rake test:system

ytbryan:
I can't reproduce it. I was using ruby 2.4.0p0, sierra 10.12.5, webpacker 2.0, rails 5.1.1.

S-YOU:
@ytbryan, I have uploaded to a repo, pretty much clean install.

https://github.com/S-YOU/webpacker-test

```
% RAILS_ENV=test rake webpacker:clobber

Webpacker is installed 🎉 🍰
Using /somewhere/webpacker-test/config/webpacker.yml file for setting up webpack paths
Removed webpack output path directory /somewhere/webpacker-test/public/packs-test

/somewhere/webpacker-test%
[...]

PR #477, opened 2017-06-06
Rakefile is already loaded

S-YOU:
I am getting duplicated task running with system tests because of this.
No idea when you need to manually load project's Rakefile inside a rake task.

Fixes https://github.com/rails/webpacker/issues/476 for me.

### DIFF
--- a/lib/webpacker/compiler.rb
+++ b/lib/webpacker/compiler.rb
@@ -14,7 +14,6 @@ module Webpacker::Compiler
     end
 
     def load_rake_task(name)
-      @load_rakefile ||= Rake.load_rakefile(Rails.root.join("Rakefile"))
       Rake::Task[name]
     end
 end


javan:
Thanks, @S-YOU. I was able to reproduce the issue using your sample app in #476. I think we still want to load the Rakefile in cases where it may not be loaded already. Here's a change that should handle both scenarios:
```diff
--- a/lib/webpacker/compiler.rb
+++ b/lib/webpacker/compiler.rb
@@ -14,7 +14,11 @@ module Webpacker::Compiler
     end
 
     def load_rake_task(name)
-      @load_rakefile ||= Rake.load_rakefile(Rails.root.join("Rakefile"))
+      load_rakefile unless Rake::Task.task_defined?(name)
       Rake::Task[name]
     end
+
+    def load_rakefile
+      @load_rakefile ||= Rake.load_rakefile(Rails.root.join("Rakefile"))
+    end
```
Mind giving that a try?

S-YOU:
@java, yeah, thats work. will be send a PR or I update this?

S-YOU:
Its a little late here, I am going to sleep.
So may be you push your patch and close this or add commits here.
As long as it is fixed, I am happy :)

PR #480, opened 2017-06-07
Change vue example in README

yeonhoyoon:
Change example for passing props into view using Vue.js to reflect changes made in #304.


gauravtiwari:
Thanks 🍰

Issue #420, opened 2017-05-25
Issue with Elm compiler


ghivert:
Hi !

I'm trying to deploy a Rails server to Heroku with Elm on frontend, but I'm encountering a bug. It seems that the Elm compiler is not found…

Here's the full log:

```
remote:        [Webpacker] Compilation Failed
remote:        yarn run v0.24.5
remote:        $ "/tmp/build_3da1d65b78c4b248a9c140d2599e5d28/node_modules/.bin/webpack" --config /tmp/build_3da1d65b78c4b248a9c140d2599e5d28/config/webpack/production.js
remote:        Hash: 789c03420c8028135ff2
remote:        Time: 1610ms
remote:        [0] ./app/frontend/javascript/image_uploader.js 452 bytes {0} [built]
remote:        [1] ./app/frontend/javascript/markdownify_elm.js 332 bytes {0} [built]
remote:        [2] ./app/frontend/elm/Event.elm 1.11 kB {0} [built] [failed] [1 error]
remote:        [3] ./app/frontend/elm/Footer.elm 1.11 kB {1} [built] [failed] [1 error]
remote:        [4] ./app/frontend/packs/event_elm.js 567 bytes {0} [built]
remote:        [5] ./app/frontend/packs/footer_elm.js 418 bytes {1} [built]
remote:        ERROR in ./app/frontend/elm/Event.elm
remote:        Module build failed: Error: Compiler process exited with error Compilation failed
remote:        events.js:163
remote:        throw er; // Unhandled 'error' event
remote:        ^
remote:        Error: spawn /tmp/build_3da1d65b78c4b248a9c140d2599e5d28/node_modules/elm/Elm-Platform/0.18.0/.cabal-sandbox/bin/elm-make ENOENT
remote:        at exports._errnoException (util.js:1050:11)
remote:        at Process.ChildProcess._handle.onexit (internal/child_process.js:193:32)
remote:        at onErrorNT (internal/child_process.js:367:16)
remote:        at _combinedTickCallback (internal/process/next_tick.js:80:11)
remote:        at process._tickCallback (internal/process/next_tick.js:104:9)
remote:        at Module.runMain (module.js:607:11)
remote:        at run (bootstrap_node.js:427:7)
remote:        at startup (bootstrap_node.js:151:9)
remote:        at bootstrap_node.js:542:3
[...]

PR #486, opened 2017-06-09
Adds postinstall rebuild Elm step to troubleshooting readme

abrahamsangha:
Addresses https://github.com/rails/webpacker/issues/420


gauravtiwari:
Thank you 🍰

Issue #489, opened 2017-06-09
Importing styles from node_modules is showing no file or directory error


johnisaac:
Hi, 

   I tried to follow the [instructions](https://github.com/rails/webpacker#from-node-modules-folder) from the wiki to import styles from node_modules. But I keep getting the following error message.

> Module build failed: Error: ENOENT: no such file or directory, open '/Users/johnisaac/code/webpacker_sample/app/javascript/styles/~boostrap/dist/bootstrap.css'

Above line is in a _hello_react.sass_ file with the following content
`@import '~boostrap/dist/bootstrap.css'`

_hello_react.sass_ file is imported in another hello_react.js file like
`import '../styles/hello_react'`

Any help is appreciated.

gauravtiwari:
@johnisaac Please don't use extension it's already added in the extensions list. 

`@import '~boostrap/dist/bootstrap`

johnisaac:
@gauravtiwari When I tried without the extension, it is showing a similar error

> File to import not found or unreadable: ~boostrap/dist/bootstrap.

gauravtiwari:
There is a typo - `~bootstrap/dist/bootstrap`

johnisaac:
I did not include `.` at the end in the source code but it is showing up in the log.

<img width="478" alt="screen shot 2017-06-09 at 7 49 07 am" src="https://user-images.githubusercontent.com/284118/26980864-3b4a2da2-4ce8-11e7-84a7-611c6efbd8d1.png">

<img width="547" alt="screen shot 2017-06-09 at 7 50 37 am" src="https://user-images.githubusercontent.com/284118/26980917-63f77714-4ce8-11e7-8844-f2abb1558cfb.png">

johnisaac:
Ah, I see you what you mean. Thanks!

PR #494, opened 2017-06-12
Fix README typo: s/boostrap/bootstrap/

freeatnet:


freeatnet:
Additionally, the only way this code worked for me (on Rails 5.1.1, default `config/webpacker.yml`) was to have `@import '~bootstrap/dist/css/bootstrap'` rather than `@import '~bootstrap/dist/bootstrap.css'`. I know this issue has been mentioned [before](https://github.com/rails/webpacker/issues/489#issuecomment-307408324). Please let me know if this should be fixed (in this or another PR).

gauravtiwari:
@freeatnet Yes please, if you could remove the extensions that would great - since we already have them in yml config file 👍

kevgathuku:
Hey @freeatnet the path to the bootstrap CSS is also wrong. I had made a separate PR for that but since they're dealing with the same issue, we might as well fix it here.
It should be `bootstrap/dist/css/bootstrap.css` instead of `boostrap/dist/bootstrap.css`

gauravtiwari:
Fixed in #514

PR #500, opened 2017-06-13
allow vue templates to be precompiled by erb (with `.vue.erb` extension)

doits:
Or is there any reason why vue templates should not be precompiled by erb?

More specifically this PR does not change that they are precompiled (every file ending in `.erb` is compiled afaik), but now they can be imported with vue-loader.


gauravtiwari:
@doits Since this isn't very common use case, it would best to add this to README. Can you please make a PR with README change?

PR #501, opened 2017-06-14
Remove Gemfile.lock

rossta:
Following typical Ruby gem convention of omitting and gitignoring the `Gemfile.lock` file.


rafaelfranca:
Thank you for the PR but this is not the Rails convention. See https://github.com/rails/rails/pull/18992

rossta:
Thanks for the link, I wasn't aware.

PR #502, opened 2017-06-14
Respect RAILS_ENV and NODE_ENV in webpack-dev-server binstub

rossta:
The following changes substitute hard-coded `'development`' usages with respective `RAILS_ENV` and `NODE_ENV` values. This would be one step in paving the way for running a separate `webpack-dev-server` in test env as opposed to lazy-compiling assets to `packs-test`.


gauravtiwari:
Thanks @rossta 👍

PR #503, opened 2017-06-14
Adds compile option to configuration

rossta:
This change introduces a new configuration option for enabling lazy compilation via `config/webpacker.yml`. 

Along with #502, this would allow for running a webpack-dev-server process in the test environment instead of enforcing compilation to the `public` directory. The fallback for a missing `compile` key keeps the default lazy-compilation behavior intact.

The `compile` key name borrows from the `assets.compile` option for the asset pipeline.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,20 @@
-## [2.0] - 2017-05-24
+## Unreleased
 
+### Fixed
+- Update `webpack-dev-server.tt` to respect RAILS_ENV and NODE_ENV values [#502](https://github.com/rails/webpacker/issues/502)
+
+### Breaking changes
+- Add `compile_missing_packs` option to `config/webpacker.yml` for configuring lazy compilation of packs when not found [#503](https://github.com/rails/webpacker/pull/503). To enable expected behavior for defaults and production, update `config/webpacker.yml`:
+
+  ```yaml
+    default: &default
+      compile_missing_packs: true
+
+    production:
+      compile_missing_packs: false
+  ```
+
+## [2.0] - 2017-05-24
 
 ### Fixed
 - Update `.babelrc` to fix compilation issues - [#306](https://github.com/rails/webpacker/issues/306)
--- a/lib/install/config/webpacker.yml
+++ b/lib/install/config/webpacker.yml
@@ -4,7 +4,7 @@ default: &default
   source_path: app/javascript
   source_entry_path: packs
   public_output_path: packs
-  compile_missing_packs: false
+  compile_missing_packs: true
 
   extensions:
     - .coffee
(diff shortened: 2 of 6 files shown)


javan:
+1 to making this configurable

> this would allow for running a webpack-dev-server process in the test environment instead of enforcing compilation to the public directory

Wouldn't running `webpack-dev-server` prevent lazy compiling on its own since it updates manifest.json? `compile_and_find!(name)` is only called when there's no entry in the manifest.

lib/install/config/webpacker.yml
```diff
@@ -4,6 +4,7 @@ default: &default
   source_path: app/javascript
   source_entry_path: packs
   public_output_path: packs
+  compile: false
```
javan: WDYT about flipping this around? Make it `true` by default and `false` in production.

lib/webpacker/configuration.rb
```diff
@@ -36,6 +36,10 @@ def source
       fetch(:source_path)
     end
 
+    def compile?
+      fetch(:compile).nil? ? Webpacker.env.test? : fetch(:compile)
```
javan: I'd drop the `Webpacker.env.test?` default condition and just use the configuration option.

lib/install/config/webpacker.yml
javan: Since compilation only happens when packs are missing (vs. stale), the name `compile` doesn't feel quite right. Perhaps something more descriptive like `compile_missing_packs`?

rossta:
@javan Thanks for the feedback.

> Wouldn't running webpack-dev-server prevent lazy compiling on its own since it updates manifest.json? 

You're right—it would've been more accurate for me to say the "lookup and compile if not found" behavior is only possible in `Webpacker.env.test` prior to these changes.

lib/install/config/webpacker.yml
rossta: 👍 Flipping the values and making the name more descriptive both sound like good changes.

lib/webpacker/configuration.rb
rossta: One concern is for folks upgrading for whom the `compile_missing_packs` entry would be missing... in this case, they would have to add in the values to get the expected behavior back in `test`. Should we add a warning or upgrade note in README?

gauravtiwari: @rossta Yes please 👍  add it to Chagelog and highlight if it's a breaking change

CHANGELOG.md
```diff
@@ -1,5 +1,20 @@
-## [2.0] - 2017-05-24
+## Unreleased
+
+### Fixed
+- Update `webpack-dev-server.tt` to respect RAILS_ENV and NODE_ENV values [#502](https://github.com/rails/webpacker/issues/502)
```
javan: Is this entry related to a change in this PR?

rossta: Loosely. It was a necessary change along with this PR for the ability to run the webpack-dev-server in test. Would you prefer a separate PR for this entry?

javan:
Thank you!

PR #528, opened 2017-06-22
Add a missing \ to angular.js

ytbryan:
Is this necessary? Seen this in other .js files


gauravtiwari:
Yepp, for escaping forward slashes 👍

Issue #542, opened 2017-06-28
`No PostCSS Config found` in production


x-yuri:
It makes sense to specify path to `.postcssrc.yml` explicitly. It's [looked for][1] in [parent directories][2] [relative][3] to [asset][4] being [built][5]. And if your asset is in `node_modules` and your `node_modules` is symlinked between releases (which is a reasonable thing to do on production), it fails with:

```js
       /home/u1/app/shared/~/css-loader!/home/u1/app/shared/~/postcss-loader!/home/u1/app/shared/~/sass-loader/lib/loader.js!/home/u1/app/shared/~/eonasdan-bootstrap-datetimepicker/build/css/bootstrap-datetimepicker.css
       Module build failed: Error: No PostCSS Config found in: /home/u1/app/shared/node_modules/eonasdan-bootstrap-datetimepicker/build/css
           at Error (native)
           at /home/u1/app/shared/node_modules/postcss-load-config/index.js:51:26
        @ /home/u1/app/shared/~/eonasdan-bootstrap-datetimepicker/build/css/bootstrap-da
tetimepicker.css 4:14-154
        @ ./app/webpack/application/posts.js
        @ ./app/webpack/packs/application.js
```

And the way to fix it is:

```js
    use: ['css-loader', {
      loader: 'postcss-loader',
      options: {
        config: '.postcssrc.yml',
      },
    }, 'sass-loader'],
```

in `config/webpack/loaders/sass.js`.

[1]: https://github.com/postcss/postcss-loader/blob/v2.0.6/lib/index.js#L68
[2]: https://github.com/postcss/postcss-loader/blob/v2.0.6/lib/index.js#L89
[3]: https://github.com/michael-ciniawsky/postcss-load-config/blob/v1.2.0/index.js#L49
[4]: https://github.com/michael-ciniawsky/postcss-load-config/blob/v1.2.0/index.js#L9
[5]: https://github.com/davidtheclark/cosmiconfig

gauravtiwari:
@x-yuri Does this work? 

```js
const ExtractTextPlugin = require('extract-text-webpack-plugin')
const path = require('path')

const { env } = require('../configuration.js')
const postcssConfig = path.resolve(process.cwd(), '.postcssrc.yml')

module.exports = {
  test: /\.(scss|sass|css)$/i,
  use: ExtractTextPlugin.extract({
    fallback: 'style-loader',
    use: [
[...]

PR #548, opened 2017-07-04
set explicit path on postcssConfig file

DavidNV:
Based on issue #542  

Without the explicit path the application.css wasn't able to be created when the sass loader started. 

This update provided by @gauravtiwari will fix the problem.


gauravtiwari:
Thanks @DavidNV 👍  🍰

PR #549, opened 2017-07-05
require => require_relative

muryoimpl:
Rails merged [pull/29638](https://github.com/rails/rails/pull/29638). I think that webpacker is a member of rails gems, so let webpacker follow the way?


rafaelfranca:
For small gems, there is no need to change. Thank you for the PR

PR #583, opened 2017-07-22
Throw an exception if NODE_ENV is not set when running webpack

soundasleep:
On a brand new install, if NODE_ENV is not already set, then 
running `bin/webpack` will fail with an obscure error message:

```
  path: resolve('public', settings.public_output_path),
                                  ^

TypeError: Cannot read property 'public_output_path' of undefined
    at Object.<anonymous> (.../config/webpack/configuration.js:26:35)
```

Since someone will never be able to run webpack without NODE_ENV set,
it makes sense to give the developer a message to help them debug.

### DIFF
--- a/lib/install/config/webpack/configuration.js
+++ b/lib/install/config/webpack/configuration.js
@@ -22,6 +22,10 @@ function formatPublicPath(host = '', path = '') {
   return `${formattedHost}/${formattedPath}/`
 }
 
+if (typeof env.NODE_ENV == 'undefined') {
+  throw "Cannot start webpack: ENV[NODE_ENV] not set." 
+}
+
 const output = {
   path: resolve('public', settings.public_output_path),
   publicPath: formatPublicPath(env.ASSET_HOST, settings.public_output_path)


gauravtiwari:
@soundasleep Thanks for this but we already set this inside binstub - https://github.com/rails/webpacker/blob/master/lib/install/bin/webpack.tt#L10

soundasleep:
I realised that later... I was having issues with yarn not playing nicely on Windows (it was trying to execute using the wrong path separators), so was having to run webpack manually, which caused this. Happy to close :)

gauravtiwari:
@soundasleep Could you please help me debug that? It seems that many people are having same problem on windows but I don't have access to a windows machine.

soundasleep:
Absolutely! I'd be happy to help, what would you like me to do? Here's some of my history:

You can't run something with a forward slash in Windows:

```
C:\path>bin/webpack
'bin' is not recognized as an internal or external command,
operable program or batch file.
```

So, I used backslash. But because it isn't a .bat or .cmd file, Windows doesn't read the first line of the script as the interpreter to use:

```
C:\path>bin\webpack
'bin\webpack' is not recognized as an internal or external command,
operable program or batch file.
```

So I then started passing along the script to Ruby. But then the path separator thing started occurring:

```
C:\path>ruby bin\webpack
yarn run v0.27.5
warning package.json: No license field
$ "C:\path\node_modules\.bin\webpack" "--config" "C:/path/config/webpack/development.js"
The filename, directory name, or volume label syntax is incorrect.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

What I've done, in the meantime as a workaround, is the following in my `bin/webpack`:

```ruby
# ... etc

newenv  = { "NODE_PATH" => NODE_MODULES_PATH.shellescape }
if Gem.win_platform?
  cmdline = ["#{NODE_MODULES_PATH}/.bin/webpack", "--config", WEBPACK_CONFIG] + ARGV
else
  cmdline = ["yarn", "run", "webpack", "--", "--config", WEBPACK_CONFIG] + ARGV
end

Dir.chdir(APP_PATH) do
  exec newenv, *cmdline
end
```

It seems like it's an issue with yarn.

gauravtiwari:
Thank you so much. This is great 👍  🍰 

Perhaps, you wanna make a PR with your solution until yarn is fixed. I saw many issues raised with same problem. Does the solution works with foreman too?

gauravtiwari:
I guess we will also need to document the README to add `ruby` prefix when running these binstubs on windows

soundasleep:
How's this? ^_^ https://github.com/rails/webpacker/pull/584

gauravtiwari:
closing in favour of #584

PR #584, opened 2017-07-22
Workarounds for Yarn not playing nicely on Windows

soundasleep:
- Bypass `yarn run` when running on Windows; instead, we can go directly to `node_modules/.bin/`
- Create batch scripts so users don't need to type in `ruby bin/webpack` manually

See discussion on https://github.com/rails/webpacker/pull/583

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@
     host: localhost
 ```
 
+- On Windows, `ruby bin/webpacker` and `ruby bin/webpacker-dev-server` will now bypass yarn, and execute via `node_modules/.bin` directly - [#584](https://github.com/rails/webpacker/pull/584)
+
 ### Breaking changes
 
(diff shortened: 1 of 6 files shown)


gauravtiwari:
@soundasleep Looks great 👍  🍰  Not sure about `.bat` scripts though. How do you run `rails` binstub on windows? `./bin/rails` or `ruby ./bin/rails`?

Can we also document this in README and Changelog please?

gauravtiwari:
Lets drop `.bat` scripts (primarily because it's not uniform with other binstubs and isn't useful for non-window users). We can document this in README to use it as `ruby ./bin/webpack`

soundasleep:
Yup! That's a great idea. Updated the MR, and removed the batch scripts.

On Windows you don't tend to use any of the binstubs anyway, you'll use `rails` directly, as there's usually only a single version globally installed. e.g. http://railsinstaller.org/en

gauravtiwari:
Awesome 🍰  🎉  Looks great, thank you :)

Yeah, that makes sense 👍

PR #595, opened 2017-07-26
Document environment variables

### LINKED ISSUE #586: Add instructions to support dotenv?
I believe dotenv is fairly popular in the Ruby community.

Here's what I did to support a "Ruby-like" dotenv with Webpack:

```
yarn add dotenv
```

```javascript
// config/webpack/dotenv.js

const dotenv = require('dotenv');

const dotenvFiles = [
  `.env.${process.env.NODE_ENV}.local`,
  '.env.local',
  `.env.${process.env.NODE_ENV}`,
  '.env'
];

function loadEnv() {
  dotenvFiles.forEach((dotenvFile) => {
    dotenv.config({ path: dotenvFile, silent: true });
  });
}

module.exports = { loadEnv };
```

```javascript
// config/webpack/shared.js

...
const dotenv = require('./dotenv');

dotenv.loadEnv();

module.exports = {

  ...

  plugins: [
    new webpack.EnvironmentPlugin(process.env),
    ...
```

And here's how to use env variables in your JS code:

```bash
# .env

STRIPE_PUBLISHABLE_KEY=...
```

```javascript
// Example JS file

Stripe.setPublishableKey(process.env.STRIPE_PUBLISHABLE_KEY);
```

Do you guys think it would be useful to add these instructions in the README? I can do a PR with the instructions above if needed.

jeromedalbert:
Closes #586. This PR documents environment variables and gives optional instructions to support dotenv.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1041,7 +1041,7 @@ console.log(process.env.FOO) // Compiles to console.log("hello")
 ```
 
 You may want to store configuration in environment variables via `.env` files,
-similarly to the [dotenv Ruby gem](https://github.com/bkeepers/dotenv).
+similar to the [dotenv Ruby gem](https://github.com/bkeepers/dotenv).
 
 In development, if you use Foreman or [Invoker](http://invoker.codemancers.com)
 to launch the Webpack server, both of these tools have basic support for a
@@ -1077,8 +1077,8 @@ module.exports = {
   ...
 ```
 
-**Warning:** using Foreman/Invoker and npm dotenv at the same can result in
-confusing behavior, in that Foreman/Invoker variables can take precedence over
+**Warning:** using Foreman/Invoker and npm dotenv at the same time can result in
+confusing behavior, in that Foreman/Invoker variables take precedence over
 npm dotenv variables.
 
 ## Extending
--- a/README.md
+++ b/README.md
@@ -61,6 +61,7 @@ in which case you may not even need the asset pipeline. This is mostly relevant 
   - [Link sprocket assets](#link-sprocket-assets)
     - [Using helpers](#using-helpers)
     - [Using babel module resolver](#using-babel-module-resolver)
+  - [Environment variables](#environment-variables)
 - [Extending](#extending)
 - [Deployment](#deployment)
   - [Heroku](#heroku)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1024,6 +1025,62 @@ import FooImage from 'assets/images/foo-image.png'
...
+
+You can then reference these variables in your javascript app code with
+`process.env`:
+
+```js
+console.log(process.env.FOO) // Compiles to console.log("hello")
+```
+
+You may want to store configuration in environment variables via `.env` files,
+similarly to the [dotenv Ruby gem](https://github.com/bkeepers/dotenv).
```
prathamesh-sonpatki: similar to

README.md
```diff
@@ -1024,6 +1025,62 @@ import FooImage from 'assets/images/foo-image.png'
...
+];
+dotenvFiles.forEach((dotenvFile) => {
+  dotenv.config({ path: dotenvFile, silent: true });
+});
+
+module.exports = {
+  ...
+```
+
+**Warning:** using Foreman/Invoker and npm dotenv at the same can result in
```
prathamesh-sonpatki: at the same time

README.md
```diff
@@ -1024,6 +1025,62 @@ import FooImage from 'assets/images/foo-image.png'
...
+confusing behavior, in that Foreman/Invoker variables can take precedence over
```
prathamesh-sonpatki: /s/can/will ? If we know that it will take precedence?

dwightwatson:
Would it perhaps be better to move the dotenv configuration to another file (rather than `shared.js`) and then call it from there? I think it would make it clearer that it's non-standard and would allow easier removal if the developer ever change their mind.

jeromedalbert:
Sure, in the linked issue I actually used a separate separate dotenv.js file. This file exports a loadenv function that is to be called in `shared.js`.

I could go even further and not have any function exported, so that just doing a require in `shared.js` would just work. Even easier removal if developers change their mind, although a little more magical.

dwightwatson:
That's what I just ended up doing, and then called `require('./dotenv')` before the `model.exports`. Feels a little more plug and play.

One other thought, could the `dotenvFiles` variable just become `paths` - I think it looks a little awkward but maybe that's just me. Also gives a little more flexibility to just pass `path` into the `config` method.

```js
const dotenv = require('dotenv')

const paths = [
  `.env.${process.env.NODE_ENV}.local`,
  '.env.local',
  `.env.${process.env.NODE_ENV}`,
  '.env'
]

paths.forEach(path => {
  dotenv.config({ path, silent: true })
})

```

gauravtiwari:
@jeromedalbert Thanks 🍰  👍

Issue #594, opened 2017-07-26
Summary of Changes to Support React on Rails


justin808:
From a long discussion on #464, this issue will summarize. I'll soon be posting proposed changes to the README.md.

Demo: https://github.com/justin808/webpacker-demo-webpacker_lite-merge

# Summary of changes

1. Move base url out from manifest.json to manifest.rb
2. Assign env variables to dev server settings so it can be overridden at runtime.
3. The keys for dev_server should use same format as of now as documented in [Paths](https://github.com/rails/webpacker#paths) on the README.md. Note that
  1. `hot` is a new setting to indicate that the dev_server is used with hot reloading, which means that CSS should be inlined to be hot loaded.
  2. The presence of dev_server means that the webpack-dev-server is used for the given env.

```yaml
development:
   # put the created files to the /public/webpack/development directory
  public_output_path: webpack/development

 # if dev_server is not provided, then dev_server is not used
 dev_server: 
   hot: true # This is a new setting
   static: false
   host: localhost
   https: false
```

CC: @gauravtiwari

justin808:
@gauravtiwari, React on Rails will handle compilation of files separately than the bin/webpack task.
Thus, the caching for compilation won't work (without some changes). 

Instead, React on Rails needs the check if the bundle is created, per [this line](https://github.com/shakacode/react_on_rails/blob/a8d19e2ea39b49521b3275b8612ed2e95d1343f8/lib/react_on_rails/utils.rb#L73).

So I have a lookup method that does not necessarily throw, and [will trigger the loading of the instance](https://github.com/shakacode/webpacker_lite/blob/master/lib/webpacker_lite/manifest.rb#L41).

Please let me know your thoughts on this one.

justin808:
@gauravtiwari @dhh,
I added a demo: https://github.com/justin808/webpacker-demo-webpacker_lite-merge. My next step is to create a beta of React on Rails that depends on this version of Webpacker. Is there anything you'd like to see me do?

Issue #464, opened 2017-06-02
Proposal to merge in changes from Webpacker Lite.


justin808:
Following up from #453 and my discussion with @gauravtiwari, and my article [Webpacker Lite: Why Fork Webpacker?](https://blog.shakacode.com/webpacker-lite-why-fork-webpacker-f0a7707fac92) and my fork [Webpacker Lite](https://github.com/shakacode/webpacker_lite).

I'd like to get some consensus on merging in the differences of Webpacker Lite before doing any work on a PR.

# YML file and LIVE 

**Webpacker**, from [installer file](https://github.com/rails/webpacker/blob/master/lib/install/config/webpacker.yml):

```yaml
development:
   public_output_path: packs
   dev_server:
     host: 0.0.0.0
     port: 8080
     https: false
```

**Webpacker Lite**, from [README](https://github.com/shakacode/webpacker_lite/blob/master/README.md):
```yaml
development:
  <<: *default
  # generated files for development, in /public/webpack/development
  webpack_public_output_dir: webpack/development
  
  # Default is localhost:3500. You can specify the protocol if needed. Defaults to http://.
  hot_reloading_host: localhost:3500
  
  # Developer note: considering removing this option so it can ONLY be turned by using an ENV value.
  # Default is false, ENV 'HOT_RELOADING' will always override 
  hot_reloading_enabled_by_default: false
```

1. Should we use host vs individual keys and nested (note, `host` should have been `hostname` in webpacker).
2. We need an option regarding whether or not using the webpack-dev-server should be the default, so this can be overridden with an env value, like `HOT_RELOADING`
3. I'm fine with `public_output_path` rather than `webpack_public_output_dir` so long as this is the full path within the public directory where the generated files go, minus the `/public` part.

# Helper differences
1. Putting minimal in the manifest.json, not putting in the port, host, etc. and putting this logic in the [Ruby helper](https://github.com/shakacode/webpacker_lite/blob/master/lib/webpacker_lite/helper.rb#L62). This minimizes the requirement on a custom Webpack config to the bare minimum mapping of bundle names to file names.
2. [stylesheet_pack_tag uses the hot reload (webpack-dev-server)](https://github.com/shakacode/webpacker_lite/blob/master/lib/webpacker_lite/helper.rb#L56) to decide if CSS should be included because you can't hot reload extracted CSS.
[...]

PR #601, opened 2017-07-28
Support React on Rails by merging Webpacker Lite

justin808:
See https://github.com/rails/webpacker/issues/594

PR to merge this into React on Rails: https://github.com/shakacode/react_on_rails/pull/908

From a long discussion on #464, this issue will summarize. I'll soon be
posting proposed changes to the README.md.

Summary of changes

* Move base url out from manifest.json to manifest.rb
* Assign env variables to dev server settings so it can be overridden at
  runtime.
* The keys for dev_server should use same format as of now as documented
  in Paths on the README.md. Note that
* hot is a new setting to indicate that the dev_server is used with hot
  reloading, which means that CSS should be inlined to be hot loaded.
  The presence of dev_server means that the webpack-dev-server is used for
  the given env.

development:
  // put the created files to the /public/webpack/development directory
  public_output_path: webpack/development

 //# if dev_server is not provided, then dev_server is not used
 dev_server:
   hot: true # This is a new setting
   static: false
   host: localhost
   https: false

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,7 +4,7 @@ gemspec
 
 gem "rails"
 gem "rake", ">= 11.1"
-gem "rubocop", ">= 0.47", require: false
+gem "rubocop", ">= 0.49", require: false
 
 group :test do
   gem "minitest", "~> 5.0"
--- a/Gemfile
+++ b/Gemfile
@@ -9,3 +9,13 @@ gem "rubocop", ">= 0.47", require: false
 group :test do
   gem "minitest", "~> 5.0"
 end
+
+if ENV["USE_PRY"]
+  gem "awesome_print"
+  gem "pry"
+  gem "pry-byebug"
+  gem "pry-doc"
+  gem "pry-rails"
+  gem "pry-rescue"
+  gem "pry-stack_explorer"
+end
--- a/Gemfile
+++ b/Gemfile
@@ -9,13 +9,3 @@ gem "rubocop", ">= 0.47", require: false
 group :test do
   gem "minitest", "~> 5.0"
 end
-
-if ENV["USE_PRY"]
-  gem "awesome_print"
-  gem "pry"
-  gem "pry-byebug"
-  gem "pry-doc"
-  gem "pry-rails"
-  gem "pry-rescue"
-  gem "pry-stack_explorer"
-end
--- a/README.md
+++ b/README.md
@@ -271,7 +271,7 @@ foreman start
 ```
 
 By default, `webpack-dev-server` listens on `0.0.0.0` that means listening
-on all IP addresses available on your system: LAN IP address, localhost, 127.0.0.1 etc. 
+on all IP addresses available on your system: LAN IP address, localhost, 127.0.0.1 etc.
 However, we use `localhost` as default hostname for serving assets in browser
 and if you want to change that, for example on cloud9 you can do so
 by changing host set in `config/webpacker.yml`.
@@ -360,10 +360,10 @@ development:
     host: 0.0.0.0
     port: 8080
     https: false
-    hot_reloading: false
+    hmr: false
 ```
 
-If you have hot_reloading turned to `true`, then the `stylesheet_pack_tag` generates no output, as you will want
+If you have hmr turned to `true`, then the `stylesheet_pack_tag` generates no output, as you will want
 to configure your styles to be inlined in your JavaScript for hot reloading. During production and testing, the
 `stylesheet_pack_tag` will create the appropriate HTML tags.
 
@@ -451,12 +451,12 @@ You can checkout these links on this subject:
 - https://webpack.js.org/configuration/dev-server/#devserver-hot
 - https://webpack.js.org/guides/hmr-react/
 
-If you are using hot reloading, you should either set the `dev_server/hot_reloading` key to TRUE or set the 
+If you are using hot reloading, you should either set the `dev_server/hmr` key to TRUE or set the
 ENV value `WEBPACKER_HMR=TRUE`. That way, your stylesheet pack tag will do **nothing** because you
 need your styles inlined in your JavaScript for hot reloading to work properly.
 
 ### Live Reloading or Static Reloading
-Live Reloading is having your assets for development provided by the webpack-dev-server. 
+Live Reloading is having your assets for development provided by the webpack-dev-server.
 
 You can override the the presence of the `dev_server` setup by setting ENV value: `WEBPACKER_DEV_SERVER=FALSE`.
 
--- a/README.md
+++ b/README.md
@@ -27,8 +27,7 @@ in which case you may not even need the asset pipeline. This is mostly relevant 
     - [Using Rails helpers in .vue files](#using-rails-helpers-in-vue-files)
   - [Elm](#elm)
 - [Binstubs](#binstubs)
-  - [Webpack dev server](#webpack-dev-server)
-  - [Webpack](#webpack)
+- [Developing with Webpacker](#developing-with-webpacker)
 - [Configuration](#configuration)
   - [Webpack](#webpack-1)
   - [Loaders](#loaders)
@@ -61,6 +60,7 @@ in which case you may not even need the asset pipeline. This is mostly relevant 
   - [Link sprocket assets](#link-sprocket-assets)
     - [Using helpers](#using-helpers)
     - [Using babel module resolver](#using-babel-module-resolver)
+  - [Environment variables](#environment-variables)
 - [Extending](#extending)
 - [Deployment](#deployment)
   - [Heroku](#heroku)
@@ -113,7 +113,8 @@ using new `--webpack` option:
 rails new myapp --webpack
 ```
 
-Or add it to your `Gemfile`, run bundle and `./bin/rails webpacker:install` or `bundle exec rake webpacker:install` (on rails version < 5.0):
+Or add it to your `Gemfile`, run bundle and `./bin/rails webpacker:install` or
+`bundle exec rake webpacker:install` (on rails version < 5.0):
 
 ```ruby
 # Gemfile
@@ -129,9 +130,8 @@ with rails version < 5.0
 
 ## Integrations
 
-Webpacker by default ships with basic out-of-the-box integration
-for React, Angular, Vue and Elm. You can see a list of available
-commands/tasks by running:
+Webpacker ships with basic out-of-the-box integration for React, Angular, Vue and Elm.
+You can see a list of available commands/tasks by running:
 
 ```bash
 # Within rails app
@@ -239,15 +239,20 @@ executable to ensure that the right configuration file and environment variables
 are loaded depending on your environment.
 
 
-### Webpack dev server
+## Developing with Webpacker
 
-In development, you'll need to run `./bin/webpack-dev-server` in a separate terminal
-from `./bin/rails server` to have your `app/javascript/packs/*.js` files compiled
-as you make changes.
+In development, Webpacker compiles on demand rather than upfront by default. This
+happens when you refer to any of the pack assets using the Webpacker helper methods.
+That means you don't have to run any separate process. Compilation errors are logged
+to the standard Rails log.
 
-`./bin/webpack-dev-server` launches the [Webpack Dev Server](https://webpack.js.org/configuration/dev-server/), which serves your pack files
-on `http://localhost:8080/` by default and supports live code reloading in the development environment. You will need to install additional plugins for Webpack if you want
-features like [Hot Module Replacement](https://webpack.js.org/guides/hot-module-replacement/).
+If you want to use live code reloading, you'll need to run `./bin/webpack-dev-server`
+in a separate terminal from `./bin/rails server`. This process will watch for changes
+in the `app/javascript/packs/*.js` files and automatically reload the browser to match.
+
+Note: The dev server serves pack files from `http://localhost:8080/` by default, which
+is a different origin than your application server. Therefore it's not compatible with
+things like Service Workers, that requires you to serve from the same origin.
 
 If you'd rather not have to run the two processes separately by hand, you can use [Foreman](https://ddollar.github.io/foreman):
 
@@ -266,7 +271,8 @@ foreman start
 ```
 
 By default, `webpack-dev-server` listens on `0.0.0.0` that means listening
-on all IP addresses available on your system: LAN IP address, localhost, 127.0.0.1 etc. However, we use `localhost` as default hostname for serving assets in browser
+on all IP addresses available on your system: LAN IP address, localhost, 127.0.0.1 etc. 
+However, we use `localhost` as default hostname for serving assets in browser
 and if you want to change that, for example on cloud9 you can do so
 by changing host set in `config/webpacker.yml`.
 
@@ -280,19 +286,6 @@ precedence over the ones already set in the configuration file.
 
 ```bash
 ./bin/webpack-dev-server --host example.com --inline true --hot false
-```
-
-### Webpack
-
-We recommend using `webpack-dev-server` during development for a better experience.
-However, if you don't want that for some reason you can always use `webpack` binstub with
-watch option, which uses webpack Command Line Interface (CLI). This will use `public_output_path` from `config/webpacker.yml`
-directory to serve your packs using configured rails server.
-
-You can pass cli options available with [Webpack](https://webpack.js.org/api/cli/):
-
-```bash
-./bin/webpack --watch --progress --colors
 ```
 
 
@@ -440,10 +433,9 @@ you don't need to do anything extra for webpacker, it just works.
 
 ### HTTPS in development
 
-You may require the `webpack-dev-server` to serve views over HTTPS in development.
-To do this, set the `https` option for `webpack-dev-server`
-to `true` in `config/webpacker.yml`, then start the dev server as usual
-with `./bin/webpack-dev-server`.
+If you're using the `webpack-dev-server` in development, you can serve views over HTTPS
+by setting the `https` option for `webpack-dev-server` to `true` in `config/webpacker.yml`,
+then start the dev server as usual with `./bin/webpack-dev-server`.
 
 Please note that the `webpack-dev-server` will use a self-signed certificate,
 so your web browser will display a warning upon accessing the page.
@@ -1041,6 +1033,62 @@ import FooImage from 'assets/images/foo-image.png'
 import 'assets/stylesheets/bar'
 ```
 
+### Environment variables
+
+Environment variables are supported out of the box in Webpacker. For example if
+you run the Webpack dev server like so:
+```
+FOO=hello BAR=world ./bin/webpack-dev-server
+```
+
+You can then reference these variables in your javascript app code with
+`process.env`:
+
+```js
+console.log(process.env.FOO) // Compiles to console.log("hello")
+```
+
+You may want to store configuration in environment variables via `.env` files,
+similar to the [dotenv Ruby gem](https://github.com/bkeepers/dotenv).
+
+In development, if you use Foreman or [Invoker](http://invoker.codemancers.com)
+to launch the Webpack server, both of these tools have basic support for a
+`.env` file (Invoker also supports `.env.local`), so no further configuration
+is needed.
+
+However, if you run the Webpack server without Foreman/Invoker, or if you
+want more control over what `.env` files to load, you can use the
+[dotenv npm package](https://github.com/motdotla/dotenv). Here is what you could
+do to support a "Ruby-like" dotenv:
+
+```
+yarn add dotenv
+```
+
+```javascript
+// config/webpack/shared.js
+
+...
+const dotenv = require('dotenv');
+
+const dotenvFiles = [
+  `.env.${process.env.NODE_ENV}.local`,
+  '.env.local',
+  `.env.${process.env.NODE_ENV}`,
+  '.env'
+];
+dotenvFiles.forEach((dotenvFile) => {
+  dotenv.config({ path: dotenvFile, silent: true });
+});
+
+module.exports = {
+  ...
+```
+
+**Warning:** using Foreman/Invoker and npm dotenv at the same time can result in
+confusing behavior, in that Foreman/Invoker variables take precedence over
+npm dotenv variables.
+
 ## Extending
 
 We suggest you don't directly overwrite the provided configuration files
@@ -1089,12 +1137,12 @@ The `javascript_pack_tag` and `stylesheet_pack_tag` helper method will automatic
 By default the output will look like this in different environments:
 
 ```html
-  <!-- In development mode with webpack-dev-server -->
-  <script src="http://localhost:8080/calendar.js"></script>
-  <link rel="stylesheet" media="screen" href="http://localhost:8080/calendar.css">
   <!-- In development mode -->
   <script src="/packs/calendar.js"></script>
   <link rel="stylesheet" media="screen" href="/packs/calendar.css">
+  <!-- In development mode with webpack-dev-server -->
+  <script src="http://localhost:8080/calendar.js"></script>
+  <link rel="stylesheet" media="screen" href="http://localhost:8080/calendar.css">
   <!-- In production mode -->
   <script src="/packs/calendar-0bd141f6d9360cf4a7f5.js"></script>
   <link rel="stylesheet" media="screen" href="/packs/calendar-dc02976b5f94b507e3b6.css">
@@ -1237,12 +1285,6 @@ manually with Ruby:
 C:\path>ruby bin\webpack
 C:\path>ruby bin\webpack-dev-server
 ```
-
-## Wishlist
-
-- HMR - [#188](https://github.com/rails/webpacker/issues/188)
-- Support rails engines - [#348](https://github.com/rails/webpacker/issues/348)
-
 
 ## License
 Webpacker is released under the [MIT License](https://opensource.org/licenses/MIT).
--- a/README.md
+++ b/README.md
@@ -367,10 +367,10 @@ development:
     host: 0.0.0.0
     port: 8080
     https: false
-    hot: false
+    hot_reloading: false
 ```
 
-If you have hot turned to `true`, then the `stylesheet_pack_tag` generates no output, as you will want
+If you have hot_reloading turned to `true`, then the `stylesheet_pack_tag` generates no output, as you will want
 to configure your styles to be inlined in your JavaScript for hot reloading. During production and testing, the
 `stylesheet_pack_tag` will create the appropriate HTML tags.
 
@@ -459,7 +459,7 @@ You can checkout these links on this subject:
 - https://webpack.js.org/configuration/dev-server/#devserver-hot
 - https://webpack.js.org/guides/hmr-react/
 
-If you are using hot reloading, you should either set the `dev_server/hot` key to TRUE or set the 
+If you are using hot reloading, you should either set the `dev_server/hot_reloading` key to TRUE or set the 
 ENV value `WEBPACKER_HMR=TRUE`. That way, your stylesheet pack tag will do **nothing** because you
 need your styles inlined in your JavaScript for hot reloading to work properly.
 
@@ -617,10 +617,7 @@ and
 
 #### React
 
-The most widely used React integration is [shakacode/react_on_rails](https://github.com/shakacode/react_on_rails) which includes support for server rendering, redux and react-router.
-
-Other alternatives include [react-rails](https://github.com/reactjs/react-rails) and 
-[webpacker-react](https://github.com/renchap/webpacker-react) for more advanced react integration. 
+If you need more advanced React-integration, like server rendering, redux, or react-router, see [shakacode/react_on_rails](https://github.com/shakacode/react_on_rails), [react-rails](https://github.com/reactjs/react-rails), and [webpacker-react](https://github.com/renchap/webpacker-react).
 
 If you're not concerned with view helpers to pass props or server rendering, can do it yourself:
 
@@ -1124,7 +1121,7 @@ Webpacker lazily compiles assets in test env so you can write your tests without
 setup and everything will just work out of the box.
 
 Note, [React on Rails] users should set configuration value `compile` to false, as React on Rails
-handle compilation for test and production environments.
+handles compilation for test and production environments.
 
 Here is a sample system test case with hello_react example component:
 
--- a/README.md
+++ b/README.md
@@ -463,6 +463,12 @@ If you are using hot reloading, you should either set the `dev_server/hot` key t
 ENV value `WEBPACKER_HMR=TRUE`. That way, your stylesheet pack tag will do **nothing** because you
 need your styles inlined in your JavaScript for hot reloading to work properly.
 
+### Live Reloading or Static Reloading
+Live Reloading is having your assets for development provided by the webpack-dev-server. 
+
+You can override the the presence of the `dev_server` setup by setting ENV value: `WEBPACKER_DEV_SERVER=FALSE`.
+
+You might do this if you are switching back and forth between statically compiling files during development and trying to get hot reloading to work.
 
 ## Linking Styles, Images and Fonts
 
--- a/README.md
+++ b/README.md
@@ -1120,9 +1120,6 @@ git push heroku master
 Webpacker lazily compiles assets in test env so you can write your tests without any extra
 setup and everything will just work out of the box.
 
-Note, [React on Rails] users should set configuration value `compile` to false, as React on Rails
-handles compilation for test and production environments.
-
 Here is a sample system test case with hello_react example component:
 
 ```js
--- a/README.md
+++ b/README.md
@@ -459,6 +459,10 @@ You can checkout these links on this subject:
 - https://webpack.js.org/configuration/dev-server/#devserver-hot
 - https://webpack.js.org/guides/hmr-react/
 
+If you are using hot reloading, you should either set the `dev_server/hot` key to TRUE or set the 
+ENV value `WEBPACKER_HMR=TRUE`. That way, your stylesheet pack tag will do **nothing** because you
+need your styles inlined in your JavaScript for hot reloading to work properly.
+
 
 ## Linking Styles, Images and Fonts
 
--- a/README.md
+++ b/README.md
@@ -367,7 +367,13 @@ development:
     host: 0.0.0.0
     port: 8080
     https: false
+    hot: false
 ```
+
+If you have hot turned to `true`, then the `stylesheet_pack_tag` generates no output, as you will want
+to configure your styles to be inlined in your JavaScript for hot reloading. During production and testing, the
+`stylesheet_pack_tag` will create the appropriate HTML tags.
+
 
 #### Resolved Paths
 
@@ -601,8 +607,12 @@ and
 
 #### React
 
-You may consider using [react-rails](https://github.com/reactjs/react-rails) or
-[webpacker-react](https://github.com/renchap/webpacker-react) for more advanced react integration. However here is how you can do it yourself:
+The most widely used React integration is [shakacode/react_on_rails](https://github.com/shakacode/react_on_rails) which includes support for server rendering, redux and react-router.
+
+Other alternatives include [react-rails](https://github.com/reactjs/react-rails) and 
+[webpacker-react](https://github.com/renchap/webpacker-react) for more advanced react integration. 
+
+If you're not concerned with view helpers to pass props or server rendering, can do it yourself:
 
 ```erb
 <%# views/layouts/application.html.erb %>
@@ -1103,6 +1113,8 @@ git push heroku master
 Webpacker lazily compiles assets in test env so you can write your tests without any extra
 setup and everything will just work out of the box.
 
+Note, [React on Rails] users should set configuration value `compile` to false, as React on Rails
+handle compilation for test and production environments.
 
 Here is a sample system test case with hello_react example component:
 
--- a/lib/install/config/webpack/shared.js
+++ b/lib/install/config/webpack/shared.js
@@ -11,7 +11,7 @@ const ManifestPlugin = require('webpack-manifest-plugin')
 const extname = require('path-complete-extname')
 const { env, settings, output, loadersDir, resolvedModules } = require('./configuration.js')
 
-const extensionGlob = `**/*{${settings.extensions.join(',')}}*`
+const extensionGlob = `**/*{${settings.extensions.join(',')}}`
 const entryPath = join(settings.source_path, settings.source_entry_path)
 const packPaths = sync(join(entryPath, extensionGlob))
 
@@ -36,7 +36,7 @@ module.exports = {
 
   plugins: [
     new webpack.EnvironmentPlugin(JSON.parse(JSON.stringify(env))),
-    new ExtractTextPlugin(env.NODE_ENV === 'production' ? '[name]-[hash].css' : '[name].css'),
+    new ExtractTextPlugin(env.NODE_ENV === 'production' ? '[name]-[contenthash].css' : '[name].css'),
     new ManifestPlugin({
       writeToFileEmit: true
     })
--- a/lib/install/config/webpack/shared.js
+++ b/lib/install/config/webpack/shared.js
@@ -27,8 +27,7 @@ module.exports = {
 
   output: {
     filename: '[name].js',
-    path: output.path,
-    publicPath: output.publicPath
+    path: output.path
   },
 
   module: {
@@ -39,7 +38,6 @@ module.exports = {
     new webpack.EnvironmentPlugin(JSON.parse(JSON.stringify(env))),
     new ExtractTextPlugin(env.NODE_ENV === 'production' ? '[name]-[hash].css' : '[name].css'),
     new ManifestPlugin({
-      publicPath: output.publicPath,
       writeToFileEmit: true
     })
   ],
--- a/lib/install/config/webpack/shared.js
+++ b/lib/install/config/webpack/shared.js
@@ -37,9 +37,7 @@ module.exports = {
   plugins: [
     new webpack.EnvironmentPlugin(JSON.parse(JSON.stringify(env))),
     new ExtractTextPlugin(env.NODE_ENV === 'production' ? '[name]-[contenthash].css' : '[name].css'),
-    new ManifestPlugin({
-      writeToFileEmit: true
-    })
+    new ManifestPlugin({ writeToFileEmit: true })
   ],
 
   resolve: {
--- a/lib/install/config/webpacker.yml
+++ b/lib/install/config/webpacker.yml
@@ -4,8 +4,7 @@ default: &default
   source_path: app/javascript
   source_entry_path: packs
   public_output_path: packs
-
-  compile: false
+  cache_path: tmp/cache/webpacker
 
   # Additional paths webpack should lookup modules
   # ['app/assets', 'engine/foo/app/assets']
@@ -29,7 +28,7 @@ default: &default
 
 development:
   <<: *default
-
+  compile: true
   dev_server:
     host: localhost
     port: 8080
@@ -38,10 +37,11 @@ development:
 
 test:
   <<: *default
-
-  public_output_path: packs-test
-  # Compile webpack assets during lookup when running tests
   compile: true
+  public_output_path: packs-test
 
 production:
   <<: *default
+
+  # Production demands on precompilation of packs prior to booting for performance.
+  compile: false
--- a/lib/install/config/webpacker.yml
+++ b/lib/install/config/webpacker.yml
@@ -34,7 +34,7 @@ development:
     host: localhost
     port: 8080
     https: false
-    hot: false
+    hot_reloading: false
 
 test:
   <<: *default
--- a/lib/install/config/webpacker.yml
+++ b/lib/install/config/webpacker.yml
@@ -33,7 +33,7 @@ development:
     host: localhost
     port: 8080
     https: false
-    hot_reloading: false
+    hmr: false
 
 test:
   <<: *default
--- a/lib/install/config/webpacker.yml
+++ b/lib/install/config/webpacker.yml
@@ -34,6 +34,7 @@ development:
     host: localhost
     port: 8080
     https: false
+    hot: false
 
 test:
   <<: *default
--- a/lib/webpacker.rb
+++ b/lib/webpacker.rb
@@ -2,14 +2,15 @@ module Webpacker
   extend self
 
   def bootstrap
-    Webpacker::Env.load
-    Webpacker::Configuration.load
-    Webpacker::Manifest.load
+    Webpacker::Env.load_instance
+    Webpacker::Configuration.load_instance
+    Webpacker::DevServer.load_instance
+    Webpacker::Manifest.load_instance
   end
 
   def compile
     Webpacker::Compiler.compile
-    Webpacker::Manifest.load
+    Webpacker::Manifest.load_instance
   end
 
   def env
@@ -19,6 +20,7 @@ end
 
 require "webpacker/env"
 require "webpacker/configuration"
+require "webpacker/dev_server"
 require "webpacker/manifest"
 require "webpacker/compiler"
 require "webpacker/railtie" if defined?(Rails)
--- a/lib/webpacker.rb
+++ b/lib/webpacker.rb
@@ -4,7 +4,6 @@ module Webpacker
   def bootstrap
     Webpacker::Env.load
     Webpacker::Configuration.load
-    Webpacker::DevServer.load
     Webpacker::Manifest.load
   end
 
--- a/lib/webpacker.rb
+++ b/lib/webpacker.rb
@@ -14,10 +14,11 @@ module Webpacker
   end
 
   def env
-    Webpacker::Env.current.inquiry
+    Webpacker::Env.current
   end
 end
 
+require "webpacker/logger"
 require "webpacker/env"
 require "webpacker/configuration"
 require "webpacker/dev_server"
--- a/lib/webpacker.rb
+++ b/lib/webpacker.rb
@@ -2,15 +2,15 @@ module Webpacker
   extend self
 
   def bootstrap
-    Webpacker::Env.load_instance
-    Webpacker::Configuration.load_instance
-    Webpacker::DevServer.load_instance
-    Webpacker::Manifest.load_instance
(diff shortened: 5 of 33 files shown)


Gemfile
```diff
@@ -8,4 +8,16 @@ gem "rubocop", ">= 0.47", require: false
 
 group :test do
   gem "minitest", "~> 5.0"
+  gem "minitest-stub-const"
+end
+
+
+if ENV["USE_PRY"]
```
justin808: I can remove the pry stuff, but it's very useful for debugging.

lib/install/config/webpack/shared.js
```diff
@@ -39,7 +38,6 @@ module.exports = {
     new webpack.EnvironmentPlugin(JSON.parse(JSON.stringify(env))),
     new ExtractTextPlugin(env.NODE_ENV === 'production' ? '[name]-[hash].css' : '[name].css'),
     new ManifestPlugin({
-      publicPath: output.publicPath,
```
justin808: @gauravtiwari Can you help me with the webpack changes? I didn't see any docs on how to test this.

Gemfile
```diff
@@ -9,3 +9,14 @@ gem "rubocop", ">= 0.47", require: false
 group :test do
   gem "minitest", "~> 5.0"
 end
+
+
+if ENV["USE_PRY"]
```
justin808: pry is super useful for debugging

lib/install/config/webpack/shared.js
```diff
@@ -28,7 +28,6 @@ module.exports = {
   output: {
     filename: '[name].js',
     path: output.path,
-    publicPath: output.publicPath
```
justin808: This should remove the subdirectory from the manifest.

lib/webpacker/configuration.rb
```diff
@@ -4,12 +4,21 @@
 
 class Webpacker::Configuration < Webpacker::FileLoader
   class << self
+    def reset
+      @defaults = nil
+      super
+    end
```
justin808: This method is used by tests.

lib/webpacker/configuration.rb
```diff
@@ -45,19 +54,30 @@ def fetch(key)
...
+        public_output_path.starts_with?("/") ? public_output_path : "/#{public_output_path}"
+      end
+    end
   end
 
   private
-    def load
-      return super unless File.exist?(@path)
+    def load_data
+      return Webpacker::Configuration.defaults unless File.exist?(@path)
```
justin808: I changed `load` to `load_instance` and `load_data`. Having just `load` is super confusing.

test/configuration_test.rb
```diff
@@ -3,31 +3,31 @@
 class ConfigurationTest < Minitest::Test
   def test_entry_path
     entry_path = File.join(File.dirname(__FILE__), "test_app/app/javascript", "packs").to_s
-    assert_equal Webpacker::Configuration.entry_path.to_s, entry_path
+    assert_equal entry_path, Webpacker::Configuration.entry_path.to_s
```
justin808: FYI -- all tests wrote expected and actual in reverse order.

gauravtiwari:
@justin808 Hey, Sorry I am away until next week and don't have access to good internet connection :( Will take a look once I am back on 6th

justin808:
@rafaelfranca @claudiob, any chance that either of you can take a look at this, given @gauravtiwari is unavailable?

@gauravtiwari: I released https://github.com/shakacode/react_on_rails/pull/908 as [version 9.0.0.beta.1](https://rubygems.org/gems/react_on_rails/versions/9.0.0.beta.1).

claudiob:
@justin808 thanks for the PR. This is a big change and I'm not that familiar with the codebase of `rails/webpacker`, so I'd be happy to have @gauravtiwari review your request once he is available.

lib/webpacker/dev_server.rb
```diff
@@ -0,0 +1,91 @@
...
+    def host
+      fetch(:host)
+    end
+
+    def port
+      fetch(:port)
+    end
+
+    def https?
+      fetch(:https)
```
leighhalliday: does `fetch(:https)` already return a boolean value?

lib/webpacker/manifest.rb
```diff
@@ -13,32 +13,95 @@ def file_path
...
+    # If no manifest file exists, then throw an error.
+    # **Used by React on Rails.**
+    def lookup_path_no_throw(name)
+      instance.confirm_manifest_exists
+      load_instance
+      unless instance
+        raise Webpacker::FileLoader::FileLoaderError.new("Webpacker::Manifest.load must be called first")
       end
+      hashed_name = instance.data[name.to_s]
+      return hashed_name if hashed_name.blank?
```
leighhalliday: if it's blank it might be `""` or `nil`, does it matter which for the return value or should it be explicitly returning `nil`?

test/helper_test.rb
```diff
@@ -7,29 +7,39 @@ def setup
...
   def test_stylesheet_pack_tag_splat
     style = %(<link rel="stylesheet" media="all" href="/packs/bootstrap-c38deda30895059837cf.css" />\n) +
       %(<link rel="stylesheet" media="all" href="/packs/application-dd6b1cd38bfa093df600.css" />)
-    assert_equal @view.stylesheet_pack_tag("bootstrap.css", "application.css", media: "all"), style
+    assert_equal style, @view.stylesheet_pack_tag("bootstrap.css", "application.css", media: "all")
+  end
+
+  def test_stylesheet_pack_tag_outputs_nothing_for_hot
+    Webpacker::DevServer.stub(:hot?, true) do
+      # Webpacker::Configuration.reset
```
leighhalliday: should these commented out lines be here?

README.md
```diff
@@ -601,8 +617,12 @@ and
 
 #### React
 
-You may consider using [react-rails](https://github.com/reactjs/react-rails) or
-[webpacker-react](https://github.com/renchap/webpacker-react) for more advanced react integration. However here is how you can do it yourself:
+The most widely used React integration is [shakacode/react_on_rails](https://github.com/shakacode/react_on_rails) which includes support for server rendering, redux and react-router.
```
dhh: Let's just keep this to the facts: `If you need more advanced React-integration, like server rendering, redux, or react-router, see [shakacode/react_on_rails](https://github.com/shakacode/react_on_rails), [react-rails](https://github.com/reactjs/react-rails), and [webpacker-react](https://github.com/renchap/webpacker-react)`.

lib/webpacker/configuration.rb
dhh: Not sure I follow? In which cases are we calling #load to mean different things?

lib/webpacker/dev_server.rb
```diff
@@ -0,0 +1,91 @@
...
+      Webpacker::Configuration.file_path
+    end
+
+    # Uses the hot_reloading_host if appropriate
+    def base_url
+      "#{protocol}://#{host}:#{port}"
+    end
+
+    private
+
```
dhh: ✂️ NL, indent everything below private.

Gemfile
```diff
@@ -9,3 +9,13 @@ gem "rubocop", ">= 0.47", require: false
...
+  gem "awesome_print"
+  gem "pry"
+  gem "pry-byebug"
+  gem "pry-doc"
+  gem "pry-rails"
+  gem "pry-rescue"
+  gem "pry-stack_explorer"
+end
```
dhh: Don't think we need this as part of the package by default.

lib/webpacker/configuration.rb
dhh: If this is solely needed for testing, then we should just do it by hand in the tests. Not keen on having test-only methods part of the public API.

lib/webpacker/dev_server.rb
```diff
@@ -0,0 +1,91 @@
+# Same convention as manifest/configuration.rb
+
+# Loads webpacker configuration from config/webpacker.yml
+
+require "webpacker/configuration"
+
+class Webpacker::DevServer < Webpacker::FileLoader
+  class << self
+    def dev_server?
```
dhh: Let's go with `enabled?` instead.

lib/webpacker/dev_server.rb
```diff
@@ -0,0 +1,91 @@
...
+      env_val = env_value("WEBPACKER_DEV_SERVER")
+
+      # override dev_server setup WEBPACKER_DEV_SERVER=FALSE
+      return false if env_val == false
+
+      # If not specified, then check if values in the config file for the dev_server key
+      !dev_server_values.nil?
```
dhh: We can boil this down to:

```ruby
def enabled?
  case ENV["WEBPACKER_DEV_SERVER"]
  when /true/i then true
  when /false/i then false
  else data.fetch(:dev_server).present?
  end
end
```

lib/webpacker/dev_server.rb
```diff
@@ -0,0 +1,91 @@
...
+    end
+
+    # read settings for dev_server
+    def hot?
+      return false unless dev_server?
+      env_val = env_value("WEBPACKER_HMR")
+      return env_val unless env_val.nil?
+      fetch(:hot)
```
dhh: Let's start by renaming to `hot_reloading?` and then we can simplify the implementation:

```ruby
if enabled?
  case ENV["WEBPACKER_HMR"]
  when /true/i then true
  when /false/i then false
  else fetch(:hot_reloading)
  end
end
```

lib/webpacker/dev_server.rb
```diff
@@ -0,0 +1,91 @@
...
+
+    def env_value(env_key)
+      if ENV[env_key].present?
+        val = ENV[env_key]
+        val_upcase = val.upcase
+        return true if val_upcase == "TRUE"
+        return false if val_upcase == "FALSE"
+        raise new ArgumentError("#{env_key} value is #{val}. Set to TRUE|FALSE")
+      end
+      # returns nil
```
dhh: We can nix this per the refactorings above.

lib/webpacker/dev_server.rb
```diff
@@ -0,0 +1,91 @@
...
+    end
+
+    def dev_server_values
+      data.fetch(:dev_server, nil)
+    end
```
dhh: Nix this.

lib/webpacker/dev_server.rb
```diff
@@ -0,0 +1,91 @@
...
+
+    def fetch(key)
+      return nil unless dev_server?
+      dev_server_values.fetch(key, dev_server_defaults[key])
```
dhh: Can use:

```ruby
if enabled?
  data[:dev_server][key] || Webpacker::Configuration.defaults[:dev_server][key]
end
```

lib/webpacker/dev_server.rb
```diff
@@ -0,0 +1,91 @@
...
+    end
+
+    def data
+      load_instance if Webpacker.env.development?
+      unless instance
+        raise Webpacker::FileLoader::FileLoaderError.new("Webpacker::DevServer.load_data must be called first")
+      end
```
dhh: Use the same single-line format as the rest of the code base: `raise Webpacker::FileLoader::FileLoaderError.new("Webpacker::DevServer.load must be called first") unless instance`

lib/webpacker/dev_server.rb
```diff
@@ -0,0 +1,91 @@
...
+      instance.data
+    end
+
+    def dev_server_defaults
+      @defaults ||= Webpacker::Configuration.defaults[:dev_server]
+    end
```
dhh: Nix this.

lib/webpacker/file_loader.rb
```diff
@@ -4,21 +4,44 @@ class NotFoundError < StandardError; end
...
 
   class << self
-    def load(path = file_path)
+    def load_instance(path = file_path)
+      # Assume production is 100% cached and don't reload if file's mtime not changed
+      cached = self.instance && # if we have a singleton
+        (env == "production" || # skip if production bc always cached
+          (File.exist?(path) && self.instance.mtime == File.mtime(path))) # skip if mtime not changed
+
+      return if cached
```
dhh: This needs to be extracted and reworked. We shouldn't need this many comments to explain the method. Let's use the Composed Method pattern.

lib/webpacker/file_loader.rb
```diff
@@ -4,21 +4,44 @@ class NotFoundError < StandardError; end
...
+      raise FileLoaderError.new("Subclass of Webpacker::FileLoader should override this method")
+    end
+
+    def reset
+      self.instance = nil
+      load_instance
+    end
+
+    private
+
```
dhh: ✂️ the NL, indent under private.

lib/webpacker/helper.rb
```diff
@@ -9,8 +9,12 @@ module Webpacker::Helper
   # In production mode:
   #   <%= asset_pack_path 'calendar.css' %> # => "/packs/calendar-1016838bab065ae1e122.css"
   def asset_pack_path(name, **options)
-    asset_path(Webpacker::Manifest.lookup(name), **options)
+    path = Webpacker::Configuration.public_output_path
+    file = Webpacker::Manifest.lookup(name)
+    full_path = "#{path}/#{file}"
```
dhh: Why wouldn't we have all this work happening in the manifest?

lib/webpacker/helper.rb
```diff
@@ -42,8 +45,25 @@ def javascript_pack_tag(*names, **options)
...
+    return "" if Webpacker::DevServer.hot?
+    stylesheet_link_tag(*asset_source(names, :stylesheet), **options)
   end
+
+  private
+    def asset_source(names, type)
+      output_path_or_url = Webpacker::Configuration.output_path_or_url
+      names.map do |name|
+        path = Webpacker::Manifest.lookup("#{name}#{compute_asset_extname(name, type: type)}")
+        "#{output_path_or_url}/#{path}"
```
dhh: Too much work happening in the helper here as well.

lib/webpacker/helper.rb
```diff
@@ -42,8 +45,25 @@ def javascript_pack_tag(*names, **options)
...
+  #
+  #   # In development mode with hot-reloading and enabled_when_hot_loading
+  #   # <%= stylesheet_pack_tag('main', enabled_when_hot_loading: true) %>
+  #   <link rel="stylesheet" media="screen" href="/public/webpack/development/calendar-1016838bab065ae1e122.css" />
+  #
   def stylesheet_pack_tag(*names, **options)
-    sources = names.map { |name|  Webpacker::Manifest.lookup("#{name}#{compute_asset_extname(name, type: :stylesheet)}") }
-    stylesheet_link_tag(*sources, **options)
+    return "" if Webpacker::DevServer.hot?
+    stylesheet_link_tag(*asset_source(names, :stylesheet), **options)
```
dhh: Don't use multiple exits from a method if you can avoid it:

```ruby
if Webpacker::DevServer.hot_reloading?
  ""
else
  stylesheet_link_tag(*asset_source(names, :stylesheet), **options)
end
```

lib/webpacker/manifest.rb
```diff
@@ -13,32 +13,95 @@ def file_path
       Webpacker::Configuration.manifest_path
     end
 
+    # Throws an error if the file is not found. If Configuration.compile? then compilation is invoked
+    # the file is missing.
+    # React on Rails users will need to set Configuration.compile? to false as compilation is configured
+    # in the package.json for React on Rails.
```
dhh: Specific configuration instructions for React on Rails should happen in React on Rails. We're not including config instructions for other integration packages here either.

lib/webpacker/manifest.rb
```diff
@@ -13,32 +13,95 @@ def file_path
...
     def lookup(name)
       if Webpacker::Configuration.compile?
         compile_and_find!(name)
       else
+        # Since load_instance is a no-op for production, this should be fine.
+        # If we're using `webpack -w` to recompile when files changes, we need
+        # to call this before find!
+        load_instance
```
dhh: Not exactly confidence inspiring that something "should be fine" 😄. We should explain clearly here why/when the #load needs to be called prior to find!

lib/webpacker/manifest.rb
```diff
@@ -13,32 +13,95 @@ def file_path
...
-      Rails.root.join(File.join(Webpacker::Configuration.public_path, lookup(name)))
+      Rails.root.join(File.join(Webpacker::Configuration.output_path, lookup(name)))
     end
 
-    private
-      def find!(name)
-        raise Webpacker::FileLoader::FileLoaderError.new("Webpacker::Manifest.load must be called first") unless instance
-        instance.data[name.to_s] || raise(Webpacker::FileLoader::NotFoundError.new("Can't find #{name} in #{file_path}. Is webpack still compiling?"))
+    # Helper method to determine if the manifest file exists. Maybe Webpack needs to run?
+    # **Used by React on Rails.**
```
dhh: Why can't React on Rails not use `Webpacker::Configuration.manifest_path.exist?` directly?

lib/webpacker/manifest.rb
```diff
@@ -13,32 +13,95 @@ def file_path
...
+    def exist?
+      path_object = Webpacker::Configuration.manifest_path
+      path_object.exist?
+    end
+
+    # Find the real file name from the manifest key. Don't throw an error if the file is simply
+    # missing from the manifest. Return nil in that case.
+    # If no manifest file exists, then throw an error.
+    # **Used by React on Rails.**
```
dhh: If this is used by React on Rails, then it should be in React on Rails.

lib/webpacker/manifest.rb
```diff
@@ -13,32 +13,95 @@ def file_path
...
       end
+      hashed_name = instance.data[name.to_s]
+      return hashed_name if hashed_name.blank?
+      Rails.root.join(File.join(Webpacker::Configuration.output_path, hashed_name))
+    end
 
-      def compile_and_find!(name)
-        Webpacker.compile
-        find!(name)
+    private
```
dhh: Indent under private.

dhh:
Added a partial code review. There's still more to do once this is cleaned up, but let's start with that.

lib/webpacker/helper.rb
justin808: @dhh The manifest should not know about where the file is on the Rails server. When in webpack-dev-server mode, there is typically no subdirectory. In other words. Let's have the manifest be a simple mapping of the bundle name to the possibly hashed version of the bundle name.

lib/webpacker/helper.rb
justin808: That's fine. However, the default for Rubocop is the [Guard Clause](http://www.rubydoc.info/github/bbatsov/rubocop/Rubocop/Cop/Style/GuardClause).

In eslint, we have [Disallow return before else (no-else-return)](http://eslint.org/docs/rules/no-else-return) and the airbnb eslint pattern is to use error for this. Thus, for those of us that do lots of JS with the airbnb linter, we're used to the pattern of the "guard clause".

That being said, since there's no `return` needed in Ruby, I think your suggestion is fine, and it's 💯 what I used to always do and have the linter complain to me.

lib/webpacker/manifest.rb
justin808: How about

```
   # Since load_instance checks a `mtime` on the manifest for a non-production env before loading 
   # in a non-production env, we should always call this before a call to `find!` since one may be using 
   # `webpack -w` and a file may have been added to the manifest since Rails first started.
```

lib/webpacker/manifest.rb
justin808: It can. The `exist?` seemed to belong here in Webpacker, as the manifest might not yet exist yet. I believe this method is called elsewhere in Webpacker.

lib/webpacker/manifest.rb
justin808: In order to put this logic into React on Rails, we'd have to expose even more to the outside world.

I'd suggest we have the methods `find!` and `lookup_path` take an optional named param (or hash param) of `throw: true` and React on Rails can call them passing `throw: false`.

lib/webpacker/manifest.rb
justin808: No, it would not matter.

lib/install/config/webpacker.yml
```diff
@@ -34,6 +34,7 @@ development:
     host: localhost
     port: 8080
     https: false
+    hot_reloading: false
```
justin808: @dhh, per the method change from `hot` to `hot_reloading`, I updated the yml file.

lib/webpacker/configuration.rb
justin808: Done.

lib/webpacker.rb
```diff
@@ -2,14 +2,15 @@ module Webpacker
...
+    Webpacker::Env.load_instance
+    Webpacker::Configuration.load_instance
+    Webpacker::DevServer.load_instance
+    Webpacker::Manifest.load_instance
   end
 
   def compile
     Webpacker::Compiler.compile
-    Webpacker::Manifest.load
+    Webpacker::Manifest.load_instance
```
gauravtiwari: Sorry but could we please drop these changes and move to another PR that way we can merge this one quickly and then address the naming issue in another PR :) ?

lib/webpacker/file_loader.rb
```diff
@@ -4,21 +4,38 @@ class NotFoundError < StandardError; end
...
+      raise FileLoaderError.new("Subclass of Webpacker::FileLoader should override this method")
+    end
+
+    private
+      def file_cached?(path)
+        File.exist?(path) && self.instance.mtime == File.mtime(path)
+      end
+
+      # Prefer the NODE_ENV to the rails env.
+      def production_env?
```
gauravtiwari: `Webpacker::Env` already does this - https://github.com/shakacode/webpacker/blob/4cf3fcab5c03203216be7ab0b7f161b591908174/lib/webpacker/env.rb#L19

so if we just call `Webpacker.env.production?` it would be same thing as this

lib/webpacker/helper.rb
```diff
@@ -42,8 +45,27 @@ def javascript_pack_tag(*names, **options)
   #   # In production mode:
   #   <%= stylesheet_pack_tag 'calendar', 'data-turbolinks-track': 'reload' %> # =>
   #   <link rel="stylesheet" media="screen" href="/packs/calendar-1016838bab065ae1e122.css" data-turbolinks-track="reload" />
+  #   # In development mode with hot-reloading
+  #   <%= stylesheet_pack_tag('main') %> <% # Default is false for enabled_when_hot_loading%>
+  #   # No output
+  #
+  #   # In development mode with hot-reloading and enabled_when_hot_loading
+  #   # <%= stylesheet_pack_tag('main', enabled_when_hot_loading: true) %>
```
gauravtiwari: Where we using this `enabled_when_hot_loading`?

lib/webpacker/manifest.rb
```diff
@@ -13,32 +13,94 @@ def file_path
...
 
+    # Why does this method exist? Testing? It's not in the README
     def lookup_path(name)
-      Rails.root.join(File.join(Webpacker::Configuration.public_path, lookup(name)))
+      Rails.root.join(File.join(Webpacker::Configuration.output_path, lookup(name)))
+    end
+
+    # Helper method to determine if the manifest file exists. Maybe Webpack needs to run?
+    # **Used by React on Rails.**
+    def exist?
```
gauravtiwari: This should be added to `react_on_rails` instead of here

lib/webpacker/manifest.rb
```diff
@@ -13,32 +13,94 @@ def file_path
...
+Webpacker can't find #{bundle_name} in your manifest at #{file_path}. Possible causes:
+  1. You are hot reloading.
+  2. You want to set Configuration.compile to true for your environment.
+  3. Webpack has not re-run to reflect updates.
+  4. You have misconfigured Webpacker's config/webpacker.yml file.
+  5. Your Webpack configuration is not creating a manifest.
+Your manifest contains:
+#{instance.data.to_json}
+        MSG
+        raise(Webpacker::FileLoader::NotFoundError.new(msg))
```
gauravtiwari: this is great 👍  🍰

lib/webpacker/manifest.rb
```diff
@@ -13,32 +13,94 @@ def file_path
...
+
+        if throw_if_missing
+          missing_file_from_manifest_error(name)
+        end
+      end
+
+      def compile_and_find!(name, throw_if_missing: true)
         Webpacker.compile
-        find!(name)
+        find(name, throw_if_missing: throw_if_missing)
```
gauravtiwari: Not sure if these changes should be included in this PR. Can we revert the behaviour of these methods as it was please? I like the error messages though so we can keep that. There is lot of out-of-the-scope changes here so it's hard to review :)

lib/webpacker/dev_server.rb
```diff
@@ -0,0 +1,73 @@
+# Same convention as manifest/configuration.rb
+
+# Loads webpacker configuration from config/webpacker.yml
+
+require "webpacker/configuration"
+
+class Webpacker::DevServer < Webpacker::FileLoader
+  class << self
+    def enabled?
+      case ENV["WEBPACKER_DEV_SERVER"]
```
gauravtiwari: Perhaps we reverse the behaviour with env variable - like `DISABLE_DEV_SERVER`? In case of webpacker this is enabled by default so not needed anyway.

lib/webpacker/dev_server.rb
```diff
@@ -0,0 +1,73 @@
...
+    # read settings for dev_server
+    def hot_reloading?
+      if enabled?
+        case ENV["WEBPACKER_HMR"]
+        when /true/i then true
+        when /false/i then false
+        else fetch(:hot_reloading)
+        end
+      else
+        false
```
gauravtiwari: Same as above `DISABLE_HMR` and then fallback to `fetch(:hot_reloading)`

```rb
ENV['DISABLE_HMR'] || fetch(:hot_reloading)
```

lib/webpacker/file_loader.rb
justin808: @gauravtiwari are you *sure* that would work since you're creating a circular dependency amongst the files. I don't think so but if you're *sure*, I'll try.

lib/webpacker.rb
justin808: Which is "this one"? If we can keep this small refactoring in here, I'd prefer it. @dhh?

lib/webpacker/manifest.rb
justin808: @gauravtiwari How about if I just remove the comment? There's 2 situations: 

1. No manifest.
2. Manifest is missing the files.

lib/webpacker/manifest.rb
justin808: @gauravtiwari this is 100% required for me, as we use a watch on Webpack rather than a rake task to compile.

lib/webpacker.rb
gauravtiwari: Oh I meant, change from `load` to `load_instance`. I am happy to see it too - just a lot of unrelated change to review in one PR :)

lib/webpacker/file_loader.rb
gauravtiwari: Oh yes, totally missed 👍

lib/webpacker.rb
dhh: Concur with @gauravtiwari on this being unrelated to this PR. I also don't think it's an improvement but we could discuss that in another ticket.

lib/webpacker/helper.rb
justin808: Correct, I'm removing that. We solved that issue with webpacker.yml.

lib/webpacker/dev_server.rb
justin808: OK with any way to solve the problem of having some developers on my team have to keep a modified version of the yml file.

1. The ENV value is totally optional
2. DISABLE_XYZ == false . ==> does that mean enable XYZ?

lib/webpacker/dev_server.rb
justin808: The current idea is that the hot reloading setting only applies if the dev server is enabled.

An alternative interpretation could be that the ENV for hot reloading overrides the dev server setting. I think that's getting too complicated.

lib/webpacker/dev_server.rb
justin808: Done.

lib/webpacker/file_loader.rb
```diff
@@ -4,21 +4,38 @@ class NotFoundError < StandardError; end
...
   private
     def initialize(path)
       @path = path
-      @data = load
+      @mtime = File.exist?(path) ? File.mtime(path) : nil
+      @data = load_data
     end
 
-    def load
+    def load_data
```
justin808: I find it very confusing to have `load` mean both load_instance and load_data, depending on if you have the static or non-static method. @dhh @gauravtiwari

lib/webpacker/manifest.rb
justin808: @dhh and @gauravtiwari Is the above simple and clear enough?

lib/webpacker/manifest.rb
justin808: This is not used elsewhere. I will remove.

lib/webpacker/manifest.rb
justin808: Removed.

lib/webpacker/manifest.rb
justin808: Done

lib/webpacker.rb
justin808: @dhh and @gauravtiwari if you want me to rename all the load_instance and load_data to `load`, I'll do that. I'll wait for your confirmation on this one.

lib/install/config/webpacker.yml
justin808: @gauravtiwari @dhh Let's see if we can come to a quick agreement on the naming of the ENV and YML values.

lib/webpacker/configuration.rb
justin808: @gauravtiwari fyi

README.md
```diff
@@ -453,6 +459,16 @@ You can checkout these links on this subject:
 - https://webpack.js.org/configuration/dev-server/#devserver-hot
 - https://webpack.js.org/guides/hmr-react/
 
+If you are using hot reloading, you should either set the `dev_server/hot_reloading` key to TRUE or set the 
+ENV value `WEBPACKER_HMR=TRUE`. That way, your stylesheet pack tag will do **nothing** because you
+need your styles inlined in your JavaScript for hot reloading to work properly.
```
dhh: Why do we need both a configuration option and a ENV value for this again?

README.md
```diff
@@ -1103,6 +1120,8 @@ git push heroku master
 Webpacker lazily compiles assets in test env so you can write your tests without any extra
 setup and everything will just work out of the box.
 
+Note, [React on Rails] users should set configuration value `compile` to false, as React on Rails
+handles compilation for test and production environments.
```
dhh: I think all integration advice for React on Rails is best presented within the React on Rails documentation.

lib/install/config/webpacker.yml
dhh: I'm fine with HMR as well, though that raises a larger question. Now that we have lazy compilation available without a separate dev server, isn't HMR the _only_ reason why you'd use the dev server? If so, maybe it should be called the webpack-hmr-server and focus on that task exclusively.

lib/webpacker.rb
dhh: Please revert all back to #load, yes. We can discuss a change elsewhere.

lib/webpacker/configuration.rb
```diff
@@ -45,19 +49,30 @@ def fetch(key)
...
       @defaults ||= HashWithIndifferentAccess.new(YAML.load(default_file_path.read)[Webpacker.env])
     end
+
+    # Uses the webpack dev server host if appropriate
+    def output_path_or_url
+      if Webpacker::DevServer.enabled?
+        Webpacker::DevServer.base_url
+      else
+        # Ensure we start with a slash so that the asset helpers don't prepend the default asset
+        # pipeline locations.
```
dhh: Why not just require that in the configuration file? Seems a bit defensive otherwise to me.

lib/webpacker/dev_server.rb
dhh: What's the use case for locally forking the configuration file?

lib/webpacker/dev_server.rb
dhh: This gets even closer to the point that perhaps the dev server should just focus exclusively on HMR now.

lib/webpacker/helper.rb
dhh: I'm not sure I follow. What's the advantage of having the dev-server use a different path than what'll be used in production? That doesn't seem like a win to me.

Also, we're going to have to support multiple paths in the future to deal with service workers, as they're scoped by the directory structure. So you could have /packs/calendar.css and /service_worker.js both be generated. Can't disambiguate between the two unless the path is part of it.

lib/webpacker/manifest.rb
```diff
@@ -13,32 +13,87 @@ def file_path
       Webpacker::Configuration.manifest_path
     end
 
-    def lookup(name)
+    # Converts the "name" (aka the pack or bundle name) to to the full path for use in a browser.
+    def asset_source(name)
+      output_path_or_url = Webpacker::Configuration.output_path_or_url
+      mapped_name = lookup(name)
+      "#{output_path_or_url}/#{mapped_name}"
```
dhh: Again, I don't think this is going to work when you have multiple directory outputs.

lib/webpacker/manifest.rb
dhh: I don't follow this either. Not a fan of the naming style as well, but let's first clarify the purpose: Are you using a watch in production? If in development, how does that square against running the dev server?

README.md
justin808: Hi @dhh. The issue I'm trying to solve is:

* We can't put Ruby in the YML file, since the YML file is read by JavaScript.
* On bigger teams, it's common to have some developers wishing to do hot reloading or the webpack-dev-server, while others prefer just to have the files built in a similar manner to production.
* We don't want some developers on a team to have the webpack.yml file modified just for their own environment.
* We want a team to have different Procfiles depending if a developer wants to use these options for hot-reloading and the webpack-dev-server.

lib/webpacker/dev_server.rb
justin808: @dhh @gauravtiwari My team at https://www.friendsandguests.com has been doing static files for development mostly, with a few developers trying hot-reloading. We never explored any possible speed differences with doing just "live-reloading" (which is using the webpack-dev-server without HMR).

lib/webpacker/manifest.rb
justin808: Hi @dhh and @gauravtiwari,

1. Nobody (as far as I know) is ever using "watch" in production.
2. We do use "watch" for development and test files. That is the default from the React on Rails generator. We need the ability to _not_ throw if a file is missing so that Webpack can be _generically_ invoked to compile assets.

lib/install/config/webpacker.yml
justin808: @dhh @gauravtiwari,

1. HMR is typically always used, **BUT**, upgrades to React often put it into an "experimental state".
2. Live-reloading is typically a good choice, but per my comment below, I'd have a really hard time getting every one of my team members to switch to the live reloading or just building the files and letting Rails load the files.

lib/webpacker/dev_server.rb
justin808: See my comments below on how different team members want "live-reloading" or just the statically built files.

README.md
justin808: Done

lib/webpacker/manifest.rb
gauravtiwari: Yepp - the pack should have a full reference (if inside sub-directory)

lib/webpacker/helper.rb
justin808: @dhh We could configure using the same subdirectory of /public for the webpack-dev-server. Do you have any more details on your service worker example? Would there be new asset helpers?

lib/webpacker/manifest.rb
justin808: @gauravtiwari are you sure that the `mapped_name` does not include the subdirectory?

lib/webpacker/configuration.rb
justin808: @dhh would you prefer an exception if the path does not begin with a slash?

```rb
    # Uses the webpack dev server host if appropriate
    def output_path_or_url
      if Webpacker::DevServer.enabled?
        Webpacker::DevServer.base_url
      else
        # Ensure we start with a slash so that the asset helpers don't prepend the default asset
        # pipeline locations.
        public_output_path.starts_with?("/") ? public_output_path : "/#{public_output_path}"
      end
    end
```

lib/webpacker/manifest.rb
gauravtiwari: Sorry - yes that should work 👍  - webpack will reference sub-directories

lib/webpacker/helper.rb
dhh: I've only just started working on the service worker angle, but it's clear from the early exploration that it won't work well to force it inside /packs/. Maybe there are other solutions to the issue, but I'd like to understand more about why the manifest couldn't just use full paths? This isn't just a digest index.

lib/install/config/webpacker.yml
dhh: Doesn't that point even further to the fact that the dev server should just be relegated to the live-reloading/hmr option? If you have team members who don't want/need that, then they can just rely on on-demand compilation, which is the default now anyway?

lib/webpacker/manifest.rb
dhh: That sounds like on-demand compilation to me, which is the default now. Is there a reason that's not sufficient here as well?

lib/webpacker/dev_server.rb
dhh: Again, as above, I think we should refocus this. The dev-server is there when you want live-reloading (and potentially HMR). On-demand compilation, the new default, is there when . you just want "static files" (which, coincidentally, is where we ended up at Basecamp as well -- we didn't like live reloading either).

README.md
dhh: Again, all the more reason to separate the two. People who want the live reloading stuff runs the dev server. People who don't want that, don't. Then we can drop these extra configuration bits.

lib/webpacker/configuration.rb
dhh: I don't think we need any particular guard there. We'll birth the file with the right defaults. If people from there remove the first slash, they'll get whatever that gives.

lib/install/config/webpacker.yml
```diff
@@ -33,7 +33,7 @@ development:
     host: localhost
     port: 8080
     https: false
-    hot_reloading: false
+    hmr: false
```
justin808: @gauravtiwari @dhh FWIW, the [docs](https://webpack.js.org/configuration/dev-server/#devserver-hot) for `devServer.hot` use `hot` and not `hmr`.

ytbryan: I think `hmr` is consistent with names like 'WEBPACKER_HMR'

Also, considering usages like ` Webpacker::DevServer.hmr?` and def `hmr?`, they are consistent.

justin808: @ytbryan I agree in the sense that `DevServer.hot?` and `WEBPACKER_HOT` are a bit ambiguous.

justin808:
@gauravtiwari @dhh What's pending? I'd like to get this merged and a beta of the gem pushed so I can release 9.0.0 of React on Rails, per this PR:  https://github.com/shakacode/react_on_rails/pull/908

gauravtiwari:
@justin808 Just finalising a final few things in regards to how to check if the dev-server is running. Basically, we want to make on-demand compilation as default so for ex - if the design team want to test something they don't have to run a dev server, it just works. However, if someone in the tech team wants to use the dev-server or watcher they can simply start the server or watcher, which would turn off on-demand compilation and everything should be served by dev-server, out of the box (no config changes required).

justin808:
@gauravtiwari Regarding your prior comment, please make no assumptions that the Rails wrappers around the executables are being used as many chose to run such executables directly from yarn scripts. Also, on-demand compilation might be from `webpack -w` rather than the Webpacker compile option.

gauravtiwari:
@justin808 Yepp, I know. `webpack -w` is a blocking process so can't be used for on-demand compilation so it's just `webpack` or `Webpacker.compile`

gauravtiwari:
Ofcourse folks are free to run either `watcher` or `dev-server` and that would turn off on-demand compilation automatically

lib/webpacker/manifest.rb
```diff
@@ -17,31 +17,86 @@ def file_path
...
+
       if Webpacker::Configuration.compile?
-        compile_and_find!(name)
+        compile_and_find!(name, throw_if_missing: throw_if_missing)
       else
-        find!(name)
+        # Since load checks a `mtime` on the manifest for a non-production env before loading,
+        # we should always call this before a call to `find!` since one may be using
+        # `webpack -w` and a file may have been added to the manifest since Rails first started.
+        load
```
dhh: We're not going to want to do a file stat in production.

lib/webpacker/dev_server.rb
```diff
@@ -0,0 +1,54 @@
+# Same convention as manifest/configuration.rb
+# Loads webpacker configuration from config/webpacker.yml
+
+class Webpacker::DevServer < Webpacker::Configuration
```
dhh: I'm trying to figure out what role this class plays that's separate from the configuration class. Most of the methods simply delegate. Is this just to overlay the ENV lookups? I don't think that's enough justification for the additional indirection.

dhh:
We can't merge this in the current form where the manifest is changed from providing full paths to providing merely digest lookups, as we discussed earlier re: having different output paths, like with service workers.

Also, I still don't follow the throw_if_missing business. The methods we're using, like find!, already imply a raise if missing. If you're wrapping these calls down the line, and you don't want the raise to make it to the user, can't you catch and log instead?

lib/webpacker/manifest.rb
justin808: @dhh There's no file stat in production. The code checks the Rails.env. Does that meet your concern here?

lib/webpacker/dev_server.rb
justin808: @gauravtiwari, I think you'll need to comment here as you suggested this. I'm happy to merge these together.

justin808:
From @dhh: 

> Also, I still don't follow the throw_if_missing business. The methods we're using, like find!, already imply a raise if missing. If you're wrapping these calls down the line, and you don't want the raise to make it to the user, can't you catch and log instead?`

Should we change the API of `lookup` to `lookup!` to indicate that it throws if either the manifest or file indicated by the manifest does not exist.

The issue that I have with the lookup method always throw if the bundle name requested doesn't exist in the manifest is as follows. Possibly you can give me advice on particular use case that may eventually apply to Webpacker:

In some cases but not always, we use a different Webpack configuration for a server bundle (server rendering). In this case, we'll typically run two `webpack -w` processes to compile the client and server bundles when files change. Since we don't ever need hashing on the server bundle, we don't need a manifest. In fact, we can't have a manifest of the same name in the same location from both the client and server webpack configurations, or else we'd have a collision. In other cases, it's "good enough" to have the client bundle file used on the server, so the file will be hashed.

Thus, for React on Rails on top of Webpacker, we'll have one of 2 scenarios:

1. Same webpack config file for client and server, and server rendering needs to use `lookup`
2. Different webpack config and a non-hashed file for server rendering.

**The current code in React on Rails will first assume that all files produced by Webapck are in the manifest, and if not found, then the non-hashed, simple bundle-name is used** Thus, for **every single server request in production, we'd be doing a try/catch with the call to lookup** as React on Rails is not tracking if the server bundle is or is not in the manifest.

If we believe that these try/catches per request are unacceptable, possible solutions include:
* If the requested bundle is missing once **during non-development**, cache that information, and never try calling `lookup` again for that bundle name. I say non-development as test, production, and any such variations should never run into the issue of a bundle file is added to the Rails server, and the server is not restarted. For development, maybe there's a tiny performance hit of the extra try/catch, but that should be at most a small number of times per request, so that should be acceptable.
* Provide a configuration option in the `/config/initializers/react_on_rails.rb` as to whether or not the server bundle will ever be in the manifest.

Another case is that for test compilation when the test environment loads Webpacker, the manifest is empty, as React on Rails has not detected that the test run needs to compile the files. In this case, it's OK to catch the catch the exception and then compile. React on Rails will then run a task defined in `package.json`. Yes, Webpacker has some code in the [lib/webpacker/compiler.rb](https://github.com/rails/webpacker/blob/master/lib/webpacker/compiler.rb#L63) to check if a file is stale and to then run webpack:

```rb
      sterr, stdout, status = Open3.capture3(webpack_env, "#{RbConfig.ruby} ./bin/webpack")
```

React on Rails users have long used the conventional `package.json` scripts to do the compilation, and this is configurable for the test and production modes in [config/initializers/react_on_rails.rb](https://github.com/shakacode/react_on_rails/blob/rails-webpacker/spec/dummy/config/initializers/react_on_rails.rb#L31)

```ruby
  # If you are using the ReactOnRails::TestHelper.configure_rspec_to_compile_assets(config)
  # with rspec then this controls what yarn command is run
  # to automatically refresh your webpack assets on every test run.
  config.npm_build_test_command = "yarn run build:test"

  # This configures the script to run to build the production assets by webpack. Set this to nil
  # if you don't want react_on_rails building this file for you.
  config.npm_build_production_command = "yarn run build:production"
```

While it might be feasible to force the React on Rails community to switch to using the `bin/webpacker` binstub, one of the goals of React on Rails is to keep the JavaScripts parts all JavaScript, and that means using the `package.json` to define tasks. Thus, I would surely run into issues trying to force this.

justin808:
> We can't merge this in the current form where the manifest is changed from providing full paths to providing merely digest lookups, as we discussed earlier re: having different output paths, like with service workers.

1. Where are the different output directories configured for service workers?
2. Please confirm there would be different manifest files for service work bundles vs. regular bundles.
3. If there are different manifest files, would the helpers need to specify both the output directory (or type of output file) as well as the bundle name?

While I don't think there's a problem, besides some inconvenience, with putting back in the `publicPath` option for React on Rails users, I'd like to be able to clearly explain to the community on why the path in the manifest needs to be the full browser path, rather than relative to the manifest. 

Doc references:

* https://webpack.js.org/configuration/output/#output-publicpath
* https://webpack.js.org/configuration/dev-server/#devserver-publicpath-
* https://github.com/danethurber/webpack-manifest-plugin#publicpath

dhh:
I believe we've now extracted all the major parts of this PR into smaller changes adopted in master. We have the dev server running in a progressive enhancement mode where if you don't start it, we'll simply do on-demand compilation. This means testing just works with nothing else running, and for many smaller apps that'll be sufficient too.

In your case, where you have some developers using the dev server and some using the watcher, all you'll need to do is turn off compilation. Then developers can run the dev server or the watcher as they please. We automatically detect a running dev server and will proxy to that if there.

I've extracted the HMR enhancement into this PR: https://github.com/rails/webpacker/pull/641.

Regarding the manifest for client and server side, I still don't fully understand your case. If you're using separate ways of generating these, and they have different expectations, why not just use two separate manifests?

Re: service workers, I'm not quite sure yet. One option could simply be to start outputs in the root, and then have a manifest like { "service_worker.js" => "/service_worker-234234.js", "application.js" => "/packs/application-234234.js" }. We'll need to explore that as we do the proper implementation. So far we've removed one important roadblock, which is to proxy for the dev server, such that there are no same-origin violations any more.

justin808:
@dhh wrote:

> Regarding the manifest for client and server side, I still don't fully understand your case. If you're using separate ways of generating these, and they have different expectations, why not just use two separate manifests?

In the case of using react-router, the server side and client side JS will be different (different entry points), and possibly some other build options. In some cases, one can use the same client side bundle, and thus that will be hashed in production, and the React on Rails code ideally will use the Webpacker function to read the manifest. If the manifest name is hard coded to be manifest.json, and the directory is set for the given ENV, then having 2 manifest files would require some API changes.

I'm not convinced we need 2 manifests as we've not come across cases where the server needs more than one bundle. While some have asked for it, nobody has presented any compelling reason. Per my my prior comment above, the React on Rails server code might be able to avoid taking the hit of an exception every time the server bundle is missing from the manifest in non-development by simply detecting this case only once, and caching the result. For development, we can probably afford the extra catch for every request.

@gauravtiwari and @dhh, I'm concerned about one case of turning off compilation and expecting watch mode to work. I can probably also look for the exception here.

React on Rails supports a custom test build command that's run. Would you be open to the compilation supporting a custom command in the `webpacker.yml` rather than the default rake task. Something like what we have in the `/config/initializers/react_on_rails.rb`:

```
  # If you are using the ReactOnRails::TestHelper.configure_rspec_to_compile_assets(config)
  # with rspec then this controls what yarn command is run
  # to automatically refresh your webpack assets on every test run.
  config.npm_build_test_command = "yarn run build:test"

  # This configures the script to run to build the production assets by webpack. Set this to nil
  # if you don't want react_on_rails building this file for you.
  config.npm_build_production_command = "yarn run build:production"
```

lib/webpacker/manifest.rb
matthewd: Forgive the drive-by comment, but framework code should not be directly checking the environment: we have a config subsystem for that. People often have a `production`-alike `staging` environment, for example.

(More locally-relevant, this should likely use the file watcher to avoid the stat [where possible] even in development.)

dhh:
@justin808 Let's try to see if the catch and cache case can work and take it from there 👍.

Re: Compilation, that's no longer going through the rake tasks, but directly via Webpacker.compiler.compile. That's only for on-demand compilation, though. If you're using the watcher or the dev server, you can just start that with whatever command you like. So that should be good 👍

dhh:
Going to close this PR now that the code is out of date and we've extracted all the major features and addressed the major concerns. Let's start smaller, feature-focused issues or PRs for whatever may be left. Thanks again for raising all these concerns.

justin808:
# 💥 [React on Rails v9](https://github.com/shakacode/react_on_rails) fully supports v3 of Webpacker. 

Big thanks to the rails/webpacker team for getting in the necessary changes. 

🙏 👏

PR #613, opened 2017-08-04
Windows-friendly webpack binstub invocation

### LINKED ISSUE #596: Invalid format: ./bin/webpack
On windows machines, it's impossible to run `webpacker:compile` task because by default windows treats `./bin/webpack` as a file without associated interpreting program. A workaround for this is to add `rails` before `./bin/webpack` at line 15 on `webpacker/compile.rake` file resulting in `rails ./bin/webpack`

msjonker:
Ensures that the call to the webpack binstub works in Windows by directly executing it from ruby.

Fixes #596

### DIFF
--- a/lib/tasks/webpacker/compile.rake
+++ b/lib/tasks/webpacker/compile.rake
@@ -1,6 +1,7 @@
 $stdout.sync = true
 
 require "open3"
+require "rbconfig"
 require "webpacker/env"
 require "webpacker/configuration"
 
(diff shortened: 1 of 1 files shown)


lib/tasks/webpacker/compile.rake
```diff
@@ -1,6 +1,7 @@
 $stdout.sync = true
 
 require "open3"
+require "rbconfig"
```
gauravtiwari: Hey, don't think we need to require `rbconfig` here. Does it not work on windows without this?

msjonker: You're right, it's not necessary. I have removed that line and it runs without issue. Created #617 to remove this line.

Thanks!

PR #622, opened 2017-08-08
[WIP] Enhance existing rake tasks in a separate file

himdel:
Right now, there's no way to prevent `webpacker:compile` and `webpacker:clobber` from calling `enhance` on `assets:precompile` and `assets:clobber`.

This splits off those `enhance` calls to a separate file, so that there's no change when all are loaded, but it is possible to skip loading `lib/tasks/webpacker/enhancements.rake` and not get those hooks.

### DIFF
--- a/lib/tasks/webpacker/clobber.rake
+++ b/lib/tasks/webpacker/clobber.rake
@@ -8,10 +8,3 @@ namespace :webpacker do
     $stdout.puts "Removed webpack output path directory #{output_path}"
   end
 end
-
-# Run clobber if the assets:clobber is run
-if Rake::Task.task_defined?("assets:clobber")
-  Rake::Task["assets:clobber"].enhance do
-    Rake::Task["webpacker:clobber"].invoke
-  end
-end
--- a/lib/tasks/webpacker/compile.rake
+++ b/lib/tasks/webpacker/compile.rake
@@ -25,14 +25,3 @@ namespace :webpacker do
     end
   end
 end
-
-# Compile packs after we've compiled all other assets during precompilation
-if Rake::Task.task_defined?("assets:precompile")
-  Rake::Task["assets:precompile"].enhance do
-    unless Rake::Task.task_defined?("yarn:install")
-      # For Rails < 5.1
-      Rake::Task["webpacker:yarn_install"].invoke
-    end
-    Rake::Task["webpacker:compile"].invoke
-  end
-end
--- a/lib/tasks/webpacker/enhancements.rake
+++ b/lib/tasks/webpacker/enhancements.rake
@@ -1,0 +1,17 @@
+# Compile packs after we've compiled all other assets during precompilation
+if Rake::Task.task_defined?("assets:precompile")
+  Rake::Task["assets:precompile"].enhance do
+    unless Rake::Task.task_defined?("yarn:install")
+      # For Rails < 5.1
+      Rake::Task["webpacker:yarn_install"].invoke
+    end
+    Rake::Task["webpacker:compile"].invoke
+  end
+end
+
+# Run clobber if the assets:clobber is run
+if Rake::Task.task_defined?("assets:clobber")
+  Rake::Task["assets:clobber"].enhance do
+    Rake::Task["webpacker:clobber"].invoke
+  end
+end


himdel:
Actually.. this doesn't work because all those files get autoloaded anyway...

himdel:
@guilleiguaran Any idea how to go on about making this optional?

(we're wrapping the task to support engines, and enhancing precompile with *that*, but there's no way to not include these files without having to run rake from rake to run this in the right engine)

dhh:
Why do we need to prevent webpacker's compile/clobber tasks from running together with the asset ones? Seems like that's what you'd want to integrate with Capistrano, Heroku, and any other integration that relies on precompiling.

dhh:
(Closing for now, since this approach didn't work regardless of rationale anyway)

himdel:
@dhh That's because the tasks only work from a directory which has `bin/webpack` under it :(.

We are using webpacker in not quite a supported way - with Rails engines:

* ManageIQ/manageiq is the app root, which is moslty models now
* ManageIQ/manageiq-ui-classic is the UI, implemented as a Rails engine
* there are/will be other repos with UI plugins implemented as Rails engines.

Given that, it makes sense that ui-classic should contain all the webpacker related stuff (no UI code in non-UI repos).
But that means we need the `webpacker:compile` tasks to run in `manageiq-ui-classic/`, which is not `Rails.root`. If we run them from the app root, webpacker fails because it can't find `manageiq/bin/webpack`.

Which made us wrap the webpacker tasks with essentially 

```
    task :compile do
      Dir.chdir ManageIQ::UI::Classic::Engine.root do
        Rake::Task["webpacker:compile"].invoke
      end
```

To make it work, and we're OK with that.

**But**: since `webpacker:compile` registers itself in `asset:precompile`, which isn't (and shouldn't be) wrapped in such a thing (and runs in the app root), suddenly `asset:precompile` fails whenever you have Webpacker included :).

So.. we need some way of making `asset:precompile` run the wrapped task which does do the chdir.
(Or, alternatively we need a binstub in the rails root, but then the binstub needs a way of resolving `ManageIQ::UI::Classic::Engine.root` to find the npm packages, and that seems impossible :(.)

himdel:
We ended up using this hack instead, but love a hackless solution :).

```
# compile and clobber when running assets:* tasks
if Rake::Task.task_defined?("assets:precompile")
  Rake::Task["assets:precompile"].enhance do
    Rake::Task["webpack:compile"].invoke
  end

  Rake::Task["assets:precompile"].actions.each do |action|
    if action.source_location[0].include?(File.join("lib", "tasks", "webpacker"))
      Rake::Task["assets:precompile"].actions.delete(action)
    end
  end
end
```

dhh:
Okay, I'm not a 100% I understand the use case, but we did just refactor all the rake tasks such that they're simply calling model classes instead. So it's not just Webpacker.compile, Webpacker.clobber, etc. And they should work regardless of which directory they're call from since they use Rails.root as the offset.

himdel:
@dhh Looking at the current state, I think this suffers the same couple issues as before..

1. Have a Rails app in `~/root/`.
1. Have a Rails engine in `~/engine/`, make your app use that engine.
1. If you go to `~/engine` and run `rake webpack:compile`, it will try running `~/engine/bin/webpack`.
1. If you go to `~/root` and run `rake webpack:compile`, it will try running `~/root/bin/webpack`.
1. If you run the app and Webpacker sees missing packs (and has `compile: true`), it will try running `~/root/bin/webpack`.

The problem we have is that:
  * we need points 3-5 to run in the same directory
  * we need that directory configurable, specifically, we need it to be an engine root instead of rails root (as an aside, would be nice it configurability was easier than [overriding Webpacker::Env and Webpacker::Configuration methods](https://github.com/ManageIQ/manageiq-ui-classic/pull/1132/files#diff-5125c00f46b3b9d290387f6820827e1e))

justin808:
https://github.com/rails/webpacker/pull/656 addresses this.

himdel:
@justin808 It really doesn't. It seems it touches on point 5, but that's it.

The basic problem is this: Rake tasks run from an engine's repo have their CWD as that engine's root.

So if you still run `bin/webpack` without chdir, you don't address this.


But I feel this is probably a discussion that should happen in an issue related to engines, right? :)

PR #625, opened 2017-08-08
Adds Webpack.logger, defaults to STDOUT

rossta:
Rake tasks currently output messages to stdout/stderr. It would be nice to have the ability to configure this. For example, Webpacker spits out its compilation messages when running Rails integration tests with the lazy compile option which adds noise to the test progress output. 

This PR is one take on providing the ability to customize output when running rake tasks. It changes behavior in that messages formerly output to stderr are now logged to stdout with log level error by default.


dhh:
Yes please. Let's set the default Webpacker logger to the Rails.logger even.

lib/webpacker/railtie.rb
```diff
@@ -12,6 +12,7 @@ class Webpacker::Engine < ::Rails::Engine
       include Webpacker::Helper
     end
 
+    Webpacker.logger = ::Rails.logger
```
dhh: I think you have to set this in a config.after_initialize style, like we do with Active Storage: https://github.com/rails/rails/blob/master/activestorage/lib/active_storage/engine.rb#L16

rossta: Got it, thanks!

lib/webpacker/railtie.rb
```diff
@@ -15,4 +15,10 @@ class Webpacker::Engine < ::Rails::Engine
     Webpacker.bootstrap
     Spring.after_fork { Webpacker.bootstrap } if defined?(Spring)
   end
+
+  initializer "webpacker.logger" do
+    config.after_initialize do |app|
+      Webpacker.logger = ::Rails.logger
```
georgeclaghorn: Won’t this prevent setting `Webpack.logger` in a Rails initializer?

lib/webpacker.rb
```diff
@@ -22,3 +22,13 @@ def env
 require "webpacker/manifest"
 require "webpacker/compiler"
 require "webpacker/railtie" if defined?(Rails)
+
+module Webpacker
+  def self.logger
+    @@logger ||= ActiveSupport::TaggedLogging.new(ActiveSupport::Logger.new(STDOUT))
```
rafaelfranca: You need to require those files

lib/webpacker/railtie.rb
rossta: If simply using `Webpacker.logger = ...` in an initializer, you're right I don't think it would work since the Rails initializer will execute before this `config.after_initialize` block. But, I believe it would be possible to wrap the Rails initializer call in a separate `config.after_initialize` block. 

As indicated in an [earlier comment](https://github.com/rails/webpacker/pull/625#discussion_r132017784), this approach appears to be used elsewhere in the Rails, e.g., https://github.com/rails/rails/blob/master/activestorage/lib/active_storage/engine.rb#L16.

georgeclaghorn: Cool. It doesn’t seem like something people are likely to override.

lib/webpacker.rb
```diff
@@ -15,6 +18,14 @@ def compile
   def env
     Webpacker::Env.current
   end
+
+  def logger
+    @@logger ||= ActiveSupport::TaggedLogging.new(ActiveSupport::Logger.new(STDOUT))
+  end
+
+  def logger=(logger)
```
gauravtiwari: @rossta May be use `mattr_accessor` instead - guess it's same thing

```rb
mattr_accessor(:logger) { ActiveSupport::TaggedLogging.new(ActiveSupport::Logger.new(STDOUT)) }
```

lib/webpacker/railtie.rb
rossta: I'd suppose this errs on the side of making sure the library's logger stays in sync with `Rails.logger` since it's more likely that `Rails.logger` would get overridden in app configuration.

dhh:
So we're working on moving the actual compilation step out of the rake task and into the compiler here: https://github.com/rails/webpacker/pull/627. When that's done, I don't actually think that the rake tasks should log to the Rails logger over STDOUT. In fact, it probably needs to explicitly set the Webpack.logger to use STDOUT rather than the log.

What we don't want is all this debug info when the compiler is triggered by tests or dev, but we DO want it when you're running the rake commands from the console.

rossta:
@dhh Sounds like we'll want a convention to set the logger appropriately in the rake tasks and possibly restore the previous logger when finished.

dhh:
Implemented that approach on https://github.com/rails/webpacker/commit/e539b4903c3fc24fd9b732e6253972acd5b50060.

dhh:
I think we got what we need from #627. Thanks for your help on this!

PR #627, opened 2017-08-08
Move actual compilation to the Compiler and call that from the rake task

dhh:
Before this on-demand compilation would exit the process if it failed. Doesn’t work for on-demand compilation in development.


javan:
❤️

PR #629, opened 2017-08-09
Enable babel-loader cache

javan:
> You can also speed up babel-loader by as much as 2x by using the cacheDirectory option. This will cache transformations to the filesystem.

– https://github.com/babel/babel-loader#babel-loader-is-slow

Shaves ~1s off Basecamp's compile time.

### DIFF
--- a/lib/install/config/loaders/core/babel.js
+++ b/lib/install/config/loaders/core/babel.js
@@ -1,5 +1,11 @@
+const { join } = require('path')
+const { settings } = require('../configuration.js')
+
 module.exports = {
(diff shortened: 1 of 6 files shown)


renchap:
Webpack's `cache-loader` (https://webpack.js.org/loaders/cache-loader/) could also help, and not only on the Babel step. Have you considered including it in the default setup ?

javan:
TIL! These changes pave the way for more caching by sharing the `cache_path`. Let's start with `babel-loader` since it provides a significant boost and continue to explore additional options.

PR #633, opened 2017-08-11
Allow blank public_output_path

alexchee:
If `public_output_path` is an empty string `public_output_path: ''`, all the packed assets would be prefixed with an extra slash (ex. `https://HOST//service-worker.js`. 

This change would allow public_output_path to be at the root of the host, by applying the outer slash removal on final formatPublicPath to remove the extra slash.

I was running into an issue trying to hook a service worker into the root path of the site to work on other URLs within the site. This might be useful for others doing the same.


dhh:
Alex, thanks for taking a stab at this. We're aware that service worker support is b0rked right now, but I think we'll want a broader solution where the packs still stay below that prefix. We should have something ready shortly for you to have a look at!

PR #636, opened 2017-08-11
Move to a single singleton

dhh:
Having a million different singletons all referring to each other was getting too much. Now there’s a single root singleton and just normal classes used through composition instead.


lib/install/template.rb
```diff
@@ -12,7 +12,7 @@
 copy_file "#{__dir__}/config/.babelrc", ".babelrc"
 
 puts "Creating javascript app source directory"
-directory "#{__dir__}/javascript", "#{Webpacker::Configuration.source}"
+directory "#{__dir__}/javascript", "app/javascript"
```
gauravtiwari: If someone re-runs `webpacker:install` this will re-create `app/javascript`if the source path is not the default we ship

dhh: Ah, I see. We can revert that then.

Issue #612, opened 2017-08-03
Issue with https


Radzhab:
ps. sorry for my english.

I have site that work over https. I add changes to webpacker to work with ssl. 

![image](https://user-images.githubusercontent.com/2556411/28912600-91b6229c-783d-11e7-9bca-d0dc57dbd2ac.png)

**development.js**
```
// Note: You must restart bin/webpack-dev-server for changes to take effect

const merge = require('webpack-merge')
const sharedConfig = require('./shared.js')
const { settings, output } = require('./configuration.js')

module.exports = merge(sharedConfig, {
  devtool: 'cheap-eval-source-map',

  output: {
    pathinfo: true
  },

  devServer: {
    https: true,
    clientLogLevel: 'none',
    host: settings.dev_server.host,
    port: settings.dev_server.port,
    contentBase: output.path,
    publicPath: output.publicPath,
    compress: true,
    headers: { 'Access-Control-Allow-Origin': '*' },
    historyApiFallback: true,
    watchOptions: {
      ignored: /node_modules/
    },
    stats: {
      errorDetails: true
    }
  }
})

```

But its raise error in browser. Its download applictions js over http? Why so?
`Mixed Content: The page at 'https://localhost:3000/' was loaded over HTTPS, but requested an insecure script 'http://localhost:8080/packs/application.js'. This request has been blocked; the content must be served over HTTPS.`

server puma.

rossta:
You'll need to create a self-signed SSL certificate for `localhost` to use with `webpack-dev-server`. There are lots of posts on the web about how to do this and for how to get some/most browsers to trust the certificate.

To make your webpack dev server config aware of the certificate, I'm using something similar to the config here:

```javascript
// config/webpack/local-ssl.js
const { resolve  } = require('path');
const { readFileSync } = require('fs');

module.exports = {
  cert: readFileSync(resolve('path', 'to', 'localhost-self-signed.pem'), 'utf8')
};
```
```javascript
// config/webpack/development.js
const { cert } = require('./local-ssl.js')
const https = { key: cert, cert: cert }

module.exports = merge(sharedConfig, {
  // ...
  devServer: {
    // ...
    https: https,
  },
//...
```
I may open a PR to add this to the README, but in the meantime, I hope that's helpful.

gauravtiwari:
thanks @rossta  🍰

Issue #550, opened 2017-07-05
Development JS often cached


terracatta:
Related to #267

I am using Webpacker 2.0 extensively in our Rails app with the default generated `development.js` config options.

Many times, my browser (Chrome 59 running on macOS) will cache the contents of the js files produced by webpacker in development (seems random, but I'm sure there are a reproducible set of steps where you could achieve the bad caching behavior).

Upon inspection I see the following webpacker behavior (which I will contrast to assets served by the Rails asset pipeline):

1. Webpacker compiled assets do not contain hashes in development (assets compiled by the asset pipeline contain a hash: `product.self-67a415712cf8c824bc5cf9518a7d2ca725ff07a88d9b7ae73b43c426275c5444.js?body=1`)

1. Webpacker assets are served by my webserver (Using vanillia Puma in dev) without _any_ cache control settings (assets compiled by the asset pipeline are served with `Cache-Control:public, max-age=31536000`)

Someone proposed adding hashes to the dev config in PR #271 but DHH closed it because he thought we should be using cache-control like the Rails asset-pipeline does. But upon actual inspection, it looks like Rails asset pipeline actually hashes the files and asks for them to be cached by the browser.

I think the current Rails Asset Pipeline hashing and caching strategy is sound, robust, and should be ported over to webpacker.

gauravtiwari:
@terracatta Are you using `./bin/webpack-dev-server` binstub or `./bin/webpack`? Does the issue persist with both?

terracatta:
> Are you using ./bin/webpack-dev-server binstub or ./bin/webpack? Does the issue persist with both?

I'm using `./bin/webpack` but we used to use the dev server and I believe the issue persisted there but I can't comment on the cache-control settings on the dev-server.

richardvenneman:
The absence of the `Cache-Control` header actually has caused us some trouble in production as well as CloudFront seems to evict the javascript pack file prematurely (we [share a CDN between staging/production](https://devcenter.heroku.com/articles/please-do-not-use-asset-sync#pipelines)).

gauravtiwari:
Fixed in #653

richardvenneman:
@gauravtiwari while #653 fixes 1, the `Cache-Control` header is still absent (2)

gauravtiwari:
@richardvenneman Right I missed (2). Make sense to add

gauravtiwari:
See #666 for context

Issue #319, opened 2017-05-03
Page requested an insecure script


bigtunacan:
I created just a basic app to test using webpacker with Vue and added the entry ` <%= javascript_pack_tag 'hello_vue' %>` to the application layout file.  

My view loads, but the request for the Vue pack is blocked with the following error message.

>  Content: 
>
>The page at 'https:/myproj-bigtunacan.c9users.io/wat/home' was loaded over HTTPS, but requested an insecure script 'http://localhost:8080/packs/hello_vue.js'. This request has been blocked; the content must be served over HTTPS.

gauravtiwari:
@bigtunacan Just setup https option in `webpack/development.server.js` - https://webpack.js.org/configuration/dev-server/#devserver-https

bigtunacan:
@gauravtiwari This is maybe just due to my unfamiliarity with Webpack, but I'm still getting the same error.

This is what my `development.server.js` file now looks like.

    // Note: You must restart bin/webpack-dev-server for changes to take effect
    
    const { resolve } = require('path')
    const merge = require('webpack-merge')
    const devConfig = require('./development.js')
    const { devServer, publicPath, paths } = require('./configuration.js')
    
    module.exports = merge(devConfig, {
      devServer: {
        host: devServer.host,
        port: devServer.port,
        compress: true,
        historyApiFallback: true,
        contentBase: resolve(paths.output, paths.entry),
        publicPath,
        https: true
      }
    })
    
I reran `rake webpacker:compile` and restarted my Rails server, but no change.

I notice at the top of the `development.server.js` file it says you need to restart bin/webpack-dev-server, but I don't see how to even do that?

gauravtiwari:
Oh yes, my bad the protocol is hard-coded so that needs to be fixed. Will make a PR

martyphee:
I changed this line in `configuration.js` to get moving
`const devServerUrl = `https://${devServer.host}:${devServer.port}/${paths.entry}/``

In order for this to work I had to try opening the file through the browser and accept the security alert.  Is there a better way?

gauravtiwari:
@martyphee Is origin also running on https? Also, you would need to assign a self signed certificate - https://github.com/gagle/node-ssl-self-signed-certificate

gauravtiwari:
I think dev-server should take care of this, but not sure how it works for the browser - https://webpack.js.org/configuration/dev-server/#devserver-https

martyphee:
[...]

PR #637, opened 2017-08-12
Add dev server config class and setup proxy

gauravtiwari:
This PR adds a few things related to webpack-dev-server: 

1. A dev server config class that exposes config options
```rb
Webpacker.dev_server.running?
```
2. A rack middleware that proxies webpacker requests to dev server so we can always serve from same-origin and the lookup works out of the box - no more paths prefixing 

```rb
   if Rails::VERSION::MAJOR >= 5
      app.middleware.insert_before 0, Webpacker::DevServerProxy, ssl_verify_none: true
    else
      config.middleware.insert_before 0, "Webpacker::DevServerProxy", ssl_verify_none: true
    end
```

screen shot 2017-08-12 at 12 16 06

3. HTTPS ssl settings for webpack-dev-server

```yml
# Absolute paths to ssl key and certificate
    ssl_key_path:
    ssl_cert_path:
```

4. Move dev-server-settings to defaults so it's transparently available in all environments

5. The host info has been removed from manifest.json, now looks like this:

```json
{
  "hello_react.js": "/packs/hello_react.js"
}
```

### DIFF
--- a/lib/install/config/webpack/configuration.js
+++ b/lib/install/config/webpack/configuration.js
@@ -24,8 +24,8 @@ function formatPublicPath(host = '', path = '') {
 
 const output = {
   path: resolve('public', settings.public_output_path),
-  publicPath: formatPublicPath(env.ASSET_HOST, settings.public_output_path),
-  dirPath: formatPublicPath(`/${settings.public_output_path}/`)
+  publicPath: `/${settings.public_output_path}/`.replace(/([^:]\/)\/+/g, '$1'),
+  publicPathWithHost: formatPublicPath(env.ASSET_HOST, settings.public_output_path)
 }
 
 let resolvedModules = [
--- a/lib/install/config/webpack/configuration.js
+++ b/lib/install/config/webpack/configuration.js
@@ -24,7 +24,8 @@ function formatPublicPath(host = '', path = '') {
 
 const output = {
   path: resolve('public', settings.public_output_path),
-  publicPath: formatPublicPath(env.ASSET_HOST, settings.public_output_path)
+  publicPath: formatPublicPath(env.ASSET_HOST, settings.public_output_path),
+  dirPath: formatPublicPath(`/${settings.public_output_path}/`)
 }
 
 let resolvedModules = [
--- a/lib/install/config/webpack/development.js
+++ b/lib/install/config/webpack/development.js
@@ -5,17 +5,6 @@ const { readFileSync } = require('fs')
 const sharedConfig = require('./shared.js')
 const { settings, output } = require('./configuration.js')
 
-const { dev_server: devServer } = settings
-let https = devServer.https
-
-// Set ssl key and certificates if supplied
-if (devServer.ssl_key_path && devServer.ssl_cert_path) {
-  https = {
-    key: readFileSync(devServer.ssl_key_path),
-    cert: readFileSync(devServer.ssl_cert_path)
-  }
-}
-
 module.exports = merge(sharedConfig, {
   devtool: 'cheap-eval-source-map',
 
@@ -25,9 +14,8 @@ module.exports = merge(sharedConfig, {
 
   devServer: {
     clientLogLevel: 'none',
-    https,
-    host: devServer.host,
-    port: devServer.port,
+    host: settings.dev_server.host,
+    port: settings.dev_server.port,
     contentBase: output.path,
     publicPath: output.publicPath,
     compress: true,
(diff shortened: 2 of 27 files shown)


gauravtiwari:
There is one extra bit of change here i.e. renaming `packs` to `packs-test` inside tests - apparently CI is setting `RAILS_ENV` when running tests and due to output folder being different in `webpacker.yml` the tests were failing however when running same tests locally the app was booted in development environment. 

So I have setup the RAILS.env to be test to avoid any discrepancy:

```rb
module TestApp
  class Application < ::Rails::Application
    config.root = File.join(File.dirname(__FILE__), "test_app")
    config.eager_load = true
    ::Rails.env = ENV["RAILS_ENV"] = "test"
  end
end
```

and renamed all folders from `packs` to `packs-test`.

lib/install/config/webpack/development.js
```diff
@@ -1,9 +1,21 @@
...
+const { dev_server: devServer } = settings
+let https = devServer.https
+
+// Set ssl key and certificates if supplied
+if (devServer.ssl_key_path && devServer.ssl_cert_path) {
+  https = {
+    key: readFileSync(devServer.ssl_key_path),
+    cert: readFileSync(devServer.ssl_cert_path)
+  }
+}
```
dhh: When do we ever need to bother with HTTPS in development?

gauravtiwari: @dhh There were some issues filed where people are using HTTPS for development, mainly for cases where they are interacting with API's that is hosted on HTTPS

dhh:
Links?

gauravtiwari:
https://github.com/rails/webpacker/issues/612 and https://github.com/rails/webpacker/issues/319

gauravtiwari:
@dhh Should we move the proxy middleware to middlewares folder, guess we gonna have one more middleware for `cache-control`.

dhh:
What do we need for cache-control?

gauravtiwari:
For expiring cached assets: https://github.com/rails/webpacker/issues/267 and https://github.com/rails/webpacker/issues/550

dhh:
Ah yes. Let's do a separate PR for that.

gauravtiwari:
Yepp 👍

dhh:
Actually, if we just use -hash in dev as well, we shouldn't need to explicitly set any headers? But not sure how that'll combine with the dev server.

gauravtiwari:
Yeah that will work 👍  I doubt this is a problem when using dev-server because it takes care of settings cache-control. I guess this is happening mainly with watcher, where assets are being served by Rails server and getting cached but -hash can totally fix it

dhh:
For SSL, we should double down on the reproxy setup. The dev server should not need to know about this. Just like people use nginx to terminate SSL before talking to a puma server.

lib/install/config/webpack/configuration.js
```diff
@@ -24,7 +24,8 @@ function formatPublicPath(host = '', path = '') {
 
 const output = {
   path: resolve('public', settings.public_output_path),
-  publicPath: formatPublicPath(env.ASSET_HOST, settings.public_output_path)
+  publicPath: `/${settings.public_output_path}/`.replace(/([^:]\/)\/+/g, '$1'),
```
justin808: On line 32 below, can you confirm that you're using the adding of the source_path to make non-JS asset inclusion easier? Or are there other reasons?

gauravtiwari: @justin808 I am going to push some cleanup for this whole webpack part in another PR. Basically, the idea was to not end up with `//` if `public_output_path` is set to `''` i.e. the assets will be outputted at root. Since we are now using proxy, there is no host for `publicPath` except for internal assets like fonts or images, that's what `publicPathWithHost` is used for. 

> On line 32 below, can you confirm that you're using the adding of the source_path to make non-JS asset inclusion easier? Or are there other reasons?

Yes, that's just adding source path to webpack module resolver.

lib/webpacker/dev_server_proxy.rb
```diff
@@ -0,0 +1,23 @@
+require "rack/proxy"
+
+class Webpacker::DevServerProxy < Rack::Proxy
+  def rewrite_response(response)
+    status, headers, body = response
+    headers.delete "transfer-encoding"
```
sfcgeorge: Hi @gauravtiwari do you by any chance remember why you needed to delete the Transfer-Encoding header? Maybe something to do with preventing Gzipping?

Doing so breaks chunked streaming as per https://github.com/rails/webpacker/pull/2196#pullrequestreview-268686577

Currently my PR conditionally deletes the header under the assumption that in some cases it does need deleting, but as suggested it would be even better to get rid of the line if not needed :)

PR #638, opened 2017-08-12
Speed up `bin/webpack*` executables by dropping `yarn run` overhead

javan:
We were already doing this on Windows only (https://github.com/rails/webpacker/pull/584). Just do it.

Before
```sh
$ time bin/webpack
yarn run v0.27.5
$ "/Users/javan/Desktop/myapp/node_modules/.bin/webpack" "--config" "/Users/javan/Desktop/myapp/config/webpack/development.js"
Hash: 9a50a6710c2068693890
Version: webpack 3.5.4
Time: 1005ms
         Asset     Size  Chunks                    Chunk Names
hello_react.js  2.05 MB       0  [emitted]  [big]  hello_react
  [82] ./app/javascript/packs/hello_react.jsx 759 bytes {0} [built]
    + 184 hidden modules
Done in 1.77s.

real	0m2.309s
user	0m2.295s
sys	0m0.210s
```

After
```sh
$ time bin/webpack
Hash: 9a50a6710c2068693890
Version: webpack 3.5.4
Time: 958ms
         Asset     Size  Chunks                    Chunk Names
hello_react.js  2.05 MB       0  [emitted]  [big]  hello_react
  [82] ./app/javascript/packs/hello_react.jsx 759 bytes {0} [built]
    + 184 hidden modules

real	0m1.803s
user	0m1.827s
sys	0m0.169s
```

### DIFF
--- a/lib/install/bin/webpack-dev-server.tt
+++ b/lib/install/bin/webpack-dev-server.tt
@@ -51,20 +51,19 @@ end
   ARGV.delete(arg)
 end
 
-newenv = {
+env = {
   "NODE_PATH" => NODE_MODULES_PATH.shellescape,
   "ASSET_HOST" => DEV_SERVER_ADDR.shellescape
-}.freeze
+}
 
-options = ["--progress", "--color", "--config", WEBPACK_CONFIG, "--host", LISTEN_HOST_ADDR, "--public", "#{HOSTNAME}:#{PORT}", "--port", PORT.to_s]
-
-if Gem.win_platform?
-  # Workaround for yarn not playing nicely with path separators
-  cmdline = ["#{NODE_MODULES_PATH}/.bin/webpack-dev-server", *options] + ARGV
-else
-  cmdline = ["yarn", "run", "webpack-dev-server", "--", *options] + ARGV
-end
+cmd = [
+  "#{NODE_MODULES_PATH}/.bin/webpack-dev-server", "--progress", "--color",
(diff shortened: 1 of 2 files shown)


dhh:
I thought it wasn't safe to go straight to NODE_MODULES/.bin by yourself?

gauravtiwari:
@dhh Should be alright however sometimes binaries might get corrupted specially after yarn install (https://github.com/yarnpkg/yarn/issues/1981), probably fixed now. Another thing `yarn run` does is multiple lookups before bailing out but we are already using the binaries directly on windows so may be do the same for others. 

@javan Have you come across this problem when using binaries directly? https://github.com/rails/webpacker/pull/364, sometimes the process isn't get killed

javan:
Executing `node_modules/.bin/*` is safe and is what `yarn run` eventually does in our case. `yarn run` can do additional things like execute `"scripts"` defined in `package.json`, but we don't need any of that.

@gauravtiwari, I can't reproduce #364:
> Run `foreman start`
>
> Start a new terminal and run `foreman stop`, and `ps aux | grep webpacker` will sometimes show the child process continue in the background.

`foreman stop`  doesn't appear to even be a valid foreman command and stopping foreman "normally" with ctrl+c works as expected.

dhh:
We don't need foreman anymore now that we're proxing to the dev server.

gauravtiwari:
@javan Yeah with foreman this isn't a problem but I guess it's happening when we run it directly.

PR #640, opened 2017-08-12
Add lookup path for looking up pack paths for server rendering

gauravtiwari:
@dhh We accidentally removed this method yesterday, this is being used here for server rendering packs - https://github.com/reactjs/react-rails/blob/f3f9576345dae018dd7d7e96466824d78b19919d/lib/react/server_rendering/webpacker_manifest_container.rb#L24

### DIFF
--- a/lib/webpacker/manifest.rb
+++ b/lib/webpacker/manifest.rb
@@ -27,6 +27,10 @@ class Webpacker::Manifest
     find name
   end
 
+  def lookup_path(name)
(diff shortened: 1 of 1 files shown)


dhh:
If this is not something used inside of Webpacker, I don't see why whatever plugin needs to use it can't just implement that themselves? All the information is exposed (public output path + asset pack path)?

gauravtiwari:
Yepp makes sense 👍 

cc:// @rmosolgo

PR #641, opened 2017-08-12
Add HMR to the dev server

dhh:
Extracted from https://github.com/rails/webpacker/pull/601/

### DIFF
--- a/lib/install/config/webpack/development.js
+++ b/lib/install/config/webpack/development.js
@@ -15,8 +15,8 @@ module.exports = merge(sharedConfig, {
   devServer: {
     clientLogLevel: 'none',
     host: settings.dev_server.host,
-    port: settings.dev_server.port
-    hmr: settings.dev_server.hmr,
+    port: settings.dev_server.port,
+    hot: settings.dev_server.hmr,
     contentBase: output.path,
     publicPath: output.publicPath,
     compress: true,
--- a/lib/install/config/webpack/development.js
+++ b/lib/install/config/webpack/development.js
@@ -1,5 +1,6 @@
 // Note: You must restart bin/webpack-dev-server for changes to take effect
 
+const webpack = require('webpack')
 const merge = require('webpack-merge')
 const sharedConfig = require('./shared.js')
 const { settings, output } = require('./configuration.js')
@@ -14,7 +15,8 @@ module.exports = merge(sharedConfig, {
   devServer: {
     clientLogLevel: 'none',
     host: settings.dev_server.host,
-    port: settings.dev_server.port,
+    port: settings.dev_server.port
+    hmr: settings.dev_server.hmr,
     contentBase: output.path,
     publicPath: output.publicPath,
     compress: true,
@@ -26,5 +28,10 @@ module.exports = merge(sharedConfig, {
     stats: {
       errorDetails: true
     }
-  }
+  },
+
+  plugins: settings.dev_server.hmr ? [
+    new webpack.HotModuleReplacementPlugin(),
+    new webpack.NamedModulesPlugin()
+  ] : []
 })
--- a/lib/webpacker/helper.rb
+++ b/lib/webpacker/helper.rb
@@ -41,6 +41,10 @@ module Webpacker::Helper
   #   <%= stylesheet_pack_tag 'calendar', 'data-turbolinks-track': 'reload' %> # =>
   #   <link rel="stylesheet" media="screen" href="/packs/calendar.css" data-turbolinks-track="reload" />
   #
+  #   # In development mode with hot module replacement:
+  #   <%= stylesheet_pack_tag 'calendar', 'data-turbolinks-track': 'reload' %> # =>
+  #   nil
+  #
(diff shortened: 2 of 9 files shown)


lib/install/config/webpack/development.js
```diff
@@ -14,7 +15,8 @@ module.exports = merge(sharedConfig, {
   devServer: {
     clientLogLevel: 'none',
     host: settings.dev_server.host,
-    port: settings.dev_server.port,
+    port: settings.dev_server.port
```
gauravtiwari: Missing trailing comma ;)

lib/install/config/webpack/development.js
```diff
@@ -14,7 +15,8 @@ module.exports = merge(sharedConfig, {
...
+    hmr: settings.dev_server.hmr,
```
gauravtiwari: Option name is `hot: settings.dev_server.hmr`

gauravtiwari:
I will push some updates for sass.js since the styles won't be extracted when HMR is on 👍  and then we should be good to go

lib/webpacker/helper.rb
```diff
@@ -42,7 +45,9 @@ def javascript_pack_tag(*names, **options)
   #   <%= stylesheet_pack_tag 'calendar', 'data-turbolinks-track': 'reload' %> # =>
   #   <link rel="stylesheet" media="screen" href="/packs/calendar-1016838bab065ae1e122.css" data-turbolinks-track="reload" />
   def stylesheet_pack_tag(*names, **options)
-    stylesheet_link_tag(*sources_from_pack_manifest(names, type: :stylesheet), **options)
+    unless Webpacker.dev_server.running? && Webpacker.dev_server.hot_module_replacing?
```
justin808: This is not correct in that React on Rails users will build static assets for test runs while the dev server is possibly running for development. Thus, we'll skip this tag when we need it for tests.

lib/webpacker/helper.rb
```diff
@@ -42,7 +45,9 @@ def javascript_pack_tag(*names, **options)
   #   <%= stylesheet_pack_tag 'calendar', 'data-turbolinks-track': 'reload' %> # =>
   #   <link rel="stylesheet" media="screen" href="/packs/calendar-1016838bab065ae1e122.css" data-turbolinks-track="reload" />
   def stylesheet_pack_tag(*names, **options)
-    stylesheet_link_tag(*sources_from_pack_manifest(names, type: :stylesheet), **options)
```
justin808: We need to update the above docs to indicate what happens in HMR, as well as for the possibility that the webpack-dev-server is not running for development mode.

lib/webpacker/helper.rb
gauravtiwari: @justin808 What are the implications? I guess dev server can be used too for serving assets in test environment and don't think that will affect any behaviour. Obviously, if someone wants to statically build and test they can turn off the dev-server and use watcher or on-demand compilation. With dev-server proxy usage is more transparent and consistent so the output will be same in views regardless of compiler used.

lib/webpacker/helper.rb
gauravtiwari: Yes, added some notes in the README. Anything missing?

lib/webpacker/helper.rb
dhh: I fixed it such that Webpacker.dev_server.running? will return false in testing by moving the dev_server configuration to the development env only in webpacker.yml. So now the tag will appear in test mode.

lib/webpacker/helper.rb
justin808: The doc for the method should be updated (or removed? with a reference to the readme.)

lib/webpacker/helper.rb
justin808: > What are the implications?
Since we have a different test setup in webpack for test compared to dev (we want more logging in dev), I'd feel uncomfortable that some of the setup is coming from a different webpack config.

lib/webpacker/helper.rb
```diff
@@ -32,6 +32,9 @@ def javascript_pack_tag(*names, **options)
   # in config/webpack/shared.js. By default, this list is auto-generated to match everything in
   # app/javascript/packs/*.js. In production mode, the digested reference is automatically looked up.
   #
+  # Note: If the development server is running and hot module replacement is active, this will return nothing.
+  # In that setup you need to configure your styles to be inlined in your JavaScript for hot reloading.
+  #
```
gauravtiwari: @justin808 You mean this doc?

justin808: Maybe adjust this line to indicate more than "development mode":
```
  #   # In development mode:
  #   <%= stylesheet_pack_tag 'calendar', 'data-turbolinks-track': 'reload' %> # =>
  #   <link rel="stylesheet" media="screen" href="/packs/calendar.css" data-turbolinks-track="reload" />

gauravtiwari: Done 👍

Issue #642, opened 2017-08-13
Add new Troubleshooting Item


techgurupezza:
Recently I got an error that was very non descript when loading webpack with this configuration.

```bash
Invalid configuration object. Webpack has been initialised using a configuration object that does not match the API schema.
20:09:18 webpacker.1 |  - configuration.entry should be one of these:
20:09:18 webpacker.1 |    object { <key>: non-empty string | [non-empty string] } | non-empty string | [non-empty string] | function
20:09:18 webpacker.1 |    The entry point(s) of the compilation.
20:09:18 webpacker.1 |    Details:
20:09:18 webpacker.1 |     * configuration.entry should NOT have less than 1 properties ({
20:09:18 webpacker.1 |         "keyword": "minProperties",
20:09:18 webpacker.1 |         "dataPath": ".entry",
20:09:18 webpacker.1 |         "schemaPath": "#/oneOf/0/minProperties",
20:09:18 webpacker.1 |         "params": {
20:09:18 webpacker.1 |           "limit": 1
20:09:18 webpacker.1 |         },
20:09:18 webpacker.1 |         "message": "should NOT have less than 1 properties",
20:09:18 webpacker.1 |         "schema": 1,
20:09:18 webpacker.1 |         "parentSchema": {
20:09:18 webpacker.1 |           "minProperties": 1,
20:09:18 webpacker.1 |           "additionalProperties": {
20:09:18 webpacker.1 |             "oneOf": [
20:09:18 webpacker.1 |               {
20:09:18 webpacker.1 |                 "description": "The string is resolved to a module which is loaded upon startup.",
20:09:18 webpacker.1 |                 "minLength": 1,
20:09:18 webpacker.1 |                 "type": "string"
20:09:18 webpacker.1 |               },
20:09:18 webpacker.1 |               {
20:09:18 webpacker.1 |                 "description": "All modules are loaded upon startup. The last one is exported.",
20:09:18 webpacker.1 |                 "$ref": "#/definitions/common.nonEmptyArrayOfUniqueStringValues"
20:09:18 webpacker.1 |               }
20:09:18 webpacker.1 |             ]
20:09:18 webpacker.1 |           },
20:09:18 webpacker.1 |           "description": "Multiple entry bundles are created. The key is the chunk name. The value can be a string or an array.",
20:09:18 webpacker.1 |           "type": "object"
20:09:18 webpacker.1 |         },
20:09:18 webpacker.1 |         "data": {}
[...]

PR #643, opened 2017-08-13
#642 Add Troubleshooting Item for invalid config

techgurupezza:
Added Troubleshooting item for ticket #642 to help with debugging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,8 +251,8 @@ in a separate terminal from `./bin/rails server`. This process will watch for ch
 in the `app/javascript/packs/*.js` files and automatically reload the browser to match.
 
 Once you start this development server, Webpacker will automatically start proxying all
-webpack asset requests to this server. When you stop the server, it'll revert to 
-on-demand compilation again. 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1243,5 +1243,56 @@ C:\path>ruby bin\webpack
 C:\path>ruby bin\webpack-dev-server
 ```
 
+#### Invalid configuration object. Webpack has been initialised using a configuration object that does not match the API schema.
+
+If you receive this error ensure your configuration is correct most likely the path to your packs folder is incorrect if you modified from the original path defined.
+
+Example error from `$ ./bin/webpack-dev-server`
```
gauravtiwari: @techgurupezza Thanks for this - don't think we need to add whole error here - just the top bit

techgurupezza: Roger that @gauravtiwari I'll trim it down. I've removed the error and went very specific in the text to ensure no confusion.

gauravtiwari:
thanks 🍰  🎈

PR #646, opened 2017-08-13
WIP: Service workers

gauravtiwari:
* Adds ServiceWorker support using SWPrecache (https://github.com/GoogleChrome/sw-precache)

TODOS:
- Proper script for registering, unregistering workers and handling when cache changes.
- Passing full base url

*Note:* This doesn't work with dev-server in development env, either use watcher or on-demand compilation for testing


dhh:
I don't think this is going to cut it. I'd like to focus on a solution for service workers that'll work in all environments and situations supported by webpacker. One solution is to proxy root requests to /assets/, another would be to enable someone to have an output actually at the root.

PR #648, opened 2017-08-13
Move to a helper and fix tests

gauravtiwari:
Moves the reloaded_config method to a helper and use that for dev_server_test as well


javan:
How about creating a base test class instead?
```ruby
# webpacker_test_helper.rb

class Webpacker::Test < Minitest::Test
  private
    def reloaded_config
      Webpacker.instance.instance_variable_set(:@config, nil)
      Webpacker.config
    end

    def with_node_env(env)
      original = ENV["NODE_ENV"]
      ENV["NODE_ENV"] = env
      yield
    ensure
      ENV["NODE_ENV"] = original
    end
end
```

gauravtiwari:
@javan Even better 👍  💯  Pushed an update

PR #650, opened 2017-08-14
Issue 464 merge webpacker lite into webpacker v2

justin808:
This branch contains the HMR changes as well as the following commit.

This fix addresses the cases where compile is disabled:

1. If React on Rails depends on Webpacker, but the React on Rails
installer has not yet generated a webpacker.yml file or a React on Rails
installation is not using Webpacker, then the gem loading of Webpacker
will crash due to the missing config file.

2. If React on Rails has its own helper to compile assets for tests,
then the manifest might need to get reloaded after Webpacker statically
loads.

Relevant change:
https://github.com/rails/webpacker/pull/650/commits/7abcc2140f01a22d5a5dea444c1e7faefaa17c96


justin808:
The following changed in React on Rails: https://github.com/shakacode/react_on_rails/pull/908/files

You can find this as https://rubygems.org/gems/react_on_rails/versions/9.0.0.beta.7

justin808:
See #656

PR #652, opened 2017-08-15
Allow to use environments not present by default

doits:
This commit allows to define non standard environments. Before it failed
to load defaults for environments not existing in the default yaml
because it tried to do a `.deep_symbolize_keys` on `nil`. By using the
constructor form it returns an empty hash now.

---

I tried to write a simple test but it got more and more complicated (because custom and default config are the same in the tests, and one would need one custom config to test it in that case.)

### DIFF
--- a/lib/webpacker/configuration.rb
+++ b/lib/webpacker/configuration.rb
@@ -60,6 +60,6 @@ class Webpacker::Configuration
 
     def defaults
       @defaults ||= \
-        YAML.load(File.read(File.expand_path("../../install/config/webpacker.yml", __FILE__)))[env].deep_symbolize_keys
+        HashWithIndifferentAccess.new(YAML.load(File.read(File.expand_path("../../install/config/webpacker.yml", __FILE__)))[env])
     end
 end


doits:
Some more explanation where this hit me: I have a `staging` env in my `config/webpacker.yml` (same config as production):

```yml
# config/webpacker.yml

production: &production
  ...

staging:
  <<: *production
```

Additionally `config/webpack/staging.js` is a symlink to `config/webpack/production.js`. This way I can test deployment on my staging env under production settings.

Of course `staging` is not defined in `/install/config/webpacker.yml` and therefore `.with_indifferent_access` broke on `nil`.

doits:
rebased on master so tests are green now

gauravtiwari:
@doits Great 👍  Feel free to close

doits:
@gauravtiwari I just meant that this PR passes all tests now (because master was fixed), but this PR fixes my issue! So I'd be glad to have this merged (or fixed by other means) :-)

gauravtiwari:
@doits Ahh yes, I see. Could you also change the `load` method as well in the same file please? Since that could also produce the same error.

PR #653, opened 2017-08-15
Use hashed assets for all env

### LINKED ISSUE #644: Dev server proxy bypassed by static asset servers
When using a local app server like [puma-dev](https://github.com/puma/puma-dev) or [pow](https://github.com/basecamp/pow) that serves static assets, starting `bin/webpack-dev-server` will have no effect if a compiled .js file already exists. The preexisting files in `public/packs` will be served directly and the requests never hit Rails.

Steps to reproduce:
1. Load a page with webpacker's `compile: true` option enabled (the default)
2. Start `bin/webpack-dev-server`
3. Change your application's js in some way and reload the page

/cc @gauravtiwari @dhh

gauravtiwari:
Fixes #644 #550

This PR enables hashed assets for all environments so we can expire them when changed

### DIFF
--- a/lib/install/config/webpack/shared.js
+++ b/lib/install/config/webpack/shared.js
@@ -14,6 +14,7 @@ const { env, settings, output, loadersDir, resolvedModules } = require('./config
 const extensionGlob = `**/*{${settings.extensions.join(',')}}`
 const entryPath = join(settings.source_path, settings.source_entry_path)
 const packPaths = sync(join(entryPath, extensionGlob))
+const isHMR = settings.dev_server && settings.dev_server.hmr
 
 module.exports = {
   entry: packPaths.reduce(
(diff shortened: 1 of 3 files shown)


javan:
Can remove `output: { filename: '[name]-[chunkhash].js' }` in `production.js` too, right?

lib/install/config/webpack/shared.js
```diff
@@ -26,7 +26,7 @@ module.exports = {
   ),
 
   output: {
-    filename: '[name].js',
+    filename: '[name]-[hash].js',
```
javan: Should this be `[name]-[chunkhash].js`? That's what `production.js` uses.

gauravtiwari: Yeah but not available with dev server and watcher actually, just production build

justin808: @gauravtiwari do you have a doc link for that?

gauravtiwari: This is not properly explained in doc but basically chunkhash isn't available with HMR
screen shot 2017-08-16 at 09 56 25

PR #654, opened 2017-08-15
Clobber assets before recompilation

gauravtiwari:
Depends on #653 

This PR will ensure that we keep cleaning old hashed assets before recompiling again

### DIFF
--- a/lib/webpacker/compiler.rb
+++ b/lib/webpacker/compiler.rb
@@ -13,6 +13,7 @@ class Webpacker::Compiler
 
   def compile
     if stale?
+      Webpacker.clobber
       record_compilation_timestamp
       run_webpack
     else
--- a/lib/webpacker/compiler.rb
+++ b/lib/webpacker/compiler.rb
@@ -13,12 +13,15 @@ class Webpacker::Compiler
 
   def compile
     if stale?
-      Webpacker.clobber
+      clobber
       record_compilation_timestamp
       run_webpack
-    else
-      logger.debug "No compiling needed as everything is fresh"
     end
+  end
+
+  def clobber
+    config.public_output_path.rmtree if config.public_output_path.exist?
+    config.cache_path.rmtree if config.cache_path.exist?
   end
 
   # Returns true if all the compiled packs are up to date with the underlying asset files.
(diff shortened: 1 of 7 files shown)


lib/webpacker/compiler.rb
```diff
@@ -13,6 +13,7 @@ def initialize(webpacker)
 
   def compile
     if stale?
+      Webpacker.clobber
```
dhh: Seems like perhaps "clobber" should be a method on the compiler which is then called from Webpacker. This kinda of reaching is odd.

gauravtiwari:
@dhh Pushed some updates. Do you think we should delegate clobber from compiler to `Webpacker` instance? Or this looks good?

lib/webpacker/compiler.rb
```diff
@@ -13,13 +13,17 @@ def initialize(webpacker)
...
       run_webpack
-    else
-      logger.debug "No compiling needed as everything is fresh"
     end
   end
 
+  def clobber
+    config.public_output_path.rmtree if config.public_output_path.exist?
+    config.cache_path.rmtree if config.cache_path.exist?
+  end
```
javan: `clean` is a more common name for this.

matthewd: Sprockets uses 'clean' for "normal usage" cleanup -- discarding older versions of compiled assets, for example -- and 'clobber' for the nuclear "trash everything, even if reusing it "shouldn't" have any effect on the output".

Thus, clean is a thing one does when things are getting messy, and clobber is a thing one only really does when trying to debug a compilation problem.

At a quick glance, this might straddle those definitions, so I'll stop at providing that history/context, and won't venture a guess at which way this should go.

gauravtiwari: In this sense clobber is what we are doing here - nuclear trash old assets 😀

dhh: Actually, why are we automatically clobbering on every compile? I don't think Sprockets does that? In production, you may well want to keep older assets around for a bit such that you don't break in-flight requests during a deploy.

gauravtiwari: @dhh Not needed if we are using dev server but if we are using on-demand compilation the output folder can get pretty big after each compilation since we are now using assets hashed. Although, this doesn't solve the same problem when using watcher.

dhh:
I don't think we should be doing automatic clobbering. Let's have it as a
rake task instead.

gauravtiwari:
Yepp we already have a rake task `bundle exec rake webpacker:clobber`

gauravtiwari:
I think we can close this one then.

dhh:
Then let's nix this auto-clobbering business 👍

PR #655, opened 2017-08-15
Fix dev_server HMR reference and refresh after compile

### LINKED ISSUE #651: Packs not available on first load in test env
I've installed latest webpacker from master (b95ca0501df8e67784973fabf880ebe083a2e4e8) but have the following problem:

If test packs are removed (`public/packs-test`) and I start my rspec test, packs cannot be found: 

```
ActionView::Template::Error:
    Can't find .... Is webpack still compiling?
```

`public/packs-test` is generated on that test run though.

When I start the test a second time, now with existing `publick/packs-test` folder, it works without problem.

To me it looks like compiling happens async/in background and the first run does not catch the newly compiled packs. This breaks my CI completely, because there webpacker files are removed before each CI run and should be compiled automatically. I can workaround this be compiling them manually before the CI run but then I still have the problem on local test runs when some files are changed - I suspect it will use the old version once if a file changed then.

This issue was not present with webpacker 2b0972a0df158c1614c5c30e591c657813f1f904.

gauravtiwari:
Fixes #651

This PR fixes a bug where `dev_server` is only available in development env but the JS is trying to reference HMR without checking if dev_server object is available.

Also it moves refresh to compile call so that the manifest is refreshed after each compile and not just in development.

### DIFF
--- a/lib/webpacker/configuration.rb
+++ b/lib/webpacker/configuration.rb
@@ -37,8 +37,8 @@ class Webpacker::Configuration
     public_output_path.join("manifest.json")
   end
 
-  def reload_manifest?
-    fetch(:reload_manifest)
+  def cache_manifest?
+    fetch(:cache_manifest)
   end
 
   def cache_path
--- a/lib/webpacker/configuration.rb
+++ b/lib/webpacker/configuration.rb
@@ -37,6 +37,10 @@ class Webpacker::Configuration
     public_output_path.join("manifest.json")
   end
 
+  def reload_manifest?
+    fetch(:reload_manifest)
+  end
+
   def cache_path
     root_path.join(fetch(:cache_path))
   end
@@ -47,11 +51,7 @@ class Webpacker::Configuration
     end
 
     def data
-      if env.production?
-        @data ||= load
-      else
-        refresh
-      end
+      @data ||= load
     end
 
     def load
--- a/lib/webpacker/configuration.rb
+++ b/lib/webpacker/configuration.rb
@@ -47,10 +47,10 @@ class Webpacker::Configuration
     end
 
     def data
-      if env.development?
-        refresh
-      else
+      if env.production?
         @data ||= load
+      else
+        refresh
       end
     end
 
--- a/lib/webpacker/manifest.rb
+++ b/lib/webpacker/manifest.rb
@@ -34,7 +34,6 @@ class Webpacker::Manifest
 
     def compile
       Webpacker.logger.tagged("Webpacker") { compiler.compile }
-      refresh
     end
 
     def find(name)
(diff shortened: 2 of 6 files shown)


lib/webpacker/manifest.rb
```diff
@@ -46,11 +47,7 @@ def handle_missing_entry(name)
     end
 
     def data
-      if env.development?
-        refresh
-      else
-        @data ||= load
-      end
+      @data ||= load
```
dhh: This will now only refresh if you're using the compile method here. Not if you're using, say, the watcher.

gauravtiwari: Oh yes 😄 , pushed an update

lib/webpacker/configuration.rb
```diff
@@ -47,10 +47,10 @@ def fetch(key)
     end
 
     def data
-      if env.development?
-        refresh
-      else
+      if env.production?
```
dhh: Rather than checking the env directly, we should have a configuration option that can be set. Maybe something like `cache_config` and `cache_manifest`. Then people can specify what they want in all sorts of different envs, like staging or beta or whatever.

justin808: @dhh is right as I've seen many rails installs where there are various flavors of production using for staging and performance tests.

gauravtiwari:
@dhh Added a new `reload_manifest` option, for config I think we can follow the current Rails convention where the config don't auto reload until the server is restarted. What do you think?

gauravtiwari:
Secondly, people won't change config often compared to manifest

lib/webpacker/manifest.rb
```diff
@@ -46,7 +46,7 @@ def handle_missing_entry(name)
     end
 
     def data
-      if env.development?
+      if config.reload_manifest?
```
dhh: Now we can also stop delegating env to the webpacker instance in this class.

gauravtiwari:
Makes sense 👍  🍰  Renamed

gauravtiwari:
@dhh We could cache config mtime in memory though. Do you think that would be good idea?

Something like this: 

```rb
  def current?
    @cached_mtime == current_mtime
  end

  def old?
    !current?
  end

  def refresh
    @cached_mtime = current_mtime
    @data = load
  end

  def cached_mtime
      @cached_mtime ||= current_mtime
    end

    def current_mtime
      File.mtime(config_path).utc.to_i
    end

   def data
      if old?
        refresh
      else
        @data ||= load
      end
    end
```

dhh:
Don't think it's necessary at this time. We should probably just have a comment in the config file reminding people to restart their server. Just like we do with Rails.

gauravtiwari:
Sounds good 👍

PR #656, opened 2017-08-16
React on Rails: last issue is the check for the binstubs crashes if the binstub does not exist

justin808:
Small changes for React on Rails based on master branch of rails/webpacker

I can move the better error message part to a separate PR tomorrow.

The only issue is the check for the binstubs crashes if the binstub does not exist.

I have updated https://github.com/shakacode/react_on_rails/pull/908

### DIFF
--- a/lib/install/config/webpacker.yml
+++ b/lib/install/config/webpacker.yml
@@ -5,6 +5,7 @@ default: &default
   source_entry_path: packs
   public_output_path: packs
   cache_path: tmp/cache/webpacker
+  custom_compile: false
 
   # Additional paths webpack should lookup modules
   # ['app/assets', 'engine/foo/app/assets']
--- a/lib/tasks/webpacker/check_binstubs.rake
+++ b/lib/tasks/webpacker/check_binstubs.rake
@@ -1,7 +1,8 @@
 namespace :webpacker do
   desc "Verifies that bin/webpack & bin/webpack-dev-server are present."
   task :check_binstubs do
-    unless File.exist?("bin/webpack") && File.exist?("bin/webpack-dev-server")
+    unless Webpacker.config.custom_compile? ||
+      (File.exist?("bin/webpack") && File.exist?("bin/webpack-dev-server"))
       $stderr.puts "Webpack binstubs not found.\n"\
            "Have you run rails webpacker:install ?\n"\
            "Make sure the bin directory or binstubs are not included in .gitignore\n"\
--- a/lib/tasks/webpacker/verify_install.rake
+++ b/lib/tasks/webpacker/verify_install.rake
@@ -3,14 +3,16 @@ require "webpacker/configuration"
 namespace :webpacker do
   desc "Verifies if webpacker is installed"
   task verify_install: [:check_node, :check_yarn, :check_binstubs] do
-    if Webpacker.config.config_path.exist?
-      $stdout.puts "Webpacker is installed 🎉 🍰"
-      $stdout.puts "Using #{Webpacker.config.config_path} file for setting up webpack paths"
-    else
-      $stderr.puts "Configuration config/webpacker.yml file not found. \n"\
+    unless Webpacker.config.custom_compile?
+      if Webpacker.config.config_path.exist?
+        $stdout.puts "Webpacker is installed 🎉 🍰"
+        $stdout.puts "Using #{Webpacker.config.config_path} file for setting up webpack paths"
+      else
+        $stderr.puts "Configuration config/webpacker.yml file not found. \n"\
            "Make sure webpacker:install is run successfully before " \
            "running dependent tasks"
-      exit!
+        exit!
+      end
     end
   end
 end
--- a/lib/tasks/webpacker/verify_install.rake
+++ b/lib/tasks/webpacker/verify_install.rake
@@ -3,16 +3,14 @@ require "webpacker/configuration"
 namespace :webpacker do
   desc "Verifies if webpacker is installed"
   task verify_install: [:check_node, :check_yarn, :check_binstubs] do
-    unless Webpacker.config.custom_compile?
-      if Webpacker.config.config_path.exist?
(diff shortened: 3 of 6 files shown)


justin808:
@dhh @gauravtiwari I placed my one commit on top of master. Please review.

lib/webpacker/manifest.rb
```diff
@@ -42,19 +42,36 @@ def find(name)
...
         @data ||= load
+      else
+        refresh
       end
     end
 
     def load
-      if config.public_manifest_path.exist?
+      if config.public_manifest_path.exist? &&
+        (@parsed_mtime.nil? ||
```
gauravtiwari: @justin808 What this does? We are already using cached timestamp

justin808: This makes sure we don't re-parse the JSON manifest every time for the helper. The cached timestamp is for re-compilation. React on Rails is not using the compile feature.

lib/install/config/webpacker.yml
```diff
@@ -5,6 +5,7 @@ default: &default
   source_entry_path: packs
   public_output_path: packs
   cache_path: tmp/cache/webpacker
+  custom_compile: false
```
justin808: @dhh @gauravtiwari I need to add something like this so that assets:precompile does not try to use the bin/webpacker command.

We can't just turn compile off, as that's used to determine if compile checking should happen. Maybe instead of true/false, we could have a third value of "custom"? Whatever you prefer is fine with me.

lib/webpacker/manifest.rb
```diff
@@ -42,19 +42,36 @@ def find(name)
...
+      msg = <<-MSG
+Webpacker can't find #{bundle_name} in #{config.public_manifest_path}. Possible causes:
+1. You are hot reloading.
+2. You want to set Configuration.compile to true for your environment.
+3. Webpack has not re-run to reflect updates.
+4. You have misconfigured Webpacker's config/webpacker.yml file.
+5. Your Webpack configuration is not creating a manifest.
+Your manifest contains:
+#{@data.to_json}
+    MSG
```
justin808: @gauravtiwari @dhh Do we want a more detailed error message? or at least what I did on line 45 to output the manifest?

lib/install/config/webpacker.yml
dhh: Not sure I understand? Should be possible for React on Rails to overwrite the webpacker:compile task or to even remove it from the assets:precompile preconditions. I think that's probably the better move.

lib/webpacker/manifest.rb
dhh: Not sure I follow. In production, we'll shortly have `cache_manifest`, so there won't be any performance penalty. In development and test, I'd be surprised if there's any big performance ding from parsing the JSON on demand as needed? This isn't going to be a huge multi-megabyte file.

lib/webpacker/manifest.rb
dhh: I think this is good 👍

lib/tasks/webpacker/verify_install.rake
```diff
@@ -3,14 +3,16 @@ require "webpacker/configuration"
 namespace :webpacker do
   desc "Verifies if webpacker is installed"
   task verify_install: [:check_node, :check_yarn, :check_binstubs] do
-    if Webpacker.config.config_path.exist?
-      $stdout.puts "Webpacker is installed 🎉 🍰"
-      $stdout.puts "Using #{Webpacker.config.config_path} file for setting up webpack paths"
-    else
-      $stderr.puts "Configuration config/webpacker.yml file not found. \n"\
+    unless Webpacker.config.custom_compile?
```
dhh: Now that we've extracted all the logic from the rake tasks into the Webpacker classes, I think the better move here, rather than trying to hack the webpacker tasks, is probably just to have your own reactonrails:* tasks that call these Webpacker classes directly. Then you can tailor output and prerequisites as you please 👍

lib/install/config/webpacker.yml
justin808: @dhh Right now, the default precompile from both Webpacker and React on Rails is getting run. Rake is additive. I don't see a good way for React on Rails to disable to the Webpacker task. That's all I want to do.

lib/tasks/webpacker/check_binstubs.rake
```diff
@@ -1,7 +1,8 @@
 namespace :webpacker do
   desc "Verifies that bin/webpack & bin/webpack-dev-server are present."
   task :check_binstubs do
-    unless File.exist?("bin/webpack") && File.exist?("bin/webpack-dev-server")
+    unless Webpacker.config.custom_compile? ||
+      (File.exist?("bin/webpack") && File.exist?("bin/webpack-dev-server"))
```
justin808: @dhh @gauravtiwari How do we work around this part for React on Rails users that will not want to have the bin/webpack file installed?

It turns out that if the automatic precompile task runs, this message prints:

`No compiling needed as everything is fresh`

as React on Rails first does the compile, so this seems OK.

lib/tasks/webpacker/verify_install.rake
```diff
@@ -3,14 +3,16 @@ require "webpacker/configuration"
 namespace :webpacker do
   desc "Verifies if webpacker is installed"
   task verify_install: [:check_node, :check_yarn, :check_binstubs] do
-    if Webpacker.config.config_path.exist?
```
justin808: The :check_binstubs is the issue, not the part below.

lib/tasks/webpacker/check_binstubs.rake
gauravtiwari: @justin808 Yeah probably should test for `config/webpacker.yml` since that's more primary than checking binstubs. 

We discussed to remove logger output - don't think it's needed when assets are fresh.

gauravtiwari:
We have 3 PR's open for same thing :)

PR #658, opened 2017-08-16
Force Rack::Proxy to use http

paulca:
Since webpack-dev-server is running as http, force a http connection.

If the frontend server is running https, Rack::Proxy will otherwise
attempt to conenct via https to the backend.

I'm using webpacker with passenger in development behind passenger, with SSL enabled for local development.

Rack::Proxy was attempting to connect to the backend server over SSL, causing nginx to 500.

### DIFF
--- a/lib/webpacker/dev_server_proxy.rb
+++ b/lib/webpacker/dev_server_proxy.rb
@@ -10,6 +10,8 @@ class Webpacker::DevServerProxy < Rack::Proxy
   def perform_request(env)
     if env["PATH_INFO"] =~ /#{public_output_uri_path}/ && Webpacker.dev_server.running?
       env["HTTP_HOST"] = Webpacker.dev_server.host_with_port
+      env["HTTPS"] = "off"
+      env["rack.url_scheme"] = "http"
       super(env)
     else
       @app.call(env)


gauravtiwari:
@paulca You might wanna use HTTPS for dev-server otherwise dev-server websockets won't connect if you terminate SSL on proxy. We probably should add back HTTPS options for dev-server

gauravtiwari:
screen shot 2017-08-13 at 14 08 48

paulca:
@guaravtiwari ah, I assumed the plan was to stick with http and just run with that since the http is hard-coded.

gauravtiwari:
@paulca Yepp it was for simplicity until we hit that issue with dev-server, which kind of kills the whole purpose of using it since you won't have live-updates or anything. Going to put back https options soon. 

Thanks for PR though 👍

paulca:
@gauravtiwari — I would recommend two things:

- add a step for generating a self-signed cert, since the cert that is generated with `webpack-dev-server` doesn't work with Chrome (after 58 anyway), and have webpacker use that one (possibly stored in `config/webpack`
- Set `rack.ssl_verify_none` to false, since it's a self-signed cert (as mentioned in the `Rack::Proxy` `README`
- Or look in to a way of using letsencrypt.org as a CA or something

😟

gauravtiwari:
@paulca Yepp that's what I ended up in the proxy PR but later thought we could just bypass the whole thing for simplicity. 

Thanks for the pointers though 🍰

paulca:
@gauravtiwari Thanks for the awesome work. This stuff is really, really great.

PR #664, opened 2017-08-18
Check for config file before attempting to read

### LINKED ISSUE #663: Errno::ENOENT: No such file or directory @ rb_sysopen config/webpacker.yml
The webpacker's master branch (rails v5.1.3, ruby 2.4) is giving this error when i run `rails webpacker:install` on an existing rails project. The immediate solution is to use `gem 'webpacker', '~> 2.0'` instead. 

> Errno::ENOENT: No such file or directory @ rb_sysopen - /Users/ytbryan/Desktop/X/vue-X/config/webpacker.yml
> /Users/ytbryan/Desktop/Doerlist/X/config/environment.rb:5:in `<top (required)>'
> /Users/ytbryan/Desktop/Doerlist/X/bin/rails:9:in `require'
> /Users/ytbryan/Desktop/Doerlist/X/bin/rails:9:in `<top (required)>'
> /Users/ytbryan/Desktop/Doerlist/X/bin/spring:14:in `require'
> /Users/ytbryan/Desktop/Doerlist/X/bin/spring:14:in `<top (required)>'
> ./bin/rails:3:in `load'
> ./bin/rails:3:in `<main>'
> Tasks: TOP => app:template => environment
> (See full trace by running task with --trace)
> 

I will look at this when I'm free. Feel free to report or PR if you find out what's wrong.

gauravtiwari:
Fixes #663 i.e. reading config file before webpacker is installed

### DIFF
--- a/lib/webpacker/configuration.rb
+++ b/lib/webpacker/configuration.rb
@@ -56,6 +56,16 @@ class Webpacker::Configuration
 
     def load
       YAML.load(config_path.read)[env].deep_symbolize_keys
+
+    rescue Errno::ENOENT => e
+      raise "Webpacker configuration file not found #{config_path}. " \
+            "Have you installed webpacker? " \
+            "Error: #{e.message}"
+
+    rescue Psych::SyntaxError => e
+      raise "YAML syntax error occurred while parsing #{config_path}. " \
+            "Please note that YAML must be consistently indented using spaces. Tabs are not allowed. " \
+            "Error: #{e.message}"
     end
 
     def defaults
--- a/lib/webpacker/configuration.rb
+++ b/lib/webpacker/configuration.rb
@@ -55,7 +55,11 @@ class Webpacker::Configuration
     end
 
     def load
-      YAML.load(config_path.read)[env].deep_symbolize_keys
+      if config_path.exist?
+        YAML.load(config_path.read)[env].deep_symbolize_keys
+      else
+        {}.freeze
+      end
     end
 
     def defaults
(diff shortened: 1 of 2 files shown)


dhh:
I'd rather figure out why Webpacker is being booted at a time prior to webpacker.yml existing. It doesn't make sense to me to silently swallow a missing config file in normal use. So I'd rather just wrap the initialization on this check.

gauravtiwari:
Oh yes, makes sense. I will investigate and update this PR 👍🙏

gauravtiwari:
Added `defined?(Rails::Server) || defined?(Rails::Console)` check prior to bootstrap, not sure if the check is robust though. The idea is to try and bootstrap whenever server or console is run and fail (compared to config file check which would ignore bootstrap in all cases), for other cases like running a rake task don't bootstrap.

lib/webpacker/configuration.rb
```diff
@@ -56,6 +56,16 @@ def data
 
     def load
       YAML.load(config_path.read)[env].deep_symbolize_keys
+
+    rescue Errno::ENOENT => e
+      raise "Webpacker configuration file not found #{config_path}. " \
+            "Have you installed webpacker? " \
```
ytbryan: I think this line should be replaced with `Please run: rails webpacker:install`. What do you think?

gauravtiwari:
@dhh Please review and merge this one as well if looks good :)

dhh:
Might want to consider @ytbryan's exception message improvement as well.

lib/webpacker/configuration.rb
gauravtiwari: @ytbryan Oh totally missed, feel free to make a PR for this 👍

Issue #660, opened 2017-08-17
DevServerProxy doesn't work when app is requested over SSL


grk:
I'm running my development app using puma-dev, which gives me the possibility to use SSL in development. This breaks the DevServerProxy on current master (7a1d306f367174e39c815bd298ab5d61cbbde72e), because it tries to connect to webpack-dev-server using SSL.

Here's the backtrace:
```
2017-08-17 16:24:32 +0200: Rack app error handling request { GET /packs/manager-782b5244ccbb002df6cc.js }
#<OpenSSL::SSL::SSLError: SSL_connect SYSCALL returned=5 errno=0 state=SSLv2/v3 read server hello A>
/Users/grk/.rbenv/versions/2.4.1/lib/ruby/2.4.0/net/protocol.rb:44:in `connect_nonblock'
/Users/grk/.rbenv/versions/2.4.1/lib/ruby/2.4.0/net/protocol.rb:44:in `ssl_socket_connect'
/Users/grk/.rbenv/versions/2.4.1/lib/ruby/2.4.0/net/http.rb:948:in `connect'
/Users/grk/.rbenv/versions/2.4.1/lib/ruby/2.4.0/net/http.rb:887:in `do_start'
/Users/grk/.rbenv/versions/2.4.1/lib/ruby/2.4.0/net/http.rb:882:in `start'
/Users/grk/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/rack-proxy-0.6.2/lib/rack/http_streaming_response.rb:71:in `session'
/Users/grk/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/rack-proxy-0.6.2/lib/rack/http_streaming_response.rb:60:in `response'
/Users/grk/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/rack-proxy-0.6.2/lib/rack/http_streaming_response.rb:29:in `headers'
/Users/grk/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/rack-proxy-0.6.2/lib/rack/proxy.rb:120:in `perform_request'
/Users/grk/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/bundler/gems/webpacker-7a1d306f3671/lib/webpacker/dev_server_proxy.rb:13:in `perform_request'
[...]

PR #665, opened 2017-08-18
Add back Https

gauravtiwari:
This PR adds  back `https` option for `webpack-dev-server` since we now have a proxy and people seems to be using `https` in development. Sadly, we can't terminate SSL since webpack-dev-server uses websockets, which won't connect until it's using same protocol. 

Fixes: https://github.com/rails/webpacker/issues/660
Supersedes: https://github.com/rails/webpacker/pull/658

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,12 @@
 
 - Add new `HMR` option for hot-module-replacement
 
+- HTTPS ssl settings to webpack-dev-server config
+  ```yml
+    # Absolute paths to ssl key and certificate
+    ssl_key_path:
+    ssl_cert_path:
+  ```
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,7 +29,7 @@
 
 - Add new `HMR` option for hot-module-replacement
 
-- HTTPS ssl settings to webpack-dev-server config
+- Configurable HTTPS ssl key and cert settings
   ```yml
     # Absolute paths to ssl key and certificate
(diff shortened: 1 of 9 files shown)


grk:
A few comments:

1) I think it would still be good to add the change from #658, take the https setting from webpacker.yml. With this branch, when I have the dev server running with https, and I request the Rails app with http, DevServerProxy tries to connect to dev server over http and fails.
2) The cert path might be different for various development setups working on the same codebase. For example, if I wanted to use the puma-dev cert, it's in ~/Library/Application Support/io.puma.dev/, but node doesn't resolve ~, so I need to use the full path, which includes the username. Possibly something like webpacker.yml.local would be a solution?
3) I'm getting `GET https://localhost:3035/sockjs-node/info?t=1503127986584 net::ERR_INSECURE_RESPONSE` in chrome console.

I think it might not be worth the effort, and just saying that ssl in development is not supported would be a good solution.

gauravtiwari:
@grk changes in #658 are basically useless considering it will block the web sockets connection which means the dev server will be useless. 

The exception you are getting is due to self signed certificate. Please open the dev server link in a separate browser tab and accept browser exception( i.e. the link you posted in 3). Then reload the view and everything should work - just one time thing.   

It’s not necessary to use same certificate but if you have certificate that’s trusted then you can use it with both Rails and dev server. Then browser won’t go mad :)

gauravtiwari:
Going to nix key and certificate configuration though - not needed

shaundr:
what is blocking this? anything I could contribute?

PR #666, opened 2017-08-18
Cache control

gauravtiwari:
Fixes: #550 (2)

This PR adds a middleware that adds `cache-control` and `etag` header on webpacker assets when served from file system. We already added hashes to the assets in all env however absence of the `Cache-Control` header might evict the assets prematurely (when sharing CDN or in browser). 

Previously: 
screen shot 2017-08-18 at 22 26 15

Now: 
screen shot 2017-08-18 at 22 22 44

cc// @richardvenneman @terracatta

### DIFF
--- a/lib/webpacker/cache_control.rb
+++ b/lib/webpacker/cache_control.rb
@@ -1,0 +1,51 @@
+class Webpacker::CacheControl
+  ALLOWED_REQUEST_METHODS = ["GET", "HEAD"].freeze
+
+  def initialize(app)
+    @app = app
+  end
+
+  def call(env)
+    if env["PATH_INFO"] =~ /#{public_output_uri_path}/
+      status, headers, body = @app.call(env)
+
+      unless ALLOWED_REQUEST_METHODS.include?(env["REQUEST_METHOD"])
+        return method_not_allowed_response
+      end
+
+      etag, new_body = Rack::ETag.new(env).send(:digest_body, body)
+
+      # Set caching headers
+      headers["Cache-Control"] = String.new("public")
+      headers["ETag"]          = %("#{etag}")
+
+      # If the request url contains a fingerprint, set a long
+      # expires on the response
+      if path_fingerprint(env["PATH_INFO"])
+        headers["Cache-Control"] << ", max-age=31536000, immutable"
+
+      # Otherwise set `must-revalidate` since the asset could be modified.
+      else
+        headers["Cache-Control"] << ", must-revalidate"
+        headers["Vary"] = "Accept-Encoding"
+      end
+
+      [status, headers, new_body]
+    else
+      @app.call(env)
+    end
+  end
+
+  private
+    def public_output_uri_path
+      Webpacker.config.public_output_path.relative_path_from(Webpacker.config.public_path)
+    end
+
+    def method_not_allowed_response
+      [ 405, { "Content-Type" => "text/plain", "Content-Length" => "18" }, [ "Method Not Allowed" ] ]
+    end
+
+    def path_fingerprint(path)
+      path[/-([0-9a-f]{7,128})\.[^.]+\z/, 1]
+    end
+end
--- a/lib/webpacker/railtie.rb
+++ b/lib/webpacker/railtie.rb
@@ -2,6 +2,7 @@ require "rails/railtie"
 
 require "webpacker/helper"
 require "webpacker/dev_server_proxy"
+require "webpacker/cache_control"
 
 class Webpacker::Engine < ::Rails::Engine
   initializer "webpacker.proxy" do |app|
@@ -10,6 +11,12 @@ class Webpacker::Engine < ::Rails::Engine
         Rails::VERSION::MAJOR >= 5 ?
           Webpacker::DevServerProxy : "Webpacker::DevServerProxy"
     end
+  end
+
+  initializer "webpacker.cache_control" do |app|
+    app.middleware.insert_before 0,
+      Rails::VERSION::MAJOR >= 5 ?
+        Webpacker::CacheControl : "Webpacker::CacheControl"
   end
 
   initializer "webpacker.helper" do |app|


dhh:
I don't follow why we need to set cache controls in development if we're using hashes? When the hash changes, the asset will be requested fresh. And for production, you're going to want to set the cache controls on nginx or whatever.

richardvenneman:
@dhh With a default Rails app, the Sprockets-generated `application.js` does get served with a `Cache-Control` header in production. This way we have browser caching by default and it also makes it really easy to setup a CDN (e.g. Cloudfront with origin cache headers) when you're on Heroku.

I don't know about other webservers and Heroku, but if you're using Passenger, getting caching control with webpacker currently involves setting [Passenger Standalone with a custom Nginx configuration template](https://www.phusionpassenger.com/library/config/standalone/intro.html#nginx-configuration-template) instead of the regular plug-and-play Passenger.

gauravtiwari:
Yep not in development just for the servers which doesn’t handle static file delivery, for example Rails app on Heroku, serving from standard rails servers like puma or passenger standalone. 

This ships by default with Sprockets but may be we can make this opt-in or even ship as another Gem.

terracatta:
> I don't follow why we need to set cache controls in development if we're using hashes? When the hash changes, the asset will be requested fresh. And for production, you're going to want to set the cache controls on nginx or whatever.

@dhh I am the creator of the original issue. My problem is solved by fingerprinting all webpacker assets in development, but I think the goal here is to match the current behavior of sprockets which _does_ serve fingerprinted assets with cache-control headers in development. I can't find the sprockets/rails PR that introduces this behavior, but my assumption is that if the asset _hasn't_ changed, it is faster to serve it from the browser cache vs from the dev server.

The hashing and caching strategy introduced here and already present in sprockets I think produce a more performant solution for developers.

dhh:
That's fair. @schneems how do we set the right cache-control for sprocket assets? Using a middleware like this as well? Seems heavy handed to me.

gauravtiwari:
@dhh Part of the code is from there actually - https://github.com/rails/sprockets/blob/d81ca88b959950954cbc1dd76c45a6fbe41b504f/lib/sprockets/server.rb#L238

schneems:
@gauravtiwari that server does not get used in production. Assets are served in production by nginx or by the `Static` middleware if you've got serve static assets config set to true.

Here's a PR of mine from 2014 to enable cache control by default https://github.com/rails/rails/pull/16465

Currently we don't set cache control for sprockets assets due to \<reasons\>. Mostly because you cannot guarantee every asset in the directory to be a managed asset. I've got an idea for how to move forwards there, but it's not 100% straightforward. 

As long as managed assets have fingerprints, I see no problems in adding max-age. Must make sure that this middleware goes before the dispatch static middleware

gauravtiwari:
@schneems Ahh I see, thanks :) 

But if an app is running on Heroku on standard Rails server, then it won't add these headers automatically right? Don't think `Static` middleware add them either https://github.com/rails/rails/blob/cfb1e4dfd8813d3d5c75a15a750b3c53eebdea65/actionpack/lib/action_dispatch/middleware/static.rb

But I think one can pass headers to it:,

```rb
config.public_file_server.headers = {
  'Cache-Control' => 'public, s-maxage=31536000, maxage=15552000',
  'Expires' => "#{1.year.from_now.to_formatted_s(:rfc822)}"
}
```

I think then we don't need this PR 😄  Just use `Static` middleware

schneems:
You can't set a cache-control header for all things in your public folder because that middleware handles all these cases: 

- Assets managed by Rails (sprockets etc.) we can safely set max-age here because they are fingerprinted.
- Assets uploaded to disk (this is the oldschool way of doing avatars etc. before S3). We cannot set maxage here but with active storage we should remove this requirement IMHO.
- Manually added assets. These are not fingerprinted and we cannot automatically determine a cache age as a result.
  - Manually added regular assets css/js/images/etc. not that common
  - Error pages 404.html and 500.html etc. EXTREMELY common, we generate these with a default rails app.
  - People can generate "non-digest" versions of managed assets via gems or their own code. We cannot set max-age here.

One down side of this is that it has to touch disk 3 times on every request even if it's not a request for an asset: https://github.com/rails/rails/blob/b497c16690c9f153cfcb99eeeb3103426a5db447/actionpack/lib/action_dispatch/middleware/static.rb#L35-L43

Basically a request for `/foo` path will check for `foo.html` and `foo/index.html` and `foo` without the `.html` extension. I would love to get rid of this behavior.

To be able to set max-age we need to split out the assets into "managed" i.e. sprockets and webpacker and "unmanaged" i.e. things people manually put on disk.

dhh:
How about a general purpose DigestAssetsLiveForever middleware that scans for the presence of the digest in a sprockets or webpack asset? Then even manually generated assets would benefit too. And public/whatever.html would not.

schneems:
I like the idea. The only conceivable downfall would be if someone names their asset something digest like, but does not change the digest when the asset changes.

I think that's a reasonable expectation. We just need to document it. 

Alternatively we could have the middleware take an array of directories where max-age is applied. So anything in `assets/` gets the header etc. Expose a configuration endpoint so that gems can declare their directories.

On the issue of slowing down web requests with a disk check, I have a PR from a long time ago, not sure if maybe there's another alternative here: https://github.com/rails/rails/pull/16464

dhh:
I think it's a fair trade-off that if people add a digest to their filename
we'll treat it as a live-forever asset. Then we also don't have to do
anything that's directory specific. You can already tweak the outputs for
Webpack, so would be good to move away from hardcoded paths and rely on a
filename signature instead 👍

On Thu, Aug 24, 2017 at 2:41 PM, Richard Schneeman <notifications@github.com

dhh:
Let's close this PR and focus on a general "file as digest, thus it'll get far-future expiry" solution in Rails itself.

antonlitvinenko:
Was there any activity on making this actually happen in Rails itself?

hrdwdmrbl:
Yes, it seems that in development I'm recompiling and reloading all JS every page-load.

PR #667, opened 2017-08-18
WIP: Run webpack compiler as subprocess

gauravtiwari:
This PR adds subprocess support so we can fork a new webpack compiler when running `rails s` using `::Rails::Server`, which also kills the subprocess properly on exit (CTRL+C).

Not sure about the config setup though, like how we should wire this together since we have a on-demand compiler too. 

screen shot 2017-08-19 at 00 04 28
screen shot 2017-08-19 at 00 04 47
screen shot 2017-08-19 at 00 03 31

Or with WebRick:

screen shot 2017-08-19 at 00 37 40

### DIFF
--- a/lib/install/config/webpacker.yml
+++ b/lib/install/config/webpacker.yml
@@ -13,6 +13,12 @@ default: &default
   # Reload manifest.json on all requests so we reload latest compiled packs
   cache_manifest: false
 
+  # Enable on-demand compilation
+  compile: true
+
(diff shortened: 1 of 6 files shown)


dhh:
Given how good our on-demand compilation option is now, I don't think this level of complexity is worth it. Especially since starting and stopping the dev server manually has a clean and automated hand-over. I'd say we skip this for now.

Issue #523, opened 2017-06-20
babel-polyfill is not included/loaded


schuylr:
I noticed that the README states that `babel-polyfill` is included, however there's no reference to it in any entry points:

http://babeljs.io/docs/usage/polyfill/#usage-in-node-browserify-webpack

Is this something we must do ourselves?

ytbryan:
It is only installed here https://github.com/rails/webpacker/blob/66115ba76ccd36e8cdd6e19b24b76f1c5f039119/lib/install/template.rb#L33

maybe you can send in a PR for this?

schuylr:
I can't seem to put a `require` or `import` anywhere for it - I still end up missing `prototype[Symbol.iterator]` for my Arrays, so I'm not sure where it should be added or if my configuration is just off.

lsanwick:
I corrected this by changing the `/config/webpack/shared.js` to have entry paths that include the polyfill:

```js
module.exports = {
  entry: packPaths.reduce((map, entry) => {
    const localMap = map
    const namespace = relative(join(entryPath), dirname(entry))
    localMap[join(namespace, basename(entry, extname(entry)))] = ['babel-polyfill', resolve(entry)]
    return localMap
  }, {}),
  ...
}
```

gauravtiwari:
You can do the same in your main entry/packs file.

gauravtiwari:
@lsanwick Doesn't this option takes care of adding correct polyfill during compilation? https://github.com/babel/babel-preset-env#usebuiltins What do you have in your babelrc?

lsanwick:
> NOTE: This does not currently polyfill experimental/stage-x built-ins like the regular "babel-polyfill" does. This will only work with npm >= 3 (which should be used with Babel 6 anyway)

I'm using generators, so I don't believe that is covered by the `useBuiltins` flag.

lsanwick:
@gauravtiwari:

Here is my .babelrc and debug output.
```json
{
  "presets": [
    [
      "env",
      {
        "modules": false,
        "targets": {
          "browsers": "> 1%",
          "uglify": true
        },
        "useBuiltIns": true,
        "debug": true
      }
    ],
    "react",
    "stage-1"
  ],
  "plugins": [
    "syntax-dynamic-import", [
      "transform-class-properties",
      {
        "spec": true
      }
    ]
  ]
}
```

```
babel-preset-env: `DEBUG` option
07:39:33 webpack.1 |
07:39:33 webpack.1 | Using targets:
[...]

Issue #532, opened 2017-06-22
ActionView::Template::Error (Can't find sales.js in /app/public/packs/manifest.json. Is webpack still compiling?)


amingilani:
I get this error in production, while the app works perfectly in development.

From my [Stack Overflow question](https://stackoverflow.com/questions/44690763/actionviewtemplateerror-cant-find-sales-js-in-app-public-packs-manifest-j):

> ```
> 2017-06-22T05:15:05.148829+00:00 app[web.1]: F, [2017-06-22T05:15:05.148717 #4] FATAL -- : [ef509b29-8637-41b3-8224-4423dbf2b2ed]
> 2017-06-22T05:15:05.148948+00:00 app[web.1]: F, [2017-06-22T05:15:05.148883 #4] FATAL -- : [ef509b29-8637-41b3-8224-4423dbf2b2ed] ActionView::Template::Error (Can't find sales.js in /app/public/packs/manifest.json. Is webpack still compiling?):
> 2017-06-22T05:15:05.149203+00:00 app[web.1]: F, [2017-06-22T05:15:05.149103 #4] FATAL -- : [ef509b29-8637-41b3-8224-4423dbf2b2ed]     1: %div#sale
> 2017-06-22T05:15:05.149206+00:00 app[web.1]: [ef509b29-8637-41b3-8224-4423dbf2b2ed]     2: = javascript_pack_tag 'sales'
> 2017-06-22T05:15:05.149297+00:00 app[web.1]: F, [2017-06-22T05:15:05.149231 #4] FATAL -- : [ef509b29-8637-41b3-8224-4423dbf2b2ed]
> 2017-06-22T05:15:05.149398+00:00 app[web.1]: F, [2017-06-22T05:15:05.149328 #4] FATAL -- : [ef509b29-8637-41b3-8224-4423dbf2b2ed] app/views/sales/new.html.haml:2:in `_app_views_sales_new_html_haml___2340659139987297731_36903000'
> 2017-06-22T05:15:07.004434+00:00 heroku[router]: at=info method=GET path="/favicon.ico" host=planetlauncher.herokuapp.com request_id=8bca0b65-0d4a-494f-88c7-a4b31c2ff6dc fwd="202.163.79.6" dyno=web.1 connect=1ms service=2ms status=200 bytes=143 protocol=https
> ```
> 
> There's only one route
> 
> ```
> Rails.application.routes.draw do
>   root 'sales#new'
>   resources :sales
> end
> ```
> 
[...]

PR #674, opened 2017-08-20
Use babel-polyfill in application.js

tai2:
This implies users need `import babel-polyfill` at entry point.

see: https://github.com/rails/webpacker/issues/523

I agree with @adamsanderson's proposal. This is it.

If there is no hint, users like me might misunderstand that polyfill is applied without any additional effort.

### DIFF
--- a/lib/install/javascript/packs/application.js
+++ b/lib/install/javascript/packs/application.js
@@ -6,5 +6,6 @@
 //
 // To reference this file, add <%= javascript_pack_tag 'application' %> to the appropriate
 // layout file, like app/views/layouts/application.html.erb
+import 'babel-polyfill'
 
 console.log('Hello World from Webpacker')


gauravtiwari:
@tai2 Thanks for this but this can still be confusing. Simply including it in `application.js` doesn't mean that certain app needs polyfilling. That's why suggested to add a section about this in the README instead.

tai2:
Okay I'll make another PR which modifies the README.

feliperaul:
@tai2 Tai, could you please reconsider reopening this? I think the goal of this gem is make Webpack approachable and capitalize on Rails convention over configuration. If that's the case, when I saw `"browsers": "> 1%"` in my .babelrc file I thought I would have the polyfills in my bundle, which is not the case; as the discussion in #532 reveals (https://github.com/rails/webpacker/issues/523#issuecomment-321880881), if `import "babel-polyfill"`is not present in the js entry file that configuration in .babelrc then `useBuiltIns` is simply ignored and no polyfill is made.

I know this is already in the docs [here](https://github.com/rails/webpacker/blob/e25b73be0f1d030bcca071bd146d2832ee5f8450/docs/es6.md), but instead of 'Don't forget to import @babel/polyfill in your main entry point', why don't we just leave it there already, but include a comment like the one I'm suggesting now:

```
// Comment the line below if you don't need pollyfilling to target older browsers. The target is defined in .babelrc file according to browserslist. 
import 'babel-polyfill'
```

Check the comment I linked in the discussion above, it already has +9 votes for including the import as a default in the application.js placeholder :)

PR #676, opened 2017-08-21
List possible causes of not finding a file in the manifest.

justin808:
Using these reasons:
1. You want to set wepbacker.yml value of compile to true for your
   environment unless you are using the `webpack -w` or the webpack-dev-server.
2. Webpack has not yet re-run to reflect updates.
3. You have misconfigured Webpacker's config/webpacker.yml file.
4. Your Webpack configuration is not creating a manifest.

Broken out of #656 [per comment from @dhh](https://github.com/rails/webpacker/pull/656/files#r133791005)

We may want to adjust the possible reasons.

### DIFF
--- a/lib/webpacker/manifest.rb
+++ b/lib/webpacker/manifest.rb
@@ -41,8 +41,20 @@ class Webpacker::Manifest
     end
 
     def handle_missing_entry(name)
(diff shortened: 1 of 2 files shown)


gauravtiwari:
thanks @justin808 🍰

PR #381, opened 2017-05-13
Check for the binstubs in verify_install (#380)

ytbryan:
Fix https://github.com/rails/webpacker/issues/380

Basically, sound off when binstubs are not found in production. 

This could happen when users include bin directory or binstub in .gitignore 

cc @christophermlne @gauravtiwari

PR #677, opened 2017-08-21
Check for webpacker being installed by having config file

justin808:
Previously checking for the bin stubs, but some frameworks, like React
on Rails, don't use the binstubs.

### DIFF
--- a/lib/tasks/webpacker/check_binstubs.rake
+++ b/lib/tasks/webpacker/check_binstubs.rake
@@ -1,10 +1,11 @@
 namespace :webpacker do
   desc "Verifies that bin/webpack & bin/webpack-dev-server are present."
   task :check_binstubs do
-    unless File.exist?("bin/webpack") && File.exist?("bin/webpack-dev-server")
-      $stderr.puts "Webpack binstubs not found.\n"\
-           "Have you run rails webpacker:install ?\n"\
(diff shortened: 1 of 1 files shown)


justin808:
@dhh @gauravtiwari this is the final change required by React on Rails.
[See this comment](https://github.com/rails/webpacker/pull/656#discussion_r133919961) from @gauravtiwari

gauravtiwari:
@justin808 It's already done here: 

https://github.com/rails/webpacker/blob/master/lib/tasks/webpacker/verify_install.rake#L6

And the `check_binstubs` change was introduced here: https://github.com/rails/webpacker/pull/381

gauravtiwari:
@justin808 Yepp I know, thinking a way around this ;). The change introduced in this PR is already present in `verify_install` task.

dhh:
I'm not sure I follow this. What's the point of using these tasks that depend on binstubs if they binstubs are not there? Why not just have an alternate set of tasks for React on Rails that doesn't have these preconditions?

justin808:
@dhh is right. I was trying to customize the compiler task in webpacker.yml to allow a custom task that would use `nps` or `yarn`. We can close this for now. The idea was that for non-js developers, compile could be set to true, and the JS would just work. However, on a giant project, the cost of checking for ever web request is too much.

PR #668, opened 2017-08-19
Check for Webpacker config_path at bootstrap

ytbryan:
Update: closing this as I'm quite satisfied with https://github.com/rails/webpacker/pull/664

---

Fixes & closes https://github.com/rails/webpacker/issues/663

Replaces https://github.com/rails/webpacker/pull/664

PR #678, opened 2017-08-21
fix(configuration): fix initialization error when default config file is not existed

dcalsky:
`config_path` variable is pointed at `config/webpacker.yml` defaultly, but there are no webpacker.yml configuration file before executing install command, which makes any operation will cause error.

A serious mistake, fix it now. Otherwise, the latest master branch does not work...

### DIFF
--- a/lib/webpacker/configuration.rb
+++ b/lib/webpacker/configuration.rb
@@ -55,7 +55,8 @@ class Webpacker::Configuration
     end
 
     def load
-      YAML.load(config_path.read)[env].deep_symbolize_keys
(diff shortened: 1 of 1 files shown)


gauravtiwari:
@dcalsky Thanks for this but there are already two PR's pending for the same issue: #664 and #668

dcalsky:
Cool, it seems that I am late ...

dcalsky:
In my opinion, you should fix this problem as soon as possible, otherwise, the latest master branch does not work for someone doesn't know it...

gauravtiwari:
@dcalsky Yepp on it. Should be fixed today 👍

PR #679, opened 2017-08-21
Check for webpacker being installed by having config file

justin808:
Previously checking for the bin stubs, but some frameworks, like React
on Rails, don't use the binstubs.

And don't crash if the webpacker.yml not there yet (https://github.com/rails/webpacker/pull/668)


gauravtiwari:
Already addressing crash problem in #664

PR #681, opened 2017-08-21
Pending changes for Beta 11 of React on Rails

justin808:


lib/webpacker/railtie.rb
```diff
@@ -29,7 +29,7 @@ class Webpacker::Engine < ::Rails::Engine
   end
 
   initializer "webpacker.bootstrap" do
-    if defined?(Rails::Server) || defined?(Rails::Console)
+    if (defined?(Rails::Server) || defined?(Rails::Console)) && Webpacker.config.config_path.exist?
```
gauravtiwari: Why we need to check config file? The idea is to fail when running server or console if none exist. For other things like running react on rails installer this block won't run.

lib/webpacker/commands.rb
```diff
@@ -11,6 +11,7 @@ def clobber
   end
 
   def bootstrap
+    return unless config.config_path.exist?
```
dhh: I don't think we want #bootstrap to be a silent no-op if the config is missing.

justin808: @dhh @gauravtiwari How about if I submit another PR with the check only here, and print a  warning message?

At this level of the code, is a "puts" to stdout OK?

dhh: I don't think we should have any check inside the method of #bootstrap. I'd rather have it where #bootstrap is called. Also don't want to do any puts anywhere inside these classes. They should all use the logger. Only the rake tasks can put.

PR #684, opened 2017-08-22
Use sha digest to track recompilation

gauravtiwari:
This PR fixes a bug when the on demand compiler won't trigger a recompile if an older file (`mtime`) is moved to the source directory. The new implementation uses SHA digest to calculate a digest based on file name and `mtime`, which would allow us to track all types of changes that should trigger a recompilation.

### DIFF
--- a/lib/webpacker/compiler.rb
+++ b/lib/webpacker/compiler.rb
@@ -1,4 +1,5 @@
 require "open3"
+require "digest/sha1"
 
 class Webpacker::Compiler
   # Additional paths that test compiler needs to watch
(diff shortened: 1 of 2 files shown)


dhh:
I don't fully understand this change. The current code checks the NEWEST file in the watched paths, no? So what difference does pulling in an old file have on that? Also, doing digest seems like it would be a lot slower than checking mtime.

gauravtiwari:
Yepp, since we only check max `mtime`, which means if we copy and paste old files from asset pipeline to webpacker and `mtime` is older than the compiled webpacker packs time, it won't trigger a recompilation. 

```yml
 app/assets/packs
  - foo.js (mtime: 101)
  - bar.js (mtime: 100)
  - legacy.js (mtime: 50)
```

ref: https://github.com/rails/webpacker/blob/master/lib/webpacker/compiler.rb#L40

Yepp compared the differences and it's more or less the same as checking max `mtime`.

gauravtiwari:
Tested it with 33000 node modules files, timing is mostly same

dhh:
Sounds good to me then 👍

PR #690, opened 2017-08-24
Fix extensionGlob suffix in webpack/shared.js

### LINKED ISSUE #689: Generated webpack config results in an error if you only configure one extension to be compiled
If your `webpacker.yml` only lists a single `extension` like [this](https://github.com/rmacklin/webpacker-issue-689-repro/blob/75688c5c9eb27e4651a26cf8698f7025d242ffbc/test_app/config/webpacker.yml#L8-L9), running `bin/webpack` will error out with:

```
Invalid configuration object. Webpack has been initialised using a configuration object that does not match the API schema.
 - configuration.entry should be one of these:
   object { <key>: non-empty string | [non-empty string] } | non-empty string | [non-empty string] | function
   The entry point(s) of the compilation.
   Details:
    * configuration.entry should NOT have less than 1 properties ({
        "keyword": "minProperties",
        "dataPath": ".entry",
        "schemaPath": "#/oneOf/0/minProperties",
        "params": {
          "limit": 1
        },
        "message": "should NOT have less than 1 properties",
        "schema": 1,
        "parentSchema": {
          "minProperties": 1,
          "additionalProperties": {
            "oneOf": [
              {
                "description": "The string is resolved to a module which is loaded upon startup.",
                "minLength": 1,
                "type": "string"
              },
              {
```

It should handle this case since the user may only want to compile `.js` files.

rmacklin:
Fixes #689

Prior to this change, the extensionGlob only supported webpacker configurations that had more than one extension. If there was only extension configured, the glob generated wouldn't match anything because it was hardcoded to include a [brace expansion] which requires at least two comma-separated strings. So, to fix it, we do not include a brace expansion if there is only one configured extension.

This change also adds an explicit error message if _no_ extensions are configured.

[brace expansion]: https://www.gnu.org/software/bash/manual/html_node/Brace-Expansion.html

### DIFF
--- a/lib/install/config/webpack/shared.js
+++ b/lib/install/config/webpack/shared.js
@@ -11,14 +11,14 @@ const ManifestPlugin = require('webpack-manifest-plugin')
 const extname = require('path-complete-extname')
 const { env, settings, output, loadersDir, resolvedModules } = require('./configuration.js')
 
-const numberOfExtensions = settings.extensions && settings.extensions.length
+const extensions = settings.extensions
 
-if (numberOfExtensions === undefined || numberOfExtensions === 0) {
+if (!extensions || extensions.length === 0) {
   throw new Error('You must configure at least one extension to compile in webpacker.yml')
 }
 
 const extensionGlobSuffix =
-  numberOfExtensions === 1 ? settings.extensions[0] : `{${settings.extensions.join(',')}}`
+  extensions.length === 1 ? extensions[0] : `{${extensions.join(',')}}`
 
 const extensionGlob = `**/*${extensionGlobSuffix}`
 const entryPath = join(settings.source_path, settings.source_entry_path)
@@ -56,7 +56,7 @@ module.exports = {
   ],
 
   resolve: {
-    extensions: settings.extensions,
+    extensions,
     modules: resolvedModules
   },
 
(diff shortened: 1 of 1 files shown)


rmacklin:
@gauravtiwari This passed CI if you actually click through to the results on Travis: https://travis-ci.org/rails/webpacker/builds/267878350?utm_source=github_status&utm_medium=notification

Not sure why the webhook didn't run to update the status on GitHub.

lib/install/config/webpack/shared.js
```diff
@@ -11,7 +11,16 @@ const ManifestPlugin = require('webpack-manifest-plugin')
 const extname = require('path-complete-extname')
 const { env, settings, output, loadersDir, resolvedModules } = require('./configuration.js')
 
-const extensionGlob = `**/*{${settings.extensions.join(',')}}`
+const numberOfExtensions = settings.extensions && settings.extensions.length
+
+if (numberOfExtensions === undefined || numberOfExtensions === 0) {
```
gauravtiwari: just need: 

```js
if(!numberOfExtensions) // it will check for undefined and 0 anyway
```

Also, lets just call it `extensions` instead of `numberOfExtensions`

rmacklin: I've pushed up a change to change the variable to be `const extensions = settings.extensions` and calling `length` in the two places that check its `length` (3a7c92cf3f6d707f3196462f032f5d97fda9000e).

I prefer this to using the name `extensions` to hold the `length` of `settings.extensions` as I think that could be a bit confusing. However, if you'd rather go with that, just let me know and I'll change it.

gauravtiwari:
Thank you, looks good 🍰  👍

rmacklin:
Thanks!

PR #692, opened 2017-08-24
Use app/webpacker instead of app/javascript

k0kubun:
As written in README,

> it is possible to use Webpacker for CSS, images and fonts assets as well, in which case you may not even need the asset pipeline.

projects like [mastodon](https://github.com/tootsuite/mastodon) don't use asset pipeline and distribute images and css using webpacker. In such case, I think `app/javascript` is not a good path for source_path because it includes images and css. So I propose to use `app/webpacker` by default.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,7 +74,7 @@ only triggers if files are changed under tracked paths.
 Following paths are watched by default -
 
   ```rb
-    ["app/javascript/**/*", "yarn.lock", "package.json", "config/webpack/**/*"]
+    ["app/webpacker/**/*", "yarn.lock", "package.json", "config/webpack/**/*"]
   ```
 
   To add more paths:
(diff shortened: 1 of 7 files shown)


gauravtiwari:
@k0kubun You can change the name of source_path from `config/webpacker.yml`

k0kubun:
Yes, but this is suggestion of changing default. Many people will use default, and using default is better for maintainability and I would be very happy if default is good.

gauravtiwari:
@k0kubun I see. 

Earlier when this path was chosen it was intended that this place is where you can write app like javascript code (`app/javascript` kinda of reading this path in reverse). Nowadays, javascript apps aren't limited to just writing javascript code, thanks to module bundlers like Webpack, we can include images and fonts inside our javascript code and it's pretty normal - most people even write css in Javascript.

What do you think?

k0kubun:
I sent this PR because I thought it's totally unreasonable and couldn't imagine why nobody changes it. But now I understood your idea that reverses "javascript apps".

I prefer app/webpacker anyway. That's because my application is not just a javascript app but rather an app having many Ruby code, so I regard the directory as javascript/css/images directory. And I personally feel "app/assets/javascript" and "app/javascript" are very similar and I feel having both of them is confusing. In that point of view, "app/webpacker" describes what's different from "app/assets/javascript" well. 

But as the idea is found to be just my personal idea, I close this and change "source_path" in my app. Thank you for your attention and quick response!

gauravtiwari:
Thank you for understanding @k0kubun and I understand your concerns as well. It's just that as default `app/javascript` still makes more sense (considering varied use cases), that's why we made the path configurable so folks can change as they like.

No problem 👍  😄

PR #694, opened 2017-08-24
Configurable compilation environment

### LINKED ISSUE #691: Pass custom environment variables to the compiler
Given deployment flow using Capistrano, one would like to pass custom environment variables, like current revision hash or some access tokens (to upload source map to Rollbar or API key for Google Maps) to Webpack, so it can compile them into generated code.

I suppose it could be done by extending [`webpack_env`](https://github.com/rails/webpacker/blame/master/lib/webpacker/compiler.rb#L71) method on `Webpacker::Compiler`?

sjchmiela:
Fixes #691, as suggested by @gauravtiwari.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,12 @@
 
 - Rack middleware proxies webpacker requests to dev server so we can always serve from same-origin and the lookup works out of the box - no more paths prefixing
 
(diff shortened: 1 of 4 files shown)


gauravtiwari:
thanks @sjchmiela 🍰

PR #700, opened 2017-08-25
Improve hello examples

gauravtiwari:
This PR improves hello examples:

- Adds hot reloading support to react example
- Add style and image again for consistency
- Move out vue app component out of entry directory so it's more clear and consistent

screen shot 2017-08-25 at 18 19 33


dhh:
I'd like to keep these hellos as simple as possible. It's neat with the logos, but I don't think it's worth it. Same for splitting the react example into two separate files. This should be just enough to prove the framework is booting and nothing more.

gauravtiwari:
Makes sense 👍  although one of the things I wanted to address was consistency between examples so we don't send wrong message to users. We received a couple of issues where people were using packs directory for everything, which resulted in slow builds and even circular dependency problem. That's why moved both react and vue example into separate folder (vue in particular has two files, both into packs directory whereas angular has separate folder). 

Perhaps I can remove the images and then we can merge. What do you think?

dhh:
Hmm, I'm not sure we can fully educate people on this issue with the packs just through splitting up the files. I'd rather have a statement at the top of the single file about it, actually. And then just keep a single file if we can.

gauravtiwari:
Cool 👍  lets address this in docs once Javan's PR is merged for `node_modules `.

Issue #649, opened 2017-08-14
Creating a node module for webpacker...


justin808:
Are we considering moving some of this code:

https://github.com/rails/webpacker/tree/master/lib/install/config/webpack

into a node module?

gauravtiwari:
Some discourse here: https://github.com/rails/webpacker/pull/264

Would be ideal though to not run installer whenever these file changes.

dhh:
See #706 for the rebooted approach.

Issue #645, opened 2017-08-13
Slim down the README


javan:
The README has accumulated a lot of baggage that document webpacker changes, gotchas, and workarounds. I propose we pare it down to the basics and move non-essential bits to the wiki.

gauravtiwari:
👍  💯

dhh:
👍👍

justin808:
Rather than a wiki, consider having a `/docs`, like we've done for https://github.com/shakacode/react_on_rails/tree/master/docs

The problem with the wiki is that you cannot change the branch and get relevant docs. Plus, doc changes can go into PRs.

gauravtiwari:
@dhh May be we can have a page like this: http://guides.rubyonrails.org/asset_pipeline.html for webpacker too 😄  😉

jeromedalbert:
Looks like this issue can be closed

PR #706, opened 2017-08-28
Node module

javan:
Moves the bulk of our webpack config into a node module instead of copying it all to apps on `webpacker:install`. There are only four files copied to `config/webpack/` now:

**1** `config/webpack/environment.js` for setting up base configuration:
```js
const { environment } = require('webpacker')
module.exports = environment
```

**2-4** `config/webpack/{development,production,test}.js`, which build off `environment.js` and are identical to start:
```js
```

---
**TODO**

**Before merging**
- [x] Relax dependency versions?
- [x] Update docs

**After merging**
- [ ] Publish to npm
  - [ ] `$ npm version major`
  - [ ] `$ npm publish --access public`
---

Fixes https://github.com/rails/webpacker/issues/649
Closes https://github.com/rails/webpacker/pull/264


gauravtiwari:
@javan Just briefly looked at your changes and it's looking really good 👍  💯 

Happy to help with docs if you like or perhaps we can do it in separate PR i.e. move all README contents into `docs/` folder or something. I can do that in parallel. 

For `node_modules`, seems like someone already got https://www.npmjs.com/package/webpacker registered 😞

javan:
@gauravtiwari, I added new documentation and touched up some of the existing bits in 34142b1040657a0f35062f4cb997401d441b84a9. Let me know if I missed anything critical.

gauravtiwari:
@javan Sure thing, reviewing now. The API looks really nice though 💯 

We haven't published the package to npm yet right?

javan:
Right. Not published yet. Will eventually be published as a scoped package: `@rails/webpacker`.

Side note: We should gut the README as soon as this is merged.

gauravtiwari:
@javan Cool 👍 

Yepp, will do. Thinking we create a `docs/` folder and create `.md` for each topic. What do you think?

javan:
We can discuss more in https://github.com/rails/webpacker/issues/645, but basically I think we should remove everything that's webpack-not-webpacker related.

javan:
@gauravtiwari, can you please try this out in a real app?

You can do:
```rb
# Gemfile
gem 'webpacker', github: 'rails/webpacker', branch: 'node-module'
```

```js
// package.json
"dependencies": {
  "webpacker": "rails/webpacker#node-module"
}
```

and then `rails webpacker:install`

gauravtiwari:
@javan Yepp doing that now. 

BTW, is code like this intentional i.e. `if` block and rest of the code? 

```js
 const { extensions } = config
  if (!extensions.length) {
    throw new Error('You must configure at least one extension to compile in webpacker.yml')
  }
  return extensions.length === 1 ? `**/${extensions[0]}` : `**/*{${extensions.join(',')}}`
```

I corrected a few and if you don't mind I can push :)

javan:
Hmm, not seeing that locally or on github
https://github.com/rails/webpacker/blob/c4bf1c2144e9ef811797f6c48e2cd167ada92c05/package/environment.js#L33-L39

gauravtiwari:
Oops sorry I meant `lb` before/after `if` block.

javan:
I prefer the spacing as-is since it's a relatively short block of code. And the linter isn't complaining. 😬

package/index.js
```diff
@@ -0,0 +1,15 @@
+/* eslint global-require: 0 */
+/* eslint import/no-dynamic-require: 0 */
+
+const Environment = require('./environment')
+const { existsSync } = require('fs')
+
+function createEnvironment() {
+  const path = `./environments/${process.env.NODE_ENV}`
```
gauravtiwari: @javan Missing file extension here

javan: Not needed. I removed them all throughout the codebase. Just like Ruby, Node's `require` doesn't need an extension.

gauravtiwari: but `existSync` won't work, it requires a path 

```
const path = resolve(__dirname, `./environments/${process.env.NODE_ENV}.js`)
```

package.json
```diff
@@ -1,12 +1,39 @@
 {
-  "name": "webpacker",
-  "version": "1.0.0",
+  "name": "@rails/webpacker",
+  "version": "2.0.0",
   "description": "Webpacker makes it easy to use the JavaScript preprocessor and bundler [Webpack](https://webpack.github.io) to manage application-like JavaScript in Rails. It coexists with the asset pipeline, as the purpose is only to use Webpack for app-like JavaScript, not images, css, or even JavaScript Sprinkles (that all continues to live in app/assets).",
-  "main": "index.js",
+  "main": "package/index.js",
```
gauravtiwari: Need to add files we are distributing from NPM so the package doesn't include ruby files. 

```
  "files": [
    "package"
  ]
```

package/index.js
gauravtiwari:
```  
const constructor = existsSync(path) ? require(path) : Environment
```

Will always evaluate to false and return default config

javan: D'oh, right you are. Great catch! Fixed in df21faead2b427c56152eca744f12ce8fcc1e0ca

package.json
javan: 👍 e0ba006f64c5f4e1f871a8bd23817c65004b3583

gauravtiwari:
Great, will take a final look now and we can merge

README.md
```diff
@@ -842,7 +843,16 @@ yarn remove prop-types
...
 
-3. Finally add `.tsx` to the list of extensions in `config/webpacker.yml`
+3. Add `ts-loader` to `config/webpack/environment.js`
+
+```js
+environment.loaders.add('ts', {
+  test:  /.(ts|tsx)$/,
+  loader: 'ts-loader'
+})
+```
```
javan: I removed this step because we're adding all loaders by default now, but didn't notice the `tsx` extension part. Maybe we should just add it to the loader?
https://github.com/rails/webpacker/blob/3cd2e49b7943ddbea9da751a86b0ecdd80f2f61a/package/loaders/typescript.js#L1-L4

gauravtiwari: @javan Oh yes, totally forgot. Yes, makes sense 👍

README.md
```diff
@@ -202,16 +202,18 @@ experiment Vue right away.
...
+// config/webpack/environment.js
+const { environment } = require('webpacker')
 
-module.exports = {
-  test: /\.vue(\.erb)?$/,
-  ...
-}
+// Update an option directly
+const vueLoader = environment.loaders.get('vue')
+vueLoader.test = /\.vue(\.erb)?$/
```
javan: Should we make this the default `test` for the vue loader?
https://github.com/rails/webpacker/blob/3cd2e49b7943ddbea9da751a86b0ecdd80f2f61a/package/loaders/vue.js#L7

gauravtiwari: Yeah we can do that but I guess then we would need to update all loaders to support `.erb` (if we want that out of the box)?

javan: We already do for ~half of the loaders. Might as well for all?

gauravtiwari: Sounds good 👍  We also received this through many PR's and issues so makes sense 💯

javan: Cool. Want to make that change (and also add `tsx` to the typescript loader)?

gauravtiwari: on it 👍

README.md
```diff
@@ -90,7 +90,7 @@ in which case you may not even need the asset pipeline. This is mostly relevant
 
 ## Features
 
-* [Webpack 2](https://webpack.js.org/)
+* [Webpack 2 and 3](https://webpack.js.org/)
```
javan: Maybe drop the `2`? We technically don't support it now: https://github.com/rails/webpacker/blob/b4ded385e4cc520cc90ef46d3491ff0927d6ada8/package.json#L37

gauravtiwari: Hmm yeah, although the config we have will work with both versions but makes sense to just say 3. 

Also, we should add that module concatenation plugin? https://github.com/webpack/webpack/tree/master/examples/scope-hoisting#webpackconfigjs

gauravtiwari: Will push an update now 👍

javan: The config may work, but you won't be able to install v2 if you have webpacker as a dependency in package.json

PR #710, opened 2017-08-30
Start moving webpack docs

gauravtiwari:
This PR moves webpack related docs to `docs/` folder so README can stay clean and concise.


docs/assets.md
```diff
@@ -0,0 +1,47 @@
+## Link sprocket assets
```
javan: Let's fix all these heading levels. Should start with `#` and step into `##`, etc.

gauravtiwari: Yepp, will do - just moving them into files now and then format 👍

javan:
LGTM. Let's minimally link to `/docs` from the README.

PR #712, opened 2017-08-30
feat(angular): bootstrap on document loaded

gdi2290:


gauravtiwari:
@gdi2290 Thought this line `platformBrowserDynamic().bootstrapModule` already does the mounting on the browser. Are you getting any errors?

gdi2290:
this waits for the document to finish being parsed by the browser before changing the dom

gauravtiwari:
Gotcha 👍  Yeah, I was bit confused because example works without it so I wondered if `platformBrowserDynamic().bootstrapModule` took care of that part. Anyway merged 🎉  thanks

javan:
Is this what the Angular docs recommend? If you're using a `<script async>` tag, your js may load *after* `DOMContentLoaded` and the callback code will never run.

gdi2290:
I work with the Angular docs team. The problem (async tag) is also an issue with the other framework(s) example code. The developer can change their code later but here's the code to work with async/hmr/initial load
```js
function onDocumentReady (callback) {
  const readyState = document.readyState
  if (readyState === 'complete' || readyState === 'interactive') {
    return setTimeout(callback, 0, readyState)
  }
  function onLoad () {
    callback(readyState)
    document.removeEventListener('DOMContentLoaded', onLoad);
  }
  document.addEventListener('DOMContentLoaded', onLoad)
}

onDocumentReady(function(readyState) {
  platformBrowserDynamic().bootstrapModule(AppModule);
})
```

PR #716, opened 2017-08-31
Fix: NoMethodError - undefined method 'tagged' for #<ActiveSupport::Logger>

curvo:
This fixes a bug would be found in versions of Rails applications (<v5) that do not have the default logger set up for tagging.  The suggested fix is to check for the method before re-assigning the default webpacker logger to be the rails.logger.

### DIFF
--- a/lib/webpacker/railtie.rb
+++ b/lib/webpacker/railtie.rb
@@ -26,6 +26,8 @@ class Webpacker::Engine < ::Rails::Engine
     config.after_initialize do |app|
       if ::Rails.logger.respond_to?(:tagged)
         Webpacker.logger = ::Rails.logger
+      else
+        Webpacker.logger = ActiveSupport::TaggedLogging.new(::Rails.logger)
       end
     end
   end
--- a/lib/webpacker/railtie.rb
+++ b/lib/webpacker/railtie.rb
@@ -24,7 +24,9 @@ class Webpacker::Engine < ::Rails::Engine
 
   initializer "webpacker.logger" do
     config.after_initialize do |app|
-      Webpacker.logger = ::Rails.logger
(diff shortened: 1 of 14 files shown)


curvo:
screenshot 2017-08-30 18 54 31

lib/webpacker/railtie.rb
```diff
@@ -24,7 +24,9 @@ class Webpacker::Engine < ::Rails::Engine
 
   initializer "webpacker.logger" do
     config.after_initialize do |app|
-      Webpacker.logger = ::Rails.logger
+      if ::Rails.logger.respond_to?(:tagged)
+        Webpacker.logger = ::Rails.logger
+      end
```
javan: Could create a tagged logger here to preserve the original logger:
```ruby
if ::Rails.logger.respond_to?(:tagged)
  Webpacker.logger = ::Rails.logger
else
  Webpacker.logger = ActiveSupport::TaggedLogging.new(::Rails.logger)
end
```

gauravtiwari: I guess we are already doing this here right? https://github.com/rails/webpacker/blob/68b9f086c1d262be895c2750cb3a25879157733a/lib/webpacker/instance.rb#L2

javan: That creates a brand new logger. Doesn't wrap your existing `Rails.logger` instance.

gauravtiwari: Ahh I see, totally missed 👍

curvo: Makes sense... I'll push that today.

gauravtiwari:
thanks  🍰

curvo:
Happy to contribute!

PR #717, opened 2017-08-31
FIX: CI Test Failure (test_lookup_exception)

curvo:
This offers a simpler assertion to ensure that the Webpacker::Manifest::MissingEntryError includes the name of the missing file when it's not found.  The exact contents of the message output may change over time, but the missing file name should always be included.  This fixes the broken test and helps make it less brittle moving forward.

### DIFF
--- a/test/manifest_test.rb
+++ b/test/manifest_test.rb
@@ -10,23 +10,7 @@ class ManifestTest < Minitest::Test
         Webpacker.manifest.lookup(asset_file)
       end
 
-      expected = <<-MSG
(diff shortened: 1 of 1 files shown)


javan:
Wups, I *just* fixed this before seeing your PR. Thank you! https://github.com/rails/webpacker/commit/257cde4955bdbf646155f65ce1516ae7b3416abd

curvo:
No problem... your fix is better.  Thanks for looking.  I have a couple others open... maybe something will be useful?

PR #718, opened 2017-08-31
Fix: updated Gemfile.lock to match master

curvo:
The current Gemfile.lock file is out-of-sync with the release of the 3.0.0 version of this gem.  This updates the Gemfile.lock to current levels.


javan:
Thanks!

Issue #505, opened 2017-06-15
no source maps


elbarto132:
If i create a project using for example "rails new myapp --webpack=react" (doesn't matter if i use react or vue) and then run ./bin/webpack-dev-server or ./bin/webpack it should generate source map files besides of the js files, right? I'm using rails 5.1.1 and with the default configuration it doesn't create any source map files. What do i have to change to get source maps?

Update:
If i change "cheap-eval-source-map" to "source-map" in config/webpack/development.js it's creating the source maps, but according to https://webpack.js.org/configuration/devtool/ that's probably not the desired option

gauravtiwari:
@elbarto132 If you run ` NODE_ENV=production ./bin/webpack` you will have sourcemaps files. The development environment uses - 

```js
//config/webpack/development.js
devtool: 'cheap-eval-source-map'
```

ref: https://webpack.js.org/configuration/devtool/#devtool

elbarto132:
@gauravtiwari that's probably true, because production.js is setting devtool: 'source-map', and i didn't know i had to set NODE_ENV=production too

I tried all of the possible options for devtools in development.js
All the "development options", eval, cheap-eval-source-map, cheap-module-eval-source-map, eval-source-map don't produce source map files, but probably they should?

gauravtiwari:
@elbarto132 Just use `devtool: 'source-map'` in development.js and it will output sourcemaps for you (a bit heavy for compiler though). No rest of the options are outputs inline with JS file.

elbarto132:
Yes, i will do this for now, i was just confused that the default configuration isn't working properly.
So maybe the default devtool option should be changed until the real issue, which seems to be this: https://github.com/webpack/webpack/issues/2145 is solved

gauravtiwari:
@elbarto132 That's default for faster compilation but feel free to change it if you feel that works better for you.

alex-felix:
On a fresh `rails new` install (5.1.4), source map won't work in development on Chrome and Firefox. To switch to another `devtool` which actually work, you can do that:

```js
// config/webpack/development.js
const environment = require('./environment')

const config = environment.toWebpackConfig();
config.devtool = 'sourcemap';

module.exports = config;
```

gauravtiwari:
Yeah, we have changed the default source map option. Will make a new release soon

edelgado:
Any idea why the source maps are showing ES5 code and not the source code as in my ES6 source?

Stafie:
[...]

Issue #390, opened 2017-05-15
Missing SourceMaps for CSS in Development


dpuscher:
I am using CSS modules in my react project which works quite well out of the box using webpacker. The only problem are the SourceMaps. In production everything is working well, but not so in development.

In production, webpack is generating `*.css`, `*.css.gz` and `*.css.map` when using
```
devtool: "source-map",
plugins: [
  new webpack.optimize.UglifyJsPlugin({
    sourceMap: true
  }),
  ...
```
in `config/webpack/production.js`.

In development I am using a different devtool setting, since "source-map" is too slow ([Documentation](https://webpack.js.org/configuration/devtool/))
```
devtool: "cheap-module-eval-source-map",
``` 
which works fine for JS-files (adding `sourceMappingURL=data:application/json;charset=utf-8;base64...` at the end of each module).
But it is sadly not working for my css/sass files...

Here my `sass` loader configuration:
```
const ExtractTextPlugin = require("extract-text-webpack-plugin");
const { dev } = require("../configuration.js");

module.exports = {
  test: /\.(scss|sass)$/i,
  use: ExtractTextPlugin.extract({
    fallback: "style-loader",
    use: [
      {
        loader: "css-loader",
        options: {
          modules: true,
          sourceMap: true,
          camelCase: "only",
          minimize: !dev && {
            autoprefixer: {
              browsers: ["> 1%", "last 2 versions", "IE > 9"],
              add: true
            }
          }
        }
      }, {
        loader: "sass-loader",
        options: {
          sourceMap: true,
          indentedSyntax: true
        }
      }
    ]
  })
};
```

gauravtiwari:
@dpuscher Were you able to get this to work? Sorry for late response.

dpuscher:
@gauravtiwari No, I couldn't get it to work... Do you have any idea?

gauravtiwari:
@dpuscher Seems to be working for me with this setup - 

```js
// sass.js loader
const ExtractTextPlugin = require('extract-text-webpack-plugin')
const { env } = require('../configuration.js')

module.exports = {
  test: /\.(scss|sass|css)$/i,
  use: ExtractTextPlugin.extract({
    fallback: 'style-loader',
    use: [
      { loader: 'css-loader', options: { minimize: env.NODE_ENV === 'production', sourceMap: true } },
      { loader: 'postcss-loader', options: { sourceMap: true } },
      'resolve-url-loader',
[...]

PR #729, opened 2017-08-31
allow devtool settings to override defaults

tyrp:
Prior to config files being pulled into the package with the v3 update, I was able to override the `devtool` setting in my `config/webpack/environment.js` file to fix source maps not working in development. (Similar to issues #505 and #390) However, now it seems any `devtool` I set is being ignored. The environments are hard coded to the defaults now. https://github.com/rails/webpacker/blob/master/package/environments/development.js#L22

Thanks!

### DIFF
--- a/package/environment.js
+++ b/package/environment.js
@@ -66,7 +66,7 @@ module.exports = class Environment {
   }
 
   toWebpackConfig() {
-    return {
+    const result = {
       entry: getEntryObject(),
 
       output: {
@@ -91,5 +91,11 @@ module.exports = class Environment {
         modules: ['node_modules']
       }
     }
+
+    if (this.devtool) {
+      result.devtool = this.devtool
+    }
+
+    return result
   }
 }
--- a/package/environments/development.js
+++ b/package/environments/development.js
@@ -19,7 +19,7 @@ module.exports = class extends Environment {
       result.output.filename = '[name]-[hash].js'
     }
     result.output.pathinfo = true
-    result.devtool = 'cheap-eval-source-map'
+    result.devtool = result.devtool || 'cheap-eval-source-map'
     result.devServer = {
       host: dev_server.host,
       port: dev_server.port,
--- a/package/environments/production.js
+++ b/package/environments/production.js
@@ -27,7 +27,7 @@ module.exports = class extends Environment {
 
   toWebpackConfig() {
     const result = super.toWebpackConfig()
-    result.devtool = 'source-map'
+    result.devtool = result.devtool || 'source-map'
     result.stats = 'normal'
     return result
   }


javan:
In 3.0 you can change your environment's config like:
```js
const config = environment.toWebpackConfig()
config.devtool = ...
module.exports = config
```

Or you can make that change in your `config/webpack/development.js` if that's the only environment you want it in.

8bitDesigner:
@javan since the `{test,development,production}.js` files call `toWebpackConfig()` on the base environment, I ended up replacing the `toWebpackConfig` with one that would return a modified config:

``` javascript
const oldToConfigMethod = environment.toWebpackConfig
environment.toWebpackConfig = function () {
  const config = oldToConfigMethod.call(environment)
  config.foo = config.bar // overrides...
  return config
}
```

javan:
> on the base environment

That's not a base `environment` instance exactly, it's an instance for the current `NODE_ENV`:
https://github.com/rails/webpacker/blob/fb86ef13264a18f7eb33c1c1373549a5dc9953dc/package/index.js#L8-L16

So when `NODE_ENV=development` it's a:
https://github.com/rails/webpacker/blob/fb86ef13264a18f7eb33c1c1373549a5dc9953dc/package/environments/development.js#L6

8bitDesigner:
Ah, right! However, `environment` as imported, is going to keep the same "shape" though, which keeps all the configuration behind a `toWebpackConfig()` method.

So if you export `module.exports = require('./environment').toWebpackConfig()` in your project's `environment.js`, then `development.js` is going to throw a reference error when it tries to call `.toWebpackConfig()` on a plain object:

https://github.com/rails/webpacker/blob/fb86ef13264a18f7eb33c1c1373549a5dc9953dc/lib/install/config/webpack/development.js#L1-L3

We've had to do this to manually add `externals` to the webpack config for all our environments.

javan:
Ah, right from me too! :)

Another option might be additing your own `config/webpack/common-config.js` (or whatever you want to name it) and modify your env js files to merge it with `Object.assign` or `webpack-merge`. Something like:

```js
// config/webpack/development.js

const commonConfig = require('./common-config')
const merge = require('webpack-merge')

module.exports = merge(environment.toWebpackConfig(), commonConfig)
```

8bitDesigner:
Yep, I think that's the way to handle it. Long term, I'd love the `Environment` class to be extensible, so we could subclass it like:

``` javascript
class MyEnvironment extends Environment {
  toWebpackConfig () {
    const config = super.toWebpackConfig()
    config.foo = 'bar'
    return config
  }
}
```

I think the only blocker here is that I can't get the right `environment` object, _from_ the `Environment` class.

Moving this `createEnvironment` method into a super method on `Environment` might do the trick

https://github.com/rails/webpacker/blob/5d45278e99d25f8b1e98687ba3491720cc73f1bb/package/index.js#L8-L12

So, my override would become

``` javascript
class MyEnvironment extends Environment {
  constructor () {
    return Environment.createEnvironment() // returns ./{development,production,environment}.js
  }

  toWebpackConfig () {
    const config = super.toWebpackConfig()
    config.foo = 'bar'
    return config
  }
}
```

gauravtiwari:
@javan I think we can also import `Environment` class as well right? https://github.com/rails/webpacker/blob/master/package/index.js#L16

So I think that's possible now.

8bitDesigner:
You definitely could, but if you're in production, importing `Environment` will give you the base Environment class, not the extended dev/prod/test classes as that is exported to `environment` by `createEnvironment`.

I've seen a few other issues about configuring the base webpack config; do you think this is worth moving the conversation to a new issue? I could take a stab at a PR to allow extending environment while respecting the dev/prod/test configs.

javan:
This works (correctly, I *think*):
```js

const CustomEnvironment = class extends environment.constructor {
  toWebpackConfig() {
    const config = super.toWebpackConfig()
    config.foo = 'bar'
    return config
  }
}

module.exports = new CustomEnvironment
```

8bitDesigner:
That works - hey, any reason why `Environment` is being exported instead
of the constructor from `createEnvironment`? I think that tidies up a
lot of things.

On Fri, Sep 1, 2017, at 13:55, Javan Makhmali wrote:
> This works (correctly, I *think*):


> // config/webpack/environment.js

> const { environment } = require('@rails/webpacker')

> const CustomEnvironment = class extends environment.constructor {
> toWebpackConfig() {  const config = super.toWebpackConfig()
> config.foo = 'bar'  return config } }

> module.exports = new CustomEnvironment> — You are receiving this because you commented. Reply to this email
> directly, view it on GitHub[1], or mute the thread[2].> 


Links:

  1. https://github.com/rails/webpacker/pull/729#issuecomment-326681853

Issue #709, opened 2017-08-29
Need option to bind server to 0.0.0.0 in development mode when using Vagrant


kiere:
Original Issue Title: **manifest.json not respecting "public" option to devServer settings (Vagrant nginx proxy)**

I am using webpacker 2.0 on Rails 4.2.  I have run `webpack:install` and `webpack:install:elm`.  Our setup uses Vagrant with the VM aliased to something similar to `myapp.test` using `vagrant-hostmanager` plugin.  We want to be able to use `foreman start` to spin up Unicorn and `webpack-dev-server` inside the VM and be able to access the Rails app from our local dev laptop (outside the VM) using `http://myapp.test`.  This all works fine except for the "packs" assets because accessing `http://0.0.0.0:8080` from the local laptop will not get through to the VM.  So, adding the `public` option to the devServer settings (through webpacker.yml, or development.js directly, or as a `--public` option to `bin/webpack-dev-server` WILL allow me access the "packs" resources from the dev laptop using the host alias (http://myapp.test:8080/packs/application.js), but the scripts inside the Rails view template will not reflect the `public` option and continues to render the `host` and `port` option.

```HTML
<script src="http://0.0.0.0:8080/packs/application.js"></script>
<script src="http://0.0.0.0:8080/packs/hello_elm.js"></script>
```

```Yaml
dev_server:
    host: 0.0.0.0
    port: 8080
    https: false
    public: myapp.test:8082  # <--- Line added
```

```JavaScript
devServer: {
    clientLogLevel: 'none',
    https: settings.dev_server.https,
    host: settings.dev_server.host,
    port: settings.dev_server.port,
    contentBase: output.path,
    public: settings.dev_server.public, // <--- Line added
    publicPath: output.publicPath,
    compress: true,
    headers: { 'Access-Control-Allow-Origin': '*' },
    historyApiFallback: true,
    watchOptions: {
      ignored: /node_modules/
    }
  }
```

For example, if I open up 3 browser windows... 1.) http://myapp.test  2.) http://myapp.test:8080/packs/application.js 3.) http://myapp.test/packs/manifest.json, I will see my app in the first window, the output of application.js in the second window, but in the 3rd window, I will see the JSON using http://0.0.0.0:8080 (instead of http://myapp.test:8080).  This will result in 2 console errors in the first tab (where my app is) referencing connection errors to http://0.0.0.0:8080.

[...]

Issue #572, opened 2017-07-18
Bind address configuration confusion & best practices


inopinatus:
I'm concerned about the way the bind address for the development server is configured.

Binding to 0.0.0.0 by default is unsafe, it puts anyone using shared/untrusted networks at risk.

Is there a good reason not to have behaviour congruent to Rack behaviour, which defaults to localhost in development unless otherwise configured? See https://github.com/rack/rack/commit/28b014484a8ac0bbb388e7eaeeef159598ec64fc and this was confirmed in Rails 4.2 release notes.

I also think 99% of people, on seeing a configuration option for "host" and "port" in a section named "something_server" will be astonished to learn that it does not configure the host, but does configure the port, for that server. Surely a mistake? There's basically no way to configure the host binding short of editing bin/webpack-dev-server.

I think webpacker should, as much as possible, follow the same defaults and conventions as rack & rails with the listeners and asset URLs, with optional config of either in webpacker.yml.

Happy to put together a PR expressing this opinion.

dhh:
Happy to see an option for this to be configured 👍

PR #731, opened 2017-08-31
Make dev server listen host to be configurable

gauravtiwari:
This PR makes dev server listen host configurable:  

Fixes: https://github.com/rails/webpacker/issues/709
Ref: https://github.com/rails/webpacker/issues/572 

Not sure if this approach is better or using an env variable?

### DIFF
--- a/lib/install/bin/webpack-dev-server.tt
+++ b/lib/install/bin/webpack-dev-server.tt
@@ -16,7 +16,7 @@ CONFIG_FILE       = File.join(APP_PATH, "config/webpacker.yml")
 NODE_MODULES_PATH = File.join(APP_PATH, "node_modules")
 WEBPACK_CONFIG    = File.join(APP_PATH, "config/webpack/#{NODE_ENV}.js")
 
-LISTEN_HOST_ADDR = NODE_ENV == 'development' ? 'localhost' : '0.0.0.0'
+DEFAULT_LISTEN_HOST_ADDR = NODE_ENV == 'development' ? 'localhost' : '0.0.0.0'
 
 def args(key)
   index = ARGV.index(key)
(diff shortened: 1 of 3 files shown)


javan:
What's the reason for CLI args generally when you can configure `webpacker.yml`?

gauravtiwari:
Current we have hardcoded listen address: https://github.com/rails/webpacker/blob/master/lib/install/bin/webpack-dev-server.tt#L19

and https://github.com/rails/webpacker/blob/master/lib/install/bin/webpack-dev-server.tt#L60

lib/install/bin/webpack-dev-server.tt
```diff
@@ -47,9 +53,10 @@ rescue Errno::EADDRINUSE
 end
 
 # Delete supplied host and port CLI arguments
-["--host", "--port"].each do |arg|
+["--host", "--port", "--listen-host"].each do |arg|
   ARGV.delete(args(arg))
   ARGV.delete(arg)
+  ARGV.delete(arg)
```
curvo: Why did you add this line?  Isn't it redundant with the one above it?

lib/install/bin/webpack-dev-server.tt
```diff
@@ -16,13 +16,19 @@ CONFIG_FILE       = File.join(APP_PATH, "config/webpacker.yml")
...
 WEBPACK_CONFIG    = File.join(APP_PATH, "config/webpack/#{NODE_ENV}.js")
 
-LISTEN_HOST_ADDR = NODE_ENV == 'development' ? 'localhost' : '0.0.0.0'
-
 def args(key)
   index = ARGV.index(key)
   index ? ARGV[index + 1] : nil
 end
 
+LISTEN_HOST_ADDR = if args('--listen-host')
```
curvo: I think this could be cleaner with something like this...

First, just rename the current LISTEN_HOST_ADDR to something like DEFAULT_LISTEN_HOST_ADDR:

```ruby
DEFAULT_LISTEN_HOST_ADDR = NODE_ENV == 'development' ? 'localhost' : '0.0.0.0'
```

Then within the begin and parallel to the other constants like (DEV_SERVER_ADDR) do this:

```ruby
  HOSTNAME        = args('--host') || dev_server["host"]
  PORT            = args('--port') || dev_server["port"]
  HTTPS           = ARGV.include?('--https') || dev_server["https"]
  DEV_SERVER_ADDR = "http#{"s" if HTTPS}://#{HOSTNAME}:#{PORT}"
  LISTEN_HOST_ADDR = args('--listen-host') || DEFAULT_LISTEN_HOST_ADDR
```

lib/install/bin/webpack-dev-server.tt
```diff
@@ -47,9 +53,10 @@ rescue Errno::EADDRINUSE
 end
 
 # Delete supplied host and port CLI arguments
```
curvo: Might want to update comment as well:

```ruby
# Delete supplied host, port and listen-host CLI arguments
```

lib/install/bin/webpack-dev-server.tt
gauravtiwari: thanks, pushed an update 👍  💯

lib/install/bin/webpack-dev-server.tt
gauravtiwari: haha, don't know - removed it 👍

lib/install/bin/webpack-dev-server.tt
gauravtiwari: Done 👍

PR #734, opened 2017-09-01
Change coffee-script npm dep to coffeescript

swrobel:
coffee-loader depends on coffeescript so this stops yarn from complaining. The coffeescript & coffee-script packages are the same, and without this change yarn complains like so on every run `warning "coffee-loader@0.8.0" has unmet peer dependency "coffeescript@>= 1.8.x".`


gauravtiwari:
Thanks!

PR #740, opened 2017-09-01
Yarn engine requirement

javan:
Adds yarn requirement to `package.json`s engines:
https://github.com/rails/webpacker/blob/09247aaf8efc093d5527a58da0ab22a90a5a9c7e/package.json#L9-L11

Updates `webpacker:check_yarn` and `webpacker:check_node` to read requirements from `package.json` so they're not duplicated (and easily forgotten):
https://github.com/rails/webpacker/blob/09247aaf8efc093d5527a58da0ab22a90a5a9c7e/lib/tasks/webpacker/check_yarn.rake#L8-L14

### DIFF
--- a/lib/tasks/webpacker/check_node.rake
+++ b/lib/tasks/webpacker/check_node.rake
@@ -4,12 +4,16 @@ namespace :webpacker do
     begin
       node_version = `node -v`
       node_version = `nodejs -v` if node_version.blank?
-      required_node_version = "6.0"
-
       raise Errno::ENOENT if node_version.blank?
 
-      if Gem::Version.new(node_version.strip.tr("v", "")) < Gem::Version.new(required_node_version)
-        $stderr.puts "Webpacker requires Node.js >= v#{required_node_version} and you are using #{node_version}"
+      pkg_path = Pathname.new("#{__dir__}/../../../package.json").realpath
+      node_requirement = JSON.parse(pkg_path.read)["engines"]["node"]
+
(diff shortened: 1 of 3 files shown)


lib/tasks/webpacker/check_node.rake
```diff
@@ -4,12 +4,16 @@ namespace :webpacker do
...
-
       raise Errno::ENOENT if node_version.blank?
 
-      if Gem::Version.new(node_version.strip.tr("v", "")) < Gem::Version.new(required_node_version)
-        $stderr.puts "Webpacker requires Node.js >= v#{required_node_version} and you are using #{node_version}"
+      pkg_path = Pathname.new("#{__dir__}/../../../package.json").realpath
+      node_requirement = JSON.parse(pkg_path.read)["engines"]["node"]
+
+      requirement = Gem::Requirement.new(node_requirement)
+      version = Gem::Version.new(node_version.strip.tr("v", ""))
```
gauravtiwari: Shall we extract this part inside version module? And then perhaps use that to check inside tasks? 

```rb
if Webpacker.required_node_version?(version)
if Webpacker.required_yarn_version?(version)
```

Or is it overkill?

javan: I'd rather not add a new API that's only used by rake tasks, but 👍 if there's ever a broader need.

gauravtiwari: Cool, makes sense 👍

PR #741, opened 2017-09-01
Remove lock files from version control

javan:
> **When developing a gem**, use the gemspec method in your Gemfile to avoid duplication. In general, a gem's Gemfile should contain the Rubygems source and a single gemspec line. **Do not check your Gemfile.lock into version control**

http://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/


benpickles:
Though Yarn's advice is the opposite:

> Framework and library authors should also check yarn.lock into source control. Don’t worry about publishing the yarn.lock file as it won’t have any effect on users of the library.

https://yarnpkg.com/lang/en/docs/yarn-lock/#toc-check-into-source-control

javan:
Good to know, but doesn't matter much in this case. From a development perspective, Webpacker only really relies on eslint and locking it to precise versions doesn't concern me. Plus this makes me happy:
screen shot 2017-09-01 at 12 29 39 pm

rossta:
I tried to open a PR for this awhile back—not sure if the policy changed? See #501.

kaspth:
Our policy is still what @rafaelfranca said in https://github.com/rails/rails/pull/18992. This should be reverted.

javan:
Is that the policy for core Rails gems or every gem under the Rails org? There are plenty with no lock file:
- https://github.com/rails/jbuilder
- https://github.com/rails/sprockets-rails
- https://github.com/rails/sprockets
- https://github.com/rails/spring

https://github.com/rails/webpacker/pull/718 is what prompted this change. I'm not sure when/how the version in Gemfile.lock is supposed to be bumped.

kaspth:
I thought we already followed it everywhere. 😕

True I haven't got it down to an exact science either. We do issue a `bundle check` before releasing though: https://github.com/rails/rails/blob/master/tasks/release.rb#L184

Issue #745, opened 2017-09-01
Getting Errno::ETIMEDOUT


mmaauu:
Hi! I've added the webpacker gem, after bundle install with the gem, I run rails s and I'm getting this error, I can't figure out what's going on.
Thanks!

``
Puma caught this error: A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond. - user specified timeout (Errno::ETIMEDOUT)
c:/RailsInstaller/Ruby2.3.0/lib/ruby/2.3.0/socket.rb:61:in `connect_internal'
c:/RailsInstaller/Ruby2.3.0/lib/ruby/2.3.0/socket.rb:139:in `connect'
c:/RailsInstaller/Ruby2.3.0/lib/ruby/2.3.0/socket.rb:636:in `block in tcp'
c:/RailsInstaller/Ruby2.3.0/lib/ruby/2.3.0/socket.rb:231:in `each'
c:/RailsInstaller/Ruby2.3.0/lib/ruby/2.3.0/socket.rb:231:in `foreach'
c:/RailsInstaller/Ruby2.3.0/lib/ruby/2.3.0/socket.rb:626:in `tcp'
c:/RailsInstaller/Ruby2.3.0/lib/ruby/gems/2.3.0/gems/webpacker-3.0.0/lib/webpacker/dev_server.rb:9:in `running?'
c:/RailsInstaller/Ruby2.3.0/lib/ruby/gems/2.3.0/gems/webpacker-3.0.0/lib/webpacker/dev_server_proxy.rb:7:in `rewrite_response'
c:/RailsInstaller/Ruby2.3.0/lib/ruby/gems/2.3.0/gems/rack-proxy-0.6.2/lib/rack/proxy.rb:57:in `call'
c:/RailsInstaller/Ruby2.3.0/lib/ruby/gems/2.3.0/gems/railties-5.1.3/lib/rails/engine.rb:522:in `call'
c:/RailsInstaller/Ruby2.3.0/lib/ruby/gems/2.3.0/gems/puma-3.10.0/lib/puma/configuration.rb:225:in `call'
c:/RailsInstaller/Ruby2.3.0/lib/ruby/gems/2.3.0/gems/puma-3.10.0/lib/puma/server.rb:605:in `handle_request'
c:/RailsInstaller/Ruby2.3.0/lib/ruby/gems/2.3.0/gems/puma-3.10.0/lib/puma/server.rb:437:in `process_client'
[...]

PR #753, opened 2017-09-04
Catch socket timeout and make it configurable

gauravtiwari:
Fixes: https://github.com/rails/webpacker/issues/745

The only side effect of Memoization here is the compiler can't be switched at runtime i.e. starting webpack-dev-server in the middle whilst it's using on-demand compiler, rails server need to be restarted. 

```rb
# with memoization
0.000000   0.010000   0.010000 (  0.002971)
0.000000   0.000000   0.000000 (  0.003402)

0.000000   0.000000   0.000000 (  0.001943)
0.000000   0.000000   0.000000 (  0.001906)

0.000000   0.000000   0.000000 (  0.004089)
0.000000   0.010000   0.010000 (  0.005296)

0.010000   0.000000   0.010000 (  0.002744)
0.000000   0.000000   0.000000 (  0.002775)

# without memoization
0.010000   0.000000   0.010000 (  0.009734)
0.010000   0.000000   0.010000 (  0.005229)

0.000000   0.000000   0.000000 (  0.005725)
0.000000   0.000000   0.000000 (  0.004190)

0.000000   0.000000   0.000000 (  0.008097)
0.000000   0.000000   0.000000 (  0.004188)

0.000000   0.010000   0.010000 (  0.004776)
0.010000   0.000000   0.010000 (  0.004454)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+## Unreleased
+
+- Allow dev server connect timeout(in seconds) to be configurable, default: 0.01
+
+```rb
+# Change to 1s
+Webpacker::DevServer.connect_timeout = 1
+```
+
 ## [3.0.1] - 2017-09-01
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,10 @@
 ## Unreleased
 
-- Allow dev server connect timeout(in seconds) to be configurable, default: 0.01
+- Allow dev server connect timeout (in seconds) to be configurable, default: 0.01
 
 ```rb
 # Change to 1s
-Webpacker::DevServer.connect_timeout = 1
+Webpacker.dev_server.connect_timeout = 1
 ```
 
 ## [3.0.1] - 2017-09-01
--- a/lib/webpacker/dev_server.rb
+++ b/lib/webpacker/dev_server.rb
@@ -6,10 +6,14 @@ class Webpacker::DevServer
   end
 
   def running?
-    Socket.tcp(host, port, connect_timeout: 1).close
-    true
-  rescue Errno::ECONNREFUSED, NoMethodError
-    false
+    return @running if defined?(@running)
+
+    @running = begin
+        Socket.tcp(host, port, connect_timeout: 0.1).close
+        true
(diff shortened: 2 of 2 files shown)


javan:
> The only side effect of Memoization here is the compiler can't be switched at runtime

That's a big step back, IMO. Especially to work around what might be a macOS High Sierra (beta) issue/bug. I'd start without memoizing the result. Or set it up so the result is only cached for a second or two.

lib/webpacker/dev_server.rb
```diff
@@ -6,10 +6,14 @@ def initialize(webpacker)
...
 
   def running?
-    Socket.tcp(host, port, connect_timeout: 1).close
-    true
-  rescue Errno::ECONNREFUSED, NoMethodError
-    false
+    return @running if defined?(@running)
+
+    @running = begin
+        Socket.tcp(host, port, connect_timeout: 0.1).close
```
javan: In my testing (https://github.com/rails/webpacker/issues/745#issuecomment-326973005), `connect_timeout` could be even shorter. Could start with `0.01` and make it configurable (with a `mattr_accessor`) to accommodate slower systems. Then memoization won't be necessary.

gauravtiwari: Yeah, I was bit afraid as well but wanted to get your take on memoization. Pushed an update with configurable timeout 👍

CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+## Unreleased
+
+- Allow dev server connect timeout(in seconds) to be configurable, default: 0.01
```
javan: Missing a space before the `(`

CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
...
+
+```rb
+# Change to 1s
+Webpacker::DevServer.connect_timeout = 1
```
javan: Prefer the instance? `Webpacker.dev_server.connect_timeout`

lib/webpacker/dev_server.rb
```diff
@@ -1,14 +1,18 @@
 class Webpacker::DevServer
+  # Configure dev server connection timeout(in seconds), default: 0.01
+  # Webpacker::DevServer.connect_timeout = 1
```
javan: Ditto for my two comments above ^^

PR #758, opened 2017-09-05
Revert "Remove lock files from version control"

javan:
Reverts rails/webpacker#741
Plays by the rules https://github.com/rails/rails/pull/18992


kaspth:
Awesome thanks ❤️

PR #770, opened 2017-09-07
Restricts the source maps generated for production

### LINKED ISSUE #769: Is enabling full source maps in production a wise default?
This ticket is about the default production configuration (both the webpack & uglify settings) in this file: https://github.com/rails/webpacker/blob/50aaf772a50e589f7ddf4917a3dcd22e42ee3adc/package/environments/production.js#L30

In my experience in (mostly Node.js-based) static asset compilation ecosystems, I haven't run across very many cases where full source maps are, by default, generated and pushed to production servers. I presume the reason for this is because that modern JS build toolchains provide a sort of accidental benefit of code obfuscation, and well-generated source maps can actually make it very easy to rip off someone else's source.

The webpack docs don't actually justify this statement, but they [say this much](https://webpack.js.org/configuration/devtool/#production) about using the `source-map` setting in production:

> You should configure your server to disallow access to the Source Map file for normal users!

I've looked through the repo and was curious if there was much discussion/justification for enabling full source-maps in production by default and whether it'd be fair to revisit the issue in a thread here.

Curious to hear your thoughts & thanks for your time! ✨

searls:
This will generate source maps only to the extent required for stack
trace reporting tools like Sentry and Rollbar. It should result in build
time savings and prevent people's unobfuscated source code from
"leaking" with every production build. 

Fixes #769

### DIFF
--- a/package/environments/production.js
+++ b/package/environments/production.js
@@ -27,7 +27,7 @@ module.exports = class extends Environment {
 
   toWebpackConfig() {
     const result = super.toWebpackConfig()
-    result.devtool = 'source-map'
+    result.devtool = 'nosources-source-map'
     result.stats = 'normal'
     return result
   }


guilleiguaran:
Thank you very much!!!

manojmj92:
Congrats on your first PR merge!

PR #774, opened 2017-09-07
Fallback to RACK_ENV environment variable before defaulting to development mode.

### LINKED ISSUE #773: binstubs should honor the RACK_ENV environment variable in addition to the RAILS_ENV environment variable
Currently, the `webpack` and `webpack-dev-server` binstubs only look at the `RAILS_ENV` environment variable to determine what the Rails environment is:

https://github.com/rails/webpacker/blob/daffa50b1a49b64d60d5f0f857f44fa134359258/lib/install/bin/webpack.tt#L6-L7

We run our production environment with `RAILS_ENV` unset, and `RACK_ENV` set to "production". This is fine because **Rails** will fallback to the `RACK_ENV` environment variable if `RAILS_ENV` is unset:

https://github.com/rails/rails/blob/v5.1.3/railties/lib/rails.rb#L71

```ruby
def env
  @_env ||= ActiveSupport::StringInquirer.new(ENV["RAILS_ENV"].presence || ENV["RACK_ENV"].presence || "development")
end
```

However, when running webpacker in our production environment, we end up with `RACK_ENV=production` and `RAILS_ENV=development`, which causes asset compilation to fail.

I will send in a PR shortly with a change to make the webpacker binstubs behave more like Rails does.

jfly:
This fixes #773.

### DIFF
--- a/lib/install/bin/webpack-dev-server.tt
+++ b/lib/install/bin/webpack-dev-server.tt
@@ -5,7 +5,7 @@ require "shellwords"
 require "yaml"
 require "socket"
 
-ENV["RAILS_ENV"] ||= "development"
+ENV["RAILS_ENV"] ||= ENV["RACK_ENV"] || "development"
 RAILS_ENV = ENV["RAILS_ENV"]
 
 ENV["NODE_ENV"] ||= RAILS_ENV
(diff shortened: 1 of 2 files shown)


gauravtiwari:
thanks @jfly 🍰

PR #776, opened 2017-09-08
Refactored defaults code for readability

januszm:
It should be easier to debug and removes unnecessary new line escape `\`


kaspth:
Think it reads better as is 😊

PR #777, opened 2017-09-08
Explicitly documents how to enable CSS modules

ezuk:
This technique is shown in the document but I wanted to make it more explicit for CSS modules, as it seems like I was not the only one confused. ;)

### DIFF
--- a/docs/webpack.md
+++ b/docs/webpack.md
@@ -80,9 +80,7 @@ module.exports = environment
 
 ### Overriding Loader Options in Webpack 3+ (for CSS Modules etc.)
 
-In Webpack 3+, if you'd like to specify additional or different options for a loader, editing the actual `config/webpack/{loader}.js` file is *not* the way to go.
-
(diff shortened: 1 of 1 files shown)


docs/webpack.md
```diff
@@ -78,6 +78,31 @@ babelLoader.options.cacheDirectory = false
 module.exports = environment
 ```
 
+### Overriding Loader Options in Webpack 3+ (for CSS Modules etc.)
+
+In Webpack 3+, if you'd like to specify additional or different options for a loader, editing the actual `config/webpack/{loader}.js` file is *not* the way to go.
```
gauravtiwari: Lets remove reference to this `config/webpack/{loader}.js`, since it no longer exists.  All base loaders are inside npm module now.

ezuk:
@gauravtiwari thank you! How's this?

gauravtiwari:
@ezuk great thanks 🍰  💯

PR #778, opened 2017-09-08
Rename public_manifest_path and fetch manifest_filename from config

getabetterpic:
I have a project where we are implementing webpacker and the devops person asked if we could rename the manifest file to something other than just `manifest.json`. I thought we would be able to but then after messing around with it for a bit realized it wasn't possible since `manifest.json` was hard-coded.

This PR fixes that by allowing a `manifest_filename` option to be set in the `webpacker.yml` config file.

Also while I was in there I renamed the `public_manifest_path` method to `public_manifest_file` since it is actually returning the path to a file, not just a path.

### DIFF
--- a/lib/install/config/webpacker.yml
+++ b/lib/install/config/webpacker.yml
@@ -5,6 +5,7 @@ default: &default
   source_entry_path: packs
   public_output_path: packs
   cache_path: tmp/cache/webpacker
+  manifest_filename: manifest.json
 
   # Additional paths webpack should lookup modules
   # ['app/assets', 'engine/foo/app/assets']
--- a/lib/webpacker/compiler.rb
+++ b/lib/webpacker/compiler.rb
@@ -37,7 +37,7 @@ class Webpacker::Compiler
 
   private
     def last_compilation_digest
-      compilation_digest_path.read if compilation_digest_path.exist? && config.public_manifest_path.exist?
+      compilation_digest_path.read if compilation_digest_path.exist? && config.public_manifest_file.exist?
     end
 
     def watched_files_digest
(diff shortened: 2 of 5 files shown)


javan:
> the devops person asked if we could rename the manifest file to something other than just `manifest.json`

Do you know why? Is there an issue with the filename?

getabetterpic:
A lot of the other tools we use also have a `manifest` file and while we aren't running into naming collisions, it can cause confusion. He wanted to make it more specific such as `webpack-manifest.json` so he would immediately know what he was looking at.

javan:
> Also while I was in there I renamed the `public_manifest_path` method to `public_manifest_file` since it is actually returning the path to a file, not just a path.

Undo that please. We, somewhat informally, use `*_path` to indicate that a `Pathname` is returned.

javan:
This configuration will need to make its way to the `ManifestPlugin` plugin as well:
https://github.com/rails/webpacker/blob/ae84f5f07f84b2e64b22ebb9b70641f0685a0327/package/environment.js#L29
Otherwise it will continue to write `manifest.json` (the default).

javan:
After thinking more, I've decided to pass on this. "Our ops guy wants a different filename" isn't a compelling enough reason to add yet another configuration option. Thank you for your contribution though.

getabetterpic:
Fair enough. Just figured we wouldn't be the only ones wanting a different filename for the manifest. And it seemed to make sense, especially since the `ManifestPlugin` supports different filenames. Thanks for your feedback!

getabetterpic:
Oh and thanks for this gem! It's really saved us by giving us a good upgrade path to go from AngularJS to Angular. We've tried several times to come up with a way to do the upgrade but kept failing until this came along.

PR #795, opened 2017-09-13
Change inquiry method to explicit call to StringInquirer

bpeschier:
webpacker can be used in a less than full installation of rails which does not
have the <string>.inquiry helper.

### DIFF
--- a/lib/webpacker/instance.rb
+++ b/lib/webpacker/instance.rb
@@ -8,9 +8,10 @@ class Webpacker::Instance
   end
 
   def env
-    (ENV["NODE_ENV"].presence_in(available_environments) ||
-      Rails.env.presence_in(available_environments) ||
-        "production".freeze).inquiry
+    ActiveSupport::StringInquirer.new(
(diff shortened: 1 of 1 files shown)


javan:
> webpacker can be used in a less than full installation of rails

Can you expand on that? How are you using webpacker?

bpeschier:
We are working with an external client who use certain elements of rails for their ruby-powered website like the routes and asset pipeline, but they do not run an entire rails installation. The enquiry-helper fails in their environment.

javan:
Can you add `activesupport/all` (or just [core_ext/string/inquiry](https://github.com/rails/rails/blob/master/activesupport/lib/active_support/core_ext/string/inquiry.rb)) to that application? Webpacker wasn't designed to work outside of Rails and fixing this one issue won't protect against future breakage.

bpeschier:
ActiveSupport is in there, which made me frown as well it did not work.

But it is a very valid point, I will investigate that approach with them.

javan:
Closing for now

PR #798, opened 2017-09-13
Fix webpacker:check_node to display version error

carlosdp:
Currently, if you have an older version of node installed that is referenced by `nodejs` instead of `node`, this code receives an Errno::ENOENT exception from the `node -v` command, bypassing the `node_version.blank?` check and telling the user Node is not installed. Instead, we should catch that exception and check for `nodejs -v` and let them know they need to upgrade versions.


ytbryan:
cc @gauravtiwari, should merge this?

gauravtiwari:
Thanks @carlosdp for the PR and @curvo @ytbryan for reviewing 🍰

PR #800, opened 2017-09-13
Better messaging for missing config files

curvo:
# Purpose
When webpacker compiles files it looks for a matching config file for each 'top_level' environment defined in webpacker.yml.  By default, there are a number of examples of this but there is no 'staging' environment example.  I needed this so I added staging to webpacker.yml and then bin/webpack would no longer compile because it was looking for `config/webpacker/staging.js `.  The current error message is generic and did not help me debug the error.  It took a bit to figure out what was going on and so this PR adds better messaging around that process.  Simple, but will hopefully save others some time going forward.

### DIFF
--- a/lib/install/bin/webpack.tt
+++ b/lib/install/bin/webpack.tt
@@ -13,9 +13,10 @@ APP_PATH          = File.expand_path("../", __dir__)
 NODE_MODULES_PATH = File.join(APP_PATH, "node_modules")
 WEBPACK_CONFIG    = File.join(APP_PATH, "config/webpack/#{NODE_ENV}.js")
 
+puts "Loading config: #{WEBPACK_CONFIG}"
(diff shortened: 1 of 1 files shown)


gauravtiwari:
@curvo You could put the config file inside same message instead of extra puts. 

```rb
puts "Webpack config #{WEBPACK_CONFIG} not found, please run 'bundle exec rails webpacker:install' to install webpacker or add a new one for your custom environment."
```

curvo:
True... I thought about that but I was thinking that the extra information might be helpful along the way since that message wouldn't show up unless there was an error.  But then again, I probably don't care until there's an error.   🤔 

I'll do that and push a fix.  Thanks

curvo:
@gauravtiwari Made the changes you requested! Thanks.

gauravtiwari:
thanks 👍

PR #793, opened 2017-09-13
Set dev server listen host and port via environment variables

monai:
Resolve #767.

PR #809, opened 2017-09-14
Support dev server `public` option

rossta:
This PR allows makes the [webpack-dev-server public option](https://webpack.js.org/configuration/dev-server/#devserver-public) configurable through `webpacker.yml`.

We currently use dnsmasq + nginx to reverse proxy requests from a custom local domain, i.e., `myapp.dev`, to our local Rails server for development. The recent addition of the `DevServerProxy` makes it easier for us to configure our setup for serving "packs" (thanks!). However, we'd still like the ability to proxy the dev server auto-refresh websocket connection, i.e., the `/sockjs-node` endpoint, through nginx as well.

To accomplish this (unless I'm missing an alternative), we would set the dev server `public` option to our local domain `myapp.dev`; however, the `public` is currently assumed to be `"#{HOSTNAME}:#{PORT}"` in the `bin/webpack-dev-server` wrapper. These changes ensure that a `public` option given in `webpacker.yml` or on the command line are passed through to the `webpack-dev-server` executable. 

I could add a section in `docs/webpack-dev-server.md` if that would be helpful for documentation.

### DIFF
--- a/lib/install/bin/webpack-dev-server.tt
+++ b/lib/install/bin/webpack-dev-server.tt
@@ -28,7 +28,7 @@ begin
 
   HOSTNAME          = args('--host') || dev_server["host"]
   PORT              = args('--port') || dev_server["port"]
-  PUBLIC_HOSTNAME   = args('--public') || "#{HOSTNAME}:#{PORT}"
+  PUBLIC_HOSTNAME   = args('--public') || dev_server["public"] || "#{HOSTNAME}:#{PORT}"
   HTTPS             = ARGV.include?('--https') || dev_server["https"]
   DEV_SERVER_ADDR   = "http#{"s" if HTTPS}://#{HOSTNAME}:#{PORT}"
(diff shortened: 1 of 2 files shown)


p0wl:
nice! needed this just yesterday and wrote a workaround.  Would love to see this merged!

gauravtiwari:
Thanks @rossta for this but we are thinking to move away from ARGV support for webpack-dev-server and instead provide env variables to achieve same behaviour. The config options currently are spread across - cli args, env and webpacker.yml, which makes it harder to reason. Please see this PR: https://github.com/rails/webpacker/pull/793 for though process but I am working on another one, which covers your needs as well :)

rossta:
@gauravtiwari Thanks for the background. Does that mean this PR pre-empted by yours or should I reimplement based on that work after it's merged?

gauravtiwari:
@rossta Yepp, the PR will cover your change here :)

gauravtiwari:
Closing in favour of #843

Issue #805, opened 2017-09-14
Incorrect source-directories in elm-package.json


janklimo:
Here's the `elm-package.json` file that gets generated when Elm installation is completed:

    {
      "version": "1.0.0",
      "summary": "helpful summary of your project, less than 80 characters",
      "repository": "https://github.com/user/project.git",
      "license": "BSD3",
      "source-directories": [
        "/Users/Admin/Playground/test_app/app/javascript/packs"
      ],
      "exposed-modules": [],
      "dependencies": {
        "elm-lang/core": "5.1.1 <= v < 6.0.0",
        "elm-lang/html": "2.0.0 <= v < 3.0.0"
      },
      "elm-version": "0.18.0 <= v < 0.19.0"
    }

As you can see, the `source-directories` are fine when working in development, but break the build when deploying to production. The following does the trick:

    "source-directories": [
      "./app/javascript/packs"
    ]

It's been a great pleasure working with Webpacker so far. Thank you for all the work!

janklimo:
I'll prepare a PR for this over the weekend.

PR #814, opened 2017-09-15
Fix elm-package.json source-directories

janklimo:
### Changes

- [x] makes the `source-directories` path relative, so it works in both development
and production (before: `"/Users/Admin/Playground/test_app/app/javascript/packs"`, after: `"app/javascript"`)
- [x] renamed elm to Elm where applicable
- [x] fixed installation messages incorrectly referring to entry file as Elm app file and
vice versa
- [x] moved `Main.elm` outside the packs folder. [Docs](https://github.com/rails/webpacker/blob/master/docs/folder-structure.md) state that packs file is only for entry files, so I believe having `Main.elm` within `packs` is incorrect and results in generating a pack we don't want.

Moving app files away form the `packs` folder also concerns Vue. Let me know if you agree with this change and I will update it as well (Angular already follows this, React example only has one file).

Resolves https://github.com/rails/webpacker/issues/805

### DIFF
--- a/lib/install/elm.rb
+++ b/lib/install/elm.rb
@@ -1,13 +1,14 @@
 require "webpacker/configuration"
 
-puts "Copying elm example entry file to #{Webpacker.config.source_entry_path}"
-copy_file "#{__dir__}/examples/elm/Main.elm", "#{Webpacker.config.source_entry_path}/Main.elm"
-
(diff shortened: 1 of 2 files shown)


gauravtiwari:
Thanks @janklimo 🍰  💯  Yes, please feel free to make a PR for Vue too 👍

PR #816, opened 2017-09-15
Include resolved paths in watched paths

maschwenk:
For people who are using resolved paths configurations, it might make sense to include these in the `default_watched_paths`. It won't affect anyone who left the resolved paths `[]` and it'll make test env re-compilation work for those who added resolved paths.

For folks who still have assets under `app/assets/javascripts/`, this will result in less hair pulling and fewer surprises. However, totally understand if you think this is better handled by an override like:
```ruby
Webpacker::Compiler.watched_paths << 'app/assets/javascripts/**/*'
```

Thanks!

### DIFF
--- a/lib/webpacker/compiler.rb
+++ b/lib/webpacker/compiler.rb
@@ -64,8 +64,12 @@ class Webpacker::Compiler
       status.success?
     end
 
+    def resolved_paths_globbed
+      config.resolved_paths.map(&:to_s).map { |p| "#{p}/**/*" }
+    end
+
     def default_watched_paths
-      ["#{config.resolved_paths}/**/*", "#{config.source_path}/**/*", "yarn.lock", "package.json", "config/webpack/**/*"].freeze
+      [*resolved_paths_globbed, "#{config.source_path}/**/*", "yarn.lock", "package.json", "config/webpack/**/*"].freeze
     end
 
     def compilation_digest_path
--- a/lib/webpacker/compiler.rb
+++ b/lib/webpacker/compiler.rb
@@ -65,7 +65,7 @@ class Webpacker::Compiler
     end
 
     def default_watched_paths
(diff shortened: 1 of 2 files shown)


maschwenk:
@gauravtiwari you wrote the original implementation of that watch list, what do you think about this?

gauravtiwari:
@maschwenk Thanks for this but looking at the changes I am not very clear. Could you please explain the usage? Do you want resolved paths to be available through webpacker.yml instead?

maschwenk:
In #539 you setup this watch list so that in the test environment, it would lazily compile changed assets. This works great, but for people who have `resolved_paths` defined in their `webpacker.yml`, they most likely want those watched as well or else this feature won't work. The default `webpacker.yml` has:
```yml
resolved_paths: []
```
so for most people, my change will not affect them at all

gauravtiwari:
@maschwenk Ahh gotcha, thanks. I think earlier, we didn't want to use webpacker.yml for ruby bits hence, explicit accessor to add watched paths but I think later we forgot to deprecate it in favour of `resolved_paths` since they both server same purpose. Perhaps, we should deprecate `Webpacker::Compiler.watched_paths`.

cc// @javan

maschwenk:
Looking at it again, I just noticed that this also affects the `dev` environment as well

gauravtiwari:
Not sure I understand?

maschwenk:
@gauravtiwari well the original commit for the `watched_paths` was mentioning lazy loading for `test` env, but I just noticed that the issue described at the top will affect both `dev` and `test`

gauravtiwari:
Ahh I see, yep it will do - it's intended to work for all environments 👍

mgrsskls:
+1 for this. I added a folder to the resolved paths as well as the watched paths, but I was wondering why these files only get compiled when I restart my rails server. This PR seems to fix it, so it would be awesome if it could be merged!

PR #817, opened 2017-09-15
Upgrade webpack manifest plugin

rossta:
The latest version of `webpack-manifest-plugin` (version `^1.3.2`) deprecates its `publicPath` option in favor of the webpack `output.publicPath`. This means the webpacker environment no longer needs to set `assetHost.publicPath` in multiple places. See https://github.com/danethurber/webpack-manifest-plugin/pull/80 for more info.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "sass-loader": "^6.0.6",
     "style-loader": "^0.18.2",
     "webpack": "^3.5.5",
-    "webpack-manifest-plugin": "^1.3.1"
+    "webpack-manifest-plugin": "^1.3.2"
   },
   "devDependencies": {
     "eslint": "^3.16.1",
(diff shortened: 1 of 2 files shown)


rossta:
@gauravtiwari Just a heads up that I made a mistake and pinned the wrong version for `webpack-manifest-plugin`. The proper version will be `^2.0.0`, due to release next week: https://github.com/danethurber/webpack-manifest-plugin/pull/80#issuecomment-330249043. 

I can add a fix when it's released but recommend holding off on publishing a new npm version of `@rails/webpacker` until then (or I can revert if necessary). Apologies for the mixup!

gauravtiwari:
@rossta No worries 😉 will revert this for now and we can bump the version when 2.0 is released.

PR #819, opened 2017-09-16
Vue example and props docs update

janklimo:
### Changes
- [x] moved Vue app file just outside the `packs` folder as discussed [here](https://github.com/rails/webpacker/pull/814#issuecomment-329772800)
- [x] extended Props documentation with an Elm section


gauravtiwari:
Great thanks @janklimo 💯  🍰

PR #825, opened 2017-09-18
Use file-loader for OTF fonts

monai:


webhive:
Why it was reverted? OTF fonts really does not work.

gauravtiwari:
@webhive It's there (on master) but not yet released, that revert commit is weird though (seems like something wrong with Github): https://github.com/rails/webpacker/blob/master/package/loaders/file.js

PR #830, opened 2017-09-19
Added #include? to the manifest to check for packs.

exegete46:
Sprockets supported finding if an asset exists, without raising an exception. This was a bit hack-ish, requiring a train wreck: Rails.application.assets.find_asset(name).present?, but it was possible. I used this to allow controller-specific Javascript / CSS to be included as-needed, rather than always being a part of the main application-level asset.

With Webpacker this feels even more hack-ish, using Webpacker.manifest.refresh[name.to_s].present?. This feels like way too much knowledge about the inner workings of Manifest.

I'd like to add a method to check if the manifest includes an entry. My first stab was to add a #include?(name) method to the public interface. This change can be found here on my fork.

### DIFF
--- a/lib/webpacker/manifest.rb
+++ b/lib/webpacker/manifest.rb
@@ -19,11 +19,21 @@ class Webpacker::Manifest
   end
 
   def lookup(name)
-    compile if compiling?
+    initialize!
     find name
   end
 
+  def include?(name)
+    initialize!
+    resolve(name).present?
+  end
+
   private
+
+    def initialize!
+      compile if compiling?
+    end
+
     def compiling?
       config.compile? && !dev_server.running?
     end
@@ -33,7 +43,11 @@ class Webpacker::Manifest
     end
 
     def find(name)
-      data[name.to_s] || handle_missing_entry(name)
+      resolve(name) || handle_missing_entry(name)
+    end
+
+    def resolve(name)
+      data[name.to_s] || nil
     end
 
     def handle_missing_entry(name)
--- a/test/manifest_test.rb
+++ b/test/manifest_test.rb
@@ -17,4 +17,12 @@ class ManifestTest < Minitest::Test
   def test_lookup_success
     assert_equal Webpacker.manifest.lookup("bootstrap.js"), "/packs/bootstrap-300631c4f0e0f9c865bc.js"
   end
+
+  def test_include_success
+    assert_equal true, Webpacker.manifest.include?("bootstrap.js")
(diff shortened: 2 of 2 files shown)


lib/webpacker/manifest.rb
```diff
@@ -23,11 +23,21 @@ def refresh
   end
 
   def lookup(name)
-    compile if compiling?
+    initialize!
     find name
```
ytbryan: imo find(name) is better

gauravtiwari:
I am not sure if I understand the use case. Could you please describe what you are after?

lib/webpacker/manifest.rb
exegete46: If you're referring to the `lookup()` method?  I believe that was pre-existing.

exegete46:
@gauravtiwari My specific use case for this is to check for CSS and/or Javascript asset associated with a specific controller.

For example, imagine a blogging site where 90% of the users do not require the CSS/JS used in the Admin area.  Rather than having all of the CSS and/or Javascript for the entire site (including Admin) bundled together in one large `application-<hash>.css`, a helper can check for any assets specific to the current controller.  So requests to `AdminController` can include an additional `<script>` tag for the Vue components in `admin.js`, while `ArticleController` would not.  Similarly, `CommentController` may have its own assets which aren't appropriate for `AdminController` or `ArticleController`.

With the asset pipeline, this was possible with `Rails.application.assets.find_asset('admin.js')`, which returned `nil` if the asset was not found, but didn't raise an exception.

However, with Webpacker, using `Webpacker.manifest.lookup('admin.js')` raises an exception if the pack is not found.  This patch adds `Webpacker.manifest.include?('admin.js')` to check for the existence of a admin Javascript file without raising an exception.

gauravtiwari:
Right I see, so you want the lookup to be automatic based on controller name instead of including `javascript_pack_tag` on each view or layout right?

exegete46:
I'm not thinking that Webpacker itself should include packs based on the controller.  That seems a bit out of Webpacker's scope.  Instead, developers should have the ability to create that functionality without catching exceptions.

gauravtiwari:
Yep, I understand 👍

gauravtiwari:
@exegete46 Made a PR here: https://github.com/rails/webpacker/pull/869 

Is that what you are looking for?

exegete46:
That looks great!  I was trying to avoid changing the public interface.  =-]

PR #832, opened 2017-09-19
Correcting typo in webpacker:check_node

sammyd:
Currently, if the node command line is `nodejs` and not `node` this rake task will never recognise that node is installed. Due to a typo on an exception name. This fixes that typo.


gauravtiwari:
thanks

PR #833, opened 2017-09-19
Binstubs

### LINKED ISSUE #831: Why not use rake to run webpack-dev-server?
I noticed over the weekend that my local version of `./bin/webpack-dev-server` was not actually up to date with the current gem in my project `webpacker@3.0.1`.  I realized then that one would need to run `rake webpacker:install` in order to update the binstubs within my project and that lead me to wondering if it would be possible to run the dev server using rake.  

I now have a fairly simple proof-of-concept working with a new rake task defined within my local version of the webpacker gem.  I can now start the webpack-dev-server using rake which means I don't need an executable file checked into my repo for that.

Was this sort of approach considered at all in the past?  And do you think it would be a possible direction to go?  

It seems to me that minimizing the steps and dependencies for using webpacker will hopefully help to reduce issues over time and make adoption that much easier.

If this is an interesting idea at all I would love to clean things up and offer a PR replacing both the webpack-dev-server and webpack binstub files with rake tasks and removing the dependency of checking those files into all rails repos.

rossta:
Converts `bin/webpack-dev-server` and `bin/webpack` to binstubs. This convention would make it easier to change the implementation of the two task runners without requiring devs to re-run the installer.

I'd expect it would be simple to merge changes from the [configurable-dev-server-options](https://github.com/rails/webpacker/tree/) branch.

Fixes #831


gauravtiwari:
thanks @rossta ❤️  💯  for the PR

Looks like rubocop isn't letting travis pass so if you can fix the offences, I will merge 👍

lib/webpacker/runner.rb
```diff
@@ -0,0 +1,34 @@
...
+    def self.run(argv)
+      $stdout.sync = true
+
+      new(argv).run
+    end
+
+    def initialize(argv)
+      @argv = argv
+
+      @app_path          = File.expand_path("../", __dir__)
```
javan: Does `__dir__` work correctly here (in a file in the gem)?

lib/webpacker/runner.rb
```diff
@@ -0,0 +1,34 @@
+require "shellwords"
+
+module Webpacker
+  class Runner
```
javan: Perhaps `Runner` should be the base class that all runner classes inherit from.

lib/webpacker/dev_server_runner.rb
```diff
@@ -0,0 +1,85 @@
...
+      def check_server!
+        server = TCPServer.new(@listen_host_addr, @port)
+        server.close
+
+      rescue Errno::EADDRINUSE
+        $stdout.puts "Another program is running on port #{@port}. Set a new port in #{@config_file} for dev_server"
+        exit!
+      end
+
+      def update_argv
```
gauravtiwari: We need this empty method here?

lib/webpacker/runner.rb
javan: Like:
```ruby
# lib/webpacker/runners/webpack.rb
class Webpacker::WebpackRunner < Webpacker::Runner

# lib/webpacker/runners/dev_server.rb
class Webpacker::DevServerRunner < Webpacker::Runner
```

lib/webpacker/runner.rb
rossta: No, good catch—error in local testing. I may need to pass the app_path into the runner which may mean keeping the install template with less code instead of using a bundler binstub.

lib/webpacker/dev_server_runner.rb
rossta: Nope, removing!

lib/webpacker/dev_server_runner.rb
```diff
@@ -0,0 +1,85 @@
...
+      new(argv).run
+    end
+
+    def initialize(argv)
+      @argv = argv
+
+      @app_path          = File.expand_path("../", __dir__)
+      @config_file       = File.join(@app_path, "config/webpacker.yml")
+      @node_modules_path = File.join(@app_path, "node_modules")
+      @webpack_config    = File.join(@app_path, "config/webpack/#{ENV["NODE_ENV"]}.js")
```
gauravtiwari: May be we can load options from `Webpacker.config` or `dev_server` itself instead of loading the yml again?

rossta: It would be great to avoid that duplication. Currently, `Webpacker.dev_server` inherits form `Webpacker.instance` which depends on `Rails` (for `Rails.root` [here](https://github.com/rails/webpacker/blob/7459f1f6a2a454aa83b68aa609ce88773cd3ee75/lib/webpacker/instance.rb#L6)) and I think it would be preferable to avoid depending on Rails to execute these scripts. Perhaps a future improvement?

lib/webpacker/runner.rb
```diff
@@ -0,0 +1,22 @@
...
+    def self.run(argv)
+      $stdout.sync = true
+
+      new(argv).run
+    end
+
+    def initialize(argv)
+      @argv = argv
+
+      @app_path          = File.expand_path(".", Dir.pwd)
```
rossta: Using `Dir.pwd` has pros and cons: we can install the executables as binstubs with bundler (which won't supply the application dir context) but we wouldn't be able to execute the binstubs from anywhere but the project root.

rossta:
I've created a test repo pointing to this branch to demonstrate behavior with the new binstubs: https://github.com/rossta/webpacker_test

lib/webpacker/runner.rb
curvo: Yeah... maybe you can leverage the root_path accessor set via webpacker itself which defaults to Rails.root?  Not sure what's best in this context?

lib/webpacker/dev_server_runner.rb
gauravtiwari: Makes sense 👍

lib/webpacker/dev_server_runner.rb
```diff
@@ -0,0 +1,72 @@
+require "shellwords"
+require "yaml"
+require "socket"
+require "webpacker/runner"
+
+module Webpacker
+  class DevServerRunner < Webpacker::Runner
+    def run
+      load_config
+      check_server!
```
gauravtiwari: may be call this `detect_port!`?

lib/webpacker/runner.rb
rossta: `Rails.root` would make it easy but I was hoping to avoid relying on loading rails since these are just thin wrapper around node executables.

rossta: Some alternatives:

* Re-instate the bin templates (backing out of the the bundler binstub change) so we can pass in the app path to the runner 
* Support an ENV var like `APP_PATH` to override the default

gauravtiwari: Yeah, lets tackle that in another PR

gauravtiwari:
@rossta Ready to merge? Looks good to me

rossta:
@gauravtiwari Ready!

gauravtiwari:
Merged 🎉  🍰

PR #836, opened 2017-09-20
add cache-loader to rules

morr:
This PR adds `cache-loader` https://github.com/webpack-contrib/cache-loader to webpack config
This change improves bundle compiling speed.

For one of my projects (a lot of coffescripts with vue) it changes compile time from 11500ms to 8900ms
For another of my projects (tiny scripts + http://coreui.io styles and fonts) it changes compile time from 12500ms to 4800ms

In `cache-loader` description is said that
> there is an overhead for saving the reading and saving the cache file, so only use this loader to cache expensive loaders

any thoughts on which loaders should be cached?

### DIFF
--- a/docs/webpack.md
+++ b/docs/webpack.md
@@ -115,6 +115,18 @@ For this to work, don't forget to use the `stylesheet_pack_tag`, for example:
 <%= stylesheet_pack_tag 'YOUR_PACK_NAME_HERE' %>
 ```
 
+### Add cache-loader
+[cache-loader](https://github.com/webpack-contrib/cache-loader) caches results of loaders on disk.
+Adding cache-loader to expensive loaders may increase your bundle compilation speed.
+
+```js
+// config/webpack/development.js
+environment.addCacheLoader() // enables on all loaders by default
+environment.addCacheLoader(['style', 'babel']) // enables on babel and style loader
+environment.addCacheLoader('style') // enables only on style
+```
+
+
 ## Plugins
 
 The process for adding or modifying webpack plugins is the same as the process
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
     "babel-plugin-transform-object-rest-spread": "^6.26.0",
     "babel-polyfill": "^6.26.0",
     "babel-preset-env": "^1.6.0",
+    "cache-loader": "^1.0.3",
     "coffee-loader": "^0.8.0",
     "coffeescript": "^1.12.7",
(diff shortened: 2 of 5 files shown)


gauravtiwari:
@morr Thanks for this. We already cache `babel loader`, which I think is the most expensive one - https://github.com/rails/webpacker/blob/master/package/loaders/babel.js#L9 but that may vary from project to project.

Earlier we tried cache loader with hello examples and there wasn't much improvement. Could you please post a bit more detailed overview? example: 10 JS files, 5 stylesheets etc. with before and after compilation time.

renchap:
`cache-loader` can be a good idea, but the Webpack team recommends to only enable it on specific slow loaders (see https://github.com/webpack-contrib/cache-loader#usage). I am not sure this is a good idea to enable it by default on all loaders, but it could be nice to have a small helper to enable it for specific loaders. Somethink like:
```es6
environment.enableCacheLoaderFor('style')
```

gauravtiwari:
Yeah, not sure how much impact it will have by default. But like the idea of exposing an option through the environment that can enable it on one or all loaders. 

```js
environment.enableCacheLoader() // enables on all loaders by default
environment.enableCacheLoader(['style', 'babel']) // enables on babel and style loader
environment.enableCacheLoader('style') // enables only on style
```

morr:
@gauravtiwari I updated PR. added environment.addCacheLoader function. added documentation.

hmm i don't know how to calculate number of js or css files considering that some files are included from npm packages :/

13927ms -> 8123ms: 394 modules
https://gist.github.com/morr/cde896515c68016c4a945bf259573c3f detailed log w/o cache
https://gist.github.com/morr/9688237d34d4efe2e1146da264fc53e3 detailed log with cache

12747ms -> 6604ms(it was 4500ms few days ago. i have no idea of what had changed): 100 modules
https://gist.github.com/morr/65ba62453f3b53cc22c6cd8d8f66523a detailed log w/o cache
https://gist.github.com/morr/aa014f6fc005b4365261286881a895c1 detailed log with cache

gauravtiwari:
@morr Thanks for this and apologies for late reply. I am bit hesitant on adding another dependency, without knowing if many people are going to use. I have been thinking on this and perhaps we can have an alternative API where people can extend any loader and add the dependency themselves instead of we shipping it as a default.

morr:
@gauravtiwari hmm I think you are right and more abstract API is more suitable in this case.
Do you want me to close PR or to change it as it is described below?
monosnap 2017-09-26 22-27-36

Issue #767, opened 2017-09-06
Dynamic port for webpacker-dev-server


monai:
Webpacker configuration file is [extensively reloaded](https://github.com/rails/webpacker/blob/master/lib/webpacker/railtie.rb#L38) and it seems there is no straight forward way to dynamically set dev server port, e.g., using Foreman to start Rails and webpacker-dev-server. Any suggestions how to do that?

gauravtiwari:
What do you mean dynamic port? You want to specify a port on runtime i.e. inside your Procfile? You can change port inside `config/webpacker.yml`, no?

monai:
I don't want to have dev server port hardcoded to `config/webpacker.yml`. I can pass dev server port via command line argument but Rails uses the port from `config/webpacker.yml`.

javan:
> I don't want to have dev server port hardcoded to `config/webpacker.yml`

Can you expand on that? It it adding friction to your development setup in some way?

gauravtiwari:
@javan We allow passing CLI arguments to binstubs, which can change host and port at runtime but since we also got proxy that's using values from webpacker.yml, CLI arguments are never used: https://github.com/rails/webpacker/blob/master/lib/install/bin/webpack-dev-server.tt#L29

One solution is we set explicit env vars based on these arguments so we can use them as well inside dev server configuration: https://github.com/rails/webpacker/blob/master/lib/webpacker/dev_server.rb#L28

```rb
# inside binstubs: 
ENV['WEBPACK_DEV_SERVER_PORT'] ||= args('--port')
```

```rb
ENV['WEBPACK_DEV_SERVER_PORT'] || fetch(:port)
```

javan:
> We allow passing CLI arguments to binstubs, which can change host and port at runtime 

I still don't understand why we do that. Why can't the sole source of configuration be `webpacker.yml`?

gauravtiwari:
Personally, I haven't really had any use for this  :) but I guess for teams where folks are using different setup, say someone using a Docker container and someone using puma-dev, it would be nice to override the settings at runtime instead of committing conflicting `config/webpacker.yml`

javan:
I'd prefer that go in `Webpacker` itself and then remove all the CLI arg wrangling in `bin/webpack-dev-server`:
```diff
--- a/lib/webpacker/dev_server.rb
+++ b/lib/webpacker/dev_server.rb
@@ -25,7 +25,7 @@ class Webpacker::DevServer
   end
 
   def port
[...]

PR #843, opened 2017-09-21
Allow overriding dev server settings using env variables

gauravtiwari:
This PR removes cli args support for webpack dev server and adds support for env variables within Webpacker to override dev server settings. 

For overriding dev server settings, simply set env variable before calling webpacker.

```rb
WEBPACKER_DEV_SERVER_HMR=true
WEBPACKER_DEV_SERVER_INLINE=true
WEBPACKER_DEV_SERVER_DISABLEHOSTCHECK=false
```

All settings for dev server inside `config/webpacker.yml` can be overriden, just make sure they are in caps.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,3 @@
+lib/*
+node_modules/*
+vendor/*
--- a/lib/install/config/webpacker.yml
+++ b/lib/install/config/webpacker.yml
@@ -33,11 +33,18 @@ development:
   <<: *default
   compile: true
 
+  # Reference: https://webpack.js.org/configuration/dev-server/
   dev_server:
+    https: false
     host: localhost
     port: 3035
+    public: localhost:3035
     hmr: false
-    https: false
+    # Inline should be set to true if using HMR
+    inline: true
+    overlay: true
+    disable_host_check: true
+    use_local_ip: false
 
 test:
   <<: *default
--- a/lib/webpacker/dev_server.rb
+++ b/lib/webpacker/dev_server.rb
@@ -17,7 +17,12 @@ class Webpacker::DevServer
   end
 
   def hot_module_replacing?
-    fetch(:hmr)
+    case fetch(:hmr)
+    when true, "true"
+      true
+    else
+      false
+    end
   end
 
   def host
@@ -29,7 +34,12 @@ class Webpacker::DevServer
   end
 
   def https?
-    fetch(:https)
+    case fetch(:https)
+    when true, "true"
+      true
+    else
+      false
+    end
   end
 
   def protocol
@@ -37,12 +47,12 @@ class Webpacker::DevServer
   end
 
   def host_with_port
-    "#{host}:#{port}"
+    fetch(:public)
   end
 
   private
     def fetch(key)
-      config.dev_server.fetch(key, defaults[key])
+      ENV["WEBPACKER_DEV_SERVER_#{key.upcase}"] || config.dev_server.fetch(key, defaults[key])
     end
 
     def defaults
--- a/lib/webpacker/dev_server_runner.rb
+++ b/lib/webpacker/dev_server_runner.rb
@@ -12,18 +12,12 @@ module Webpacker
     end
 
     private
-
       def load_config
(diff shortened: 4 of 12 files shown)


javan:
Do we need to support dotenv with this change? I'm hesitant to add another dependency, and would prefer to start by just reading `ENV` (Ruby) / `process.env` (Node) directly.

gauravtiwari:
Yeah, I was too but later felt since we are encapsulating configs inside a node module it would make sense to provide `env` support out of the box. Obviously, we are adding another dependency but I guess the usefulness of this is greater and most of the apps will find this useful. What do you think?

javan:
I think people using `dotenv` (or any env manager for that matter) should just ensure their env vars are set before calling into webpacker.

gauravtiwari:
Yeah, totally but it's not that trivial to setup. Here is our doc about supplying env variable to Webpacker: https://github.com/rails/webpacker/blob/master/docs/env.md (which basically requires overriding Environment plugin).

guilleiguaran:
```
Warning: using Foreman/Invoker and npm dotenv at the same time can result in confusing behavior, in that Foreman/Invoker variables take precedence over npm dotenv variables.
```
I think this warning explain why we shouldn't add a default env manager and leave it to the users.

gauravtiwari:
Makes sense 👍

gauravtiwari:
@guilleiguaran Could you please review and merge this one? I have dropped `dotenv`

timkrins:
I can understand using ENV variables, but why specifically remove support for ARGV also?

Unable to use https certificates that I would usually expect to pass to webpack-dev-server like so:

`webpack-dev-server --https --cert=/home/user/selfsigned.crt --key=/home/user/selfsigned.key`

gauravtiwari:
Hey @timkrins 

The previous implementation was limited. Webpacker uses dev server options (webpacker.yml or ENV) for an internal rake middleware and it was a bit complex to allow three sources for overriding dev server configuration - cli, webpacker.yml and ENV

This could be unified but haven't had time to look into this properly. For now, you could supply certs through `config/webpack/development.js`

For example: 

```js
// development.js
environment.config.merge({
   devServer: {
    https: {
      key: fs.readFileSync('/path/to/server.key'),
      cert: fs.readFileSync('/path/to/server.crt'),
      ca: fs.readFileSync('/path/to/ca.pem'),
    }
  }
})
```

timkrins:
Hi @gauravtiwari ,

Thanks for the explanation.

I think as the webpacker `webpack` binstub does accept ARGV I was not expecting the `webpack-dev-server` binstub to ignore them.

I am aware I can do what you recommended, however, it is just a bit annoying not to be able to have access to do this via the commandline, when I know that `webpack-dev-server` supports it.

Is there a potential conflict adding ARGV to the `execute_cmd` function? (just like in `webpack_runner.rb`)
ie. 

        cmd = [
          "#{@node_modules_path}/.bin/webpack-dev-server",
          "--config", @webpack_config
        ] + @argv

timkrins:
Ah, I can see I'm not the only one confused by this - just found https://github.com/rails/webpacker/pull/1672

tbtalbottjr:
Can someone explain how to set the port for the webpacker dev server so that it is used by both the server and the client using a .env file or similar (as requested in https://github.com/rails/webpacker/issues/767)? I'm missing something simple, I'm sure, but I've been able to change the devServer port, but the client continues to use the port out of the webpacker.yml file. The only thing that has worked so far is to physically change the webpacker.yml file. I want to avoid that because I want to be able to run two versions of the app server at the same time and don't want to be changing a file that gets checked into the repository

Rails 5.1.7
Webpacker 4.02

Running app server through puma_dev.

Thanks.

minimul:
Same here. 

ENV var overrides such as:
```
WEBPACK_DEV_SERVER_PORT=3011 bin/webpack-dev-server
WEBPACKER_DEV_SERVER_HOST=0.0.0.0 ./bin/webpack-dev-server
```
**are not working**.

Rails 6.0.3.2
Webpacker 5.1.1

PR #844, opened 2017-09-22
Exposing attribute for configured file extensions

lmcgrath:
We can't get the extensions list from the config object in version 3.0, so this PR adds it as an attribute.

Backstory: In version 2.0 we were using the config object to get the configured file extensions and globbing for everything in the source directory with those extensions. This enabled us to fingerprint the files and determine at build-time whether we needed to rebuild assets or just use the ones already in CDN (also it made our builds faster.)

### DIFF
--- a/lib/webpacker/configuration.rb
+++ b/lib/webpacker/configuration.rb
@@ -45,6 +45,10 @@ class Webpacker::Configuration
     root_path.join(fetch(:cache_path))
   end
 
+  def extensions
+    fetch(:extensions)
+  end
+
   private
     def fetch(key)
       data.fetch(key, defaults[key])
--- a/test/configuration_test.rb
+++ b/test/configuration_test.rb
@@ -27,8 +27,8 @@ class ConfigurationTest < Webpacker::Test
   end
 
   def test_extensions
-    webpacker_yml = YAML.load_file('lib/install/config/webpacker.yml')
-    assert_equal Webpacker.config.extensions, webpacker_yml['default']['extensions']
+    webpacker_yml = YAML.load_file("lib/install/config/webpacker.yml")
+    assert_equal Webpacker.config.extensions, webpacker_yml["default"]["extensions"]
   end
 
   def test_cache_manifest?
--- a/test/configuration_test.rb
+++ b/test/configuration_test.rb
@@ -26,6 +26,11 @@ class ConfigurationTest < Webpacker::Test
     assert_equal Webpacker.config.cache_path.to_s, cache_path
   end
 
+  def test_extensions
+    webpacker_yml = YAML.load_file('lib/install/config/webpacker.yml')
+    assert_equal Webpacker.config.extensions, webpacker_yml['default']['extensions']
+  end
+
   def test_cache_manifest?
     with_node_env("development") do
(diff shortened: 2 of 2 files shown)


gauravtiwari:
@lmcgrath Hey, thanks for this. You can access it like so: `const config = require('@rails/webpacker/package/config'). Is that what you want? 

We will expose the configs in next release.

lmcgrath:
That's not quite what I want. I was accessing it from a rake task with Ruby.

gauravtiwari:
@lmcgrath 
> This enabled us to fingerprint the files and determine at build-time whether we needed to rebuild assets or just use the ones already in CDN (also it made our builds faster.)

We already do this now: https://github.com/rails/webpacker/blob/master/lib/webpacker/compiler.rb#L45 

Or am I interpreting it wrong?

lmcgrath:
Here's what's happening:

* CI performs a deployment-ready build
* CI computes fingerprint and compares with fingerprint of files in CDN
* If fingerprint is different
  * webpack is run to compile the assets
  * assets uploaded to CDN with new fingerprint

We're testing whether our files should be recompiled based on the fingerprint of the last-performed compilation. Those files and that fingerprint exist remotely so we have to fetch the fingerprint with each build and compare against the computed fingerprint of the local files. The [`Webpacker::Compiler#watched_files_digest()`](https://github.com/rails/webpacker/blob/master/lib/webpacker/compiler.rb#L45) method only tests files against a local fingerprint.

I would prefer to use `watched_files_digest()`, are you open to exposing it as a public method?

gauravtiwari:
Right, I see so you are doing two things here: 

> 1. webpack is run to compile the assets 

which won't run anyway with new compiler, since it caches the compilation timestamp in `tmp/cache/webpacker`

> 2. assets uploaded to CDN with new fingerprint

but I guess for this you have no way to figure out. Is that right?

Perhaps you could just use `fresh?` or `stale?` public method.

```rb
if Webpacker.compiler.fresh?
  # upload to CDN
end

# or you could send message to compiler object and get watched files digest,
Webpacker.compiler.send(: watched_files_digest)
```

Would that work?

lmcgrath:
No, that would not work. I actually also just checked to see if `watch_files_digest()` would work but because it operates on files' last modified times it won't compute a correct fingerprint in CI (CI clones the git repo, which sets all the modified times to current timestamp.)

gauravtiwari:
Ahh gotcha, makes sense 👍

gauravtiwari:
merged, thanks for your patience 👍

lmcgrath:
No worries! Thanks so much for the merge and helping to making sure this was reasonable 👍

gauravtiwari:
Btw, just realised, you could do it like so to get all paths (probably simpler than extensions): 

```rb
Dir[*Webpacker.compiler.send(:default_watched_paths), *Webpacker.compiler.watched_paths]
```

lmcgrath:
Actually that does look better, as it covers any changes in dependencies and configuration. Mind if I submit another PR to expose a method like `all_watched_paths()`? I'd like to avoid tossing around `send()`s so I can worry less about the API changing unexpectedly and breaking things.

Issue #794, opened 2017-09-13
Option to install or export default webpack config from @rails/webpacker?


jnfeinstein:
Greetings,

I just upgraded webpacker from v2 to v3 (fastest moving rails lib I know of!).  One of the notable changes is that the configuration/loaders/plugins are all managed by `@rails/webpacker`.  I'm guessing this was done such that the gem and configurations stay in sync (i.e. at 3.0.1, both look to `webpacker.yml` for configuration, but this may change in the future).  Makes a great deal of sense and it's a nifty feature.

My Angular 1 application has a bunch of custom configuration, such as a custom loader and the `ngAnnotate` plugin.  The best I could come up with in webpacker v3 world was:
```
const babelLoader = environment.loaders.get('babel')
babelLoader.use = [
  // Trying to maintain the default behavior, I trust the gurus of webpacker know better than I do
  { loader: babelLoader.loader, options: babelLoader.options },
  { loader: join(__dirname, 'utils', 'angular-loader.js') }
];
delete babelLoader.options;
delete babelLoader.loader;
```
and (since plugin order matters)
```
const UglifyJsPlugin = environment.plugins.get('UglifyJs');
environment.plugins.delete('UglifyJs');

environment.plugins.set(
  'ngAnnotate',
  new ngAnnotatePlugin({
    add: true
  })
);

Object.assign(UglifyJsPlugin, {
  mangle: true,
  sourceMap: false
});

environment.plugins.set('UglifyJs', UglifyJsPlugin);
```
To an extent these defeat the purpose of having the default configuration package, since they rely on the configuration being consistent.  `babelLoader` has to exist, otherwise my angular loader wont be installed properly, but AFAIK the internals of `@rails/webpacker` don't necessarily adhere to any API.  I also don't trust maps to be ordered, so that's not an ideal data structure for a feature where order matters.

My gut reaction was to just go back to my v2 configuration, which gave me the ultimate flexibility in configuration, but I don't know if anything was changed w/ regards to binding to the gem in v3 (like did the format of `webpacker.yml` change?).

Options I can think of (that may already exist and I'm unaware of):
* Copy the default configuration files present in `@rails/webpacker` to `config/webpacker` so I can manipulate them manually.  Whenever I bump the gem version I can regenerate these files and diff them.  I do this often for other packages.
[...]

PR #846, opened 2017-09-22
Update documentation with how to change config in the top level readm…

### LINKED ISSUE #802: No documentation on how to work with/edit configuration
We use expose-loader and some resolve aliases, but the latest webpacker doesn't have any documentation on how to use these with environment.js to edit the environment before toWebpackConfig()...in fact, looking at the source code this doesn't seem possible, so maybe the call toWebpackConfig() should be in environment.js and not in each separate environment so we can make config changes in one spot for global items?

BookOfGreg:
…e. Also reveal how to make use of internal functions

Takes comment from here: https://github.com/rails/webpacker/issues/794#issuecomment-329749982 on how to export internal functions for use.

Fixes #802

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,7 +181,7 @@ production environments in `config/webpack/*.js`. You can configure each individ
 environment in their respective files or configure them all in the base
 `config/webpack/environment.js` file.
 
-By default, you shouldn't have to make any changes to `config/webpack/*.js`
+By default, you don't need make any changes to `config/webpack/*.js`
 files since it's all standard production-ready configuration.
 See [docs/Webpack](docs/webpack.md) for modifying the configuration and loaders API.
 
--- a/README.md
+++ b/README.md
@@ -173,6 +173,55 @@ when running `./bin/webpack-dev-server` binstub:
 
 **Note:** Don't forget to prefix `ruby` when running these binstubs on windows
 
+### Configuration
+
+
+Webpacker gives you a default set of configuration files for test, development and
+production environments in `config/webpack/*.js`. You can configure each individual
+environment in their respective files or configure them all in the base
+`config/webpack/environment.js` file.
+
+By default, you shouldn't have to make any changes to `config/webpack/*.js`
+files since it's all standard production-ready configuration. However,
+if you do need to customize or add a new loader, this is where you would go.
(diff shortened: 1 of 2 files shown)


gauravtiwari:
Thanks for this. Please could you move all webpack related doc to `docs/webpack.md`. You can keep the doc link though inside main README. The idea is to keep the README slimmer 😉

BookOfGreg:
@gauravtiwari No problem. I left in the intro paragraph and link over to `webpack.md` as that's the minimum that reads well and feels appropriate. It's still quite thin this way.
Is this OK or shall I remove more of it?

docs/webpack.md
```diff
@@ -36,6 +36,14 @@ const customConfig = require('./custom')
 module.exports = merge(environment.toWebpackConfig(), customConfig)
 ```
 
+If you need access to functions within Webpacker's configuration, you can import them like this:
```
gauravtiwari: Shall we call it configs instead of functions?

README.md
```diff
@@ -173,6 +174,17 @@ when running `./bin/webpack-dev-server` binstub:
...
 **Note:** Don't forget to prefix `ruby` when running these binstubs on windows
 
+### Configuration
+
+Webpacker gives you a default set of configuration files for test, development and
+production environments in `config/webpack/*.js`. You can configure each individual
+environment in their respective files or configure them all in the base
+`config/webpack/environment.js` file.
+
+By default, you shouldn't have to make any changes to `config/webpack/*.js`
```
gauravtiwari: `you shouldn't have` => `you don't need to`

gauravtiwari:
Thanks Greg 👍  Looks good, just some minor change to wording

BookOfGreg:
Tweaked both Readme and Webpack so they are consitent. Thanks for the speedy feedback 👍

gauravtiwari:
Thank you. Totally missed the duplicate content between webpack and README so I have taken liberty to remove that. If you could please update TOC with doctoc, I will merge 👍

BookOfGreg:
Done, Thank you 👍

gauravtiwari:
thanks 🍰  🍬

PR #849, opened 2017-09-23
Remove binstub checks & compiler dependency on ./bin/webpack

curvo:
Now that webpacker no longer needs locally checked in binstubs, we no longer need to check for those when running `rake webpacker:verify_install`.  This removed the previous `webpacker:check_binstubs` task and updates the docs.  There's also a lingering dependency within the compiler class that should probably leverage Webpack::WebpackRunner.  So this connects those things as well.


curvo:
@gauravtiwari I still have some cleanup to do here and this update would effectively break windows and I'm not sure how to fix that.  Actually windows running of binstubs is probably already broken in master.  Anyway, wanted to push this for consideration as it should probably be in the next release if you are considering one to support High Sierra.

curvo:
FYI - I'm working on this.  Have everything working now with assets:precompile but broke the context with the binstubs.  Will work out the rest tomorrow and push updates to this.

README.md
```diff
@@ -127,7 +127,8 @@ you would need to link them in your "pack" or entry file.
 
 ### Development
 
-Webpacker ships with two binstubs: `./bin/webpack` and `./bin/webpack-dev-server`.
+Webpacker ships with two binstubs which are available after installing the gem:
```
aselder: We should probably refer to this as binaries rather than binstubs, since there are not binstubs anymore.

curvo:
@gauravtiwari I've got this all working on another branch now.   @aselder will be takin a look tomorrow and then hopefully I'll actually open it up here as a replacement to this PR then.  Take a look if you'd like but I'll hopefully get it over here sometime Sunday.

https://github.com/usertesting/webpacker/pull/2

curvo:
I'm opening a proper issue around this and will reference the UT PR there for now.  I'll get that PR into the rails/webpacker repo for consideration sometime Sunday.  Thanks!

PR #853, opened 2017-09-24
rake webpacker:install:upgrade

curvo:
After a gem bump it's not completely natural to remember to potentially bump the binstubs or yarn packages.  

This PR adds a new rake task to help with that process.

```bash
bundle exec rails webpacker:install:upgrade

# OR (on rails version < 5.0)
bundle exec rake webpacker:install:upgrade
```


curvo:
@gauravtiwari Not sure what you'll think but having something like this and documentation to support it might make the upgrading process a bit easier?  Please take a look!

gauravtiwari:
@curvo Thanks for this, looking good. However, I am wondering since the upgrade only involves two tasks, may be we just document that instead under `upgrade` title?

```bash
bundle binstubs webpacker --force
yarn upgrade @rails/webpacker --latest
```

curvo:
True.  I'll make the change but I guess I'm wondering if there's anyway to automate this side of an upgrade eventually?  Like, would it be possible to create a hook that's run by bundle during a gem bump?  I've read about that kind of thing.  I'm not sure it's fully supported but it sounds like it may be possible to leverage the "installing native extensions" moment to bump yarn side as well?  Just a thought.  Let me pull the rake task out and update that little docs snippet.  Maybe then I'll have contributed some little thing out of all the tangents I went on this weekend! 😅

curvo:
@gauravtiwari all good!

curvo:
thanks

gauravtiwari:
@curvo I understand and appreciate your contribution so far 💯  👍 

Yeah, there is a  `postinstall` hook. You mean to automatically upgrade dependencies of webpacker or just `@rails/webpacker`?

gauravtiwari:
@curvo Could you please update TOC? You will need to install: https://github.com/thlorenz/doctoc and then run `doctoc README.md` from terminal.

curvo:
Will do tonight! 

Sent from my iPhone

On Sep 25, 2017, at 11:29 AM, Gaurav Tiwari <notifications@github.com> wrote:

@curvo Could you please update TOC? You will need to install: https://github.com/thlorenz/doctoc and then run doctoc README.md from terminal.

curvo:
@gauravtiwari Updated.  For the postinstall hook I mean to bump the @rails/webpacker to match the gem version.  It's just not an intuitive thing for folks coming from Rails, especially Rails 4 and adding webpacker for the first time.  It's the first gem with that kind of dependency that I've seen anyway.  It would just be a nice thing to automate as part of upgrading the gem.  Then you would end up with changes to Gemfile, Gemfile.lock, packages.json and yarn.lock all at once.

gauravtiwari:
@curvo Right, yeah makes sense. Lets figure that out together 👍

PR #857, opened 2017-09-25
Passing Config file through ERB first in order to better reference configs

Gbeschbacher:
This would allow us to use secrets.yml as follows:
`<%= { Rails.env.to_s => Rails.application.secrets[:webpacker].deep_stringify_keys }.to_yaml %>` thus resulting in a single config file and therefore better maintainability

### DIFF
--- a/lib/webpacker/configuration.rb
+++ b/lib/webpacker/configuration.rb
@@ -55,7 +55,7 @@ class Webpacker::Configuration
     end
 
     def load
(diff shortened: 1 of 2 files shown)


gauravtiwari:
@Gbeschbacher Thanks for this. It would be great if we could do this but we are using same config file for JS side of things, which won't parse ERB for us like Ruby.

Issue #746, opened 2017-09-01
bin/webpack -p fails to process jest snapshot files


ktusznio:
`bin/webpack` works fine, but `NODE_ENV=production RAILS_ENV=production bin/webpack -p` fails:

```
» RAILS_ENV=production NODE_ENV=production bin/webpack -p
Hash: 9b023edef321541c6d98
Version: webpack 3.5.5
...

ERROR in application-e06269f751f2d02e6bcf.js from UglifyJs
Unexpected character '`' [./app/javascript/apps/account/components/__snapshots__/AccountPage.test.js.snap:3,0][application-e06269f751f2d02e6bcf.js:1,279057
```

The offending file:

`./app/javascript/apps/account/components/__snapshots__/AccountPage.test.js.snap`:
```js
// Jest Snapshot v1, https://goo.gl/fbAQLP

exports[`AccountPage renders 1`] = `
<section
  className="AccountPage__Wrapper-orVeb jTAEQy"
>
</section>
`;
```

A few things are weird about this:

* webpack's `uglifyjs-webpack-plugin` uses "uglify-es" which should be able to process ES6 files.
* `uglifyjs-webpack-plugin@^0.4.6` tests against `/.js($|?)/i` files, which should not match `AccountPage.test.js.snap` -- why is UglifyJs processing this file?
* I'm using `ignore-loader` with the following config -- again, why is UglifyJs processing the snapshot file?

```js
module.exports = {
  test: /.test.js.snap$/,
  loader: 'ignore-loader'
}
```

ktusznio:
Strangely, `RAILS_ENV=production NODE_ENV=production bundle exec rails webpacker:compile` does not throw an error:

```
Webpacker is installed 🎉 🍰
Using ./config/webpacker.yml file for setting up webpack paths
Compiling…
Compiled all packs in ./public/packs
```

javan:
I'm not sure what's causing the error, but Webpacker already configures production builds so there's no need for the `-p` flag.
https://github.com/rails/webpacker/blob/fb8dd8c09ccc9d93ef3e084838e6e0f0683aaeb0/package/environments/production.js#L9-L25

gauravtiwari:
Yep we do and hence the error - technically you are uglifying twice. 

For ignore loader, you can just setup with extension: 

```js
module.exports = {
  test: /\.snap$/,
  loader: 'ignore-loader'
}
```

ktusznio:
Gotcha, thank you!

Issue #395, opened 2017-05-17
Compiles fail with no output


schneems:
From https://github.com/heroku/heroku-buildpack-ruby/issues/562

I'm getting a non-trivial number of people getting compile failures, but with no debug output of any kind. Here's an example:

```
remote:        Webpacker is installed 🎉 🍰
remote:        Using /tmp/build_c3975d9e573c1bf0b4e6b532daa48fde/config/webpack/paths.yml file for setting up webpack paths
remote:        Compiling webpacker assets 🎉
remote:  !
remote:  !     Precompiling assets failed.
remote:  !
remote:  !     Push rejected, failed to compile Ruby app.
remote: 
remote:  !     Push failed
remote: Verifying deploy....
remote: 
remote: !	Push rejected to [my app].
```

Is there anything else we can do when this happens? Any way to add debug info? Maybe toggle an environment variable for extra logging?

Have you seen silent failures like this before? Any ideas?

ytbryan:
Thanks for filing this, Richard. 

@flybayer  Which version of webpacker are you using? Master branch? Can I confirm it's rails 5.1.0 and not 5.1.1? 
@fuffi Same question to you

gauravtiwari:
@schneems Yepp, I have seen it before and most probably it's related to `node-sass` binary either not available or corrupted - related to https://github.com/sass/node-sass/issues/1918 

One more thing I noticed and mentioned in (https://github.com/rails/webpacker/issues/169#issuecomment-288296033) was that with regular ruby buildpack `npm` isn't available alongside yarn. I think it would be useful to have that too on heroku.  

```json
"scripts": {
"postinstall": "npm rebuild node-sass"
}
```

I think this is a bug we need to fix for compile task. Currently we are using a non-standard way to get output from webpack compile task, which is unhelpful for debugging. I started adding js scripts in #264. Perhaps, I can just make a PR with only scripts so it improves the debugging workflow.

flybayer:
@ytbryan I'm using Rails 5.1.1 (5.1.0 didn't work either), webpacker latest master branch, ruby buildpack v160.

flybayer:
Now precompiling is failing every single time regardless of cache state, including the exact code I was using the other week in a brand new dokku app. * _scratches head_ *

Precompiling locally in production mode works flawlessly.

flybayer:
Update: `node-sass` was my problem even though there was **no** logging indicating this.

Adding 
```json
"postinstall": "npm rebuild node-sass"
[...]

PR #860, opened 2017-09-25
Add support for node_modules to be included in babel

adambutler:
I've found dependencies that are ES6 fail during `UglifyJs` due to not first being handled by babel:

For example, my code has:

```js
import { Foundation } from 'foundation-sites/js/foundation.core';
import { Reveal } from 'foundation-sites/js/foundation.reveal';
import { Tabs } from 'foundation-sites/js/foundation.tabs'
```

When I run:

```
$ rm -Rf public/packs
$ NODE_ENV=production RAILS_ENV=production rails webpacker:compile --trace
```
I get the following error:

```
ERROR in application-37763722852138e61493.js from UglifyJs
Unexpected character '`' [./node_modules/foundation-sites/js/foundation.util.core.js:24,0][application-37763722852138e61493.js:11323,124]

ERROR in Modals-8541f74f3cf2b70f4a0f.js from UglifyJs
Unexpected token: name (matchMedia) [./node_modules/foundation-sites/js/foundation.util.mediaQuery.js:21,0][Modals-8541f74f3cf2b70f4a0f.js:102,4]

ERROR in TabbedExamples-b873f916d365855bfd5d.js from UglifyJs
Unexpected character '`' [./node_modules/foundation-sites/js/foundation.util.core.js:24,0][TabbedExamples-b873f916d365855bfd5d.js:100,124]
```

This change adds the ability to include `node_modules` to be compiled by `babel`. It defaults to the current behaviour, which I'm not entirely sure is correct anyway.

Related: #395
Possibly related: #746 

Copying in @dhh @gauravtiwari who have both touched `package/loaders/babel.js`.

### DIFF
--- a/docs/webpack.md
+++ b/docs/webpack.md
@@ -190,3 +190,13 @@ Now, add these files to your `layouts/application.html.erb`:
 ```
 
 More detailed guides available here: [Webpack guides](https://webpack.js.org/guides/)
+
+## Including `node_modules` in babel
+
+If you have dependencies that are >ES6 you may want to include them to be processed by babel. A common case where you might need this is if you are finding the UglifyJS is raising errors during production complications. For example:
+
+```
+Unexpected character '`'
+```
+
+To include these dependencies in your pipeline change the value in `babel.exclude_node_modules` in `config/webpacker.yml` to `false`.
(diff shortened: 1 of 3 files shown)


gauravtiwari:
@adambutler Thanks for this but if you want to include `node_modules` for your project, you can update the loader yourself in your `environment.js`

```js

// Update an option directly
const babelLoader = environment.loaders.get('babel')
babelLoader.exclude = []

module.exports = environment
```

PR #861, opened 2017-09-25
Patch for invalid response headers coming from express

washu:
under node 8.5.0, rails 5.1 i noticed express is returning invalid http headers. this patch corrects it.
Example response:
{"x-powered-by"=>["Express"], "access-control-allow-origin"=>["*"], "accept-ranges"=>["bytes"], "content-type"=>["application/json; charset=UTF-8"], "content-length"=>["61"], "etag"=>["W/\"3d-5mD5tX4ZoI4R3XJKABPz5FxFGro\""], "vary"=>["Accept-Encoding"], "date"=>["Mon, 25 Sep 2017 21:12:36 GMT"]}

during HMR


gauravtiwari:
Is it common to send headers like this i.e. Array of values? Perhaps you wanna file an issue on express repo.

washu:
the RFC says are supposed to be single strings. and this fails rack compliance as well, header values are strings

PR #864, opened 2017-09-26
Document Nginx setup for development

rossta:
I thought it would be useful to add the following notes about getting Webpack dev server live reloading to work when using Nginx and custom domain in development.


gauravtiwari:
Thanks for this but will hold off merging this in until we have a new release, to avoid confusion. Are you using master and all working okay now - public option?

rossta:
@gauravtiwari Thanks, no problem. 

The changes to support the `public` option almost gets us what we were hoping for but I'd like to propose a slight change—I'll open a PR.

rossta:
See #867

PR #869, opened 2017-09-27
Manifest lookups without exception

gauravtiwari:
Much like Rails finders: `find_by` and `find_by!`

```rb
Webpacker.manifest.lookup('foo.js')
# => nil
Webpacker.manifest.lookup!('foo.js')
# => raises Webpacker::Manifest::MissingEntryError
```

Closes: #830


gauravtiwari:
👍  😄

Issue #859, opened 2017-09-25
coffescript2


morr:
Hi! Do you know that coffescript2 is released? http://coffeescript.org/announcing-coffeescript-2/
Its main feature is that coffee compiles now into es6 and it also supports some of es6 syntax (import, =>, async/await etc).
To make it work with webpack in my project I had to add `babel-loader` into `coffe-loader`

Here is the code
```js
// config/webpack/environment.js
babel = environment.loaders.get('babel')
coffee = environment.loaders.get('coffee')

coffee.use = [
  { loader: babel.loader, options: babel.options },
  coffee.loader
]
delete coffee.loader
```

Do you want me to make PR with such changes in webpacker?
```js
// config/webpack/environment.js
environment.supportCoffeescript2() // adds support for coffescript2

// webpacker/package/environment.js
module.exports = class Environment {
// ...
  supportCoffeescript2() {
    babel = this.loaders.get('babel')
    coffee = this.loaders.get('coffee')

    coffee.use = [
      { loader: babel.loader, options: babel.options },
      coffee.loader
    ]
    delete coffee.loader
  }
// ...
}
```

My first thought was to add coffescript version check into `webpacker/package/loaders/coffee.js` but it does not work because `yarn` takes coffee not from my project dependencies but from webpacker dependencies.
Webpacker depends on coffee1 and because coffee2 is not backwards compatible with coffee1 I can't just upgrade coffee version in webpacker `package.json`.

renchap:
I am using Coffeescript 2 for some time, and I got it working simply by using this in `environment.js`:

```es6
const babelLoader = environment.loaders.get('babel')
babelLoader.test = /\.(coffee|js)(\.erb)?$/
```

It seems simpler, and should also work with CS 1 (albeit a bit slower because Babel will try to transpile non ES6 code, which is useless).

morr:
@renchap, could you please elaborate how it works?
does webpack do
1) `coffee` -> `es6` via `coffee-loader`
2) `es6` -> `js` via `babel-loade`

or

1) `coffee` -> `coffee with js` via `babel-loader`
2) `coffee with js` -> `js` via `coffee-loader`

renchap:
Loaders are applied in the reverse order they appear in Webpack config, and to files matching the `test` attribute.

So here, `file.coffee` will be processed by `coffee-loader` (outputing ES6), and then by `babel-loader` (taking the previous ES6 output), as it matches those 2 loader `test` attributes.

[...]

Issue #868, opened 2017-09-27
Coffeescript dependency does not accept Coffeescript 2.0


renchap:
`@rails/webpacker` depends on `coffeescript@^1.12.7`
Coffeescript 2 has been released with a lot of new features: http://coffeescript.org/announcing-coffeescript-2/ (JSX support, …).
Would it be ok to either bump the dependancy to 2.0, or at least accept it?

CS2 outputs ES6, so it needs to be transpiled using Babel (see #859).

Currently I have both 1.12.7 installed and 2.0.1 (from my own `package.json`), but `coffee-loader` loads the 1.12.7 version. It worked fine previously but just broke, and I can not figure out why (it loaded 2.0.1 before, and suddently decided it would revert to loading 1.12.7).

gauravtiwari:
@renchap Apologies, totally forgot to reply on #859 

I am not sure really, considering there will be people who are still stuck on legacy code and people who want to use CS2. That's why I felt that moving all this to user space would be lot easier like React and other integrations, so we don't have to bump deps.

But anyway, for now I think we can conditionally bootstrap coffee loader like we do with `style` loader (hmr vs non-hmr) so it works with both CS1 and CS2.

renchap:
Would it work to not have `coffeescript` in dependencies, so the user can choose which version he wants? Or put it in `peerDependancies` maybe?

gauravtiwari:
Yeah, that's what I was thinking - to not have `coffeescript` as dependency like React and other stuff. We just have to make it clear to install `coffeescript`. In Rails world, we do this for gems like `mysql` or `pg` - that's all in user space.

renchap:
Ok. Is it a way to check if a package is installed to conditionally add `coffee-loader`?

gauravtiwari:
Made a PR here: https://github.com/rails/webpacker/pull/870. Please take a look

PR #870, opened 2017-09-27
Make coffeescript a peer dependency

gauravtiwari:
Make coffeescript a peer dependency

Fixes: #859 and #868

### DIFF
--- a/lib/install/template.rb
+++ b/lib/install/template.rb
@@ -25,7 +25,7 @@ EOS
 end
 
 puts "Installing all JavaScript dependencies"
-run "yarn add @rails/webpacker"
+run "yarn add @rails/webpacker coffeescript@1.12.7"
 
 puts "Installing dev server for live reloading"
 run "yarn add --dev webpack-dev-server"
--- a/package.json
+++ b/package.json
@@ -19,7 +19,6 @@
     "babel-polyfill": "^6.26.0",
     "babel-preset-env": "^1.6.0",
     "coffee-loader": "^0.8.0",
-    "coffeescript": "^1.12.7",
     "compression-webpack-plugin": "^1.0.0",
     "css-loader": "^0.28.5",
     "extract-text-webpack-plugin": "^3.0.0",
@@ -44,6 +43,9 @@
     "eslint-plugin-import": "^2.2.0",
     "eslint-plugin-jsx-a11y": "^4.0.0",
     "eslint-plugin-react": "^6.10.0"
+  },
+  "peerDependencies": {
+    "coffeescript": ">= 1.12.7 || >= 2.x"
   },
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
--- a/yarn.lock
+++ b/yarn.lock
@@ -1122,10 +1122,6 @@ coffee-loader@^0.8.0:
   dependencies:
     loader-utils "^1.0.2"
 
-coffee-script@^1.12.7:
-  version "1.12.7"
-  resolved "https://registry.yarnpkg.com/coffee-script/-/coffee-script-1.12.7.tgz#c05dae0cb79591d05b3070a8433a98c9a89ccc53"
-
 color-convert@^1.3.0, color-convert@^1.8.2, color-convert@^1.9.0:
   version "1.9.0"
   resolved "https://registry.yarnpkg.com/color-convert/-/color-convert-1.9.0.tgz#1accf97dd739b983bf994d56fec8f95853641b7a"


package/loaders/coffee.js
```diff
@@ -1,4 +1,6 @@
+const { VERSION } = require('coffeescript')
+
 module.exports = {
   test: /\.coffee(\.erb)?$/,
-  loader: 'coffee-loader'
+  use: parseInt(VERSION, 10) >= 2 ? ['babel-loader', 'coffee-loader'] : ['coffee-loader']
```
renchap: If you include `babel-loader` here, you loose the options passed to `babel-loader` here: https://github.com/rails/webpacker/blob/master/package/loaders/babel.js (or added in `environment.js`).

What about either patching the `babel` loader entry to add support for `.coffee` or (probably safer) copy the options passed to the babel loader?

gauravtiwari: Oh yes, lets just copy since there isn't many.

renchap: If you copy, it means you will need to remember to change it here if it changes in the `babel.js` file or if the user wants to change `babel-loader` options, not sure it is the best way.

gauravtiwari: Yeah, this doesn't actually quite work, since it will fail if `coffeescript` package is not found. So, I guess for now we officially support CS1 and since the deps are now released to user space, folks who want to use CS2 can update the loader. 

```js
environment.loaders.set('coffee', //* ..... CS2 loader */)
```

Jascha-Sundaresan:
heh.. I need this

renchap:
If you do this, `coffeescript` will no longer be installed when installing Webpacker and `webpack` will fail because `coffee-loader` will not find the `coffeescript` package.
To solve this, maybe add `coffeescript` here: https://github.com/rails/webpacker/blob/master/lib/install/template.rb#L28 and a message explaining that it can be changed to CS2?

Also this PR will need an entry in the changelog as users will need to run `yarn add coffeescript` when updating (except if this is done in `webpack:install`)

Jascha-Sundaresan:
if you just "yarn add coffeescript" in the template, it'll default to installing 2 at this point, so a message explaining that it can be changed to CS1 seems more appropriate.

renchap:
Or you can `yarn install coffeescript@1.12.7` :)

Jascha-Sundaresan:
Lol, but that's the past! :^)

Jascha-Sundaresan:
as a temporary solution, adding this line to my package.json works.

It throws a warning, but that's preferable over not building.
```
"resolutions":{
    "coffeescript": "2.0.1"
  },
```

gauravtiwari:
@renchap Yes, that's expected - it's now more of a user responsibility to install and upgrade `coffeescript`. Will update CHANGELOG and README before release 👍

renchap:
@gauravtiwari I think it is a bad idea to require the user to manually install `coffeescript` after `webpacker:install` and have Webpack compilation fail if he does not do it. It is much better to install a default version (as it is required anyway) and point him on how to switch to the other one.

gauravtiwari:
Right, makes sense. Was trying to cut an extra dependency for folks who don't want coffeescript but I guess either way they now have option to remove it 👍

renchap:
Another way to do it is to not configure `coffee-loader` by default, and let the user run `yarn install coffeescript coffee-loader` and add the loader in `config/webpack/environment.js`.
We can also provide a default `coffee-loader` entry but not require it by default, and have something like `environment.enableLoader('coffee')` in environment.js, with a `webpacker:install:coffee` task that installs the packages and add the line to `environment.js` if you want an easy way.

I guess this could be extended to other non-essential loaders (ERB, …).

gauravtiwari:
Yeah, that's what we have done here earlier but lets keep this simple and do that as next step in another PR 👍

gauravtiwari:
@renchap How about using a separate extension for es6 version of coffee? `.es6.coffee` or something? That way we have another loader to take care of CS2 without messing up with existing loader.

renchap:
This will not work as `coffee-loader` will still try to load `coffeescript` 1.12 because it is a dependency for `@rails/webpacker` and latest yarn versions resolve `coffeescript` to this version even if you have CS2 in your own `package.json` (this changed with yarn 1.1.0 and I am not sure why).

The current workaround is the one mentionned by @Jascha-Sundaresan above (https://github.com/rails/webpacker/pull/870#issuecomment-332906199) but it generates a warning.

For CS2 to be able to be used with Webpacker, 2 things are required:
- Relax or remove the `coffeescript` dep on `@rails/webpacker`
- Have babel process the output of `coffee-loader`

gauravtiwari:
We already addressed this, didn't we?

> For CS2 to be able to be used with Webpacker, 2 things are required:

> Relax or remove the coffeescript dep on @rails/webpacker
> Have babel process the output of coffee-loader

Coffeescript is now a peer dependency. Babel will automatically process es6 output don't need to add other loader to it 👍

gauravtiwari:
@renchap Could you give this a try please 👍

Issue #348, opened 2017-05-08
How to use this gem from within rails engines?


fiedl:
I'm not sure if this is the right approach and maybe I've not understood this new part of the pipeline, yet, but:

Suppose, the major part of the app code lives inside a rails engine including all javascripts. And there are several main apps using that engine. The main apps only contain some layout changes and some minor patches.

Therefore, in order to integrate the webpacker gem into this setup, I'm trying to do the heavy lifting inside the engine, i.e. keep the work that has to be done inside all the main apps as little as possible.

How would I do that? Any pointers or suggestions are appreciated.

I'm not sure if this is a duplicate of https://github.com/rails/webpacker/issues/21. But in any case, I'd like to conclude this issue with a step-by-step guide how to approach this, for others facing the same use case.

**What to do in the engine**

1. Include `webpacker` in the `*.gemspec` file.
2. `require 'webpacker'` in the `lib/foo/engine.rb`.
3. ...

**What to do in each main app**

1. `bundle install`
2. Include `./bin/webpack-dev-server` in the `Procfile` if using [Foreman](https://ddollar.github.io/foreman).
3. ...

gauravtiwari:
@fiedl Out-of-the-box webpacker doesn't support engines yet. There are quite a few moving pieces that needs to be considered for this setup to work properly and usually people will have different use cases. I started doing some work on this but it's going to take time. 

Feel free to leave any ideas/suggestions you might have for this work 👍

fiedl:
@gauravtiwari Yes, after diving into this, I see your point. One would have to decide for each component where it belongs---the engine, the main app, or both. 

Some thoughts on this:

- The `bin/webpack-dev-server` is needed in the main app. One could add it to the `Procfile` of the main app to start the dev server with [Foreman](https://ddollar.github.io/foreman). Is there a way to have the webpack dev server started by [Pow](https://github.com/basecamp/pow)?
- But one would need the ability to add a package from the engine.
- The `app/javascript` files have to be loaded from both, the engine and the main app. The main app files need to have the final say. I guess if there are several engines, the files from all engines need to be loaded.
- There are some files like `.babelrc`, `.postcssrc.yml` that would seem right in the main app root folder; but I'm not sure what these files are doing, yet.
[...]

PR #875, opened 2017-09-28
Enables multiple entry points nested in rails root

lazylester:
With this change, webpacker config file can contain (e.g.) source_path: "**/app/javascript". This creates packs for all entry points below the rails roots, and specifically for the engines in vendor/gems.

If this change is not present, then the keys of manifest.json are malformed due to the globbed source path.

I would be happy to submit a test for this, if you can suggest how to test it. It doesn't break the existing tests, but the affected code may not be covered by the test suite.

### DIFF
--- a/package/environment.js
+++ b/package/environment.js
@@ -45,7 +45,7 @@ function getEntryObject() {
   const paths = sync(join(rootPath, glob))
   paths.forEach((path) => {
     const namespace = relative(join(rootPath), dirname(path))
-    const name = join(namespace, basename(path, extname(path)))
+    const name = basename(path, extname(path))
     result[name] = resolve(path)
   })
   return result


mltsy:
Hmm... I'm confused as to why prepending the rails root directory would result in an invalid manifest.json... isn't `/rails/root/**/app/javascripts` a valid glob path that resolves into the same number of valid keys as `**/app/javascripts`?  Is there a way you could easily demonstrate this failure with a unit test? (does this method output something that would be invalid in manifest.json when using globs?  Or is the failure caused by the output here interacting with something else?)

gauravtiwari:
@lazylester Thanks for this, but I am unsure what you are trying to achieve here. The code you removed is there to enable nested namespaces within packs so probably this needs bit more thinking.

lazylester:
@gauravtiwari I inferred namespaces from the name of the variables. Seems like a good idea. Since my change didn't break any tests, I could not look in the test suite to see what the intended behaviour should be.

What I'm trying to achieve is for packs to be compiled from engines within vendor/gems, and the compiled packs to be available in public/packs and correctly referenced by keys in manifest.json.

I imagined that declaring the source_path in config/webpacker.yml as "**/app/javascripts" would achieve my objective, and it nearly does. It compiles all the packs that the "**/app/javascripts" finds (main app, and engines within vendor/gems). However the keys of the manifest produced by the compilation do not resolve to paths within root/public. Without my proposed change, the manifest.json keys point to file paths within the engines, whereas they should point to file paths within public, if my understanding is correct.

lazylester:
@gauravtiwari sorry I didn't articulate very clearly in my last response.
_without my proposed change_ manifest.json has keys that look like (for example, in my app that has an engine called Complaint) this:
`"../../../../vendor/gems/complaints/app/javascript/packs/complaints.js"`

but when I wish to insert this file with a javascript_pack_include tag, I should just call 'complaints', as I would if it were a javascript_include tag.

Hopefully that's clearer!

mltsy:
This is the issue we're trying to solve: https://github.com/rails/webpacker/issues/348#issuecomment-332537307
(A way to make modules from an engine using webpacker automatically compile when the parent app's packs are compiled, and available to `javascript_include` as mentioned). 

@lazylester's idea was to use `**/app/javascripts` in the `sources` option of webpacker.yml, but that apparently produced an invalid manifest.json.  Another option would be to allow multiple entries for the sources option - or perhaps automatically find/include packs from any engines using webpacker?  Do you have any other suggestion for how to do this?

gauravtiwari:
@lazylester Thanks for explaining the use case however, I am afraid that this change isn't enough (due to varied use cases) and is also potentially breaking the name-spacing feature under packs so, I am going to close this. Hopefully will take a look at engine integration soon.

PR #878, opened 2017-09-29
Make `DevServer.running?`more resilient to errors (fixes #876)

### LINKED ISSUE #876: Errno::EBADF in `DevServer.running?
See similar issue https://github.com/rails/webpacker/issues/745.

I'm running on `jruby 9.1.13.0` and getting a `Errno::EBADF` when the webpack dev server is not running. Perhaps it's appropriate to catch all exceptions here and return a false? I can submit a PR if the solution is acceptable.

https://github.com/rails/webpacker/blob/a9a6a67fe34768ca7390eb93aa6a7574a746ce25/lib/webpacker/dev_server.rb#L12-L17

ketan:
Catch all exceptions in `DevServer.running?` and return false.
This is required because some ruby implementations (particularly
`jruby 9.1.13.0` throws a `Errno::EBADF`) which ends up being uncaught
and blows up the stack with no indication as to why the error happened.

### DIFF
--- a/lib/webpacker/dev_server.rb
+++ b/lib/webpacker/dev_server.rb
@@ -12,7 +12,7 @@ class Webpacker::DevServer
   def running?
     Socket.tcp(host, port, connect_timeout: connect_timeout).close
     true
-  rescue Errno::ECONNREFUSED, Errno::ETIMEDOUT, NoMethodError
+  rescue
     false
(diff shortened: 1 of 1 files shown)


gauravtiwari:
Thanks @ketan 💯  🍰

PR #879, opened 2017-09-29
Add ERB parsing to configuration

djpate:
I'm using Docker and need to pass in the Webpacker host from an ENV variable to the webpacker conf.
Right now, we don't do any ERB parsing making this pretty hard.

This is just adding simple ERB parsing when we load the webpack.yml file.


djpate:
Just saw the dup in the already closed issues.

Issue #850, opened 2017-09-24
Add API to update an environment's resolve.modules


n0nick:
Webpack config's `resolve.modules` is determined in an [un-exposed function within `environment.js`](https://github.com/rails/webpacker/blob/v3.0.1/package/environment.js#L54-L60), that is called from `toWebpackConfig()`.
Therefore it is always "too late" within an app's `config/webpack/*.js` files to customize this setting.

I realize I have `resolved_paths` in `webpacker.yml` for that, but the problem arises when I needed to determine path values dynamically on build time. Specifically for me, the case was adding paths to assets within gems.

I ended up wrapping `environment.toWebpackConfig()` like so:
https://gist.github.com/n0nick/b18a7978b8f27f56fe1bda33ff11ebd7
Which works, but it might be nice to have a clean API to add to `resolve.modules`.

molefrog:
Have the same problem trying to extend `resolve.plugins` (it's rarely used, but that's the only way to plug in `DirectoryNamedWebpackPlugin`). For sure, as a workaround I could write a function that wraps every `environment.toWebpackConfig()` invocation and merges the config, but this makes the whole point of having smart Environment object pointless.

Problem is: we can add support for both `resolve.modules` and `resolve.plugins` using the same approach they used for loaders and plugins and even submit a PR, but chances are they will get more requests from people to support more fields like that. So that's what I'm proposing: add a mechanism to extend the config in the Enviroment object using custom rules (middlewares? layers?).

n0nick:
I agree, this is partly why I didn't submit a PR - there are probably many other use cases. So I'd appreciate a more generic API to update an environment's properties.

gauravtiwari:
@molefrog Yeah, probably best to expose the base config object itself so anything can be added to that directly before calling `toWebpackConfig()`.  Would that work?

molefrog:
@gauravtiwari Generally, yes. Ideally, I would like to do something like: `environment.merge({ resolve: { plugins: [ ... ] }})`, but as I understand you're not planning to add `webpack-merge` as a dependency (and it totally makes sense). So maybe something like:

```javascript
// Part of development.js, webpack-merge is optional here
const merge = require('webpack-merge')

environment.transformBaseConfig(config => merge(config, {}))
```

Anyways, even if you just make base config accessible via `environment.baseConfig` that would work too! Thanks 👍

gauravtiwari:
[...]

PR #884, opened 2017-10-01
Extendable base config

gauravtiwari:
This PR exposes a base webpack config and a few new API's for updating loaders, plugins and configs so it can be extended from environment instead of separate environment files. 

### Issues this PR solves: 

1. At the moment if we have to update a config option we need to add that option to all environments instead of just the base environment. This PR adds a base config property on environment itself so we can set common options from environment itself. 

```js
environment.config.set('resolve.extensions', ['.foo', '.bar'])
environment.config.set('output.filename', '[name].js')
environment.config.merge({ output: {
    filename: '[name].js'
  }
})
```

2. The order of loaders and plugins in webpack config matters and currently we don't have a mechanism to enforce order when adding a loader or plugin. This PR addresses this by providing new API's apart from regular set method. 

```js
environment.loaders.prepend('json', jsonLoader)
environment.loaders.insert('json', jsonLoader, { after: 'style'} )
```
3. Expose `resolve.modules` through environment so it can be updated like loaders and plugins. 

#### 1. List API

- prepend
- append
- insert

#### 2. Object API

- set
- get 
- delete

### Examples

```js
// Object API
environment.config.set('output', {
  filename: '[name]-[hash].js'
})

// Using string path notation
environment.config.set('resolve.extensions', ['.foo', '.bar'])
environment.config.set('output.filename', '[name].js')

// Get an object
console.log(environment.config.get('resolve'))

// delete an object
environment.config.delete('output.chunkFilename')

// Loaders list API 
const jsonLoader =  {
  test: /\.json$/,
  exclude: /node_modules/,
  loader: 'json-loader'
}

// (first append and then prepends)
environment.loaders.set('json', jsonLoader)
environment.loaders.prepend('json', jsonLoader)
environment.loaders.insert('json', jsonLoader, { after: 'style'} )

// Updating a loader object
const jsonLoaderConfig = environment.loaders.get('json')
delete jsonLoaderConfig.loader
jsonLoaderConfig.use = ['json-loader']

// Plugins list API - prepend, append, insert
environment.plugins.set('CommonChunkVendor',
  new webpack.optimize.CommonsChunkPlugin({
    name: 'vendor', // Vendor code
    minChunks: (module) => module.context && module.context.indexOf('node_modules') !== -1
  })
)

// This will override previous set
environment.plugins.insert('CommonChunkVendor',
  new webpack.optimize.CommonsChunkPlugin({
    name: 'vendor', // Vendor code
    minChunks: (module) => module.context && module.context.indexOf('node_modules') !== -1
  })
, { before: 'manifest' })

// Insert at position API
const babelLoader = environment.loaders.get('babel')
environment.loaders.insert('coffee', {
  test: /\.coffee(\.erb)?$/,
  use:  babelLoader.use.concat(['coffee-loader'])
})

// Or
environment.loaders.insert('coffee', {
  test: /\.coffee(\.erb)?$/,
  use:  babelLoader.use.concat(['coffee-loader'])
}, { before: 'babel' })


// Resolved modules list API - prepend, append, insert 
environment.resolvedModules.set('vendor', 'vendor')

// Update a plugin
const manifestPlugin = environment.plugins.get('Manifest')
manifestPlugin.opts.writeToFileEmit = false
```

There are some other chores done here:  
- like converting all functions to use ES6 syntax
- Rename loaders to rules for consistency with webpack convention
- Use one syntax for rules i.e. convert `loader` to `use`
- Expose rules, paths config and merge through environment

Fixes: https://github.com/rails/webpacker/issues/850 and https://github.com/rails/webpacker/issues/794

Closes: https://github.com/rails/webpacker/pull/836

Please review

cc // @molefrog @n0nick @renchap @rossta


gauravtiwari:
Probably need an API for updating plugin options as well.

package/asset_host.js
```diff
@@ -1,11 +1,10 @@
 const config = require('./config')
 const { resolve } = require('path')
 
-function removeOuterSlashes(string) {
-  return string.replace(/^\/*/, '').replace(/\/*$/, '')
-}
+const removeOuterSlashes = string =>
```
javan: Do you prefer this syntax for any particular reason? Personally, I find it much harder to read. And `=>`  != ES6 function. Their behavior is slightly different although it doesn't matter in this case.

To me, this is like changing:
```js
class Environment {
  constructor() {
    ...
  }
}
```
to:
```js
const Environment = class {
  constructor() {
    ...
  }
}
```

rossta:
@gauravtiwari With this strategy, how would we modify one of the default plugins. For example, we need to modify the options for the `ManifestPlugin`, using [the workaround cited](https://github.com/rails/webpacker/issues/435#issuecomment-304453303) in #435.

package/asset_host.js
gauravtiwari: Nope, just little less curly's `{}` and parans `()` for one liner functions 😄

gauravtiwari: Yeah, for some cases it doesn't do much though.

gauravtiwari: Happy to change it back if you don't like it 😉  although at some places we have fat arrows so would be ideal to make it consistent in that sense

gauravtiwari:
@rossta The plugins are bit complicated since they don't share same API for defining options, which makes merging hard. So at the moment, one has to re-add that plugin with new options and it will override the previous plugin. I am investigating to see if it's possible to add similar API for plugins too.

WoH:
@gauravtiwari maybe store the loaders and plugins as an Array of Objects with a name and a value property. I experimented with something like that [here](https://github.com/WoH/webpacker/commit/9bc996379261189629f8ba8cb61fca9b0198f320), maybe it is actually a viable approach.

This way, you can add plugins/loaders/other properties via an API, but still have a way to access the predefined rules or plugins webpack does using a name and not relying on webpacker to change the order of the items in the array.

gauravtiwari:
Thanks @WoH I haven't got chance to look at this again but will do at some point this week 👍

gauravtiwari:
Thanks @WoH for pointers. Pushed some updates and updated the description, please take a look 👍 

@rossta Plugins can now be added, updated and removed (as loaders and resolvedModules)

```js
const manifestPlugin = environment.plugins.get('Manifest')
manifestPlugin.opts.writeToFileEmit = true
```

rossta:
@gauravtiwari Thanks for sharing this work. I like that we can configure plugins now. One concern I had with the current `Map` implementation is that plugin order couldn't be guaranteed and it indeed matters in Webpack; looks like you've addressed the issue with the new `ConfigList` API.

ytbryan:
Thanks for this @gauravtiwari 

Which one will take precedence? 
environment.loaders.set('json', jsonLoader, 'top')
environment.loaders.set('json', jsonLoader, 1)

gauravtiwari:
@ytbryan The one set later, `environment.loaders.set('json', jsonLoader, 1)` will be added as second item in the list since you are initialising same loader twice, the first will be simply deleted from the list. 

```js
environment.loaders.set('json', jsonLoader, 'top') // now at 0th position in array i.e. head
environment.loaders.set('json', jsonLoader, 1) // now at 1st position in array
// [firstLoader, jsonLoader, ...moreLoaders]
```

gauravtiwari:
@rossta Yes, now the order can be controlled any way you like, not just inserting at the top or bottom but at any position in the list.

javan:
Reading through your examples, I'm finding the 3rd positioning argument unclear.
```js
// Add json loader at top
environment.loaders.set('json', jsonLoader, 'top')
// Or 3rd in the list
environment.loaders.set('json', jsonLoader, 3)
```
I'd prefer to see a new set of methods with descriptive names like:
```js
// Insert at the top
environment.loaders.prepend('json', jsonLoader) 

// Insert at the bottom
environment.loaders.append('json', jsonLoader)

// Insert before/after an existing loader
environment.loaders.insert('json', jsonLoader, { before: 'typescript' })
environment.loaders.insert('json', jsonLoader, { after: 'typescript' })
```
wdyt?

gauravtiwari:
Yeah, I like it and had similar ideas earlier but didn't wanted to introduce a breaking change. Ahh I see, you mean extra set of methods apart from `set`?

javan:
Yeah, `append` would be the same as `set` and rest are new.

gauravtiwari:
@javan Updated the PR and description with new API changes (see examples in PR description)

gauravtiwari:
@ytbryan Could you please give this a try and see if it works as intended?

package/config_types/config_object.js
```diff
@@ -0,0 +1,49 @@
+const objectify = require('../utils/objectify')
+const deepAssign = require('../utils/deep_assign')
+const deepMerge = require('../utils/deep_merge')
+const { isStrPath, prettyPrint } = require('../utils/helpers')
+
+/**
+  * @class
+  * @extends { Object }
+*/
+class ConfigObject extends Object {
```
javan: Extending native `Object` (and `Array` above) could have unexpected consequences. I don't have any references to point to, but it's generally not something I do. Do you have any thoughts?  Alternatively, these could be bare classes that manage an object internally.

javan:
It's worth restating what problems / issues these changes resolve in the PR description.

There are a lot of nice API addditions here, but also a lot of changes and increase to the overall footprint. It might be nice to split this into two pull requests: one to resolve the issues, and a second to modernize and refactor (like your `package/loaders/*` → `package/rules/*` updates).

gauravtiwari:
Yeah, makes sense. I will do this first and then modernisation later 👍

package/config_types/config_object.js
gauravtiwari: Yeah, I read a bit about it but it seems for our use case it's safe to use since the usage is limited only to webpack config and unlike `prototype` it doesn't change native objects. 

I tried using bare classes with internal object but ran into some gotchas. We basically need native types with some custom methods on it. 

Do you think it's bad for our use case?

gauravtiwari:
Going to merge this as is, looks good to me 👍

package/config_types/config_list.js
```diff
@@ -0,0 +1,79 @@
...
+*/
+class ConfigList extends Array {
+  get(key) {
+    const index = this.getIndex(key, true)
+    return this[index].value
+  }
+
+  set(key, value) {
+    return this.add({ key, value })
+  }
```
javan: I think we should deprecate `set` in favor of `append` since `append` / `prepend` have nice symmetry. For now, let's have `set` call `append` instead of the other way around.

package/config_types/config_list.js
```diff
@@ -0,0 +1,79 @@
...
+
+  append(key, value) {
+    return this.set(key, value)
```
javan: This should call `add` directly instead of `set`.

package/config_types/config_list.js
gauravtiwari: Sure, will make a PR

package/config_types/config_list.js
gauravtiwari: 👍

PR #893, opened 2017-10-04
Removes env['SCRIPT_NAME'] when proxying requests to dev_server

marcelokanzaki:
In the following situation:

- Running an app in development under a **sub-uri** (e.g. /myapp)
- webpack-dev-server **running**

Webpacker proxies the pack's request to the dev_server with wrong name:

```/myapp/packs/application-ef71d57d8f408a7eef48.js```

Causing a **502 Bad Gateway** or **Incomplete response received from application** response.


gauravtiwari:
We can manually specify the `env["REQUEST_PATH"]` as `packs/something-hash.js` then we don't need to delete anything.

PR #897, opened 2017-10-05
Fix proxy scheme

jacobdam:
Currently, if we run the rails app (with webpacker) and proxy it under an HTTPS server. The webpacker will forward asset requests to webpack-dev-server as HTTPS (`env['HTTP_X_FORWARDED_PROTO']`). The correct protocol should be the protocol should be `Webpack.dev_server.protocol`.

### DIFF
--- a/lib/webpacker/dev_server_proxy.rb
+++ b/lib/webpacker/dev_server_proxy.rb
@@ -11,6 +11,9 @@ class Webpacker::DevServerProxy < Rack::Proxy
   def perform_request(env)
     if env["PATH_INFO"] =~ /#{public_output_uri_path}/ && Webpacker.dev_server.running?
       env["HTTP_HOST"] = env["HTTP_X_FORWARDED_HOST"] = env["HTTP_X_FORWARDED_SERVER"] = Webpacker.dev_server.host_with_port
+      env.delete("HTTPS")
+      env.delete("HTTP_X_FORWARDED_SSL")
+      env["HTTP_X_FORWARDED_PROTO"] = env["HTTP_X_FORWARDED_SCHEME"] = Webpacker.dev_server.protocol
 
       super(env)
     else


jacobdam:
It seems there are some related errors in `Webpacker::DevServerProxy#perform_request` (#893, #861). I tried to replace `rack-proxy` by `rack-reverse-proxy`. It works well for me.

gauravtiwari:
Webpack dev server also uses websockets so the current setup works since it always matches the host protocol. If we try to communicate from `http` to `https` or vice-versa, socket connection will fail. 

BTW, did you encountered any problems with the current setup?

jacobdam:
Yes @gauravtiwari, my Rails app is an iframe of another app (similar to Facebook Tabs Page). I run my app on `localhost:3000` and expose it to the world via `ngrok http 3000`. Ngrok create a tunnel forward all requests from `https://<subdomain>.ngrok.io` to my localhost:3000. `Webpacker::DevServerProxy` should proxy all requests from `localhost:3000` to webpack dev server (at `localhost:3035`) as HTTP. But it resolved the protocol of webpack dev server as HTTPS (the same protocol of rails app).

I understand that there are websockets (at http://localhost:3035/sockjs-node) for hot module reload  feature and webpacker does not proxy them. I am seeking for a solution for this problem. Perhaps I have to create another ngrok channel for that.

jacobdam:
For those who got into this problem: we can create a proxy from webpack dev server to Rails instead of Rails to webpack. Change `development.dev_server` in `config/webpacker.yml` as below:

```yaml
development:
  ...
  dev_server:
    ...
    proxy:
      /: http://localhost:3000/
```

Then open http://localhost:3035/ for development.

j-clark:
@jacobdam and others who may have this issue — i had the same issue, but realized it's possible to just run the webpack dev server to accept https requests.

if you change `development.dev_server.proxy` to `true` in `config/webpacker.yml`  like so

```yml
development:
  dev_server:
    https: true
```

then webpack-dev-server will start a server listening on `https://localhost:3035`, which should solve the problem

Issue #371, opened 2017-05-12
Typescript Integration (with React)


btc:
Is there an easy way to get started with Typescript without Angular?

gauravtiwari:
@btc You can remove the angular example and modules from package.json except typescript and it should work?

gauravtiwari:
Here is an example setup - https://www.typescriptlang.org/docs/handbook/react-&-webpack.html

btc:
For others who follow in my steps, here are the steps I took to get Typescript working with a Rails project generated using `--webpacker=react`:

1. Add new dependencies to `package.json` by executing `yarn add ts-loader typescript @types/react @types/react-dom`.
2. Run `bin/yarn` to update `yarn.lock`
3. Copy `tsconfig.json` from a Rails project generated using `--webpacker=angular` or from the code block below:
``` json
{
  "compilerOptions": {
    "declaration": false,
    "emitDecoratorMetadata": true,
    "experimentalDecorators": true,
    "lib": ["es6", "dom"],
    "module": "es6",
    "moduleResolution": "node",
    "sourceMap": true,
    "target": "es5"
  },
  "exclude": [
    "**/*.spec.ts",
    "node_modules",
    "public"
  ],
  "compileOnSave": false
}
```
4. Add option `"jsx": "react"` to the `compilerOptions` section of `tsconfig.json`.
5. Create `config/webpack/loaders/typescript.js` with the following contents:
``` javascript
module.exports = {
    test: /.(ts|tsx)$/,
    loader: 'ts-loader'
}
```
6. Rename your generated `hello_react.js` to `hello_react.tsx` and make it valid typescript:

```
// Run this example by adding <%= javascript_pack_tag 'hello_react' %> to the head of your layout file,
// like app/views/layouts/application.html.erb. All it does is render <div>Hello React</div> at the bottom
// of the page.

import * as React from 'react';
import * as ReactDOM from 'react-dom';

interface HelloProps { name: string; }

const Hello = (props: HelloProps) => (
  <div>Hello {props.name}!</div>
)

document.addEventListener('DOMContentLoaded', () => {
  ReactDOM.render(
    <Hello name="React" />,
    document.body.appendChild(document.createElement('div')),
  )
})
```

et Voila...

btc:
Given that Typescript and React work so well together, how would maintainers feel about making Typescript-flavored React available as a generator option? e.g. `--webpack=react-typescript`

gauravtiwari:
[...]

PR #908, opened 2017-10-07
Add missing instruction to create typescript loader

kureikain:
We are missing an extra steps in document: https://github.com/rails/webpacker/issues/371#issuecomment-301215682

### DIFF
--- a/docs/typescript.md
+++ b/docs/typescript.md
@@ -40,6 +40,14 @@ and rename your generated `hello_react.js` using react installer
 to `hello_react.tsx` and make it valid typescript and now you can use
 typescript, JSX with React.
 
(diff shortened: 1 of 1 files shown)


gauravtiwari:
@kureikain Post 3.0+ this loader is added by default. What version of webpacker are you using?

kureikain:
@gauravtiwari That's weird. I was using 3.0.2

```
cat Gemfile.lock| grep webpack
    webpacker (3.0.2)
  webpacker
```

Any idea?

gauravtiwari:
And npm module? `@rails/webpacker` It's in node package by default: https://github.com/rails/webpacker/blob/master/package/loaders/typescript.js

PR #914, opened 2017-10-08
Use `say` instead of `puts`

yhirano55:
In the application template, I think it's better to use [Thor::Shell::Basic#say](https://github.com/erikhuda/thor/blob/master/lib/thor/shell/basic.rb#L83-L96) instead of `puts`. 

And I colorized the finish message on green that means success.

image


gauravtiwari:
Great, thanks 💯

PR #920, opened 2017-10-09
Fix Vue example. HMR not working.

### LINKED ISSUE #894: HMR not working (fresh install)
I setup a new Rails project with `--webpack=vue`, enable `hmr`, render the provided example `hello_vue` component. Initial render works fine but HMR updates are failing with

```
TypeError: Cannot read property 'extend' of undefined....
```

The error is in `node_modules/vue-hot-reload-api/dist/index.js`. I don't think this library is responsible for it.

```
var newCtor = record.Ctor.super.extend(options) // line 141
```

Any idea what could be the reason for this?

[Here](https://github.com/haffla/rails5-vue) is the Rails app. First commit is just `rails new...`. Second commit is setting everything up.

haffla:
Never seen a Vue constructor be invoked like this. It's amazing that it kind of works, but breaks HMR. Fixes #894

### DIFF
--- a/lib/install/examples/vue/hello_vue.js
+++ b/lib/install/examples/vue/hello_vue.js
@@ -10,7 +10,9 @@ import App from '../app.vue'
 
 document.addEventListener('DOMContentLoaded', () => {
   document.body.appendChild(document.createElement('hello'))
-  const app = new Vue(App).$mount('hello')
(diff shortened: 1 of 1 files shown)


gauravtiwari:
Thanks 🍰

PR #936, opened 2017-10-12
Yarn Integrity Check: New Rails initializer

curvo:
This adds a new initializer step into the Rails initialization process that runs a yarn integrity check by default for the development environment.  It is also configurable via the standard environment configuration files (`config/environments/development.rb`, etc).  

At User Testing we have found this check to be a really valuable addition to our local Rails setup.  It leverages the standard `yarn check --integrity` command which compares the exact local install state of npm packages with the expected state.  This check when it passes takes only an additional 20-30ms to complete.  Because Rails does not currently offer any way to ensure that npm packages are up-to-date, our team of over fifty engineers have run into multiple instances where having an incorrect set of local javascript packages can cause webpack to fail without reliable messaging.  In a majority of these cases, the problems are resolved by running `yarn install`.

This update also adds more info about this initializer and instructions for configuring it to `README.md`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,7 +192,7 @@ yarn upgrade @rails/webpacker --latest
 
 ### Yarn Integrity
 
-By default in development, webpacker runs a yarn integrity check to ensure that all local npm packages are up-to-date.  This is similar to what bundler does currently in Rails, but for Javascript packages.  If your system is out of date, then Rails will not initialize and you will be asked to upgrade your local npm packages by running `yarn install`.
+By default in development, webpacker runs a yarn integrity check to ensure that all local npm packages are up-to-date. This is similar to what bundler does currently in Rails, but for JavaScript packages. If your system is out of date, then Rails will not initialize and you will be asked to upgrade your local npm packages by running `yarn install`.
 
 To turn off this option, you will need to override the default by adding a new config options to your Rails development environment configuration file (`config/environment/development.rb`):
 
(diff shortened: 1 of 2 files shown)


curvo:
@gauravtiwari - Take a look at this one.  We have this running a different way locally for our team and it has been really valuable to developer productivity.  Let me know what you think!

gauravtiwari:
Looks good to me, thanks 👍 . We probably need a similar check for checking `gem` version and `npm` package version so people don't run into odd behaviours.

README.md
```diff
@@ -189,6 +190,22 @@ bundle update webpacker
 yarn upgrade @rails/webpacker --latest
 ```
 
+### Yarn Integrity
+
+By default in development, webpacker runs a yarn integrity check to ensure that all local npm packages are up-to-date.  This is similar to what bundler does currently in Rails, but for Javascript packages.  If your system is out of date, then Rails will not initialize and you will be asked to upgrade your local npm packages by running `yarn install`.
```
yhirano55: Could you replace `Javascript` with `JavaScript` ?

yhirano55: It seems there are double spaces before `If your system is ...`.

lib/webpacker/railtie.rb
```diff
@@ -4,6 +4,44 @@
 require "webpacker/dev_server_proxy"
 
 class Webpacker::Railtie < ::Rails::Railtie
+  # Allows webpacker config values to be set via rails env config files
```
yhirano55: `Allows Webpacker config values to be set via Rails env config files`

dwbutler:
Would there be any disadvantage to running `yarn install` automatically if the integrity check fails? My team is constantly adding/updating dependencies.

PR #941, opened 2017-10-13
Skip check integrity if yarn.lock not exist

yhirano55:
Currently, when it doesn't exist `yarn.lock`, we cannot execute `rails webpacker:install`.

### Case

1. When It doesn't have `yarn.lock` file in Rails application, and it runs `bin/rails webpacker:install`.
2. Then it returns error messages:

```bash
$ bin/rails webpacker:install
yarn check v1.2.1
error Couldn't find an integrity file
error Found 1 errors.
info Visit https://yarnpkg.com/en/docs/cli/check for documentation about this command.


========================================
  Your Yarn packages are out of date!
  Please run `yarn install` to update.
========================================


To disable this check, please add `config.webpacker.check_yarn_integrity = false`
to your Rails development config file (config/environments/development.rb).
```

3. And it runs `yarn install`:

```bash
$ yarn install
yarn install v1.2.1
info No lockfile found.
[2/4] 🚚  Fetching packages...
[3/4] 🔗  Linking dependencies...
[4/4] 📃  Building fresh packages...

info Lockfile not saved, no dependencies.
✨  Done in 1.61s.
```

4. It doesn't save `yarn.lock`, and we cannot pass `yarn check --integrity` 😢 
5. If it doesn't exist `yarn.lock`, it's better not to check integrity.

### DIFF
--- a/lib/webpacker/railtie.rb
+++ b/lib/webpacker/railtie.rb
@@ -6,6 +6,7 @@ require "webpacker/dev_server_proxy"
 class Webpacker::Railtie < ::Rails::Railtie
   # Allows Webpacker config values to be set via Rails env config files
   config.webpacker = ActiveSupport::OrderedOptions.new
+  config.webpacker.check_yarn_integrity = false
 
   # ================================
   # Check Yarn Integrity Initializer
@@ -23,7 +24,7 @@ class Webpacker::Railtie < ::Rails::Railtie
   #     - edit config/environments/production.rb
   #     - add `config.webpacker.check_yarn_integrity = false`
   initializer "webpacker.yarn_check" do |app|
-    if app.config.webpacker[:check_yarn_integrity] || (!app.config.webpacker.key?(:check_yarn_integrity) && Rails.env.development?)
+    if File.exist?("yarn.lock") && app.config.webpacker.check_yarn_integrity
       ok = system("yarn check --integrity")
 
       if !ok
--- a/lib/install/template.rb
+++ b/lib/install/template.rb
@@ -16,6 +16,18 @@ directory "#{__dir__}/javascript", Webpacker.config.source_path
 say "Installing binstubs"
 run "bundle binstubs webpacker"
 
+say "Adding configurations"
+environment <<CONFIG, env: :development
+# Verifies that versions and hashed value of the package contents in the project's package.json
+config.webpacker.check_yarn_integrity = true
+
+CONFIG
+environment <<CONFIG, env: :production
+# Verifies that versions and hashed value of the package contents in the project's package.json
+config.webpacker.check_yarn_integrity = false
+
+CONFIG
+
 if File.exists?(".gitignore")
   append_to_file ".gitignore", <<-EOS
 /public/packs


curvo:
@yhirano55 Why didn't it save a yarn.lock file for you?  I thought that was the default behavior when you ran `yarn install`.  I just tried to reproduce the issue locally and couldn't reproduce it.

curvo:
Otherwise this seems like a reasonable update anyway...

curvo:
OK... I was just able to reproduce this if you are adding webpacker to an existing rails project for the first time.  It definitely fails there so thank you for catching this.  Looks good to me!

yhirano55:
@curvo When it is just generated Rails application, its `package.json` is this:

```json
{
  "name": "test_app",
  "private": true,
  "dependencies": {}
}
```

curvo:
yeah... i saw that.  Not sure if it could be considered a bug in yarn that a blank yarn.lock file isn't created anyway to cover a blank state, but I like your work-around.

yhirano55:
Thanks. It may be better to report this problem (bug?) to [yarn issues](https://github.com/yarnpkg/yarn/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aopen%20integrity).

lib/webpacker/railtie.rb
```diff
@@ -24,7 +24,7 @@ class Webpacker::Railtie < ::Rails::Railtie
   #     - add `config.webpacker.check_yarn_integrity = false`
   initializer "webpacker.yarn_check" do |app|
     if app.config.webpacker[:check_yarn_integrity] || (!app.config.webpacker.key?(:check_yarn_integrity) && Rails.env.development?)
-      ok = system("yarn check --integrity")
+      ok = !File.exist?("yarn.lock") || system("yarn check --integrity")
```
gauravtiwari: Bit confused with amount of conditional we have here. Lets just use `app.config.webpacker[:check_yarn_integrity]` and set it to true for development using installer. 

```rb
environment \
  "# Verifies that versions and hashed value of the package contents in the project’s package.json .\n" +
  "  # config.webpacker[:check_yarn_integrity] = true \n",
  env: 'development'

environment \
  "# Verifies that versions and hashed value of the package contents in the project’s package.json .\n" +
  "  # config.webpacker[:check_yarn_integrity] = false \n",
  env: 'production'
```

```rb
# Set it default in Railtie itself
# after config.webpacker = ActiveSupport::OrderedOptions.new
config.webpacker.check_yarn_integrity = false

if File.exist?('yarn.lock') && app.config.webpacker.check_yarn_integrity
  # ... run initializer 
end
```

yhirano55: I'll change this following your suggestion. Thanks!

yhirano55: I fixed it and confirmed:

#### After `webpacker:install`

```diff
diff --git a/config/environments/development.rb b/config/environments/development.rb
index 376aa7f..85b9ced 100644
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -1,4 +1,7 @@
 Rails.application.configure do
+  # Verifies that versions and hashed value of the package contents in the project's package.json
+  config.webpacker.check_yarn_integrity = true
+
   # Settings specified here will take precedence over those in config/application.rb.

   # In the development environment your application's code is reloaded on
diff --git a/config/environments/production.rb b/config/environments/production.rb
index 0a7c662..a05bc32 100644
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -1,4 +1,7 @@
 Rails.application.configure do
+  # Verifies that versions and hashed value of the package contents in the project's package.json
+  config.webpacker.check_yarn_integrity = false
+
   # Settings specified here will take precedence over those in config/application.rb.

   # Code is not reloaded between requests.
```

And on booting server process, it's ok to check integrity:

```bash
% bundle exec rails s
=> Booting Puma
=> Rails 5.2.0.alpha application starting in development
=> Run `rails server -h` for more startup options
yarn check v1.2.1
success Folder in sync.
✨  Done in 0.10s.
Puma starting in single mode...
* Version 3.10.0 (ruby 2.4.2-p198), codename: Russell's Teapot
* Min threads: 5, max threads: 5
* Environment: development
* Listening on tcp://0.0.0.0:3000
Use Ctrl-C to stop
```

yhirano55: @gauravtiwari Would you please code-review this again?

gauravtiwari:
Thanks @yhirano55 🍰  💯

PR #944, opened 2017-10-13
Enable test with edge Rails

yhirano55:
Enabled tests with edge Rails to notice bugs or problems early.

But edge Rails has many changes, so I think it's better to allow failure on edge Rails.

---

It was able to configure without any problems :)

---

And I noticed test failure on edge Rails (Maybe related [Rails#30744](https://github.com/rails/rails/pull/30744).)

But I'm going to open another pull request to fix this problem.

---

### Failure logs

```bash
$ bundle exec rake test
/home/travis/.rvm/rubies/ruby-2.2.8/bin/ruby -w -I"lib:test:lib" -I"/home/travis/.rvm/gems/ruby-2.2.8/gems/rake-12.1.0/lib" "/home/travis/.rvm/gems/ruby-2.2.8/gems/rake-12.1.0/lib/rake/rake_test_loader.rb" "test/command_test.rb" "test/compiler_test.rb" "test/configuration_test.rb" "test/dev_server_test.rb" "test/helper_test.rb" "test/manifest_test.rb" "test/rake_tasks_test.rb" 
Run options: --seed 7154
# Running:
......E
Error:
HelperTest#test_stylesheet_pack_tag_splat:
NoMethodError: undefined method `send_early_hints' for nil:NilClass
    /home/travis/.rvm/gems/ruby-2.2.8/bundler/gems/rails-680b853ddbdd/actionview/lib/action_view/helpers/asset_tag_helper.rb:143:in `stylesheet_link_tag'
    lib/webpacker/helper.rb:42:in `stylesheet_pack_tag'
    test/helper_test.rb:37:in `test_stylesheet_pack_tag_splat'
    minitest (5.10.3) lib/minitest/test.rb:107:in `block (3 levels) in run'
    minitest (5.10.3) lib/minitest/test.rb:204:in `capture_exceptions'
    minitest (5.10.3) lib/minitest/test.rb:104:in `block (2 levels) in run'
    minitest (5.10.3) lib/minitest/test.rb:255:in `time_it'
    minitest (5.10.3) lib/minitest/test.rb:103:in `block in run'
    minitest (5.10.3) lib/minitest.rb:350:in `on_signal'
    minitest (5.10.3) lib/minitest/test.rb:275:in `with_info_handler'
    minitest (5.10.3) lib/minitest/test.rb:102:in `run'
    minitest (5.10.3) lib/minitest.rb:839:in `run_one_method'
    minitest (5.10.3) lib/minitest.rb:324:in `run_one_method'
[...]

PR #952, opened 2017-10-15
Fix HelperTest on edge-rails

yhirano55:
Fixed helper_test.rb on edge-rails. Related #944 

image

### DIFF
--- a/test/helper_test.rb
+++ b/test/helper_test.rb
@@ -1,39 +1,44 @@
 require "test_helper"
 
 class HelperTest < ActionView::TestCase
+  tests Webpacker::Helper
+
+  attr_reader :request
+
   def setup
-    @view = ActionView::Base.new
-    @view.extend Webpacker::Helper
+    @request = Class.new do
+      def send_early_hints(links) end
+    end.new
   end
 
   def test_asset_pack_path
-    assert_equal @view.asset_pack_path("bootstrap.js"), "/packs/bootstrap-300631c4f0e0f9c865bc.js"
-    assert_equal @view.asset_pack_path("bootstrap.css"), "/packs/bootstrap-c38deda30895059837cf.css"
+    assert_equal "/packs/bootstrap-300631c4f0e0f9c865bc.js", asset_pack_path("bootstrap.js")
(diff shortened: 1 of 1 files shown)


test/helper_test.rb
```diff
@@ -1,39 +1,44 @@
...
-    @view.extend Webpacker::Helper
+    @request = Class.new do
+      def send_early_hints(links) end
+    end.new
   end
 
   def test_asset_pack_path
-    assert_equal @view.asset_pack_path("bootstrap.js"), "/packs/bootstrap-300631c4f0e0f9c865bc.js"
-    assert_equal @view.asset_pack_path("bootstrap.css"), "/packs/bootstrap-c38deda30895059837cf.css"
+    assert_equal "/packs/bootstrap-300631c4f0e0f9c865bc.js", asset_pack_path("bootstrap.js")
```
gauravtiwari: Should be `expected` and `actual` right? 

```rb
assert_equal asset_pack_path("bootstrap.js"), "/packs/bootstrap-300631c4f0e0f9c865bc.js"
```

test/helper_test.rb
```diff
@@ -1,39 +1,44 @@
...
+    assert_equal "/packs/bootstrap-c38deda30895059837cf.css", asset_pack_path("bootstrap.css")
   end
 
   def test_javascript_pack_tag
     assert_equal \
       %(<script src="/packs/bootstrap-300631c4f0e0f9c865bc.js"></script>),
-      @view.javascript_pack_tag("bootstrap.js")
+      javascript_pack_tag("bootstrap.js")
```
gauravtiwari: Ahh, looks like it's written the opposite way everywhere

yhirano55: Oh, My understanding is... "expected" is left position, and "actual" is right position.

https://github.com/seattlerb/minitest/blob/master/lib/minitest/assertions.rb#L159-L188

But If I need to modify all test cases, I'll do it. I'll respect to your opinion :)

gauravtiwari: Ahh, yes that's right 👍  kinda of confusing though, depending on how one interprets `expected` and `actual`

gauravtiwari:
Thanks 👍  🍰

PR #953, opened 2017-10-16
Fix indentation for environment setting

ta1kt0me:
Caller adds an indent to the first line only.


yhirano55:
@ta1kt0me Thanks for the contribution. But from Rails 5.2, these indentations are optimized on environment method. Please see [here](https://github.com/rails/rails/blob/master/railties/lib/rails/generators/actions.rb#L94-L120) and [here](https://github.com/rails/rails/blob/master/railties/lib/rails/generators/actions.rb#L319-L328).

Issue #927, opened 2017-10-11
Enabling HMR disables CSS extraction even when webpack-dev-server is not being used


bessey:
Per discussion in #905. 

## Reproduction:
In webpack.yml set development to `compile: true` and dev_server to `hmr: true`

## Expected outcome:
When using `bin/webpack-dev-server`, pages load fine.

When  using `bin/rails s` without the dev server, pages still load fine, because `compile: true`

## Actual outcome:
✅  When using `bin/webpack-dev-server`, pages load fine.

🚫  When using `bin/rails s` without the dev server, CSS is not extracted, and Rails 500s complaining of missing CSS assets.

gauravtiwari:
This is a simple fix - style loader needs to be aware of both options instead of just `hmr`. Feel free to make a PR to fix this 👍

jacortinas:
@gauravtiwari I've been running into the same issue and trying to solve it, wouldn't it mean that the style loader needs to know if the dev server is running or not? I think it's the same config, with the only thing changing is whether the dev server is up or not.

gauravtiwari:
@jacortinas Yep, almost identical to what ruby part does.

PR #957, opened 2017-10-16
Only extract CSS with dev server running and HMR enabled

jacortinas:
Resolves https://github.com/rails/webpacker/issues/927.

We only want to try to extract CSS when the dev server is up and running and HMR is enabled. This matches the logic in the [`stylesheet_pack_tag` view helper](https://github.com/rails/webpacker/blob/master/lib/webpacker/helper.rb#L41), so we don't run into the issue of having a `link` tag on the page pointing to a stylesheet that is not being extracted.

Tested by running the normal rake tests, but most importantly using it in an app that I'm currently working on that is running into the same problems. Ran the same webpack config with both a dev server on and off, correctly extracted vs. inlined.

### DIFF
--- a/package/loaders/style.js
+++ b/package/loaders/style.js
@@ -4,7 +4,8 @@ const { dev_server: devServer } = require('../config')
 
 const postcssConfigPath = path.resolve(process.cwd(), '.postcssrc.yml')
 const isProduction = process.env.NODE_ENV === 'production'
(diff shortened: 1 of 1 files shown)


gauravtiwari:
thanks 🍰

gauravtiwari:
Oops, totally forgot we need the logic here too: https://github.com/rails/webpacker/blob/master/package/loaders/vue.js

PR #958, opened 2017-10-16
Default environment to 'development' when NODE_ENV isn't set

### LINKED ISSUE #329: Upgrading from 1.0 to 1.2 caused a bunch of lint warnings
We are using eslint with the `eslint-import-resolver-webpack` package, and everything was working great until we upgraded webpacker. After running `webpacker:install` to update the config files, we started getting lots of strange eslint warnings like this: 

```
  1:1   error  Resolve error: Cannot read property 'host' of undefined                                           import/no-unresolved
  1:1   error  Resolve error: Cannot read property 'host' of undefined                                           import/no-extraneous-dependencies
  1:1   error  Resolve error: Cannot read property 'host' of undefined                                           import/no-duplicates
  1:1   error  Resolve error: Cannot read property 'host' of undefined                                           import/extensions
  1:1   error  Resolve error: Cannot read property 'host' of undefined                                           import/no-absolute-path
  1:1   error  Resolve error: Cannot read property 'host' of undefined                                           import/no-named-as-default
  1:1   error  Resolve error: Cannot read property 'host' of undefined                                           import/no-named-as-default-member
  2:1   error  'components' should be listed in the project's dependencies. Run 'npm i -S components' to add it  import/no-extraneous-dependencies
```

We eventually figured out that all these errors were caused by using absolute imports for files in our projects (e.g. `import foo from 'components/foo'`). My hunch is that this is related to the way the webpack config loads yaml files, but I'm not sure how to fix it.

dpikt:
Resolves #329 - issue caused when trying to compile without a `NODE_ENV` set.

### DIFF
--- a/package/config.js
+++ b/package/config.js
@@ -3,7 +3,8 @@ const { safeLoad } = require('js-yaml')
 const { readFileSync } = require('fs')
 
 const filePath = resolve('config', 'webpacker.yml')
-const config = safeLoad(readFileSync(filePath), 'utf8')[process.env.NODE_ENV]
+const environment = process.env.NODE_ENV || 'development'
+const config = safeLoad(readFileSync(filePath), 'utf8')[environment]
 
 const isBoolean = str => /^true/.test(str) || /^false/.test(str)
 


gauravtiwari:
thanks 🍰

PR #968, opened 2017-10-22
Fix Vue example import path

hzchirs:
Fix example relative path of app.vue

### DIFF
--- a/lib/install/examples/vue/hello_vue.js
+++ b/lib/install/examples/vue/hello_vue.js
@@ -33,7 +33,7 @@ document.addEventListener('DOMContentLoaded', () => {
 
 
 // import Vue from 'vue/dist/vue.esm'
-// import App from './app.vue'
+// import App from '../app.vue'
 //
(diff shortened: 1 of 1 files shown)


ytbryan:
The relative path is correct in the latest hello_vue.js 

https://github.com/rails/webpacker/blob/master/lib/install/examples/vue/hello_vue.js
 
cc @gauravtiwari

nholden:
I needed to make this change in my project to get started. The PR looks good to me!

> The relative path is correct in the latest hello_vue.js

The relative path is correct in [the first example](https://github.com/rails/webpacker/blob/master/lib/install/examples/vue/hello_vue.js#L9) but not in [the second example](https://github.com/rails/webpacker/blob/master/lib/install/examples/vue/hello_vue.js#L36).

ytbryan:
oops, misread..  this should be merged cc @gauravtiwari when you have time.

gauravtiwari:
Apologies, I was away. Merged 🎉

PR #976, opened 2017-10-26
uses last digest file per environment

gkemmey:
If you're relying on the lazy compilation, bundles can get out of sync while running tests with the rails server running because test and dev compile to two different places (`public/packs-test` vs `public/packs`).

If you change a bundled Javascript file and run a system test it'll have the latest change. But if you refresh your browser, it won't. Here's a sample app that shows that, with steps to reproduce in the readme: https://github.com/gkemmey/webpacker_test

This is a potential fix that uses a `.last-compilation-digest` per environment.

### DIFF
--- a/lib/webpacker/compiler.rb
+++ b/lib/webpacker/compiler.rb
@@ -69,7 +69,7 @@ class Webpacker::Compiler
     end
 
     def compilation_digest_path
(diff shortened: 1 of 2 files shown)


gauravtiwari:
thanks 👍

PR #985, opened 2017-10-31
Revert "Use Railtie instead of Engine"

guilleiguaran:
assets:precompile task is not being enhanced and therefore not executing
webpacker:compile when referencing webpacker master branch on Gemfile,
reverting back to Engine fix the problem immediately.

This reverts commit 6586ef7a969716abb44d80acf8d32659405c0e10 and 2c0c7dc12d2f9d6fc8b68ba2dcf857797d663955.

### DIFF
--- a/lib/webpacker/railtie.rb
+++ b/lib/webpacker/railtie.rb
@@ -3,7 +3,7 @@ require "rails/railtie"
 require "webpacker/helper"
 require "webpacker/dev_server_proxy"
 
-class Webpacker::Railtie < ::Rails::Railtie
+class Webpacker::Engine < ::Rails::Engine
   # Allows Webpacker config values to be set via Rails env config files
   config.webpacker = ActiveSupport::OrderedOptions.new
   config.webpacker.check_yarn_integrity = false
@@ -76,10 +76,5 @@ class Webpacker::Railtie < ::Rails::Railtie
       Webpacker.bootstrap
       Spring.after_fork { Webpacker.bootstrap } if defined?(Spring)
     end
-  end
-
-  rake_tasks do
-    tasks_path = File.expand_path("../tasks", __dir__)
-    Dir.glob("#{tasks_path}/**/*.rake").sort.each { |ext| load ext }
   end
 end


rafaelfranca:
How about fixing the loading of the rake task? `rake_tasks` is a valid method on railtie. The test_unit railtie uses it for example to add `rake test`.

guilleiguaran:
Strangely `assets:precompile` task isn't enhanced when referencing master:

```ruby
gem "webpacker", git: "https://github.com/rails/webpacker.git"
```

but cloning webpacker to a local folder and pointing to it in Gemfile works as expected:

```ruby
gem "webpacker", path: "~/src/webpacker"
```

guilleiguaran:
Found something, adding explicitly to `sprockets-rails` on Gemfile before of `webpacker` makes this work:

This **doesn't work**:
```
gem "webpacker", git: "https://github.com/rails/webpacker.git"
gem "sprockets-rails"
```

This **work**:
```
gem "sprockets-rails"
gem "webpacker", git: "https://github.com/rails/webpacker.git"
```

And again changing Webpacker to an Engine makes this work regardless of the order.

This **work**:
```
gem "webpacker", git: "https://github.com/rails/webpacker.git", branch: "revert-to-engine"
gem "sprockets-rails"
```

guilleiguaran:
Note that in all cases the Webpacker rake tasks are correctly loaded is just enhancing that doesn't work because [Rake::Task.task_defined?("assets:precompile")](https://github.com/rails/webpacker/blob/master/lib/tasks/webpacker/compile.rake#L26) returns false in the failing cases.

matthewd:
As a matter of principle it does feel slightly more correct for this to be a Railtie instead of an Engine. I don't feel strongly about that, but do worry about whether this Railtie-vs-Engine difference is by design, or is just luck that it ends up running at a 'better' time.

However we decide to fix it, it sounds like we need a test to prove that the task gets enhanced regardless of the load order.

rafaelfranca:
The difference between an engine and railtie regardless the load order is:

The railtie executes the rake_tasks blocks in order of definition.
The engine loads the `lib/tasks` files after all railties are loaded.

I don't think it is by design. It looks like an implementation detail.

I also thing an Railtie is more correct, but I don't have strong opinion either.

gauravtiwari:
@guilleiguaran Guess we fixed this with `rake_tasks` method on railtie right? I can't reproduce this on master.

guilleiguaran:
@gauravtiwari I'm going to test this again and report results here

gauravtiwari:
Sure 👍  thanks

gauravtiwari:
@guilleiguaran Were you able to test this one? Thinking to make a new release this weekend.

guilleiguaran:
I can reproduce this as described in https://github.com/rails/webpacker/pull/985#issuecomment-340677364 and it still being fixed immediately switching to `revert-to-engine` branch.

I'll merge this for now and check if we can do something else about the loading order issue.

PR #990, opened 2017-11-02
Fix binstubs path in webpacker:check_binstubs rake task

monad-y:
Fix `rails webpacker:check_binstubs` rake task bug.
In new rails project, this rake task fails even if `rails webpacker:install` has already executed. 

----

`bin/webpack-dev-server` and `bin/webpack` is converted to binstubs by #833 . 
These files are not in `bin/` directory but paths of them are remained around this repository yet.
we might as well remove all paths.


rossta:
I'm not able to reproduce the failure you describe. I just created a new rails app with webpacker and ran `rails webpacker:install`. When I `rails webpacker:check_binstubs`, I don't see any failures.

The install task installs the binstubs here: https://github.com/rails/webpacker/blob/05bf821ce983a2ad88fba0da476023e67f8efe43/lib/install/template.rb#L17. 

Could you provide more info about the issue you're seeing and how to reproduce?

monad-y:
> Could you provide more info about the issue you're seeing and how to reproduce?

Sorry, I have not provided important information.
If bundle bin path is changed manually by `bundle install --path=vendor/bundle --binstubs=vendor/bundle/bin`, `rails webpacker:check_binstubs` will fail.
(I think it happens also by `bundle binstubs --path=vendor/bundle/bin`)

```sh
% mkdir rails_app && cd rails_app
% bundle init
Writing new Gemfile to /***/rails_app/Gemfile
% nano Gemfile
% cat Gemfile
# frozen_string_literal: true

source "https://rubygems.org"

git_source(:github) {|repo_name| "https://github.com/#{repo_name}" }

gem "rails"
% bundle install --path=vendor/bundle --binstubs=vendor/bundle/bin
(omit)
% rails new . --webpack=react
(omit)
        run  yarn add @rails/webpacker coffeescript@1.12.7 from "." 
(omit)
Webpacker successfully installed 🎉 🍰
       rails  webpacker:install:react
Webpack binstubs not found.
Have you run rails webpacker:install ?
Make sure the bin directory or binstubs are not included in .gitignore
Exiting!
         run  bundle exec spring binstub --all
* bin/rake: spring inserted
* bin/rails: spring inserted

% rails webpacker:check_binstubs
Webpack binstubs not found.
Have you run rails webpacker:install ?
Make sure the bin directory or binstubs are not included in .gitignore
Exiting!
```
Or, should not we change bundle bin path in rails app?

cjlarose:
I'm experiencing a similar problem. When running `bundle exec webpacker:install` for the first time, I'm getting a message like

```
Skipped webpack and webpack-dev-server since they already exist.
If you want to overwrite skipped stubs, use --force.
```

even when files do not exist at `./bin/webpack` or `./bin/webpack-dev-server` under the Rails root.

Here's a Dockerfile that demonstrates a repro

```
FROM ruby:2.3.5-alpine

RUN apk update && apk add --no-cache build-base nodejs sqlite-dev
RUN npm install -g yarn
RUN gem install rails

WORKDIR /app

RUN rails new /app
RUN sed -i "s/gem 'tzinfo-data'.*//" Gemfile
RUN bundle
RUN echo "gem 'tzinfo-data'" >> Gemfile
RUN echo "gem 'webpacker', '~> 3.0'" >> Gemfile
RUN bundle

RUN bundle exec rails webpacker:install

CMD bundle exec rails webpacker:check_binstubs
```

Build the image and run the container

```
$ docker build -t bug .
$ docker run --rm -ti bug
Webpack binstubs not found.
Have you run rails webpacker:install ?
Make sure the bin directory or binstubs are not included in .gitignore
Exiting!
```

If you get into the container, you'll find that no files were written to `/app/bin/webpack` or `/app/bin/webpack-dev-server`, but instead executables are available at `/usr/local/bundle/bin/webpack` and `/usr/local/bundle/bin/webpack-dev-server`.

In other words, the installation process creates executables alongside ofter gem executables (like `rails` and `rake`), but the task `webpacker:check_binstubs` seems to expect files in the `./bin` subdirectory of the Rails root.

gauravtiwari:
Thanks @monad-y 👍

PR #1011, opened 2017-11-14
Cross-origin warning in React error handler

MrLeebo:
Hi! I recently came across this project and got the opportunity to apply it to an application I'm working on. I want to start by saying I am really impressed with how easy you guys made it to set this up. Great job!

I noticed when I was setting up for React that the error handler was getting masked by a cross-origin warning in development. For more information, check out React's documentation on [Webpack Cross-origin Errors](https://reactjs.org/docs/cross-origin-errors.html#webpack).

The warning occurs because the webpack bundle is wrapping the code in an `eval` statement which browsers treat as cross-origin. This PR changes the default webpack devtool to `cheap-module-source-map`.

### DIFF
--- a/lib/install/config/webpack/development.js
+++ b/lib/install/config/webpack/development.js
@@ -1,3 +1,6 @@
 const environment = require('./environment')
 
-module.exports = environment.toWebpackConfig()
+const config = environment.toWebpackConfig()
+config.devtool = 'cheap-module-source-map'
+
+module.exports = config


MrLeebo:
Not sure why this change would cause any ruby tests to fail? They all pass locally.

gauravtiwari:
Thanks for the PR. We choose `cheap-eval-source-map` for speed and quality: https://webpack.js.org/configuration/devtool/#development but if this is causing cross-origin errors, please consider changing it in your own project. The default option is more generic here suited for various types of projects and use cases but the API is there to change if you need to :)

PR #1012, opened 2017-11-15
Remove check for webpacker binstubs. (Fixes #988)

### LINKED ISSUE #988: bin/webpacker-dev-server not optional
Although there are many references to the webpacker-dev-server being optional, the presence of `bin/webpacker-dev-server` is [enforced](https://github.com/rails/webpacker/blob/105c9e8d11a70b8f6f5f7d5232537665599bfc4a/lib/tasks/webpacker/check_binstubs.rake#L4) by the `check_binstubs` rake task.

Is there interest in making this guard more intelligent? Just remove it? Or be OK with folks adding dummy binstubs?

ketan:


gauravtiwari:
Thanks @ketan Lets just make `webpack-dev-server` optional and leave `webpack` binstub as required one - since `assets:precompile` and on-demand compiler depends on it.

ketan:
Isn't it just easier to leave a `script` in `package.json` and call that instead of calling a ruby script (which `exec`s `node`) from within a rake task?


```json
// package.json
{
  "scripts": {
    "webpack-dev": "NODE_ENV=development webpack --config config/webpack/development.js",
    "webpack-prod": "NODE_ENV=production webpack --config config/webpack/production.js",
  },
  "dependencies": {
    ...
  },
  "devDependencies": {
    ...
  }
}
```

ketan:
I'm happy to submit a patch to the `run_webpack` method, since we've monkey patched that method as well because the default webpack binary was not good enough for us.

https://github.com/rails/webpacker/blob/deafb46f6e0b9075ee3bedebd0c036dabe0be981/lib/webpacker/compiler.rb#L53-L65

gauravtiwari:
> Isn't it just easier to leave a script in package.json and call that instead of calling a ruby script (which execs node) from within a rake task?

Yep, we can do that but the binstubs make things easier when it comes to wrapping env variables with tools like foreman or invoker. Plus, binstubs are more conventional to Ruby and Rails folks. It would be ideal to use that though. 

> webpack binary was not good enough for us.

What issues have you encountered?

ketan:
> > webpack binary was not good enough for us.
> 
> What issues have you encountered?

Sorry if this seems like a rant, it is not :)

One of the issues we face (because of severe customisation of webpack configuration we've made) comes from how `require('@rails/webpacker')` works.

That object that `require('@rails/webpacker')` returns is an `Environment` object, which allows for limited (perhaps opinionated, and, by design?) customisation, this did not work out-of-the-box for us because of lack of support for context paths, where the rails app is hosted under `http://example.com/foo` instead of `http://example.com/`, also we needed to set a very specific order of [`resolve.modules`](https://webpack.js.org/configuration/resolve/#resolve-modules).

Another important thing with how this is currently setup, we've built multiple chunks using the `CommonsChunkPlugin`, which outputs several chunks to keep individual JS file sizes low and ensure optimal usage of browser caches. However the plugin does not understand dependencies between the entrypoints and the chunks — the expectation being that the html template should `stylesheet_pack_tag` the entrypoint, along with all its dependencies. The old-school rails-webpack gem on the other hand, had nice support for manifest written out by the (https://www.npmjs.com/package/stats-webpack-plugin)[stats-webpack-plugin] which wrote out the dependency tree for each entrypoint. We've "ported" some of this code in our rails app as well.

In hind-sight, it would have been nice, if `require('@rails/webpacker')` returned an opinionated JS pojo that could be directly fed into webpacker, this would allow developers to customize it to their needs without resorting to work-arounds.

Again, a lot of these problems are because of some complexity we've inherited, some we've built ourselves. This seems like the only gem that can very easily put rails and webpack together, and we're trying to make the best of it, I'm hoping to improve it as we go along.

gauravtiwari:
No worries, Ahh I see. Does this PR solves some or all of your problems? Please see PR description.  https://github.com/rails/webpacker/pull/884

gauravtiwari:
> I'm hoping to improve it as we go along.

Yeah, lets do that. Feel free to suggest ideas.

ketan:
I can see how #884 is useful, any thoughts about using the [stats-webpack-plugin](https://www.npmjs.com/package/stats-webpack-plugin) instead of `webpack-manifest-plugin`, or providing an option to change between them?

gauravtiwari:
Don't know what `stats-webpack-plugin ` does. Does it print like file size for each entry with deps? Can they not be used together or they are both same?

gauravtiwari:
If you do need more fine grained stats, you could tweak stats option per your need although I find it bit annoying to read through all that ;)  https://webpack.js.org/configuration/stats/

screen shot 2017-11-15 at 21 54 49
screen shot 2017-11-15 at 21 54 58

gauravtiwari:
Fixed in #1024

Issue #1021, opened 2017-11-18
Object.entries not available on Node 6


stevenharman:
I'm trying to use `master` to get the new base config object functionality (from #884), and things were working fine locally. But not when I tried deploying to Heroku. I believe I've traced the issue down to a usage of `Object.entries` in the `environment.js` module. When #884 was merged into `master` it introduced the following line:

https://github.com/rails/webpacker/blob/d6f7be8b8de10bb571b3f5615cfb8aeb1ce74ed3/package/environment.js#L19

I then noticed locally I was running Node `8.3`, but the default Node for the Heroku Ruby build pack is `6.11.1`. Upon upgrading my Heroku Node version to `^8.0.0`, things are working again. Since Webpacker claims to work on Node `6.0.0`, this seems like a bug.

The error on Heroku was:

```text
       Webpacker is installed 🎉 🍰
       Using /tmp/build_d5a48d880470f84bda4a424a5a17b9ab/my-app-web-96e7976d5146d189dc085531dba87635fe06f492/config/webpacker.yml file for setting up webpack paths
       Compiling…
       Compilation failed:
       /tmp/build_d5a48d880470f84bda4a424a5a17b9ab/my-app-web-96e7976d5146d189dc085531dba87635fe06f492/node_modules/@rails/webpacker/package/environment.js:19
       Object.entries(rules).forEach(([key, rule]) => result.set(key, rule))
       ^
       TypeError: Object.entries is not a function
       at getLoaderList (/tmp/build_d5a48d880470f84bda4a424a5a17b9ab/my-app-web-96e7976d5146d189dc085531dba87635fe06f492/node_modules/@rails/webpacker/package/environment.js:19:10)
       at Environment (/tmp/build_d5a48d880470f84bda4a424a5a17b9ab/my-app-web-96e7976d5146d189dc085531dba87635fe06f492/node_modules/@rails/webpacker/package/environment.js:82:20)
       at /tmp/build_d5a48d880470f84bda4a424a5a17b9ab/my-app-web-96e7976d5146d189dc085531dba87635fe06f492/node_modules/@rails/webpacker/package/environments/production.js:7:5
       at createEnvironment (/tmp/build_d5a48d880470f84bda4a424a5a17b9ab/my-app-web-96e7976d5146d189dc085531dba87635fe06f492/node_modules/@rails/webpacker/package/index.js:14:10)
[...]

PR #1022, opened 2017-11-18
Deprecate set and remove object.entries

gauravtiwari:
Fixes: #1021 

Deprecates `set`


stevenharman:
@gauravtiwari should probably mention the deprecation in the Changelog, no? And maybe call out examples of what will need to change?

gauravtiwari:
Yeah, good call. Will push that in final changelog before release. There is warning message though whenever set is called.

Issue #967, opened 2017-10-21
rake assets:precompile without running webpacker:compile


deesx:
(not sure how to label it, but it'd be "help wanted")

It's kind of all in the title, to expose the context : We have a single rails app, containing the backend — not really using webpacker, standard pipeline *and* the frontend — new webpacker pipeline. Those end up being deployed as two different apps in production. Not ideal, but we'd like to stick to it.

We would need to compile our assets via :
- [NOT OK] : `rake assets:precompile` for the backend, *without* running `webpacker:compile`
- [OK] : `webpacker:compile` only  when deploying the frontend

 [Looking here](https://github.com/rails/webpacker/blob/28d453121d7ff73c31feb58a80c7a30271f904ff/lib/tasks/webpacker.rake#L148-L152), feels like you can't skip running `webpacker:compile` when running `rake assets:precompile`, and I'm not sure to be able to ovveride it cleanly. Or maybe there's an option available to skip `webpacker:compile` ?

Thx a lot if any help

rossta:
The use case you're describing sounds interesting! Also, kind of unique, so I'm not sure personally how useful this feature would be.

One thing you can try is recreating the Sprockets-Rails rake task in the deployed environment where you want to omit Webpacker. You just need to do it conditionally after the bundled rake tasks are loaded in the `Rakefile`. I've only just played around with this in a sample app locally (YMMV), but here's what it could look like without conditionals:

```ruby
# Add your own tasks in files placed in lib/tasks ending in .rake,
# for example lib/tasks/capistrano.rake, and they will automatically be available to Rake.

require_relative 'config/application'

Rails.application.load_tasks

# Redefine `assets:precompile` and related tasks
require 'sprockets/rails/task'
Sprockets::Rails::Task.new(Rails.application)
```
Initializing `Sprockets::Rails::Task` will clobber any pre-existing definition of the `asset:precompile` task (and its relatives) along with any enhancements like the Webpacker compilation task ([source](https://github.com/rails/sprockets-rails/blob/58473062db8ccf5f49d18a5ed16d67336d34f850/lib/sprockets/rails/task.rb#L52)).

deesx:
thx a lot for this pointer! well without discussing the use case, once you feel the need you get a bit trapped as there is no optional exit

[...]

Issue #811, opened 2017-09-14
skip webpacker compile in certain environments?


mlennie:
We host an app on two servers. One is our webapp and the other is a worker. Both use the same code. 

For the worker, we do not need to use webpacker and it even is currently interfering with deploying on our worker server. 

Is there a way to skip webpacker compiling? Such as through an environmental variable?

ytbryan:
related? https://github.com/rails/webpacker/issues/905

gauravtiwari:
Made a PR - #1023

Issue #618, opened 2017-08-07
Only re-compile assets that have changed?


catskull:
We're using webpacker in production. Every time we deploy our servers, assets:precompile runs on each server adding ~30s * the number of servers we have. As we continue to scale, deploys will keep growing. Additionally, assets:precompile eats up a lot of CPU so doing a production deploy mid-day nearly halts production.

Is there any good way to only re-compile assets that have changed? Or any other strategies for dealing with this issue? Ideally I'd like to run assets:precompile locally and just deploy that with the rest of the code, no idea if that's feasible.

Any input is appreciated!

gauravtiwari:
@catskull This feature is only available in testing environment. Perhaps makes sense to add it to production as well. You compile locally and commit your assets to source control though.

catskull:
@gauravtiwari thanks! We were using `capistrano-faster-assets` but it 100% doesn't work with webpacker. Currently looking into `capistrano-local-precompile` which may do just what we need! I'll report back with what we find as I assume others are/will run into this issue.

javan:
I suggest running `assets:precompile` on a single server and then `rsync`'ing the compiled assets to the rest of your servers. That's what we do at Basecamp using a couple small Cap tasks.

catskull:
@javan thanks for the tip! I spent about a solid day fighting `capistrano-local-precompile`, but ultimately gave up. Just couldn't get it to work right for webpack assets.

Any chance you might share your cap tasks? I'm by no means a devops guy.

javan:
I'm not the author of these so they're presented without guarantee or support:
```ruby
after 'deploy:bundle', 'assets:symlink_cache'
after 'assets:symlink_cache', 'assets:precompile'
after 'assets:precompile', 'assets:sync'

namespace :assets do
  task :precompile, roles: :asset_precompiler, only: { primary: true } do
    run "mkdir -p #{shared_path}/assets && ln -nfs #{shared_path}/assets #{release_path}/public/assets"
    run "cd #{release_path} && #{elapsed} RAILS_ENV=#{rails_env} RAILS_GROUPS=assets #{rake} assets:precompile --trace"
  end

  task :sync, roles: :asset_users, on_no_matching_servers: :continue do
    run "mkdir -p #{shared_path}/assets && ln -nfs #{shared_path}/assets #{release_path}/public/assets"
    precompiler = roles[:asset_precompiler].detect { |server| server.options[:primary] }
[...]

PR #1023, opened 2017-11-18
Allow skipping webpacker compile when running assets:precompile

gauravtiwari:
Fixes: https://github.com/rails/webpacker/issues/967 and https://github.com/rails/webpacker/issues/811

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,16 @@ environment.resolvedModules.set('vendor', 'vendor')
 - Enable sourcemaps in `style` and `css` loader
 
 
+### Added
+
+- Allow skipping webpacker compile using an env variable
+
+```bash
+SKIP_WEBPACKER_COMPILE=yes|true|y|t
+SKIP_WEBPACKER_COMPILE=true bundle exec rails assets:precompile
+```
+
+
 ## [3.0.2] - 2017-10-04
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,8 +56,8 @@ environment.resolvedModules.set('vendor', 'vendor')
 - Allow skipping webpacker compile using an env variable
 
 ```bash
-SKIP_WEBPACKER_COMPILE=yes|true|y|t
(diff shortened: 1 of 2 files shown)


javan:
How about inverting it? `WEBPACKER=false rails assets:precompile`

gauravtiwari:
Ahh yes, makes sense. Will push an update.

gauravtiwari:
Although not sure, if just using `WEBPACKER` as env variable is fine. May be `WEBPACKER_PRECOMPILE`

javan:
Could the issues referenced above just omit `webpacker` from their `Gemfile` using an env var of their choosing?

```ruby
# Gemfile
gem 'webpacker', '~> 3.0' unless ENV['WORKER_APP']
```

javan:
I imagine that both of those issues would benefit from compiling assets on a single server and then `rsync`ing them to the rest. See https://github.com/rails/webpacker/issues/618#issuecomment-322053127 and https://github.com/rails/webpacker/issues/618#issuecomment-322062852.

gauravtiwari:
> Could the issues referenced above just omit webpacker from their Gemfile using an env var of their choosing?

Yeah that would work but I guess this solution might produce some edge cases - like app has some code related to webpacker but the Gem isn't installed for that environment and build is failing.

> I imagine that both of those issues would benefit from compiling assets on a single server and then rsyncing them to the rest. See #618 (comment) and #618 (comment).

Yep but I think having an option to skip enhancement would be better, in case folks wants to have more control over how they deploy their apps. 

Perhaps, in future we can totally ditch enhancements and just allow invoking `webpacker:compile` during deployment. For Heroku, I guess this can be inside ruby buildback to detect if webpacker (which it does anyway) is present run that hook like `assets:precompile`

For other deployments, like capistrano this can be easily added in deployment script. 

What do you think?

javan:
One more idea, and I'm not sure how to do this with rake, but we could alias the original `assets:precompile` task before enhancing it as `assets:precompile:without_webpacker` or something.

gauravtiwari:
Yeah, great idea 👍  although it seems since rake lazily loads all tasks, after boot the alias task is enhanced too

Issue #992, opened 2017-11-03
Warnings: has unmet peer dependency


phlegx:
I use the latest Rails (5.1.4), Yarn (1.2.1) and Webpacker (3.0.2) version to test Vue. I have created a project with the following commands:

```
$ gem install bundler rails
$ rails new example_app --webpack=vue
```

I get this (4 has unmet peer dependency) warnings:

```
$ cd example_app
$ yarn install --check-files
yarn install v1.2.1
[1/4] Resolving packages...
[2/4] Fetching packages...
info fsevents@1.1.2: The platform "linux" is incompatible with this module.
info "fsevents@1.1.2" is an optional dependency and failed compatibility check. Excluding it from installation.
[3/4] Linking dependencies...
warning "postcss-cssnext@3.0.2" has unmet peer dependency "caniuse-lite@^1.0.30000697".
warning "vue-loader@13.3.0" has unmet peer dependency "css-loader@*".
warning "webpack-dev-server@2.9.4" has unmet peer dependency "webpack@^2.2.0 || ^3.0.0".
warning "webpack-dev-middleware@1.12.0" has unmet peer dependency "webpack@^1.0.0 || ^2.0.0 || ^3.0.0".
[4/4] Building fresh packages...
Done in 8.07s.
```

My package.json looks like this:

```json
{
  "name": "example_app",
  "private": true,
  "dependencies": {
    "@rails/webpacker": "^3.0.2",
    "coffeescript": "^2.0.2",
    "vue": "^2.5.2",
    "vue-loader": "^13.3.0",
    "vue-template-compiler": "^2.5.2",
  },
  "devDependencies": {
    "webpack-dev-server": "^2.9.4"
  }
}
```

gauravtiwari:
Please feel free to bump up/down your deps. We will try to lock down deps in next release 👍

hovancik:
I'm trying the same (inside docker) and my install does not have `bin/webpack-dev-server`. Could it be caused by that? How to set deps correctly?

hovancik:
@gauravtiwari is there a way to start a new project now or should I wait for next release? I'm trying `rails webpacker:install` and `rails webpacker:install:vue`but I' always getting: 

```
Webpack binstubs not found.
Have you run rails webpacker:install ?
Make sure the bin directory or binstubs are not included in .gitignore
```

gauravtiwari:
@hovancik No, you can start one now. Is this a new project you are getting this error?

hovancik:
[...]

PR #1024, opened 2017-11-18
Make dev server binstub optional

gauravtiwari:
Fixes: #988 and https://github.com/rails/webpacker/issues/992

Since we now have proper binstubs.

### DIFF
--- a/lib/webpacker/compiler.rb
+++ b/lib/webpacker/compiler.rb
@@ -53,7 +53,7 @@ class Webpacker::Compiler
     def run_webpack
       logger.info "Compiling…"
 
-      sterr, stdout, status = Open3.capture3(webpack_env, "#{RbConfig.ruby} ./bin/webpack")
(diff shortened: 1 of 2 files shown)


lib/webpacker/compiler.rb
```diff
@@ -53,7 +53,7 @@ def record_compilation_digest
     def run_webpack
       logger.info "Compiling…"
 
-      sterr, stdout, status = Open3.capture3(webpack_env, "#{RbConfig.ruby} ./bin/webpack")
+      sterr, stdout, status = Open3.capture3(webpack_env, "bundle exec webpack")
```
javan: `./bin/*` binstubs aren't really needed when using `bundle exec`, right? If so, the `check_binstubs` name isn't really accurate. We only care that webpacker the gem is installed.

gauravtiwari: Yeah, I guess the whole`./bin` thing is now irrelevant. Will clean this up in another PR

PR #1025, opened 2017-11-18
Deprecate as of 3.0.3 since it does exist in 3.0.2

stevenharman:
`ConfigList#set` was deprecated _after_ the 3.0.2 release, making the current deprecation message a bit misleading. Or perhaps it's confusing as "deprecated since" can be interpreted two ways?

Maybe a better solution would be to re-word the message?

> deprecated after the 3.0.2 release and will be removed in the next major release

Is that better?

### DIFF
--- a/package/config_types/config_list.js
+++ b/package/config_types/config_list.js
@@ -9,7 +9,7 @@ class ConfigList extends Array {
   }
 
   /**
-    * @deprecated Since 3.0.2 and will be removed in next major release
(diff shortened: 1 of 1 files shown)


gauravtiwari:
@stevenharman Thanks, `after` sounds better. 

> deprecated after the 3.0.2 release and will be removed in the next major release

stevenharman:
OK, just pushed an amended commit with the _"after"_ version.

gauravtiwari:
Great thanks 👍  🍰

Issue #801, opened 2017-09-13
Webpacker uses ASSET_HOST even when config.action_controller.asset_host isn't set


jcbpl:
**Expected result:** Webpacker doesn't prefix asset URLs (via file-loader) when config.action_controller.asset_host isn't set

**Actual result:** If the compilation environment includes an ASSET_HOST variable, Webpacker prefixes asset URLs even when config.action_controller.asset_host isn't set

In other words, even if you haven't configured Rails to use an asset host, Webpacker behaves as if you had if an ASSET_HOST variable exists.

**Relevant files:**

https://github.com/rails/webpacker/blob/master/lib/webpacker/compiler.rb#L76
https://github.com/rails/webpacker/blob/master/package/asset_host.js#L20

**Possible fixes:** 

* Prefix the env var that webpacker sets & reads to be more specific (`WEBPACKER_ASSET_HOST`)?

* Use something other than an env var to provide `ActionController::Base.helpers.compute_asset_host` to the compilation env

PR #1027, opened 2017-11-18
Use WEBPACKER_ASSET_HOST to avoid conflict

gauravtiwari:
Fixes: https://github.com/rails/webpacker/issues/801


gauravtiwari:
Nope, just this at the moment so, we are good

Issue #176, opened 2017-03-20
Making Webpacker to run on Cloud 9


BigChief45:
As previously discussed in PR #107, I am having trouble making webpack work with Rails in my development environment, which is [Cloud 9](https://c9.io).

The problem seems to be that Cloud 9 only allows the `0.0.0.0` interface and a few ports (8080, 8081, etc.). 

Basically, I am experiencing the same problem as in [this issue](https://github.com/webpack/webpack-dev-server/issues/230). It seems fixed they problem by adding `webpack-dev-server/client?https://0.0.0.0:8080` as an entry point to `webpack.config.js`. However, in the setup with Rails I am not sure where to go about in adding this modification, since I do not see any `webpack.config.js` file. I am assuming that `config/webpack/development.js` would be represent something similar though.

gauravtiwari:
@BigChief45 I think this PR will fix the problem - https://github.com/rails/webpacker/pull/153
Provides some configuration options around setting up host and port. If you want to add another entry you can do that by creating an entry option in development.server.js - https://github.com/gauravtiwari/webpacker/blob/d73ca357663c3007544d70a31e1de4de25d05fc5/lib/install/config/webpack/development.server.js

ytbryan:
@BigChief45  are you able to test Gaurav's PR on cloud9? 

You will need to run it on a fresh app with `gem 'webpacker', github: 'gauravtiwari/webpacker', branch: 'feature/static-assets' because this PR has some structural changes. 

Follow the instruction here #153

After you are done, you can change the host and port at `config/webpack/development.server.yml`

gauravtiwari:
The PR is now merged so, you can try this out using master - `gem 'webpacker', github: 'rails/webpacker'`. You probably have to re-run installers as a lot has changed. 

`bundle exec rails webpacker:install` and if you are using react or vue or angular
`bundle exec rails webpacker:install:[name]`

BigChief45:
Sorry guys, was a bit busy these past days.

I gave it a try with the new changes. I think I have the right setup now but I think C9 is not allowing the requests from Rails to Webpack if they are not through HTTPS.

![webpacker_c9](https://cloud.githubusercontent.com/assets/6859247/24540113/5454aad4-1624-11e7-9575-ff5c8bad4827.JPG)

gauravtiwari:
You could set `https: true` in devserver config? https://webpack.js.org/configuration/dev-server/#devserver-https

BigChief45:
[...]

PR #1033, opened 2017-11-23
Add doc article about webpack dev server on Cloud9

rbf:
As suggested by @gauravtiwari in https://github.com/rails/webpacker/issues/176#issuecomment-345532309.

I've adapted and reduced the text of the original article to make it more appropriate for this documentation.

### DIFF
--- a/docs/cloud9.md
+++ b/docs/cloud9.md
@@ -61,28 +61,8 @@ fair, the [changelog of v3.0.2] properly mentions the change:
 ## Quick solution
 
 If you are working alone, the easiest way to fix the configuration of
-the [`webpack-dev-server`] is to add an entry to the `/etc/hosts` and
-to modify the `development.dev_server` entry of the
-`config/webpacker.yml` file.
-
-### `/etc/hosts` entry
-
-As mentioned in some of the [sources](#sources) below, you have to add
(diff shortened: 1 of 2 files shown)


docs/cloud9.md
```diff
@@ -0,0 +1,327 @@
...
+`config/webpacker.yml` file.
+
+### `/etc/hosts` entry
+
+As mentioned in some of the [sources](#sources) below, you have to add
+an entry to [`/etc/hosts`](https://askubuntu.com/a/183187) to alias
+the hostname of your [Cloud9] workspace to `0.0.0.0`:
+
+```bash
+echo "0.0.0.0 ${C9_HOSTNAME}" | sudo tee -a /etc/hosts
```
ssemakov: Hi, thanks for putting this article together!

I think this step is not necessary if you keep `host: localhost` in `webpacker.yml` (line 111 in this PR). Note that setting `public:` (line 113) to a value of `C9_HOSTNAME` is necessary.

rbf: Thanks for pointing this out, @ssemakov! I've tested it in two of my Cloud9 workspaces and it works as you describe. I've updated the doc and the PR to reflect your suggestion. Cheers.

PR #1053, opened 2017-11-29
add staging environment configs

tongueroo:
add staging environment configs


rafaelfranca:
Thank you so much for the pull request but staging is not a default environment in Rails applications.

guilleiguaran:
In Webpacker we follow the same reasoning as Rails main project for environments: https://github.com/rails/rails/issues/20576

tongueroo:
Oh wow, did not realize that. Thanks for letting me know!  Thanks for your excellent work on webpacker.

Issue #1042, opened 2017-11-27
How to profile this


manveru:
Is there any way to use `--progress --profile` with this gem? Our production build takes about 10 minutes and I'd like to know what it spends its time on.

renchap:
Have you tried `RAILS_ENV=production ./bin/webpacker --progress --profile`?

manveru:
Does that differ from what `rake assets:precompile` does?

manveru:
So after trying it, seems to work. But I'm not sure why our build times increased so much:

```
52367ms building modules                                                                  
281ms sealing
31ms optimizing
0ms basic module optimization 
18ms module optimization
20ms advanced module optimization
0ms basic chunk optimization 
0ms chunk optimization 
2ms advanced chunk optimization 
12639ms building modules                                                                 
76ms module and chunk tree optimization
0ms chunk modules optimization 
0ms advanced chunk modules optimization 
32ms module reviving
1ms module order optimization 
26ms module id optimization
49ms chunk reviving
4ms chunk order optimization 
53ms chunk id optimization
268ms hashing
1ms module assets processing 
1143ms chunk assets processing
8ms additional chunk assets processing 
19ms recording
170867ms additional asset processing
81324ms chunk asset optimization
2582ms asset optimization
787ms emitting
Hash: e4608e001b671df6e407
Version: webpack 3.8.1
Time: 322663ms
```

Major offenders seem to be the SCSS files, but unless they are compiled multiple times, it shouldn't cause so much increase.

```
[4] ./node_modules/css-loader?{"minimize":true}!./node_modules/postcss-loader/lib?{"sourceMap":true,"config":{"path":"/mnt/big/github/referral-manager/eqipia/.postcssrc.yml"}}!./node_modules/resolve-url-loader!./node_modules/sass-loader/lib/loader.js?{"sourceMap":true}!./frontend/app/admin/css/application.scss 389 kB {0} [built]
            factory:139ms building:12433ms = 12572ms
```

I don't actually need expensive source maps for production, and that setting got lost in our porting, so I'll bench again after fixing them and report back.

renchap:
I am also seeing high compile times (but not as high as yours), even in dev with HMR, and I also suspect the SCSS pipeline. I havent got time to investigate yet but I am interested in your findings.
It seems to have appeared some time ago but I dont know if its caused by a Webpacker config change or a dependancy update.

gauravtiwari:
[...]

Issue #932, opened 2017-10-11
url() Path Resolution in SASS/SCSS/CSS Stylesheets Exhaustively Searches Entire Project Root


mvastola:
So this is potentially actually a security concern and I actually first brought it up to the Rails Security Team, but it was decided to move this over to a GitHub issue as the impact is fairly low and it's unclear how best to address this issue.

### The Issue

In short, if you were to create a new rails app with `rails new . --webpack` and create `app/javascript/application.scss` with:

```css
body {
  background: url('images/pretty-picture.jpg');
}
```

..and replaced `app/javascript/packs/application.js` so it contained only:

```javascript
import '../application.scss'
```

..and placed an image named `pretty-picture.jpg` at say `private/images/pretty-picture.jpg` (relative to the project root), and then ran `rake webpacker:compile`, `webpack(er)` would grab the image from what should ostensibly be a non-public directory and place it in `public/packs`.

### The 'Why'

This has obvious implications in that normally inaccessible content (or content ordinarily behind by an access control mechanism within the app -- such as attachments to models) could be accidentally made public if there should be a file naming conflict.

Additionally, this is also not expected behavior as:
- only the directories specified by `webpacker.yml`'s `source_path` and `resolved_paths` directives should be searched for files
- the css `url()` directive (in the absence of an absolute uri) is supposed to be taken as relative to either the scss/sass file itself, or the target css file.

### The Cause

This concerning path resolution is actually due to the merging of PR #387 (resolving issue #384), which added the package `resolve-url-loader`, for which this exhaustive path searching is actually intended behavior. (Which is what makes the solution here less than straightforward.)

Per `resolve-url-loader`'s [`README.md`](https://github.com/bholloway/resolve-url-loader#how-it-works):
> Usually the asset is found relative to the original source file `O(1)`. However in some cases there is no immediate match (*cough* bootstrap *cough*) and we so we start searching both deeper and shallower from the starting directory `O(n)`.

### The Solution

There is no option to turn off this 'feature' within `resolve-url-loader`.

The [`README.md` for `sass-loader`](https://github.com/webpack-contrib/sass-loader#problems-with-url) provides additional background on why `resolve-url-loader` is needed.

[...]

Issue #918, opened 2017-10-09
Images not working ?


JerryGreen:
**Main problem:**

My images (used in css or imported from js too) are not showing.

**My little research:**

Here's my manifest.json

<pre>
{
  "_/_/frontend/src/components/shared/act-card-template/loading.svg": "/packs/_/_/frontend/src/components/shared/act-card-template/loading-6e55ac3085d4cd8651da32d87be226d1.svg",
  "_/_/frontend/src/components/shared/act-card-template/successBlack.svg": "/packs/_/_/frontend/src/components/shared/act-card-template/successBlack-de0a3c628cd7278df54be55bddac8fe2.svg",
  "_/_/frontend/src/components/shared/act-card-template/successGreen.svg": "/packs/_/_/frontend/src/components/shared/act-card-template/successGreen-3c4c4898f2f924212a0a11f69cd7600e.svg",
  "_/_/frontend/src/components/shared/act-card-template/successWhite.svg": "/packs/_/_/frontend/src/components/shared/act-card-template/successWhite-974fc1c1cbe6751ee469923a7ad5514d.svg",
  "_/_/frontend/src/components/shared/act-card/act-content/save-button/starBlack.svg": "/packs/_/_/frontend/src/components/shared/act-card/act-content/save-button/starBlack-b7b4f99072162243573dc36373a40cde.svg",
  "_/_/frontend/src/components/shared/act-card/act-content/save-button/starDark.svg": "/packs/_/_/frontend/src/components/shared/act-card/act-content/save-button/starDark-4173cc2a32cbae7f7c1a607564cc2a4c.svg",
  "_/_/frontend/src/components/shared/act-card/act-content/save-button/starGray.svg": "/packs/_/_/frontend/src/components/shared/act-card/act-content/save-button/starGray-7ef082685573660caa281c9a88782263.svg",
  "_/_/frontend/src/components/shared/act-card/act-content/update-profile/info.svg": "/packs/_/_/frontend/src/components/shared/act-card/act-content/update-profile/info-ca4b7a13cf8b16e9db2d1327cd62b574.svg",
  "_/_/frontend/src/components/shared/act-card/act-content/update-profile/user.svg": "/packs/_/_/frontend/src/components/shared/act-card/act-content/update-profile/user-8fc21994702eeed04eeb0e58f7cf4ff7.svg",
[...]

PR #1058, opened 2017-12-01
Tidy up webpack config

gauravtiwari:
This PR fixes a couple of things related to webpack config: 

- consistent structure for loaders
- deterministic order for loaders, it will start matching modules from top and then first loader that matches the test will be used. This opens possibility for adding different loaders for say processing a svg image instead of using default file loader. 
- add url loader to process small static assets so it can be embedded directly as data uri instead of a file. https://github.com/webpack-contrib/url-loader
- add case sensitive plugin to enforce casing on different platforms: https://github.com/Urthen/case-sensitive-paths-webpack-plugin
- remove resolve-url-loader see #1042 After testing with standard libraries like bootstrap it seems this loader is redundant and adds more complexity. We also had some security concerns earlier: https://github.com/rails/webpacker/issues/932


gauravtiwari:
The build time is slightly faster too with standard hello examples: 

screen shot 2017-12-01 at 18 17 33

vs

screen shot 2017-12-01 at 18 17 06

lib/install/config/webpacker.yml
```diff
@@ -43,8 +43,15 @@ development:
...
     overlay: true
+    compress: true
     disable_host_check: true
     use_local_ip: false
+    quiet: false
+    headers:
+      'Access-Control-Allow-Origin': '*'
+    watch_options:
+      ignored: /node_modules/
+
```
javan: `webpacker.yml` is intended for config that needs to be read by both Ruby and JavaScript (webpack). Seems fine to add additional dev server options here, but we should be careful of bloat.

package/environment.js
```diff
@@ -76,6 +79,14 @@ const getBaseConfig = () =>
...
     resolveLoader: {
       modules: ['node_modules']
+    },
+
+    node: {
+      dgram: 'empty',
+      fs: 'empty',
+      net: 'empty',
+      tls: 'empty',
+      child_process: 'empty'
```
javan: What is this for?

package/environment.js
```diff
@@ -93,7 +104,10 @@ module.exports = class Environment {
       entry: this.entry.toObject(),
 
       module: {
-        rules: this.loaders.values()
+        strictExportPresence: true,
+        rules: [
+          { oneOf: this.loaders.values() }
+        ]
```
javan: What do these options change?

package/environments/production.js
```diff
@@ -10,11 +10,16 @@ module.exports = class extends Environment {
...
+      },
       compress: {
-        warnings: false
+        warnings: false,
+        comparisons: false
       },
       output: {
-        comments: false
+        comments: false,
+        ascii_only: true
```
javan: What do these changes do?

package/rules/file.js
```diff
@@ -3,7 +3,17 @@ const { source_path } = require('../config')
...
+    /\.(js|jsx)?(\.erb)?$/,
+    /\.coffee(\.erb)?$/,
+    /\.(ts|tsx)?(\.erb)?$/,
+    /\.vue(\.erb)?$/,
+    /\.elm(\.erb)?$/,
+    /\.(scss|sass)$/i,
+    /\.(css)$/i,
+    /\.html$/,
+    /\.json$/
+  ],
```
javan: Looks like this could be combined to a single regex.

javan:
> The build time is slightly faster too

Where does the speed improvement come from?

lib/install/config/webpacker.yml
gauravtiwari: Yeah, just added the important ones that folks would like to change 👍

package/environment.js
gauravtiwari: So, some libraries  that import nodejs (like `fs`) modules but don't use them in the browser. Here, we are telling webpack to provide empty mocks for them so importing works.

package/environment.js
gauravtiwari: Previously, we were adding loaders directly to rules array but that had some weird limitations like say, if I want to use some other loader for processing a svg file and we also by default ship the file loader, which also processes svg files. There was no away to ensure the new svg loader is used for svg files, and not file loader. With this change we are saying, use the first loader that matches the criteria (i.e matches test regex or exclude regex). 

Also, the order of processing loaders is now from top to bottom i.e. fixed.

package/environments/production.js
gauravtiwari: https://github.com/mishoo/UglifyJS2/tree/harmony#output-options

package/rules/file.js
gauravtiwari: Yep, will do 👍

gauravtiwari:
Most probably from removing resolve-url-loader, which has some magic file lookups (remember that security issue raised earlier #932). See #1042 

The usage of `resolve-url-loader` is for very exceptional cases where a node module is using relative imports in css files and doesn't have that resource directly next to it, which means `url()` lookup won't work. The resolve url loaders, adds some kinda of file searching magic to lookup missing files if they aren't available directly in the path. 

I have tested and most standard css libraries works without it, like - bootstrap and foundation. If someone wants to use, they can update the loader themselves.

gauravtiwari:
Another place, where we might be reaping speed benefits is from using `oneOf`, since it doesn't go through all the loaders and only applies first matching loader: https://webpack.js.org/configuration/module/#rule-oneof

I will probably benchmark this and post some results here.

package/environment.js
gauravtiwari: https://webpack.js.org/configuration/node/#other-node-core-libraries

package/environment.js
gauravtiwari: https://webpack.js.org/configuration/module/#rule-oneof

gauravtiwari:
So, just tested the `oneOf` change doesn't really help with performance improvement. 

Webpack: 3.5.5

### Before 

screen shot 2017-12-02 at 10 42 08

screen shot 2017-12-02 at 10 42 13

screen shot 2017-12-02 at 10 42 18

### After

screen shot 2017-12-02 at 10 42 29

screen shot 2017-12-02 at 10 42 35

screen shot 2017-12-02 at 10 42 41

package/environment.js
javan: > **`Rule.oneOf`**
> An array of Rules from which only the first matching Rule is used when the Rule matches.

What if people are relying on files being processed by multiple loaders?

package/environments/production.js
javan: I mean, why are you adding/changing these options? Do they resolve any known issues?

package/environment.js
```diff
@@ -25,6 +26,8 @@ const getLoaderList = () => {
 const getPluginList = () => {
   const result = new ConfigList()
   result.append('Environment', new webpack.EnvironmentPlugin(JSON.parse(JSON.stringify(process.env))))
+  result.append('NamedModules', new webpack.NamedModulesPlugin())
```
javan: > This plugin will cause the relative path of the module to be displayed when HMR is enabled. Suggested for use in development. (https://webpack.js.org/plugins/named-modules-plugin/)

Should we be using this in all environments?

package/environment.js
gauravtiwari: Then whenever compiler encounters different format it will traverse again:  

**For example:**  *Started with a `.js` file -> js file has a css import -> look for a loader for css -> then css file has a image linked -> look for a for .png file -> so on....*

package/environments/production.js
gauravtiwari: No, we didn't received any but just standardising the options for now (recommended ones).

package/environment.js
gauravtiwari: Good spot 👁 , will fix 👍

package/environment.js
javan: What if you have `.js` files (or any type) that you want processed by multiple loaders? I assume that's a thing.

package/environments/production.js
javan: Recommended where?

package/environment.js
gauravtiwari: Ahh right, then that loader will offload to another loader, for example this is a rule for CS2: `use: `['babel-loader', 'coffee-loader']`

package/environments/production.js
gauravtiwari: Just noticed a lot of bugs here related to emojis and regex and it seemed like a good option to be on since emojis are everywhere :): https://github.com/mishoo/UglifyJS2/search?p=1&q=ascii_only&type=Issues&utf8=%E2%9C%93 

They haven't turned this on by default because build css size increases or something.

gauravtiwari: If you are not happy I can turn it off and wait until an issue is reported. Since the standard config is now hidden, thought it would make sense to add options that commonly causes issues.

javan: > Just noticed a lot of bugs here related to emojis and regex and it seemed like a good option to be on since emojis are everywhere :)

👍

> They haven't turned this on by default because build css size increases

CSS size❓

> If you are not happy I can turn it off and wait until an issue is reported

I'm ok with it as long as it doesn't bloat *JS* file sizes significantly

gauravtiwari: Just tested, turning it on doesn't increase the build size (all hello examples).

gauravtiwari: Oops typo there :) (ignore css)

package/environment.js
gauravtiwari: Each rule defines it's loaders so, this not totally depend on file extension `(.js or .ts)` but underlying loader definition `use`. Some loaders are more implicit like, `vue-loader` it doesn't need that extra `babel-loader` there (figures out itself)

javan: If you feel confident about this change then 👍. I haven't seen it used in the wild and I wouldn't be surprised if it breaks something for someone.

gauravtiwari: Yeah, I had some doubts too but after testing this extensively with `hello` examples we ship and the new config seems to work fine 😄  🤞 

Btw, this issue lead this change: https://github.com/rails/webpacker/issues/918#issuecomment-345074512

Previously, it seemed that webpack was using all loaders regardless of order i.e. if you have two loader that has same test extension it won't use the first one but both, atleast that was the case with svg's. This fixes it by enforcing order and only the first matching loader will be used from the array.

package/rules/url.js
```diff
@@ -0,0 +1,13 @@
+const assetHost = require('../asset_host')
+
+module.exports = {
```
farthinker: @gauravtiwari How can I reference small static images in Rails view template after using this url loader?

gauravtiwari: All images under 10KB would use url loader by default, this includes png, jpg, gif and bmp

farthinker: Then the url loader will break images under 10KB in projects with server rendering

package/environment.js
farthinker: I have an issue about test coverage config after this change. The rule condition for `istanbul-instrumenter-loader` is different from the rule condition for 'coffee-loader'. The `istanbul-instrumenter-loader` should exclude test files, but the 'coffee-loader' should include the test file. And with `oneOf` option, the two loaders  have to be in the same 'coffee' rule.

charlesvallieres: Hi, this change breaks all combinations like `.coffee.erb` the coffee-loader doesn't use the erb-loader and the erb-loader doesn't use the coffee loader.

As for :
> Previously, we were adding loaders directly to rules array but that had some weird limitations like say, if I want to use some other loader for processing a svg file and we also by default ship the file loader, which also processes svg files.

I guess the workaround for that would be to overwrite the file-loader rule like we do for SASS, something like :

```
const SASSRule = environment.loaders.get("sass");
const SASSLoader = SASSRule.use.find(el => el.loader === "sass-loader");

// Add resolve-url-loader after SASS
SASSRule.use.splice(SASSRule.use.indexOf(SASSLoader), 0, {
  loader: "resolve-url-loader"
});
```

---

This is how I fixed it if anybody runs into the same problem :
```
Environment.prototype._toWebpackConfig = Environment.prototype.toWebpackConfig;

Environment.prototype.toWebpackConfig = function() {
  const webpackConfig = this._toWebpackConfig();

  if (webpackConfig.module.rules[0].oneOf) {
    webpackConfig.module.rules = webpackConfig.module.rules[0].oneOf;
  }

  return webpackConfig;
}

module.exports = environment;
```

guilleiguaran: > Hi, this change (oneOf) breaks all combinations like .coffee.erb the coffee-loader doesn't use the erb-loader and the erb-loader doesn't use the coffee-loader.

Can you report a new issue with this?

charlesvallieres: Oh didn't really look into issues I was looking at the code to figure out why my config was breaking. Thanks @guilleiguaran!

gauravtiwari: Sorry about this folks. There is a PR that should fix all these edge cases. Will merge and make a release as soon as it’s merged and issues are addressed.

package/environment.js
Pezmc: @gauravtiwari Is this still needed? Neither [the docs](https://webpack.js.org/configuration/node/#other-node-core-libraries) or [NodeStuffPlugin](https://github.com/webpack/webpack/blob/master/lib/NodeStuffPlugin.js) seem to use it!

Issue #1061, opened 2017-12-04
asset_pack_path and aliases


timonbandit:
Hey guys!
I have a problem.
We trying to migrate from sprockets to webpacker and it's a bit tricky in our case.
We have a big application and 3 different themes.
Our folder structure looks like
```
app/themes/theme1
    /assets
        /images/theme1
        /javascript/theme1
        /stylesheets/theme1        
app/themes/theme2
    /assets
        /images/theme2
        /javascript/theme2
        /stylesheets/theme2        
app/themes/theme3
    /assets
        /images/theme3
        /javascript/theme3
        /stylesheets/theme3        

```
Sometimes I need to get my images in views. in sprockets, it was like
 `<%= asset_path('theme1/myimage.png') %>`
But now I have to write:
`<%=asset_pack_path("_/themes/theme1/assets/images/theme1/myimage.png") %>`
It looks terrible :(

I added 
`resolved_paths: ['app/assets/', 'app/themes/theme1/assets/', 'app/themes/theme2/assets/', 'app/themes/theme3/assets/']` to webpacker.yml 
and i can import images more/less comfortable `import 'images/theme1/myimage.png';` in my js files.

Is it possible to use aliases in asset_pack_path helper?

gauravtiwari:
@timonbandit `asset_pack_path` is pretty limited and only works if a pack i.e. an image is under packs or imported in one of the packs. Perhaps, consider using a hybrid approach - sprockets for image linking and webpacker for rest of the stuff.

Obviously, you can alias on webpacker side for easy import - https://github.com/rails/webpacker/blob/master/docs/assets.md#using-babel-module-resolver. 

Would that work?

timonbandit:
Thank you @gauravtiwari !
But there is another one problem:
```asset_path``` not working now ;(
```SyntaxError: Unexpected token: name ($)``` 
in my .erb
```background: #000 url(<%=asset_path('theme1/logo.png') %>) no-repeat;```

**Comment by gauravtiwari** [comment]
@timonbandit Is this under webpacker assets? or Rails views?

```yml
app/assets: 
  images: 
    # shared images here ie. used in views

app/javascript: 
   images: 
      # webpacker specific images here
```

timonbandit:
@gauravtiwari I have no separate webpack assets. Because i'm migrating existing project and don't want to change directory structure.
Cause I have to use images from directories I pointed above

gauravtiwari:
[...]

PR #1067, opened 2017-12-06
Added info about necessity of commenting uglifier

timonbandit:
Hey @gauravtiwari 
PR relating to #1061


guilleiguaran:
I think this should be mentioned in the [troubleshooting docs](https://github.com/rails/webpacker/blob/master/docs/troubleshooting.md) and that this is a problem only when you want to share files between Sprockets and Webpacker so it isn't always the case.

timonbandit:
@guilleiguaran Thank you.
You're right, i will make the new one commit

timonbandit:
@guilleiguaran - updated!

gauravtiwari:
@timonbandit Pushed a small update. Does that look alright to you?

timonbandit:
@gauravtiwari Yes. looks good! ;)

gauravtiwari:
great, thanks 🍰

Issue #1097, opened 2017-12-13
TypeError: Class constructor ConfigList cannot be invoked without 'new'


rossta:
System:

• `@rails/webpacker` version: `3.1.1`
• node version: `v6.11.5`
• lodash version: `4.14.2`

Steps to reproduce:

```javascript
const { cloneDeep } = require('lodash');
const { environment } = require('@rails/webpacker');

cloneDeep(environment.toWebpackConfig(), {});
```

Sample output:

```txt
/dev/node_modules/lodash/lodash.js:5763
          result = array.constructor(length);
                         ^

TypeError: Class constructor ConfigList cannot be invoked without 'new'
    at initCloneArray (/dev/node_modules/lodash/lodash.js:5763:26)
    at baseClone (/dev/node_modules/lodash/lodash.js:2396:18)
    at /dev/node_modules/lodash/lodash.js:2439:34
    at arrayEach (/dev/node_modules/lodash/lodash.js:485:11)
    at baseClone (/dev/node_modules/lodash/lodash.js:2433:7)
    at /dev/node_modules/lodash/lodash.js:2439:34
    at arrayEach (/dev/node_modules/lodash/lodash.js:485:11)
    at baseClone (/dev/node_modules/lodash/lodash.js:2433:7)
    at cloneDeep (/dev/node_modules/lodash/lodash.js:10650:14)
    at Object.<anonymous> (/dev/tmp/webpacker_clonedeep.js:4:18)
```

Attempting a deep clone of the object returned by `environment.toWebpackConfig()` results in a `TypeError`. I came across this error by using `webpack-merge`, which delegates to lodash `cloneDeep`, to update the Webpack config. 

Perhaps it makes sense to ensure that once `toWebpackConfig` is called, instances of Webpacker base type subclasses, e.g., `ConfigList`, `ConfigObject`, are converted back to their base types to avoid surprising results.

guilleiguaran:
Thanks for opening this!!

If you have an idea about how to fix this we would love to review a PR ❤️

rossta:
@guilleiguaran Looking into it

gauravtiwari:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/@@species

gauravtiwari:
We just need to return native types back, didn't realise it's going to be used outside of webpacker though :)

gauravtiwari:
@rossta Would be interesting to know your use case with clonedeep

[...]

PR #1099, opened 2017-12-13
Adds tests for environment.toWebpackConfig

rossta:
As a follow up to #1097 and #1098, I've taken a stab at adding tests for `environment.toWebpackConfig`. 

Because the `environment` module assumes the location of `config/webpacker.yml`, adding tests for this module presents a challenge since this location does not exists relative to the root of the project. Possible approaches include simply adding `config/webpacker.yml` to the root of the project (and potentially other app files, like `app/javascript/packs`), generating files during the test run, or using a mock filesystem. 

The approach I landed on makes it possible to customize the `RAILS_ROOT` location so that the environment paths resolve relative to `test/test_app` in tests. While approach adds flexibility, it also means ensuring that relative paths, e.g. `node_modules`, resolve relative to the given root if it exists.

Let me know what you think of this approach.


package/root.js
```diff
@@ -0,0 +1,11 @@
+const { resolve, join } = require('path')
+const root = process.env.RAILS_ROOT || '';
```
gauravtiwari: What do you think about using `process.cwd()`?

package/environment.js
```diff
@@ -77,7 +78,7 @@ const getBaseConfig = () =>
     },
 
     resolveLoader: {
-      modules: ['node_modules']
+      modules: [join(root, 'node_modules')]
```
gauravtiwari: Don't think this is necessary??

rossta:
Turns out we can change the working directory of the test process, taking advantage of how we execute requires of the default and project webpacker.yml files, without changing application code. 

I'm not sure how useful the original approach of introducing a configurable Rails root would be. In any case, it should probably be a separate PR rather than a side effect of trying to test existing app code.

package/utils/__tests__/objectify.js
```diff
@@ -1,4 +1,5 @@
 /* global test expect */
+console.log('objectify test cwd', process.cwd());
```
gauravtiwari: Is this intentional?

gauravtiwari:
Awesome @rossta 💯  🍰  

Just one minor comment and then we can merge. Thank you :)

package/utils/__tests__/objectify.js
rossta: No, that was a debugging line that should be removed (and looks like you've done that on master already).

rossta:
@gauravtiwari 👍

Issue #1071, opened 2017-12-07
Consider adding asset_pack_url?


Spone:
It would be useful to have access to the full URL to an image using `asset_pack_url`, the same way I can access its path with `asset_pack_path`

This would follow Rails convention for routes and Sprockets helpers.

maxence33:
Hi guys, 
I am having the following error while using the URL and Path helpers.

ActionView::Template::Error (undefined method `image_pack_url' for #<#<Class:0x0000559e269a5af0>:0x0000559e269930a8>
Did you mean?  image_pack_tag):

But `image_pack_tag` works flawlessly. 

I made sure I have `config.action_controller.asset_host = "http://localhost:3000"` in my dev env.
Is there something I forgot about..

leoplct:
This works for me `asset_pack_url("media/images/social/fb.png")`

PR #1102, opened 2017-12-14
Add asset_pack_url helper

dwightwatson:
As referenced in #1071, this adds a helper method for `asset_pack_url` which returns an absolute URL to an asset, much like `asset_url` would.


gauravtiwari:
thanks @dwightwatson 🍰  👍

Issue #1081, opened 2017-12-11
Module build failed: SyntaxError: Unexpected token for ERB with webpacker 3.1.1


doits:
I just upgraded to webpacker 3.1.1 and cannot compile `.js.erb` files anymore.

If have a file `./app/javascript/packs/components/event_fields.js.erb` with erb out (`<%= ... %>`) and it tells me:

```
ERROR in ./app/javascript/packs/components/event_fields.js.erb
Module build failed: SyntaxError: Unexpected token (47:22)

  45 |     },
  46 |     possible_appointments(ea) {
> 47 |       let intervals = <%= EventAppointment::INTERVALS.to_json %>
     |                       ^
  48 |
  49 |       if(!ea.end_at || !ea.start_at) { return [] }
```

Has been working fine on 3.0.2.

I followed the upgrade guide and basically freshly installed webpacker (my `webpacker.yml` has all options default).

gauravtiwari:
Thanks for report, will take a look 👍

gauravtiwari:
As a temporary fix, you can add following to your environment.js: 

```js
const { environment } = require('@rails/webpacker')
const erb = environment.loaders.get('erb')

// Re-insert back again but before babel loader
environment.loaders.insert('erb', erb, { before: 'babel' })
```

doits:
Thanks, works for my `.js.erb` now!

I think the same happens for other loaders, too? Now my `.js.erb` compiles, but my `.vue.erb` is not recognized by vue:

```
ERROR in ./app/javascript/packs/components/event_custom_question_answers_form.vue.erb
Module parse failed: Unexpected token (1:0)
You may need an appropriate loader to handle this file type.
| <template>
```

gauravtiwari:
Yeah, will push a fix later in the evening. To fix all loaders using erb:

```js
environment.config.merge({
  module: {
    rules: [
      environment.loaders.get('erb')
    ]
  }
})

environment.loaders.delete('erb')
```

gauravtiwari:
Sorry for trouble folks, will fix new bugs and make a new release soon.

PR #1103, opened 2017-12-14
Move out integration loaders

gauravtiwari:
This PR moves integration loaders like react out of the core package and instead ships through installer task. Per webpack docs (https://webpack.js.org/guides/build-performance/#bootstrap), it makes sense to only ship minimal amount of loaders and plugins necessary to get started and if someone needs to add react or coffeescript support, they are free to add it via installer. 

Plus, we also loose the liability of maintaining these loaders, for example: Coffeescript now has two versions so, offloading this loader to app makes it possible for app owner to use any version and tweak loader as fit. 

- Adds coffee and erb installer
- Removes integration loader and moves it to installer

Fixes https://github.com/rails/webpacker/issues/1081


README.md
```diff
@@ -323,6 +325,25 @@ The Elm library and core packages will be added via Yarn and Elm itself.
 An example `Main.elm` app will also be added to your project in `app/javascript`
 so that you can experiment with Elm right away.
 
+### Coffeescript
+
+To add coffeescript[Coffeescript](http://coffeescript.org/) support,
```
muryoimpl: coffeescript before [Coffeescript] is unnecessary and CoffeeScript is right, I think.

Issue #954, opened 2017-10-16
rake webpacker:check_node errors out if no node installed


tdegrunt:
Version: webpacker 3.0.2
Ruby: 2.4.2

```
rake webpacker:check_node --trace                                                                                                     ⏎
** Invoke webpacker:check_node (first_time)
** Execute webpacker:check_node
/Users/tdegrunt/.rbenv/versions/2.4.2/bin/rake: No such file or directory - node
rake aborted!
NoMethodError: undefined method `strip' for nil:NilClass
/Users/tdegrunt/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/webpacker-3.0.2/lib/tasks/webpacker/check_node.rake:16:in `block (2 levels) in <top (required)>'
/Users/tdegrunt/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/rake-12.1.0/lib/rake/task.rb:251:in `block in execute'
/Users/tdegrunt/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/rake-12.1.0/lib/rake/task.rb:251:in `each'
/Users/tdegrunt/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/rake-12.1.0/lib/rake/task.rb:251:in `execute'
/Users/tdegrunt/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/rake-12.1.0/lib/rake/task.rb:195:in `block in invoke_with_call_chain'
/Users/tdegrunt/.rbenv/versions/2.4.2/lib/ruby/2.4.0/monitor.rb:214:in `mon_synchronize'
/Users/tdegrunt/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/rake-12.1.0/lib/rake/task.rb:188:in `invoke_with_call_chain'
/Users/tdegrunt/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/rake-12.1.0/lib/rake/task.rb:181:in `invoke'
/Users/tdegrunt/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/rake-12.1.0/lib/rake/application.rb:153:in `invoke_task'
/Users/tdegrunt/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/rake-12.1.0/lib/rake/application.rb:109:in `block (2 levels) in top_level'
[...]

PR #1105, opened 2017-12-14
Gracefully handle missing node

brasic:
The `webpacker:check_node` task blows up when node is not installed.  The code that rescues `Errno::ENOENT` doesn't behave as expected when rails is loaded because [ActiveSupport monkeypatches Kernel#`]( https://github.com/rails/rails/blob/7d75599c875b081f63fc292e454b25e8e42eb60e/activesupport/lib/active_support/core_ext/kernel/agnostics.rb) to suppress raising Errno::ENOENT.  With this monkeypatch in place the code never reaches the fallback branch for legacy node versions introduced in https://github.com/rails/webpacker/pull/798.

We can fix this by manually raising the exception that the activesupport monkeypatch has suppressed.

Before:
```
$ bundle exec rake webpacker:check_node
[...]/gems/ruby-2.4.2/bin/rake: No such file or directory - node
rake aborted!
NoMethodError: undefined method `strip' for nil:NilClass
[...]/gems/webpacker-3.1.1/lib/tasks/webpacker/check_node.rake:16:in `block (2 levels) in <main>'
```

After:
```
$ bundle exec rake webpacker:check_node
[...]/gems/ruby-2.4.2/bin/rake: No such file or directory - node
[...]/gems/ruby-2.4.2/bin/rake: No such file or directory - nodejs
Node.js not installed. Please download and install Node.js https://nodejs.org/en/download/
```

Fixes https://github.com/rails/webpacker/issues/954

### DIFF
--- a/lib/tasks/webpacker/check_node.rake
+++ b/lib/tasks/webpacker/check_node.rake
@@ -4,6 +4,7 @@ namespace :webpacker do
     begin
       begin
         node_version = `node -v`
+        raise Errno::ENOENT if node_version.blank?
       rescue Errno::ENOENT
         node_version = `nodejs -v`
         raise Errno::ENOENT if node_version.blank?
--- a/lib/tasks/webpacker/check_node.rake
+++ b/lib/tasks/webpacker/check_node.rake
@@ -2,13 +2,8 @@ namespace :webpacker do
   desc "Verifies if Node.js is installed"
   task :check_node do
     begin
-      begin
-        node_version = `node -v`
-        raise Errno::ENOENT if node_version.blank?
-      rescue Errno::ENOENT
-        node_version = `nodejs -v`
-        raise Errno::ENOENT if node_version.blank?
-      end
+      node_version = `node -v || nodejs -v`
+      raise Errno::ENOENT if node_version.blank?
 
       pkg_path = Pathname.new("#{__dir__}/../../../package.json").realpath
       node_requirement = JSON.parse(pkg_path.read)["engines"]["node"]


lib/tasks/webpacker/check_node.rake
```diff
@@ -4,6 +4,7 @@ namespace :webpacker do
     begin
       begin
         node_version = `node -v`
+        raise Errno::ENOENT if node_version.blank?
       rescue Errno::ENOENT
         node_version = `nodejs -v`
         raise Errno::ENOENT if node_version.blank?
```
gauravtiwari: How about if we re-raise it outside of rescue then I guess we don't need to raise twice?

gauravtiwari:
```rb
   begin
      node_version = `node -v`
   rescue Errno::ENOENT
      node_version = `nodejs -v`
    end

   raise Errno::ENOENT if node_version.blank?
```

brasic: @gauravtiwari that approach would fix the "node not installed" issue but it would still leave the change in #798 nonfunctional since the first block would never raise so the `nodejs` branch would never be tried.

brasic: Part of the issue is that I'm not sure we can depend one way or the other on whether or not the activesupport monkeypatch will be in place.  This approach works whether Kernel#` is monkeypatched or not.

brasic: This simplifies things a lot and works the same whether or not ActiveSupport has monkeypatched Kernel#`: e49d10a0cac36fdb3206802206328418e4fb7f13

gauravtiwari: Awesome 🍰  💯

gauravtiwari:
Thanks @brasic 👍  💯

PR #1123, opened 2017-12-19
Fix checking 'webpack' binstub on Windows

hron:
webpacker:check_binstubs doesn't work on Windows because `Bundler.which("webpack")` returns false on this platform. I'm not sure if it should be fixed on webpacker side, but anyway this error doesn't allow Windows users to run any of webpacker related tasks.


gauravtiwari:
Thanks @hron

PR #1131, opened 2017-12-19
silence yarn output if checking is successfull

doits:
on every `rake` or `rails` call I get this output:

```
yarn check v1.3.2
warning package.json: No license field
warning No license field
success Folder in sync.
✨  Done in 0.18s.
```

why not silence it if everything is in sync and only display it in case of errors?


lib/webpacker/railtie.rb
```diff
@@ -37,6 +37,10 @@ class Webpacker::Engine < ::Rails::Engine
         warn "To disable this check, please add `config.webpacker.check_yarn_integrity = false`"
         warn "to your Rails development config file (config/environments/development.rb)."
         warn "\n\n"
+        warn "Yarn output while checking for integrity:"
```
gauravtiwari: Do you think we need this like?

doits: maybe not, I removed it

PR #1144, opened 2017-12-22
Don't mutate config object

gauravtiwari:


M-Dahab:
why is this not reflecting when I create a new app?

PR #1167, opened 2018-01-10
Updated instructions to set loaders and plugins

iainbeeston:
webpack.md seems to be out of date: it has examples of loaders and plugins being added with `.insert()`, `.append()` and `.prepend()`, whereas in webpacker 3 they should be added with `.set()` instead.

Forgive me if it's not correct, I found this out by trial and error, and by reading some of the other documentation files, when trying to get plugins to work.

### DIFF
--- a/docs/webpack.md
+++ b/docs/webpack.md
@@ -68,17 +68,10 @@ yarn add json-loader
 // config/webpack/environment.js
 const { environment } = require('@rails/webpacker')
 
-environment.loaders.append('json', {
+environment.loaders.set('json', {
   test: /\.json$/,
   use: 'json-loader'
 })
-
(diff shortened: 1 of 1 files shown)


renchap:
This is wrong for Webpacker >= 3.1.
`insert()`, `append()`, … has been introduced in this version, and it is working. You can see the implementation here: https://github.com/rails/webpacker/blob/master/package/config_types/config_list.js

If you are using Webpacker 3.0.2, you can see the docs here: https://github.com/rails/webpacker/blob/v3.0.2/docs/webpack.md (and they cover `set()`).

javan:
@renchap is correct. The docs are accurate for the latest `@rails/webpacker` version.

Issue #1171, opened 2018-01-12
asset_pack_path + HMR


timonbandit:
Hi!
i can't use ```asset_pack_path``` with HMR on my dev machine due of 
```
Webpacker can't find theme1.css in /var/www/app/public/packs/manifest.json.
```
it's works for js, but not for css, which i'm using for this:
```
<script>
  loadCSS("<%= asset_pack_path('theme1.css') %>");
</script>
```
i think problem is in helper.rb
```ruby
  def asset_pack_path(name, **options)
    asset_path(Webpacker.manifest.lookup!(name), **options)
  end
```

stylesheet_pack_tag helper is checking if it on dev server or HMR:
```ruby
def stylesheet_pack_tag(*names, **options)
    unless Webpacker.dev_server.running? && Webpacker.dev_server.hot_module_replacing?
      stylesheet_link_tag(*sources_from_pack_manifest(names, type: :stylesheet), **options)
    end
  end
```
Is it ok to send a PR?

gauravtiwari:
Please do 👍

timonbandit:
@gauravtiwari done :+1:

maykefreitas:
Hey guys, I know that it's merged, but, how I'll use `asset_pack_path` for images in development environment? After upgrading to the version 3.2.1, I can't see my images in development env with dev server and hmr enabled.

gauravtiwari:
Yeah this introduced a bug for other asset types. Will fix during the weekend.

PR #1172, opened 2018-01-12
asset_pack_path + HMR fix; Closes rails/webpacker#1171

timonbandit:
Fixed error with asset_pack_path and asset_pack_url helpers problem with CSS when HMR is enabled.
rails/webpacker#1171

### DIFF
--- a/lib/webpacker/helper.rb
+++ b/lib/webpacker/helper.rb
@@ -5,9 +5,15 @@ module Webpacker::Helper
   #
   # Example:
   #
+  #   # In development mode with hot module replacement:
+  #   <%= asset_pack_path 'calendar.css' %>  # => nil
+  #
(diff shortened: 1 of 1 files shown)


gauravtiwari:
thanks 🍰

javan:
This change is going to impact production apps negatively. Every `asset_pack_{path,url}` will make a network request via https://github.com/rails/webpacker/blob/c46d751b12cc7e24661bc76ad5ebebd0e08c7dda/lib/webpacker/dev_server.rb#L12-L13

javan:
Also, it doesn't feel appropriate to disable `asset_pack_{path,url}` to fix a *CSS* issue. Unless those helpers are only intended for CSS asset paths? If so they should be renamed as such?

PR #1179, opened 2018-01-14
Return false in production environments

### LINKED ISSUE #1173: Asset helpers should not call Webpacker.dev_server.running? in production
`Webpacker.dev_server.running?` opens a socket connection: https://github.com/rails/webpacker/blob/c46d751b12cc7e24661bc76ad5ebebd0e08c7dda/lib/webpacker/dev_server.rb#L12-L17

That's fine for development mode, but no good for production.
https://github.com/rails/webpacker/blob/c46d751b12cc7e24661bc76ad5ebebd0e08c7dda/lib/webpacker/helper.rb#L13-L14
https://github.com/rails/webpacker/blob/c46d751b12cc7e24661bc76ad5ebebd0e08c7dda/lib/webpacker/helper.rb#L30-L31
https://github.com/rails/webpacker/blob/c46d751b12cc7e24661bc76ad5ebebd0e08c7dda/lib/webpacker/helper.rb#L64-L65

Introduced in https://github.com/rails/webpacker/pull/641 and https://github.com/rails/webpacker/pull/1172.

/cc @dhh @gauravtiwari

gauravtiwari:
Fixes #1173 

This PR enables bypassing of dev server check in production environment since the assets are precompiled in production environments.

### DIFF
--- a/lib/webpacker/dev_server.rb
+++ b/lib/webpacker/dev_server.rb
@@ -10,11 +10,11 @@ class Webpacker::DevServer
   end
 
   def running?
-    if env.production?
-      false
-    else
+    if config.dev_server.present?
       Socket.tcp(host, port, connect_timeout: connect_timeout).close
       true
+    else
+      false
     end
   rescue
     false
--- a/lib/webpacker/dev_server.rb
+++ b/lib/webpacker/dev_server.rb
@@ -3,15 +3,19 @@ class Webpacker::DevServer
   # Webpacker.dev_server.connect_timeout = 1
   cattr_accessor(:connect_timeout) { 0.01 }
 
-  delegate :config, to: :@webpacker
+  delegate :config, :env, to: :@webpacker
 
   def initialize(webpacker)
     @webpacker = webpacker
   end
 
   def running?
-    Socket.tcp(host, port, connect_timeout: connect_timeout).close
-    true
+    if env.production?
+      false
+    else
+      Socket.tcp(host, port, connect_timeout: connect_timeout).close
+      true
+    end
   rescue
     false
   end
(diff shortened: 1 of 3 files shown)


lib/webpacker/dev_server.rb
```diff
@@ -3,15 +3,19 @@ class Webpacker::DevServer
...
 
   def running?
-    Socket.tcp(host, port, connect_timeout: connect_timeout).close
-    true
+    if env.production?
+      false
+    else
+      Socket.tcp(host, port, connect_timeout: connect_timeout).close
+      true
+    end
```
javan: I think this should be inverted and check `env.development?` instead. We have production-like envs (`staging`, `beta`) that we want skipped too.
```ruby
if env.development?
  Socket.tcp(host, port, connect_timeout: connect_timeout).close
  true
else
  false
end
```

gauravtiwari: All production like environments will have `NODE_ENV=production` and that will evaluate `env.production?` to `true` so the logic will work for all production like environments as long as NODE_ENV is set. Would that work? 

```rb
  def env
    (ENV["NODE_ENV"].presence_in(available_environments) ||
      Rails.env.presence_in(available_environments) ||
        "production".freeze).inquiry
  end
```

Reason for scoping to production was that, in case someone wants to use dev server in test environment. What do you think?

javan: Perhaps we can check if `dev_server` is configured? Since, by default, it's only in development: https://github.com/rails/webpacker/blob/6695445dc820b2f9f8adc354c3254a224c88747a/lib/install/config/webpacker.yml#L33-L38

gauravtiwari: Ahh yes, good idea 👍

gauravtiwari: Done 👍

Issue #1148, opened 2017-12-26
Integrations still appear in webpacker.yml


iainbeeston:
In #1103 integrations were moved out of the default webpacker install, and now have to be added manually. However, the default webpacker.yml still includes extensions for `.jsx`, `.coffee`, `.vue` and `.erb`. Should these be removed from the default webpacker.yml file, and be added by the appropriate installer rake task?

iainbeeston:
Just to be clear, the webpacker.yml file I'm looking at is https://github.com/rails/webpacker/blob/7738e6b2a25b5e35874cf5290a6e680ec10794b0/lib/install/config/webpacker.yml

gauravtiwari:
Good spot 👍 Yes they should be moved to installer.

PR #1181, opened 2018-01-14
Update default extensions and move to installer

gauravtiwari:
Fixes: #1148

This PR moves out extensions from webpacker.yml to specific installer.

### DIFF
--- a/lib/install/coffee.rb
+++ b/lib/install/coffee.rb
@@ -12,6 +12,9 @@ insert_into_file Rails.root.join("config/webpack/environment.js").to_s,
   "environment.loaders.append('coffee', coffee)\n",
   before: "module.exports"
 
+say "Updating webpack paths to include .coffee file extension"
+insert_into_file Webpacker.config.config_path, "    - .coffee\n", after: /extensions:\n/
+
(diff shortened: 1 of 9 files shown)


lib/install/coffee.rb
```diff
@@ -12,6 +12,9 @@
   "environment.loaders.append('coffee', coffee)\n",
   before: "module.exports"
 
+say "Updating webpack paths to include .coffee file extension"
+insert_into_file Webpacker.config.config_path, "    - .coffee\n", after: /extensions:\n/
```
javan: Inserting YAML like this at the corrected indentation level seems a little brittle. Do we do it this way elsewhere in Rails?

gauravtiwari: Couldn't find any reference but this is how it's done in Rails: https://github.com/rails/rails/blob/63f0c04850dd0bcdc7d35266e81fa1a7778570a8/railties/lib/rails/generators/actions.rb#L326

May be we can borrow that method.

javan:
Do we use these extensions on the Ruby side anywhere?

gauravtiwari:
Nope we aren't using it internally.

gauravtiwari:
Merging this for now :)

PR #1187, opened 2018-01-17
Fixes Vue root instance initialization

nerdcave:
Initializes Vue the more common way by passing in `el` instead of mounting explicitly. Calling `$mount()` is a little weird and API-y.


gauravtiwari:
thanks

PR #1194, opened 2018-01-21
Update uglify plugin

gauravtiwari:
This PR updates uglifyjs plugin so we can support ES7 and ES8 - also improves build performance


kortirso:
Yes, it works

Issue #1193, opened 2018-01-20
Should yarn-error.log be added to .gitignore file?


wiredin:
Should yarn-error.log be added to rails .gitignore automatically during webpacker setup?

inkhrt:
No

guilleiguaran:
yarn-error.log is added automatically in new Rails apps except if yarn is skipped, doing it in webpacker will add repeated entries to the .gitignore file.

See: https://github.com/rails/rails/blob/master/railties/lib/rails/generators/rails/app/templates/gitignore.tt#L31

/cc @gauravtiwari

gauravtiwari:
@guilleiguaran Ahh thanks :). Fixed: https://github.com/rails/webpacker/pull/1197

wiredin:
Thanks everyone! That makes sense, my app was originally rails version 5.0 so it did not already have yarn-error.log in the gitignore.

PR #1195, opened 2018-01-21
Gitignore

gauravtiwari:
Fixes: https://github.com/rails/webpacker/issues/1193

### DIFF
--- a/lib/install/template.rb
+++ b/lib/install/template.rb
@@ -27,6 +27,8 @@ if File.exists?(".gitignore")
 /public/packs
 /public/packs-test
 /node_modules
+yarn-debug.log*
+yarn-error.log*
 EOS
 end
 
--- a/lib/install/template.rb
+++ b/lib/install/template.rb
@@ -29,6 +29,7 @@ if File.exists?(".gitignore")
 /node_modules
 yarn-debug.log*
 yarn-error.log*
+.yarn-integrity
 EOS
 end
 
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,6 @@
 node_modules
 .byebug_history
 /test/test_app/tmp
+yarn-debug.log*
+yarn-error.log*
+.yarn-integrity


lib/install/template.rb
```diff
@@ -27,6 +27,9 @@
 /public/packs
 /public/packs-test
 /node_modules
+yarn-debug.log*
+yarn-error.log*
+.yarn-integrity
```
sikachu: I'm sorry for coming back to this old issue. However, I have a question about the decision to add `.yarn-integrity` to the gitignore template file.

So, this means that by default user won't be able to turn on integrity check on their remote server as when webpacker runs `yarn check --integrity` it would fail.

Is this the intended behavior, e.g. is it recommended to not check this file into source control (despite it seems like it should be shared across machine)?

Thank you.

gauravtiwari: @sikachu No worries! The reason for this is the integrity file would change between environments since the dependencies vary - development vs production(https://yarnpkg.com/lang/en/docs/cli/install/#toc-yarn-install-production-true-false). So, if you run `yarn install` in development the integrity file would be different vs `NODE_ENV=production yarn install` (unless I am mistaken).

gauravtiwari: Production: 

```json 
{
  "systemParams": "darwin-x64-57",
  "modulesFolders": [
    "node_modules"
  ],
  "flags": [
    "production"
  ],
  "linkedModules": [
    "@rails/webpacker"
  ],
```

vs 

Development: 

```json
{
  "systemParams": "darwin-x64-57",
  "modulesFolders": [
    "node_modules"
  ],
  "flags": [],
  "linkedModules": [
    "@rails/webpacker"
  ],
```

sikachu: Ah, I see. Thank you very much.

Very interesting, as I thought that `yarn check --integrity` with `.yarn-integrity` file is the way to check that the package hasn't been tampered with when you try to install the same package on your production server by comparing to what you have locally, but I guess it's already doing that with `packages.json`?

Anyway, I guess I'll just turn of the check on production server instead of checking this file in then, with the hope that yarn is already _doing the right thing_. 💯

gauravtiwari: Your welcome :)

Yeah, it does that by comparing `package.json` and `yarn.lock` file. I guess this check is really helpful in development when you have added some dependencies but forgot to install or using an outdated version as specified in yarn.lock or installed the dependencies in a different environment as the original. 

> Anyway, I guess I'll just turn off the check on production server instead of checking this file in then, with the hope that yarn is already doing the right thing. 

Yeah, you don't need to run this check on production. Haven't run into any problem so far.

Issue #1174, opened 2018-01-12
Webpacker gem generates incorrect manifest.json on windows 10


cshupp1:
I am using:
```
    webpacker (3.2.0)
    rails (5.1.4)
    JRuby 9.1.15.0
```

Suppose in /app/javascript/packs/hello I have:
```
hello_react.jsx
```

On Windows the manifest generated is:
```
{
  "application.js": "/packs/application-0dc18709a85406b5e3f5.js",
  "application.js.map": "/packs/application-0dc18709a85406b5e3f5.js.map",
  "hello\\hello_react.js": "/packs/hello\\hello_react-5a67faa5e91dd316f476.js",
  "hello\\hello_react.js.map": "/packs/hello\\hello_react-5a67faa5e91dd316f476.js.map"
}
```

This means in my erb file this does not work:
```
<%=javascript_pack_tag 'hello/hello_react'  %>
```

I must do this:
```
<%=javascript_pack_tag 'hello\hello_react'  %>
```
for a lookup to succeed.

Workaround:
```
class String
  def os_path!
      self.gsub!('/', java.io.File::separator)
      self.gsub!('\\', java.io.File::separator)
  end
end
```

in erb:

```
<%=javascript_pack_tag 'hello/hello_react'.os_path!  %>
```

gauravtiwari:
Thanks @cshupp1 for filing this. I don't have access to a windows machine so don't know what the problem is. If it's okay please could you take a look and submit a PR please?

cshupp1:
Will do.

cshupp1:
@gauravtiwari 
I have everything ready to go, all tests are passing.  My attempt to create a branch to make a pull request fails.

C:\work\VSO\webpacker\webpacker>git push
remote: Permission to rails/webpacker.git denied to cshupp1.
fatal: unable to access 'https://github.com/rails/webpacker/': The requested URL returned error: 403

Can you help?

Thanks,
Cris

gauravtiwari:
@cshupp1 Please fork webpacker repo and then make a PR using that repo. https://help.github.com/articles/creating-a-pull-request-from-a-fork/

cshupp1:
@gauravtiwari thanks for that.  Submitted.

guilleiguaran:
This has been fixed in `webpack-manifest-plugin` https://github.com/danethurber/webpack-manifest-plugin/pull/78 but we need to wait for the final release of the plugin (currently the fix is available only in 2.0.0-rc.1)

[...]

PR #1205, opened 2018-01-22
Allow look ups to succeed even if the build was done on Windows

cshupp1:
This is a fix for:

https://github.com/rails/webpacker/issues/1174

Originally, when I wrote up the defect I assumed the gem was writing the manifest.json file.  By holding a lock to manifest.json the stack trace reveals:

```
Error: EBUSY: resource busy or locked, open 'C:\work\VSO\vso\public\packs\manifest.json'
    at Object.fs.openSync (fs.js:646:18)
    at Object.fs.writeFileSync (fs.js:1291:33)
    at Object.outputFileSync (C:\work\VSO\vso\node_modules\fs-extra\lib\output\index.js:26:29)
    at ManifestPlugin.<anonymous> (C:\work\VSO\vso\node_modules\webpack-manifest-plugin\lib\plugin.js:158:11)
    at Compiler.applyPluginsAsyncSeries (C:\work\VSO\vso\node_modules\tapable\lib\Tapable.js:206:13)
    at Compiler.emitAssets (C:\work\VSO\vso\node_modules\webpack\lib\Compiler.js:354:8)
    at onCompiled (C:\work\VSO\vso\node_modules\webpack\lib\Compiler.js:58:19)
    at applyPluginsAsync.err (C:\work\VSO\vso\node_modules\webpack\lib\Compiler.js:510:14)
    at next (C:\work\VSO\vso\node_modules\tapable\lib\Tapable.js:202:11)
    at Compiler.<anonymous> (C:\work\VSO\vso\node_modules\webpack\lib\CachePlugin.js:78:5)
```

That the manifest file is written in node code by fs-extra.  A cursory examination leads me to believe that the code is technically correct (when running on Windows) since it is converting a file object to a string.

An alternative approach might be to get the node dependent library to support a flag to force UNIX style slashes.

The hash duplicates keys so, if running on a windows box, both look ups succeed:
'''
    <%=stylesheet_pack_tag 'stylesheets/layout' %>
'''
and
'''
    <%=stylesheet_pack_tag "stylesheets\\layout" %>
'''
Although I doubt any windows people would take the second approach (but both were tested in a deployed war file running in production mode).

I don't have a unix box at the moment to do that level of sanity testing.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,6 +6,8 @@ gem "rails"
 gem "rake", ">= 11.1"
 gem "rubocop", ">= 0.49", require: false
 gem "rack-proxy", require: false
+# Windows does not include zoneinfo files, so bundle the tzinfo-data gem
+gem 'tzinfo-data', platforms: [:mingw, :mswin, :x64_mingw, :jruby]
 
 group :test do
   gem "minitest", "~> 5.0"
--- a/Gemfile
+++ b/Gemfile
@@ -7,7 +7,7 @@ gem "rake", ">= 11.1"
 gem "rubocop", ">= 0.49", require: false
 gem "rack-proxy", require: false
 # Windows does not include zoneinfo files, so bundle the tzinfo-data gem
-gem 'tzinfo-data', platforms: [:mingw, :mswin, :x64_mingw, :jruby]
+gem "tzinfo-data", platforms: [:mingw, :mswin, :x64_mingw, :jruby]
 
 group :test do
   gem "minitest", "~> 5.0"
--- a/lib/webpacker/manifest.rb
+++ b/lib/webpacker/manifest.rb
@@ -28,24 +28,24 @@ class Webpacker::Manifest
   end
 
   private
-  def compiling?
-    config.compile? && !dev_server.running?
-  end
+    def compiling?
+      config.compile? && !dev_server.running?
+    end
 
-  def compile
-    Webpacker.logger.tagged("Webpacker") { compiler.compile }
-  end
+    def compile
+      Webpacker.logger.tagged("Webpacker") { compiler.compile }
+    end
 
-  def find(name)
-    data[name.to_s].presence
-  end
+    def find(name)
+      data[name.to_s].presence
+    end
 
-  def handle_missing_entry(name)
-    raise Webpacker::Manifest::MissingEntryError, missing_file_from_manifest_error(name)
-  end
+    def handle_missing_entry(name)
+      raise Webpacker::Manifest::MissingEntryError, missing_file_from_manifest_error(name)
+    end
 
-  def missing_file_from_manifest_error(bundle_name)
-    <<-MSG
+    def missing_file_from_manifest_error(bundle_name)
+      <<-MSG
 Webpacker can't find #{bundle_name} in #{config.public_manifest_path}. Possible causes:
 1. You want to set webpacker.yml value of compile to true for your environment
    unless you are using the `webpack -w` or the webpack-dev-server.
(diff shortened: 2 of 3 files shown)


lib/webpacker/manifest.rb
```diff
@@ -28,24 +28,24 @@ def lookup!(name)
   end
 
   private
-    def compiling?
-      config.compile? && !dev_server.running?
-    end
+  def compiling?
```
gauravtiwari: Please can you revert indentation change. All rails gems use this style: 

```rb
private
  def compiling?
  end

Comment (gauravtiwari) [reply lib/webpacker/manifest.rb]: May be just run `bundle exec rubocop -a` to fix

File: lib/webpacker/manifest.rb [hunk 2]
```diff
@@ -54,22 +54,32 @@ def missing_file_from_manifest_error(bundle_name)
...
+          data[dup_key] = data[key]
+        end
       end
+      data
+    else
+      {}
     end
-end
+  end
+end
```
gauravtiwari: A new line in the end please

Gemfile
```diff
@@ -6,6 +6,8 @@ gem "rails"
 gem "rake", ">= 11.1"
 gem "rubocop", ">= 0.49", require: false
 gem "rack-proxy", require: false
+# Windows does not include zoneinfo files, so bundle the tzinfo-data gem
+gem 'tzinfo-data', platforms: [:mingw, :mswin, :x64_mingw, :jruby]
```
gauravtiwari: Don't think we need this inside gem

cshupp1: People contributing on windows cannot run 'bundle exec rake' w/o it.

lib/webpacker/manifest.rb
cshupp1: OK - will do

lib/webpacker/manifest.rb
```diff
@@ -54,22 +54,32 @@ def missing_file_from_manifest_error(bundle_name)
...
-      if config.public_manifest_path.exist?
-        JSON.parse config.public_manifest_path.read
-      else
-        {}
+  def load
+    if config.public_manifest_path.exist?
+      data = JSON.parse config.public_manifest_path.read
+      #if the build was done on a windows box manifest.json will have backslashes as keys
+      #we will ensure lookups continue to succeed
+      data.keys.each do |key|
```
gauravtiwari: Shall we wrap this code inside? `if Gem.win_platform?`

lib/webpacker/manifest.rb
cshupp1: Will do

Gemfile
gauravtiwari: Ahh okay makes sense 👍  Double quotes please

```rb
gem "tzinfo-data", platforms: [:mingw, :mswin, :x64_mingw, :jruby]
```

lib/webpacker/manifest.rb
cshupp1: No, it has nothing to do where the code is running.  Only where the build was run.

gauravtiwari: Oh yes ofcourse 👍

lib/webpacker/manifest.rb
cshupp1: RubyMine get whiny about double quotes when singles will do... Interesting

guilleiguaran:
I would prefer to get this fixed upgrading the upstream library causing the problem (https://github.com/danethurber/webpack-manifest-plugin/pull/78) instead of fixing it on webpacker side.

cshupp1:
Me too, hence my comment:

**An alternative approach might be to get the node dependent library to support a flag to force UNIX style slashes.**

I don't suppose we can keep this bug open until the dependent library  (with the fix) is mated to this gem so I know when to update?

guilleiguaran:
Closing this but #1174 will leave open until we upgrade the dependency.

PR #1208, opened 2018-01-23
Install some npm libraries as devDependencies

ohbarye:
Hello maintainers.

When I tried to use some installers, I found that some of npm libraries which should locate in `devDependencies` are in `dependencies` section in package.json.

So I just tweaked the installation scripts.

### DIFF
--- a/lib/install/coffee.rb
+++ b/lib/install/coffee.rb
@@ -20,6 +20,6 @@ copy_file "#{__dir__}/examples/coffee/hello_coffee.coffee",
   "#{Webpacker.config.source_entry_path}/hello_coffee.coffee"
 
 say "Installing all Coffeescript dependencies"
-run "yarn add coffeescript@1.12.7 coffee-loader"
+run "yarn add --dev coffeescript@1.12.7 coffee-loader"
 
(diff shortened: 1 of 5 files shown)


javan:
I believe these are needed as `dependencies` for compiling on the server (aka deploying).

gauravtiwari:
Thanks @javan 

Yeah, please see this PR https://github.com/rails/webpacker/issues/117 and related issue for more info.

gauravtiwari:
Ahh I think `elm-hot-loader` can be under devDepencies so if you wanna update the PR with only that change I am happy to merge 👍

ohbarye:
@javan @gauravtiwari 
I didn't know the background and now understand that so let me close this.

> Ahh I think elm-hot-loader can be under devDepencies so if you wanna update the PR with only that change I am happy to merge, thank you 👍

`elm-hot-loader` has originally been under `devDependencies`. So I have nothing to do for that. 😅 

https://github.com/rails/webpacker/pull/1208/files#diff-70f863ebe5733123329d857ce42200ebL25

Thank both of you!

PR #1209, opened 2018-01-23
Install and require initializers

### LINKED ISSUE #1207: Add app/javascript/initializers and require all files by default in packs/application.js
Lots of frameworks need a generic initialization before they're ready for use. This goes for Action Cable, Active Storage, Rails UJS, and Turbolinks, to name a few. To be able to programmatically add these initializers via gems or other means, it'd be nicer if the initialization happened in individual files, and all those files are just required by pack/application.js. Just like we do with config/initializers on the Ruby side.

For example:
```javascript
// app/javascript/initializers/actioncable.js
import ActionCable from 'actioncable'
window.Cable = ActionCable.createConsumer()

// app/javascript/initializers/activestorage.js
import * as ActiveStorage from 'activestorage'
ActiveStorage.start()

// app/javascript/initializers/rails_ujs.js
import Rails from 'rails-ujs'
Rails.start()

// app/javascript/initializers/turbolinks.js
import Turbolinks from 'turbolinks'
Turbolinks.start()

// app/javascript/packs/application.js
// Code that requires all the files in app/javascript/initializers/*.js
```

rossta:
Fixes #1207 

This is perhaps the most straightforward way I know of to require all files in a directory. There are Webpack plugins that may provide similar functionality, but would also require the developer to learn extra configuration to modify.

https://webpack.js.org/guides/dependency-management/#require-context

### DIFF
--- a/lib/install/javascript/initializers/index.js
+++ b/lib/install/javascript/initializers/index.js
@@ -1,0 +1,2 @@
+const importAll = (r) => r.keys().forEach(r)
+importAll(require.context('.', true, /\.js$/));
--- a/lib/install/javascript/initializers/hello.js
+++ b/lib/install/javascript/initializers/hello.js
@@ -1,0 +1,1 @@
+console.log('Hello World from a Webpacker initializer')
--- a/lib/install/javascript/packs/application.js
+++ b/lib/install/javascript/packs/application.js
@@ -7,4 +7,5 @@
 // To reference this file, add <%= javascript_pack_tag 'application' %> to the appropriate
 // layout file, like app/views/layouts/application.html.erb
 
+import '../initializers';
 console.log('Hello World from Webpacker')


jerrymetz60:
than1

lib/install/javascript/initializers/index.js
```diff
@@ -0,0 +1,2 @@
+const importAll = (r) => r.keys().forEach(r)
+importAll(require.context('.', true, /\.js$/));
```
javan: Will this end up importing itself (`./index.js`)?

rossta: Yes, though my understanding is that is [supported in ES6-style module imports](http://exploringjs.com/es6/ch_modules.html#sec_cyclic-dependencies).  

I created a [gist](https://gist.github.com/rossta/5a9edcd7ba37416f8c6f7ed383200b0d) that includes an excerpt of the Webpack output to show that the result would be similar to placing `import './index'` and `import './hello'` in `initializers/index.js`, which does complie and execute both modules.

If preferred, I can move the `require.context` code to a separate file.

gauravtiwari:
@rossta I guess this is part of Rails 6 now so we can skip it? https://github.com/rails/rails/pull/33079/files#diff-3c2cf0be9cee71452d40f7ec007ced85

Issue #1200, opened 2018-01-22
Update documentation on Rails 5.2 security policy


savroff:
Original issue from Rails repo: https://github.com/rails/rails/issues/31754#event-1434257847

I think that update needed in the documentation of Webpacker. Right now if you read docs, you will think that for React everything works out of the box. 

You can see info about 5.2 security updates only on [Vue](https://github.com/rails/webpacker#vue) and [Angular](https://github.com/rails/webpacker#vue) topics.

Plus nothing there about web sockets (they should work by default with Webpacker to do a hot reload)

Also I wonder if Rails 5.2 with Webpacker gem should pre-generate right security policy for dev mode?

gauravtiwari:
Thanks @savroff Please feel free to make a PR to update documentation. 

> Also I wonder if Rails 5.2 with Webpacker gem should pre-generate right security policy for dev mode?

Sounds like a great idea 👍

guilleiguaran:
> Also I wonder if Rails 5.2 with Webpacker gem should pre-generate right security policy for dev mode?

Is a good idea, I tried to do it but isn't trivial to do since we must deal not only with default generated CSP config but also with potential changes done by users to the file.

savroff:
I have several ideas, wanna try them

guilleiguaran:
> Right now if you read docs, you will think that for React everything works out of the box.

Does React requires `unsafe_eval` or do we need it for Hot Module Reloading?

savroff:
I totally missed this line.

savroff:
@guilleiguaran I think we need someone who using react here :)))) I'm using vue

guilleiguaran:
> I totally missed this line.

Yup, I think it isn't visible enough in the docs right now, maybe it's a better idea to add a new section about CSP to the docs.

savroff:
@guilleiguaran We can rewrite CSP like this:

```ruby
Rails.application.config.content_security_policy do |p|
  p.font_src    :self, :https, :data
  p.img_src     :self, :https, :data
  p.object_src  :none
  p.style_src   :self, :https, :unsafe_inline


  p.script_src :self, :https
  p.default_src :self, :https

  # Specify URI for violation reports
  # p.report_uri "/csp-violation-report-endpoint"

  # standard until this moment
  if Rails.env.development?
    p.script_src :self, :https, :unsafe_eval
    p.connect_src :self, :https, 'http://localhost:3035', 'ws://localhost:3035'
  end
end
```

[...]

PR #1210, opened 2018-01-23
CSP warning for Rails 5.2

savroff:
https://github.com/rails/webpacker/issues/1200

screenshot 2018-01-23 11 17 56


guilleiguaran:
For Angular and Vue we can mention to `p.script_src` also

gauravtiwari:
Thanks @savroff I guess we can keep it simple and remove `-----` around code - an extra lb would do just fine. What do you think?

savroff:
@guilleiguaran I thought about this too :) nice

@gauravtiwari  sure! Just did it to have more readable code example

gauravtiwari:
Also, don't think we need to add same message in all installers unless it's specific to that installer (vue and angular). Since base installers runs first it's fine to just add `connect_src` block there.

savroff:
Seems like everything should be ok now

savroff:
@gauravtiwari thank you! 🎉

guilleiguaran:
Thanks for working on this!!!

Issue #1225, opened 2018-01-26
Import matches wrong file


IgorDmitriev:
I was upgrading project to most recent webpacker version `3.2.1` (from `3.0.2`) and got hit by this new behavior:
```
|- Hello.css
|- Hello.jsx
|- index.js
```

When we have stylesheets file name the same as a javascript file name and using import without specifying an extension, for some reason now it would import CSS file.

Expected behavior: if we are not specifying extension, lookup for files string with javasctipt files `js`->`jsx`-> styles -> other files.

Here is a sample code for these files, to reproduce.

```
// index.js
import React from 'react';
import ReactDOM from 'react-dom';
import Hello from './Hello';

document.addEventListener('DOMContentLoaded', () => {
  ReactDOM.render(
    <Hello name="React" />,
    document.body.appendChild(document.createElement('div')),
  );
});
```

```
// Hello.jsx
import React from 'react';
import './Hello.css';

const Hello = ({ name }) => <h1>Hello, {name}!</h1>;

export default Hello;
```

```
/* Hello.css */
body {
  background-color: #222;
  color: #fff;
}
```

I was able to reproduce it with clean repo after `rails new myapp --webpack=react`.

gauravtiwari:
Can you please share how extensions are configured in webpacker.yml?

```yml
  extensions:
    - .js
    - .jsx
    - .sass
    - .css
    - .png
    - .svg
    - .gif
    - .jpeg
    - .jpg
```

IgorDmitriev:
We have these extensions mentioned there, I think it's a default. I see the same list in a new rails app.
```yml
  extensions:
    - .jsx
    - .js
    - .sass
    - .scss
    - .css
    - .png
    - .svg
    - .gif
    - .jpeg
    - .jpg
```

I already tried to remove `scss`, `css`, `sass` from this list didn't help. Same with rearrangement. 

Also, I noticed, when I intentionally trying to import some file that doesn't exist, I'm getting this stack trace with the same ordering no matter how we set it in `webpacker.yml` file.

```
[webpacker-test/app/javascript/hello/HelloS]
[webpacker-test/app/javascript/hello/HelloS.js]
[webpacker-test/app/javascript/hello/HelloS.sass]
[webpacker-test/app/javascript/hello/HelloS.scss]
[webpacker-test/app/javascript/hello/HelloS.css]
[webpacker-test/app/javascript/hello/HelloS.png]
[webpacker-test/app/javascript/hello/HelloS.svg]
[...]

PR #1231, opened 2018-01-29
Fix config to resolve extensions as specified in webpacker.yml

IgorDmitriev:
Fixes https://github.com/rails/webpacker/issues/1225

### Problem
When we are deep-merging `defaultConfig.resolve.extensions` with `appConfig.resolve.extensions`, it ignores ordering of extensions specified in `appConfig` and always includes all `defaultConfig.resolve.extensions` and if there is something new presented in `appConfig` it will add it at the end of the array.
It results in some unexpected behavior as in this issue https://github.com/rails/webpacker/issues/1225.

### Solution
Overwrite default `config.resolve.extensions` if it presented in user's `appConfig`.

### DIFF
--- a/package/__tests__/environment.js
+++ b/package/__tests__/environment.js
@@ -70,5 +70,22 @@ describe('Environment', () => {
       expect(config.plugins).toBeInstanceOf(Array)
       expect(config.plugins).not.toBeInstanceOf(ConfigList)
     })
+
+    test('should return resolve extensions as listed in app config', () => {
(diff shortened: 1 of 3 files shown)


IgorDmitriev:
The failing ruby test is testing that `resolve.extensions` is the same in `test_app` as in install/default config.

Is this test still valuable? How should I resolve it?

gauravtiwari:
Thanks @IgorDmitriev for your contribution. Took inspiration from your work and pushed something here #1253

PR #1232, opened 2018-01-29
Removed unnecessary activesupport explicit dependency.

vavgustov:
railties >= 4.2 already has activesupport runtime dependency, https://github.com/rails/rails/blob/master/railties/railties.gemspec


rafaelfranca:
Thank you for the pull request, even that railties depends on activesupport this gem also depends on it. If we remove it now, if railties one day remove the dependency this gem will be broken. That said this gemspec is correct.

Issue #1106, opened 2017-12-15
asset paths inside assets when deploying to sub-url


ipmsteven:
#### Problem 
Asset paths inside assets when deploying to sub-URI have issue when resolving to right url

#### Example
When app is mounted under `/portal`, having a component like `Logo` below:

```javascript
import BookLogo from '../images/book_logo.jpg';

 export class Logo extends Component {
    render(){
        return (
          ...
          <img src={BookLogo} />
          ...
        )
    } 
}
```
The correct path `/portal/assets/packs/images/book_logo.jpg` is expected but a wrong path `/assets/packs/images/book_logo.jpg` is generated for src of img on production environment.

#### Question
Is there a config I can pass like how asset pipeline does by using `relative_url_root`?
#### Reference
https://github.com/rails/rails/pull/3946

gauravtiwari:
@ipmsteven What happens if you set public output path as: 

```js
__webpack_public_path__ = `/portal/packs`
```

```jsx
__webpack_public_path__ = `/portal/packs`
import BookLogo from '../images/book_logo.jpg';

 export class Logo extends Component {
    render(){
        return (
          ...
          <img src={BookLogo} />
          ...
        )
    } 
}
```

And then build and deploy to production?

ipmsteven:
Hi @gauravtiwari , Thanks for you reply, that would work but have some manual work involved.

since `webpacker:compile` is part of asset pipeline, could we let `webpacker:compile`
respect the `relative_url_root` setting from rails?

bingneef:
If `__webpack_public_path__ = '/portal/packs'` does not work, make sure to have at least webpacker 3.2.0 in your package.json (`yarn add @rails/webpacker`), I stared at this for ages since I only updated the gem, not the NPM version.

linstantnoodles:
@ipmsteven Have you tried configuring your `webpacker.yml` file to have
`public_output_path: portal/packs` so that all of your pack assets go to `portal/`?

ipmsteven:
@linstantnoodles yes, that could work, but it's not a good solution for my use case.

In my situation `portal` is a configurable value,  it is stored in a configuration file and different environments has different settings.

Say for dev: it's '/', for test: it's '/test', for stage, it's '/blah', etc.
but no matter what's environments you are in, it's should always the same with relative_url_root(where the app is mounted to)
That's the reason I create this PR  https://github.com/rails/webpacker/pull/1236

Does that make sense to you @linstantnoodles ?

[...]

PR #1236, opened 2018-01-30
add relative_url_root to webpacker config

ipmsteven:
This PR will set `WEBPACKER_RELATIVE_URL_ROOT` as `ActionController::Base.relative_url_root` which partially solve this issue https://github.com/rails/webpacker/issues/1106

would you mind taking a look? @gauravtiwari 
Thanks

### DIFF
--- a/lib/webpacker/compiler.rb
+++ b/lib/webpacker/compiler.rb
@@ -78,6 +78,7 @@ class Webpacker::Compiler
     end
 
     def webpack_env
-      env.merge("NODE_ENV" => @webpacker.env, "WEBPACKER_ASSET_HOST" => ActionController::Base.helpers.compute_asset_host)
+      env.merge("NODE_ENV" => @webpacker.env, "WEBPACKER_ASSET_HOST" => ActionController::Base.helpers.compute_asset_host,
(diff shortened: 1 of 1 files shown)


brunoOSI:
Please add this in, it is a bummer that you only can ship precompiled asset paths. It should allow some configuration!

gauravtiwari:
Thanks @ipmsteven for this. Sorry for the delay. 
I think this requires a bit more wiring to function correctly like appending relative root to publicPath. Will take a look at this soon.

ipmsteven:
Hi @gauravtiwari 

You are right, this PR only make `relative_url_root` available during webpack compile time.  A bit more wiring need to be done in webpacker level to construct `publicPath` that respect `relative_url_root`.
I am thinking we could start with it and delegate the plumbing work to the user of webpacker for now.

so with that being said above, If you agree with the proposal of adding `WEBPACKER_RELATIVE_URL_ROOT` to webpacker compile environment, 

we could either
Plan 1:
1. merge this PR that only ships `WEBPACKER_RELATIVE_URL_ROOT` first
2. have separate PR to do the wiring work as you mentioned above

OR

Plan 2: Have one PR handle both 1 and 2

I am fine with either one, what do you think? :)

ipmsteven:
thanks! @gauravtiwari

PR #1250, opened 2018-02-07
bump version number in README

waruboy:
Hello!

Thanks for the gem, I think I will use it a lot.

I see that the version listed on the README instruction is not up to date yet. On the old 3.0 version webpacker:install failed to create node_modules directory. Took me awhile to figure out I'm using the old version. Hopefully this simple change can save some little frustration on future newcomer


gauravtiwari:
thanks @waruboy

PR #1257, opened 2018-02-08
Hash files based on output rather than source

agirdler-lumoslabs:
By default webpack hashes files based on their source contents, when doing any sort of uglification or transpiling this allows the output to change without the digest changing.

This is a known issue(https://github.com/webpack/webpack/issues/4659) and has an actively maintained plugin written specifically to alleviate that issue (https://www.npmjs.com/package/webpack-plugin-hash-output).

This patch set adds a dependency on `webpack-plugin-hash-output` and adds it to the default plugin list so that any projects which depend on webpacker's environment will have properly generated hashes that will work with caching.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,8 @@
     "eslint-plugin-import": "^2.8.0",
     "eslint-plugin-jsx-a11y": "^6.0.3",
     "eslint-plugin-react": "^7.6.1",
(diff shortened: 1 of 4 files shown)


gauravtiwari:
This doesn't seem like an issue per https://github.com/webpack/webpack/issues/4659#issuecomment-385949875

PR #1258, opened 2018-02-09
use NODE_ENV for development

tkamenick:
The code currently requires the process name to match 'webpack-dev-server' in order to not use the ExtractTextPlugin loader.

This changes the code to check for  `NODE_ENV === 'development'` and sets the non `ExtractTextPlugin` versions of the loaders unless the env is production.

This allows the usage of other applications, like Storybook, to reuse the same configuration in development that work with HMR.

I'm guessing that there was a reason why `NODE_ENV` couldn't be used for this. If that's the case, I understand. But, there still should be a way to override the settings to allow the non extracted versions of the loaders to be used, outside of modifying the process name.

### DIFF
--- a/package/rules/css.js
+++ b/package/rules/css.js
@@ -4,8 +4,9 @@ const { dev_server: devServer } = require('../config')
 
 const postcssConfigPath = path.resolve(process.cwd(), '.postcssrc.yml')
 const isProduction = process.env.NODE_ENV === 'production'
+const isDevelopment = process.env.NODE_ENV === 'development'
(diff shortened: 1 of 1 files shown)


tkamenick:
For additional context: my goal is to use the same webpack configuration for both webpack and storybook.

Without this change, the webpacker configuration will use the `ExtractTextPlugin` versions of the loaders, even if the NODE_ENV is set to development. However, it seems like the configuration should not use ExtractTextPlugin unless the env is production, or maybe there should be some other way of specifying. Happy to modify this accordingly!

PR #1263, opened 2018-02-12
Fix compare link for 3.2.2

rmehner:
No biggie, just stumbled over this while trying to debug something :)


gauravtiwari:
thanks 👍

PR #1278, opened 2018-02-16
README English flow + grammar updates

hiimtaylorjones:
Hey y'all!

Was looking through the README today and found a lot of places that could use a bit more thought in consideration around how they're worded. Let me know if you have questions or concerns with the changes made!


gauravtiwari:
thanks @hiimtaylorjones

Issue #1282, opened 2018-02-22
 bundler: failed to load command: webpack


dpaola2:
Hi, I am following the installation instructions and when I attempt to deploy to Heroku, I see this:

```
remote:        Running: rake assets:precompile
remote:        I, [2018-02-22T05:05:54.811401 #412]  INFO -- : Writing /tmp/build_7ffe41733c668d5375f4c7e16f34a7c9/public/assets/subscribe-e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855.js
remote:        I, [2018-02-22T05:05:54.813303 #412]  INFO -- : Writing /tmp/build_7ffe41733c668d5375f4c7e16f34a7c9/public/assets/subscribe-e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855.js.gz
remote:        Webpacker is installed 🎉 🍰
remote:        Using /tmp/build_7ffe41733c668d5375f4c7e16f34a7c9/config/webpacker.yml file for setting up webpack paths
remote:        Compiling…
remote:        Compilation failed:
remote:        bundler: failed to load command: webpack (/tmp/build_7ffe41733c668d5375f4c7e16f34a7c9/vendor/bundle/ruby/2.2.0/bin/webpack)
remote:        The latest bundler is 1.16.1, but you are currently running 1.15.2.
remote:        To update, run `gem install bundler`
remote:        Errno::ENOENT: No such file or directory - /tmp/build_7ffe41733c668d5375f4c7e16f34a7c9/node_modules/.bin/webpack
remote:        /tmp/build_7ffe41733c668d5375f4c7e16f34a7c9/vendor/bundle/ruby/2.2.0/gems/webpacker-3.2.2/lib/webpacker/webpack_runner.rb:11:in `exec'
remote:        /tmp/build_7ffe41733c668d5375f4c7e16f34a7c9/vendor/bundle/ruby/2.2.0/gems/webpacker-3.2.2/lib/webpacker/webpack_runner.rb:11:in `block in run'
remote:        /tmp/build_7ffe41733c668d5375f4c7e16f34a7c9/vendor/bundle/ruby/2.2.0/gems/webpacker-3.2.2/lib/webpacker/webpack_runner.rb:10:in `chdir'
[...]

PR #1283, opened 2018-02-22
Update Heroku instructions to include buildpack sequence

dpaola2:


ytbryan:
Is there necessary? Heroku seems to have it installed by default, correct me if i'm wrong.

dpaola2:
I was following the instructions on a (nearly) vanilla rails app and ran into #1282 . This was my issue. Just trying make sure others behind me don't hit the same (extremely frustrating) obstacle.

p0wl:
We had to do this by hand as well.

matthewd:
cc @hone

joecorkerton:
I ran into the same problem, thanks for the fix. I also created an issue in the ruby buildpack repo as I believe that is where this should be fixed.

https://github.com/heroku/heroku-buildpack-ruby/issues/712

sunnyrjuneja:
We also had to do this. It's also helpful to specify the version of node / yarn in package.json. I think it's worth mentioning in the PR.

guilleiguaran:
Looks like this has been fixed already on the Heroku buildpack.

PR #1286, opened 2018-02-22
Make executables local

### LINKED ISSUE #1281: Why does this hijack my default webpack bin?
It's NOT okay to hijack my default webpack command, I have other projects where I'm either not using RoR or they don't use webpacker, and it keeps asking for the config:

> Webpack config /Users/csergiu/[...]/config/webpack/development.js not found, please run 'bundle exec rails webpacker:install' to install webpacker with default configs or add the missing config file for your custom environment.

How do I uninstall this? I don't want to run `node ./node_modules/webpack/bin/webpack.js` everytime. Thanks.

rossta:
Fixes #1281

Potentially breaking change as previously installed binstubs may be referencing a gem executable that may not be installed on the machine. Developers should re-run the install task to update their binstubs. They may also want to uninstall and reinstall the gem to remove the gem executables from their machines.

### DIFF
--- a/lib/tasks/webpacker/check_binstubs.rake
+++ b/lib/tasks/webpacker/check_binstubs.rake
@@ -1,7 +1,7 @@
 namespace :webpacker do
   desc "Verifies that webpack & webpack-dev-server are present."
   task :check_binstubs do
-    unless File.exist?("bin/webpack") && File.exist?("bin/webpack-dev-server")
+    unless File.exist?("bin/webpack")
       $stderr.puts "webpack binstubs not found.\n"\
            "Have you run rails webpacker:install ?\n"\
            "Make sure the bin directory or binstubs are not included in .gitignore\n"\
--- a/lib/tasks/webpacker/check_binstubs.rake
+++ b/lib/tasks/webpacker/check_binstubs.rake
@@ -1,7 +1,7 @@
 namespace :webpacker do
   desc "Verifies that webpack & webpack-dev-server are present."
   task :check_binstubs do
-    unless Bundler.which(Gem.win_platform? ? "webpack.bat" : "webpack")
+    unless File.exist?("bin/webpack") && File.exist?("bin/webpack-dev-server")
       $stderr.puts "webpack binstubs not found.\n"\
            "Have you run rails webpacker:install ?\n"\
            "Make sure the bin directory or binstubs are not included in .gitignore\n"\
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,9 @@
 **Please note that Webpacker 3.1.0 and 3.1.1 has some serious bugs so please consider using either 3.0.2 or 3.2.0**
+## Unreleased
+
+### Breaking changes
+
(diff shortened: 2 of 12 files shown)


lib/tasks/webpacker/check_binstubs.rake
```diff
@@ -1,7 +1,7 @@
 namespace :webpacker do
   desc "Verifies that webpack & webpack-dev-server are present."
   task :check_binstubs do
-    unless Bundler.which(Gem.win_platform? ? "webpack.bat" : "webpack")
+    unless File.exist?("bin/webpack") && File.exist?("bin/webpack-dev-server")
```
gauravtiwari: @rossta I think it's fine to leave out check for dev server since people might like to .gitignore it for production.

gauravtiwari:
Thanks for the PR, looks great. There are couple of places we need to replace `bundle exec webpack` executable as well - compiler.rb and docker.md readme

lib/tasks/webpacker/check_binstubs.rake
rossta: Ah, right, I was reverting to older behavior.

rossta:
@gauravtiwari I can update those references if you'd prefer them to use `./bin/webpack`, but `bundle exec webpack` will actually still work within projects since Bundler will look in the `bin` directory first for a matching executable.

gauravtiwari:
Ahh right but getting this: 

(cached executable call)
```
bundler: failed to load command: webpack (/Users/admin/.rbenv/versions/2.5.0/lib/ruby/gems/2.5.0/bin/webpack)
Gem::Exception: can't find executable webpack for gem webpacker
  /Users/admin/.rbenv/versions/2.5.0/lib/ruby/gems/2.5.0/gems/bundler-1.16.1/lib/bundler/rubygems_integration.rb:458:in `block in replace_bin_path'
  /Users/admin/.rbenv/versions/2.5.0/lib/ruby/gems/2.5.0/gems/bundler-1.16.1/lib/bundler/rubygems_integration.rb:478:in `block in replace_bin_path'
  /Users/admin/.rbenv/versions/2.5.0/lib/ruby/gems/2.5.0/bin/webpack:23:in `<top (required)>'
```
which makes sense because we have removed the executables from gemspec so bundler has no way to know that an executable exists, no?

If I remove the file, then running the same command gives,

```
rbenv: webpack: command not found

The `webpack' command exists in these Ruby versions:
  2.4.2
  2.4.3
```

rossta:
@gauravtiwari I forgot that I'm using [the technique described here](https://robots.thoughtbot.com/but-i-dont-want-to-bundle-exec#but-what-if-i-don39t-want-to-bundle-exec) to add the local bin directory to `$PATH`.

gauravtiwari:
Thanks for the update - the last bit is prefixing all the ./bin/webpack or dev server calls (if any) with `RbConfig.ruby` so it works on windows too. 

```rb
Open3.capture3(webpack_env, "#{RbConfig.ruby} ./bin/webpack")
```

rossta:
@gauravtiwari Good call. I believe the only functional usage is the one I just updated.

gauravtiwari:
Looks good. Thanks @rossta 🍰 👍

PR #1287, opened 2018-02-22
Add binstubs task

rossta:
Adds a new task `rails webpacker:binstubs` for copying the webpacker executables into Rails projects.

Allows devs to generate binstubs without having to run the whole install task, effectively replacing `bundle binstubs webpacker`. 

Depends on #1286 and may need to be modified slightly if that gets merged.


gauravtiwari:
This looks good too, if you can fix the conflicts then we can merge.

PR #1304, opened 2018-02-27
Consistently use NODE_ENV and support non-standard env

gauravtiwari:
Closes #1265

This PR fixes issue for non-standard environments

- Adds separate env class for gem
- Consistent gem and npm package behaviour
- Use NODE_ENV for consistency
- Removes deprecated set from `ConfigList` type
- Adds tests 

cc// @rmehner @IgorDmitriev @doits @hiromi2424 do you wanna try out this branch please?

### DIFF
--- a/lib/webpacker.rb
+++ b/lib/webpacker.rb
@@ -19,6 +19,7 @@ module Webpacker
 end
 
 require "webpacker/instance"
+require "webpacker/env"
 require "webpacker/configuration"
 require "webpacker/manifest"
 require "webpacker/compiler"
--- a/lib/webpacker/dev_server_runner.rb
+++ b/lib/webpacker/dev_server_runner.rb
@@ -14,7 +14,7 @@ module Webpacker
     private
       def load_config
         @config_file = File.join(@app_path, "config/webpacker.yml")
-        dev_server = YAML.load_file(@config_file)[ENV["RAILS_ENV"]]["dev_server"]
+        dev_server = YAML.load_file(@config_file)[ENV["NODE_ENV"]]["dev_server"]
 
         @hostname          = dev_server["host"]
         @port              = dev_server["port"]
--- a/lib/webpacker/env.rb
+++ b/lib/webpacker/env.rb
@@ -1,0 +1,39 @@
+class Webpacker::Env
+  DEFAULT = "production".freeze
+
+  delegate :config_path, :logger, to: :@webpacker
+
+  def self.inquire(webpacker)
+    new(webpacker).inquire
+  end
+
+  def initialize(webpacker)
+    @webpacker = webpacker
+  end
+
+  def inquire
+    fallback_env_warning unless current
+    (current || DEFAULT).inquiry
+  end
+
+  private
+    def current
+      (ENV["NODE_ENV"] || Rails.env).presence_in(available_environments)
+    end
+
+    def fallback_env_warning
+      logger.info "NODE_ENV=#{ENV["NODE_ENV"]} and RAILS_ENV=#{Rails.env} environment is not defined in config/webpacker.yml, falling back to #{DEFAULT} environment"
+    end
+
+    def available_environments
+      if config_path.exist?
+        YAML.load(config_path.read).keys
+      else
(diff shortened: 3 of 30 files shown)


lib/webpacker.rb
```diff
@@ -2,6 +2,9 @@
 require "active_support/logger"
 require "active_support/tagged_logging"
 
+ENV["RAILS_ENV"] ||= ENV["RACK_ENV"] || "development"
+ENV["NODE_ENV"] ||= ENV["RAILS_ENV"]
```
javan: We probably don't want to set any `ENV` vars here since this file gets required by applications. It was OK to do in bin/* files since those are separate one-off processes.

javan:
Does this allow `RAILS_ENV` and `NODE_ENV` to be set separately / independently? I imagine there are cases where you want `NODE_ENV` to be `production` and `RAILS_ENV` to be something else. Like deploying your Rails app to a staging environment.

gauravtiwari:
Yep it was possible before as well except that we were falling back to RAILS_ENV if there was no NODE_ENV set. I removed that case for clarity so it always checks for NODE_ENV first and then falls back to 'production'. What do you think?

lib/webpacker.rb
gauravtiwari: Ahh right, I was thinking setting here will set it in all environments (like in rails console) but I guess that's not ideal.

javan:
What happens when there's no config/webpack/[environment].js file matching the env? I can't tell at a glance.

gauravtiwari:
So if the config file is missing then binstub will raise an error (same as Rails env config):

screen shot 2018-02-28 at 08 10 34

If the config file is present matching environment but the settings isn't present inside webpacker.yml for that environment: 

screen shot 2018-02-28 at 08 11 00
 
If both config file and env is defined in webpacker.yml then: 

screen shot 2018-02-28 at 08 13 38

hiromi2424:
My Circle CI test was failed with following message:

```
Webpacker is installed 🎉 🍰
Using /******/config/webpacker.yml file for setting up webpack paths
rails aborted!
Using webpacker requires that you set NODE_ENV environment variables. Valid values are default, development, test, staging, production
```

We expect webpacker to fallback `RAILS_ENV`, that ours is 'test' 😢 

If this is breaking change please update minor(or major) version..

gauravtiwari:
@hiromi2424 Ahh I see, what is NODE_ENV set to `test`?

hiromi2424:
@gauravtiwari  It passed with `NODE_ENV: test` .

`RAILS_ENV` fallback also allow us to separate webpacker env and application env, e.g.:

RAILS_ENV = staging

webpacker.yml contains:
development test production

It fallbacks production config for webpacker. yay!

I think raising error is not comfortable decision since webpacker is for rails application that requires `RAILS_ENV` but not `NODE_ENV`.

sunnyrjuneja:
I highly endorse the idea of using RAILS_ENV as a fallback for NODE_ENV.

gauravtiwari:
alright makes sense, will push an update 👍

hiromi2424:
Huge thanks! @gauravtiwari
Now CI and staging build of my project has been succeeded!!!

Systho:
@gauravtiwari 
We still can't use NODE_ENV=staging RAILS_ENV=staging (Or just RAILS_ENV=staging)?

What would be the expected way to provide a different webpacker config in custom environment ?

The typical use case would be using different JS API keys.

Currently I'm using different sections in `webpacker.yml` to set the values then the DefinePlugin like this in `config/environment.js`

```js
environment.plugins.append('Define', new webpack.DefinePlugin({
  "GA_PROPERTY_ID":        JSON.stringify(config.ga_property_id),
  ....
}));
```

gauravtiwari:
@Systho Sorry about that, I am working on a PR to fix  this.

Systho:
No problem, It's great that you're aware of the problem. 
Since this issue is closed I was mostly wondering whether I was doing something wrong. That's why I provided my use case :)

Keep up the good work and thank you !

PR #1248, opened 2018-02-05
Add CSS Module support.

tai2:
Treat some file names like `.module.css` as CSS Modules.

This naming convention is adopted by some other OSSs such [create-react-app](https://github.com/facebook/create-react-app/pull/2285) or  [poi](https://poi.js.org/#/home?id=standalone-css-files).

PR #1305, opened 2018-02-28
Fix the supported suffix

dwightwatson:
Based upon #1248 it looks as though the supported suffix should be `module` singular, rather than plural. It looks correct in other parts of the documentation, just not here.


gauravtiwari:
thanks 👍

PR #1345, opened 2018-03-13
Freeze lockfile when installing in production mode

maschwenk:
If yarn installing is generating a lockfile diff that's a problem

### DIFF
--- a/lib/tasks/webpacker/yarn_install.rake
+++ b/lib/tasks/webpacker/yarn_install.rake
@@ -1,6 +1,6 @@
 namespace :webpacker do
   desc "Support for older Rails versions. Install all JavaScript dependencies as specified via Yarn"
   task :yarn_install do
-    system "yarn install --no-progress --production"
+    system "yarn install --no-progress --frozen-lockfile --production"
   end
 end


maschwenk:
@gauravtiwari any opinions on this?

guilleiguaran:
Note that this rake task is invoked only in the older versions of Rails that don't include a yarn install task by default so we need to fix also the task included in Rails

maschwenk:
Already filed! https://github.com/rails/rails/pull/32243

cdmo:
> Note that this rake task is invoked only in the older versions of Rails (< 5.1) that don't include a yarn install task by default so we need to fix also the task included in Rails

Doesn't this task run no matter what? Because of the code at the bottom of `compile.rake`:

```rb
# Compile packs after we've compiled all other assets during precompilation
if Rake::Task.task_defined?("assets:precompile")
  skip_webpacker_precompile = %w(no false n f).include?(ENV["WEBPACKER_PRECOMPILE"])
  enhance_assets_precompile unless skip_webpacker_precompile
else
  Rake::Task.define_task("assets:precompile" => ["webpacker:yarn_install", "webpacker:compile"])
end
```

PR #1349, opened 2018-03-14
fix for envs not specified in default configuration

haffla:
I have a staging environment in my config. There is no equivalent in the default configuration.
The PR adds a check that prevents the compilation process to crash (`delete null`).

Cheers


haffla:
Couple of hours earlier resolved by #1350

PR #1351, opened 2018-03-14
Ignore `/log` directory from git

tricknotes:
`bundle exec rake test` generated `log` directory.
However we want to ignore it.


gauravtiwari:
thanks 🍰

PR #1353, opened 2018-03-14
Add extension note to `webpacker.yml`.

### LINKED ISSUE #1348: [Suggestion] Add warning/note comment to webpacker.yml about extension order.
I'm using BootstrapVue in a project, and with Webpacker v3.2.1 I didn't have any problems when the extensions were in alphabetical order inside `webpacker.yml`.

However, after upgrading to Webpacker v3.2.2, some of the BootstrapVue components broke because Webpack was loading .css files before .js files. I'm not sure what changed between Webpacker v3.2.1 and v3.2.2 to suddenly have this issue, but reverting the order of the extensions to the default order from a fresh Rails project, the BootstrapVue components are working again.

My suggestion is to add a comment before the extensions list in `webpacker.yml` warning/noting that extension order matters.

```yaml
# Note: extension order is important.
extensions:
  - .js
  # ...
```

jpickwell:
Fix #1348

### DIFF
--- a/lib/install/config/webpacker.yml
+++ b/lib/install/config/webpacker.yml
@@ -13,6 +13,7 @@ default: &default
   # Reload manifest.json on all requests so we reload latest compiled packs
   cache_manifest: false
 
+  # Note: extension order is important.
   extensions:
     - .js
     - .sass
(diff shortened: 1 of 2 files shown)


swrobel:
Maybe a link to the correct order would be helpful? Not sure what it's providing to the user just to say order is "important"...

jpickwell:
I wouldn't call it "correct". You mean the default? A link could be useful.

gauravtiwari:
Agree, adding just this doesn't quite help. Let's document this in Readme instead. By default, webpacker ships extensions in the order of priority.

PR #1350, opened 2018-03-14
Allow custom env for Webpacker

tricknotes:
The Rails defined custom env is allowed by https://github.com/rails/webpacker/pull/1304.
However Webpacker custom env is not allowed for now.

```
/Users/tricknotes/dummy-app/node_modules/@rails/webpacker/package/config.js:16
    delete defaults.extensions
           ^

TypeError: Cannot convert undefined or null to object
```

This commit allows using user defined env for Webpacker.

PR #1355, opened 2018-03-15
Support to extend default webpacker env for non-standard env

tricknotes:
Currently, the non-standard env is allowed to use but any webpacker default config is ignored.
Especially, the developer who use "staging" expects that the environment is almost the same as "production".
    
This commit allows to specify the default webpacker env to be extended by non-standard env.
    
If someone want to extend webpacker env without this commit, the following change is required.

``` diff
diff --git a/config/webpack/staging.js b/config/webpack/staging.js
index d7346478..77bb3afa 100644
--- a/config/webpack/staging.js
+++ b/config/webpack/staging.js
@@ -1,3 +1,4 @@
-const environment = require('./environment');
+const Environment = require('@rails/webpacker/package/environments/production');
+const environment = new Environment();
    
module.exports = environment.toWebpackConfig();
```

This PR depends on https://github.com/rails/webpacker/pull/1350 due to support for non-standard env :<

### DIFF
--- a/README.md
+++ b/README.md
@@ -432,6 +432,14 @@ by adding new paths to `watched_paths` array. This is much like Rails' `autoload
 Webpacker::Compiler.watched_paths << 'bower_components'
 ```
 
+## Non standard ENV
+
+If you want to use non-standard ENV (such as "staging"), you can specify the webpacker standard ENV to be extended.
+
+``` yml
+staging:
+  extend_default_env: production
+```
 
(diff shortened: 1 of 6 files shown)


gauravtiwari:
Thanks @tricknotes for your contribution. Just merged #1359, which solves this issue.

tricknotes:
Ah, right!
Thanks, it makes sense. :-)

PR #1359, opened 2018-03-15
Detangle NODE_ENV and RAILS_ENV

gauravtiwari:
In earlier versions of webpacker we have used Rails env as fallback env, which isn't ideal because both `NODE_ENV` and `RAILS_ENV` has different meaning for different environments. 

For example: setting `NODE_ENV=production` means the webpacker will compile assets in production mode however at same time `RAILS_ENV` can be staging. Webpack and node libraries only understands development and production environments. However, it's normal to have many custom `RAILS_ENV`. 

This PR makes `RAILS_ENV` and `NODE_ENV` independent of each other for clarity. 

------------

Examples: 

```bash
# will compile in production mode but will use staging settings from webpacker.yml (if present or else fallback to production config)
RAILS_ENV=staging bundle exec rails webpacker:compile
# will use foo env settings from webpacker.yml but will compile in production mode
RAILS_ENV=foo bundle exec rails webpacker:compile
```

By default binstubs will use `NODE_ENV=development` however rake compile task will compile in production mode: 

```bash
# Compiles in development mode by default
./bin/webpack
./bin/webpack-dev-server
```

```bash
# Compiles in production mode
NODE_ENV=production ./bin/webpack
NODE_ENV=production ./bin/webpack-dev-server
```

```bash
# compiles in production mode by default (unless NODE_ENV is supplied) - same as sprockets
bundle exec rails assets:precompile
bundle exec rails webpacker:compile
```

### TODO:

- [x] Fix tests


lib/webpacker/dev_server_runner.rb
```diff
@@ -14,14 +14,14 @@ def run
     private
       def load_config
         @config_file = File.join(@app_path, "config/webpacker.yml")
-        dev_server = YAML.load_file(@config_file)[ENV["NODE_ENV"]]["dev_server"]
+        dev_server = YAML.load_file(@config_file)[ENV["RAILS_ENV"]]["dev_server"]
```
Systho: THIS ! 
Thank you very much for having thought of this :)

tricknotes:
@gauravtiwari 
This PR is pretty awesome!
Could you release a new version with this change?

gauravtiwari:
Made a new release 👍

PR #1364, opened 2018-03-18
Added config for react-hot-loader to documentation

raldred:
Out the box HMR is great, but with react-hot-leader things are even better.
Added the config required to maintain react state between HMR patches.

### DIFF
--- a/docs/webpack-dev-server.md
+++ b/docs/webpack-dev-server.md
@@ -23,8 +23,27 @@ Checkout this guide for more information:
 
 - https://webpack.js.org/configuration/dev-server/#devserver-hot
 
-To support HMR with React you would need to add `react-hot-loader`. Checkout this guide for
-more information:
+To support HMR with React you would need to add `react-hot-loader`.
+
+Update your development.js to look like this following:
+
+```
+const environment = require('./environment')
+const merge = require('webpack-merge')
+
+const babelLoader = environment.loaders.get('babel').use.find(el => el.loader === 'babel-loader')
+babelLoader.options = merge(babelLoader.options, { plugins: ['react-hot-loader/babel'] })
+
+const entries = environment.entry
+Object.keys(entries).forEach((entry) => {
+  entries.set(entry, ['react-hot-loader/patch', entries[entry]])
+});
+
+module.exports = environment.toWebpackConfig()
+
+```
+
+Checkout this guide for more information:
 
 - https://gaearon.github.io/react-hot-loader/getstarted/
 


renchap:
This is outdated. react-hot-loader 4 has been released recently and makes it much easier, you only need to add a line in your `.babelrc` and mark your components as hot loadable: https://github.com/gaearon/react-hot-loader#getting-started

AirWick219:
Not sure if that just me ... But I tried react-hot-loader 4 and added that line in ```.babelrc``` along with wrapping my root component with hot. Then got the following error once i fire up the ```./bin/webpack-dev-server``` with rails server

```
Module build failed: ReferenceError: Unknown plugin "react-hot-loader/babel"
at /node_modules/babel-core/lib/transformation/file/options/option-manager.js:180:17
```

gauravtiwari:
@AirWick219 Have you installed react-hot-loader? `yarn add react-hot-loader`

AirWick219:
Yes .. i did .. That was the weird part .... I followed from https://github.com/gaearon/react-hot-loader#getting-started

gauravtiwari:
@raldred Sorry for the late response. With latest webpack, you don't need to add react hot loader (AFAIK). It simply works with webpack HMR API - https://webpack.js.org/guides/hot-module-replacement/

We should remove the reference to react hot loader.

gauravtiwari:
> react-hot-loader is still needed with Webpack 4

Are you sure? Because the above works for me without `react-hot-loader` since Webpack v3.x. Obviously, react-hot-loader has nice API for making modules hot reloadable but it's just a HOC. 

hot-reload

renchap:
Yes. Your JS code is hot-reloaded, but your components are thrown away and re-created, meaning you loose all your state, which is not practical in any non-trivial app.

raldred:
Yes, maintaining state when hot reloading is really important like you say @renchap

gauravtiwari:
Thanks, gotcha 👍 (haven't used HMR much)

gauravtiwari:
Were you able to get react hot loader to work per documentation @raldred?

raldred:
@gauravtiwari yes, it's much simpler now, although I also use [webpacker-react](https://github.com/renchap/webpacker-react) to simplify rendering so it's slightly different, my pack looks like....

```
import { hot } from 'react-hot-loader';
import ShowcaseSlider from 'components/showcase_slider';
import OwnersForm from 'components/owners_form/index';
import WebpackerReact from 'webpacker-react';

WebpackerReact.setup({ ShowcaseSlider: hot(module)(ShowcaseSlider) });
WebpackerReact.setup({ HubspotForm: hot(module)(HubspotForm) });
```

gauravtiwari:
Awesome. Thanks for sharing 👍

Issue #1321, opened 2018-03-05
Heroku: can't find executable webpack for gem webpacker (Gem::Exception)


jonhue:
After upgrading to the master branch of this repository, I get the following error when compiling on Heroku:

```
       Compiling…
       Compilation failed:
       /tmp/build_d4da645e5840bc32169cdef3aecaed44/jonhue-jonhue-97c0597/vendor/bundle/ruby/2.3.0/gems/bundler-1.15.2/lib/bundler/rubygems_integration.rb:432:in `block in replace_bin_path': can't find executable webpack for gem webpacker (Gem::Exception)
       from /tmp/build_d4da645e5840bc32169cdef3aecaed44/jonhue-jonhue-97c0597/vendor/bundle/ruby/2.3.0/gems/bundler-1.15.2/lib/bundler/rubygems_integration.rb:463:in `block in replace_bin_path'
       from ./bin/webpack:29:in `<main>'
 !
 !     Precompiling assets failed.
```

Jehops:
We are also seeing this without Heroku:
```
rails assets:precompile
...
/usr/local/lib/ruby/gems/2.4/gems/bundler-1.16.1/lib/bundler/rubygems_integration.rb:458:in `block in replace_bin_path': can't find executable webpack for gem webpacker (Gem::Exception)
        from /usr/local/lib/ruby/gems/2.4/gems/bundler-1.16.1/lib/bundler/rubygems_integration.rb:489:in `block in replace_bin_path'
        from ./bin/webpack:17:in `<main>'
```
I notice the FreeBSD package, `rubygem-webpacker-rails5` starting with version 3.3.0 no longer installs `bin/webpacker` and `bin/bin/webpack-dev-server` under `$PREFIX`.

gauravtiwari:
Please run `bundle exec rails webpacker:binstubs` and commit binstubs to heroku.

vishthemenon:
I had the same issue as well when upgrading to webpacker 3.3.0. You need regenerate the binstubs for webpacker (and webpack-dev-server if you are using hot reloading in development).

Jehops:
Thanks.  Could you elaborate on what command need to be run to fix our `rubygem-webpacker-rails5` package on FreeBSD (no Heroku)?  I assume we need to run something to generate those two executable scripts that are now missing.

jonhue:
@gauravtiwari I did re-generate the binstubs using the `:install` generator. Your solution worked though. Any chance, they're not generating the same files?

[...]

PR #1365, opened 2018-03-20
Add troubleshooting docs for "can't find executable webpack exception"

jdlubrano:
After I upgraded the Webpacker gem in a Rails 5.1 project from 3.0.2 to 3.3.1, I saw the following error locally when I would run `RAILS_ENV=test NODE_ENV=test bin/webpack`:

```
/Users/<me>/.gem/ruby/2.4.2/gems/bundler-1.16.1/lib/bundler/rubygems_integration.rb:458:in `block in replace_bin_path': can't find executable webpack for gem webpacker (Gem::Exception)
	from /Users/<me>/.gem/ruby/2.4.2/gems/bundler-1.16.1/lib/bundler/rubygems_integration.rb:489:in `block in replace_bin_path'
	from bin/webpack:17:in `<main>'
```

I was hoping to add some troubleshooting documentation to save someone some time in the future.

Credit where credit is due, I actually found the solution to my problem in [another issue in this repo](https://github.com/rails/webpacker/issues/1321#issuecomment-370526887).

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -117,3 +117,12 @@ environment.plugins.append('ContextReplacement',
   )
 )
 ```
+
+# Rake assets:precompile or bin/webpack fails. Can't find executable webpack for gem webpacker (Gem::Exception)
+
+This can happen if you need to update your webpack and webpack-dev-server
+binstubs.  You can update your binstubs by running:
+
+```
+bundle exec rails webpacker:binstubs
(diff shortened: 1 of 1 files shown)


rossta:
@jdlubrano There are breaking changes documented in the changelog. https://github.com/rails/webpacker/blob/master/CHANGELOG.md#breaking-changes

Issue #956, opened 2017-10-16
Allow ERB in config/webpacker.yml


robertgrimm:
Any chance we could get ERB support in the config/webpacker.yml file?  I'd like to dynamically set resolved_paths.

gauravtiwari:
Unfortunately no, since we use same yml file for npm package as well, which doesn't parse erb like Ruby.

gauravtiwari:
>  I'd like to dynamically set resolved_paths.

Why you would like do this?

robertgrimm:
So I can support a large Rails engine we use.  The engine is installed as a gem and thus I'd want to use the installed gem path in production and a relative path in development.  My only option right now is to hard-code the gem's installed path, which includes the gem's version.  It would be nice to dynamically determine that so I don't have to update this file every time my engine's version is updated.

robertgrimm:
Looks like wildcards in resolved_paths aren't supported right now.  I was able to get it working by hacking `getModulePaths` in `package/environment.js` to run the resolved_paths elements through `sync`.  This could be a useful thing to add to webpacker.

robertgrimm:
I have a much better solution now that doesn't require any modifications to the webpacker gem.  In case future readers are interested, I wound up modifying `resolve.modules` in `config/webpack/production.js`.  That way, I can use `sync` to get the path of my Rails engine.

shlima:
@robertgrimm could you please provide an example of yours `config/webpack/production.js` ?

robertgrimm:
@shlima Sure

```JavaScript
const environment = require('./environment')                                                                                                                                                                               
const { sync } = require('glob')                                                                                                                                                                                           
                                                                                                                                                                                                                           
var env = environment.toWebpackConfig()                                                                                                                                                                                    
                                                                                                                                                                                                                           
// Add the installed myengine gem path to resolve.modules.  This is done                                                                                                                                                   
// here, instead of in config/webpacker.yml, because we need to use 'sync'                                                                                                                                                 
// to get the installed path.  The alternative is to use config/webpacker.yml                                                                                                                                              
// and hard-code the ruby and myengine gem version in the path which                                                                                                                                                       
// isn't desirable.                                                                                                                                                                                                        
const myenginepath = '/usr/local/lib/ruby/gems/*/gems/myengine-*/app/javascript';                                                                                                                                          
env.resolve.modules = env.resolve.modules.concat(sync(myengine))                                                                                                                                                           
                                                                                                                                                                                                                           
[...]

PR #1370, opened 2018-03-23
Support gem assets importing

rhyzx:
```
// app/javascript/packs/application.js
import 'rails-ujs'
import 'activestorage'
import 'turbolinks'
```

Related issues #956

### DIFF
--- a/lib/install/bin/webpack
+++ b/lib/install/bin/webpack
@@ -6,6 +6,7 @@ ENV["NODE_ENV"]  ||= ENV["NODE_ENV"] || "development"
 require "pathname"
 ENV["BUNDLE_GEMFILE"] ||= File.expand_path("../../Gemfile",
   Pathname.new(__FILE__).realpath)
+ENV["RAILS_LOAD_PATH"] = $LOAD_PATH.join(";")
 
 require "rubygems"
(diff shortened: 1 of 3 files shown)


rossta:
Could you elaborate on the use case? Each of the examples you listed are available on NPM, so you could add them via `yard add <package>`, no?

rhyzx:
@rossta This ensures that the gem and assets versions are consistent and they can be used in the same way as the sprockets.

gauravtiwari:
@rhyzx You could use npm packages with sprockets now so I would say just install these via yarn and use them consistently everywhere (unless I am missing something).

rossta:
@rhyzx We add selected `node_modules` directories to the Sprockets load path (as opposed to all node_modules which would add too many unneeded directories to the search space). Here's an example for `foundation-sites`, in which case, we're importing foundation javascript in Webpack and foundation scss files through Sprockets:

```ruby
# config/initializers/assets.rb
Rails.application.config do
  # ...

  config.assets.paths << Rails.root.join("node_modules", "foundation-sites", "scss")
end
```

rhyzx:
@rossta thanks, it's also a viable option.

PR #1376, opened 2018-03-23
Add image_pack_tag helper

maschwenk:
I migrated my project to use packs for static images and found that image tag references because very cumbersome and wordy. I can understand if people think this belongs in per-project helpers rather than in the library, but figured I'd open a PR if that wasn't the case.

Will add tests/documentation if we decide to move forward

### DIFF
--- a/lib/webpacker/helper.rb
+++ b/lib/webpacker/helper.rb
@@ -33,6 +33,22 @@ module Webpacker::Helper
     end
   end
 
+  # Creates a image tag that references the named pack file.
+  # This will use asset_pack_path internally, so most of their behaviors will be the same.
+  #
+  # Example:
+  #
+  #   # In development mode with hot module replacement:
+  #   <%= image_pack_tag 'logo.png' %> # => nil
+  #
+  #   # In production mode:
+  #   <%= image_pack_tag 'logo.png' %> # => "http://example.com/packs/logo-1016838bab065ae1e122.png"
+  def image_pack_tag(name, **options)
+    unless Webpacker.dev_server.running? && Webpacker.dev_server.hot_module_replacing?
+      image_tag(asset_pack_path(name), **options)
+    end
+  end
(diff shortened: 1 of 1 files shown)


gauravtiwari:
@maschwenk Thanks, `asset_pack_path` doesn't work?

maschwenk:
It's not hugely critical, I just found myself writing
```
= link_to image_tag(asset_pack_path('my_iage.png')), root_path
```
which seems really wordy

maschwenk:
You merged before I could write a test! 🤣

guilleiguaran:
@maschwenk please send a PR!!

lib/webpacker/helper.rb
```diff
@@ -33,6 +33,22 @@ def asset_pack_url(name, **options)
...
+  #
+  # Example:
+  #
+  #   # In development mode with hot module replacement:
+  #   <%= image_pack_tag 'logo.png' %> # => nil
+  #
+  #   # In production mode:
+  #   <%= image_pack_tag 'logo.png' %> # => "http://example.com/packs/logo-1016838bab065ae1e122.png"
+  def image_pack_tag(name, **options)
+    unless Webpacker.dev_server.running? && Webpacker.dev_server.hot_module_replacing?
```
gauravtiwari: @maschwenk image tag doesn't need to respect hot reloading because images aren't hot reloaded.

maschwenk: will adjust in PR with specs

guilleiguaran:
@maschwenk I rushed to press the merge button on this.

I've reverted this meanwhile, feel free to send a new PR with fixes and tests and ping me and I'll review/merge it 😊

lib/webpacker/helper.rb
gauravtiwari: thanks 👍

PR #1385, opened 2018-03-27
Remove version specification of coffeescript

ohbarye:
Hello.

Is there any reason to stick CoffeeScript's version to 1.12.7? If nothing, it's better to remove the version specification. 

Also `coffee-loader` already supports CoffeeScript 2.
ref: https://github.com/webpack-contrib/coffee-loader/pull/30

### DIFF
--- a/lib/install/coffee.rb
+++ b/lib/install/coffee.rb
@@ -20,6 +20,6 @@ copy_file "#{__dir__}/examples/coffee/hello_coffee.coffee",
   "#{Webpacker.config.source_entry_path}/hello_coffee.coffee"
 
 say "Installing all Coffeescript dependencies"
-run "yarn add coffeescript@1.12.7 coffee-loader"
+run "yarn add coffeescript coffee-loader"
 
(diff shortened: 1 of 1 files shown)


ytbryan:
cc @gauravtiwari

gauravtiwari:
@ohbarye Thanks, the idea behind coffeescript was that by default it will install 1.x to avoid any surprises and folks who want to use 2.x can install 2.x using `yarn add coffeescript` since it's a peer dependency. 

2.x also requires setting up babel loader if I am not mistaken?

ohbarye:
@gauravtiwari 
> 2.x also requires setting up babel loader if I am not mistaken?

In my understanding, to be strict, it's not mandatory because the official site says below.

> For older browsers or older versions of Node, however, transpilation is required. 
> http://coffeescript.org/#compatibility

Having said that, considering the current situation, many folks would need traspilation in most cases. So I believe the idea you clarified is fair enough! 

Thanks for taking your time for this.

PR #1402, opened 2018-04-03
Drop `"uglify": true` from babel targets

ojab:
uglify-es (that supports es6) is used since 3.2.1, so we shouldn't downgrade js features to the ones supported by non-es uglify.

### DIFF
--- a/lib/install/config/.babelrc
+++ b/lib/install/config/.babelrc
@@ -3,8 +3,7 @@
     ["env", {
       "modules": false,
       "targets": {
-        "browsers": "> 1%",
-        "uglify": true
+        "browsers": "> 1%"
       },
       "useBuiltIns": true
     }]


danielpowell4:
At very least with this PR, it would be better to use `"forceAllTransforms": true,` over `"uglify": true`

Reference:
https://babeljs.io/docs/plugins/preset-env/#targetsuglify

ojab:
This option is not available in `babel-preset-env-1.x` (which is used right now).
I assume that `babel-preset-env-2.0` w/ `"uglify"` removed and `"forceAllTransforms"` added will be released along `babel-7` and this option basically mirrors `"uglify": true`, downgrading js features to es5.

gauravtiwari:
@ojab Please correct me if I misunderstood the change, from plugin docs page, `uglify` option is available in 1.x but deprecated and replaced with `forceAllTransforms` in 2.x. We are still using 1.6.1 so the `uglify` option is still valid.

Anyway, I think what you want to change here is remove uglify option since we support es6 syntax by default right?

ojab:
Yep, correct.

PR #1408, opened 2018-04-05
Add changelog URI to gemspec

greysteil:
In line with the approach taken in [Rails](https://github.com/rails/rails/blob/master/activemodel/activemodel.gemspec#L23-L26).

Makes it easier for automated tools like Dependabot to get the right changelog (since the one on the 3.4-stable branch is now different to the one on master.


gauravtiwari:
thanks @greysteil

PR #1412, opened 2018-04-06
Update check_binstubs.rake to do as what it says

ytbryan:
"Verifies that webpack & webpack-dev-server are present."

Stumble onto this while fixing someone's project


gauravtiwari:
@ytbryan Don't think we need to check for the dev server - we reverted this a while back since folks removed this from the production environment. `webpack-dev-server` is a optional binstub.

ytbryan:
I see. thanks for the clarification

PR #1414, opened 2018-04-08
Increase yarn install timeout

pawurb:
Hi. I keep running into an issue described here https://github.com/yarnpkg/yarn/issues/4890 .

Increasing the timeout resolved a problem.

### DIFF
--- a/lib/tasks/webpacker/yarn_install.rake
+++ b/lib/tasks/webpacker/yarn_install.rake
@@ -1,6 +1,6 @@
 namespace :webpacker do
   desc "Support for older Rails versions. Install all JavaScript dependencies as specified via Yarn"
   task :yarn_install do
-    system "yarn install --no-progress --frozen-lockfile --production"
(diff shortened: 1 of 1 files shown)


lib/tasks/webpacker/yarn_install.rake
```diff
@@ -1,6 +1,6 @@
 namespace :webpacker do
   desc "Support for older Rails versions. Install all JavaScript dependencies as specified via Yarn"
   task :yarn_install do
-    system "yarn install --no-progress --frozen-lockfile --production"
+    system "yarn install --no-progress --frozen-lockfile --production --network-timeout 1000000"
```
abhinavmsra: How about making the rake task accept arguments and appending them to the `system` call. 

Would be more flexible IMO.

pawurb: This fails during deploy using a buildpack. I am not sure if it is possible to specify a concrete rake task arguments when using a buildpack.

PR #1415, opened 2018-04-09
Enable devserver proxy for custom environments

mchalandon:
It should be usefull to enable the devserver proxy for custom environements ; for projects with severall development environements.

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -4,15 +4,13 @@ require "rails"
 require "rails/test_help"
 require "byebug"
 
+require_relative "test_app/config/environment"
+
 Rails.env = "production"
 
-# Webpacker.instance paths hacks need to be done before application initialization
-require "webpacker"
 Webpacker.instance = Webpacker::Instance.new \
   root_path: Pathname.new(File.expand_path("test_app", __dir__)),
   config_path: Pathname.new(File.expand_path("./test_app/config/webpacker.yml", __dir__))
-
-require_relative "test_app/config/environment"
 
 class Webpacker::Test < Minitest::Test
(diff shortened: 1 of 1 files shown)


lib/webpacker/dev_server.rb
```diff
@@ -60,6 +60,6 @@ def fetch(key)
     end
 
     def defaults
-      config.send(:defaults)[:dev_server]
+      config.send(:defaults)[:dev_server] || {}
```
jasl: what about `config.dig(:defaults, :dev_server)`?

mchalandon: config is not a hash, but a Webpacker::Configuration which has a 'defaults' private method.

lib/webpacker/railtie.rb
```diff
@@ -46,7 +46,9 @@ class Webpacker::Engine < ::Rails::Engine
   end
 
   initializer "webpacker.proxy" do |app|
-    if Webpacker.config.dev_server.present?
+    insert_middleware = Webpacker.config.dev_server.present? rescue nil
+    insert_middleware ||= false
+    if insert_middleware
```
jasl: `nil` is equal to `false` in `if` statement
image

lib/webpacker/dev_server.rb
jasl: oh i see

lib/webpacker/railtie.rb
mchalandon: you're right ! thanks 
ed6743a170242e7611795a21b9581b7992a6d061

PR #1416, opened 2018-04-10
Rails webpacker:info helps to print out environment info to aid in debugging 

ytbryan:
Will you want this? 

Try: 
`rails webpacker:info`

Output: 

screenshot 2018-04-10 12 03 56


gauravtiwari:
thanks  @ytbryan 🍰

PR #1417, opened 2018-04-10
Update references in README to Webpack 4.x.x

returnlytom:


gauravtiwari:
Thanks, @returnlytom but let's wait until 4.x stable is released :)

PR #1421, opened 2018-04-11
From the open3 docs:

cshupp1:
# Open3.capture3 captures the standard output and the standard error of a command.
  #
  #   stdout_str, stderr_str, status = Open3.capture3([env,] cmd... [, opts])

  The order is fixed.


gauravtiwari:
thanks @cshupp1

PR #1422, opened 2018-04-12
Remove extra spaces in template.rb

szTheory:
Currently it adds an extra indent level in a new webpacker Rails project's development.rb and production.rb, so the `config` statement doesn't line up with the others.


gauravtiwari:
thanks

PR #1425, opened 2018-04-12
Use correct protocol

Systho:
Currently we can run webpacker in either http or https mode.
But if we run webpacker in `http` mode but access it through a reverse proxy handling `https`, then it will break because the reverse proxy will have set `HTTP_X_FORWARDED_PROTO` to `https` and webpack-dev-server will try to understand the request as a secured one.

We must override `X_FORWARDED_PROTO` in the same way as we override the other headers.

In the meantime I'm using this initializer as a workaround  doing exactly the same work as this PR :

```ruby
# config/initializers/webpacker.rb
require "webpacker/dev_server_proxy.rb"

module Webpacker::DevServerProxy::ProtoOverride
  def perform_request(env)
      env["HTTP_X_FORWARDED_PROTO"] = Webpacker.dev_server.protocol
      super(env)
  end
end
Webpacker::DevServerProxy.include Webpacker::DevServerProxy::ProtoOverride

```


chrisjohnson:
Glad to see I'm not crazy! Funny that we both ran into the same issue around the same time

gauravtiwari:
thanks @Systho

PR #1428, opened 2018-04-13
Include `RAILS_RELATIVE_URL_ROOT` environment variable in `publicPath`.

### LINKED ISSUE #1357: manifest.json doesn't include relative_url_root.
For production sandbox deployment I have the app "mounted" under a subdirectory; e.g., `example.com/subdir`.

Currently, webpacker doesn't include the `RAILS_RELATIVE_URL_ROOT` in the `manifest.json`, even if it's set.

jpickwell:
Fix #1357.

### DIFF
--- a/package/config.js
+++ b/package/config.js
@@ -20,14 +20,14 @@ if (isArray(app.extensions) && app.extensions.length) delete defaults.extensions
 
 const config = deepMerge(defaults, app)
 config.outputPath = resolve('public', config.public_output_path)
-config.publicPath = `/${config.public_output_path}/`
 
+let publicPath = `/${config.public_output_path}/`
 // Add prefix to publicPath.
 if (process.env.RAILS_RELATIVE_URL_ROOT) {
-  config.publicPath = `/${process.env.RAILS_RELATIVE_URL_ROOT}${config.publicPath}`
+  publicPath = `/${process.env.RAILS_RELATIVE_URL_ROOT}${publicPath}`
 }
 
 // Remove extra slashes.
-config.publicPath = config.publicPath.replace(/(^\/|[^:]\/)\/+/g, '$1')
+config.publicPath = publicPath.replace(/(^\/|[^:]\/)\/+/g, '$1')
 
 module.exports = config
--- a/package/config.js
+++ b/package/config.js
@@ -20,6 +20,14 @@ if (isArray(app.extensions) && app.extensions.length) delete defaults.extensions
 
 const config = deepMerge(defaults, app)
 config.outputPath = resolve('public', config.public_output_path)
-config.publicPath = `/${config.public_output_path}/`.replace(/([^:]\/)\/+/g, '$1')
+config.publicPath = `/${config.public_output_path}/`
+
+// Add prefix to publicPath.
+if (process.env.RAILS_RELATIVE_URL_ROOT) {
+  config.publicPath = `/${process.env.RAILS_RELATIVE_URL_ROOT}${config.publicPath}`
+}
+
+// Remove extra slashes.
+config.publicPath = config.publicPath.replace(/(^\/|[^:]\/)\/+/g, '$1')
 
 module.exports = config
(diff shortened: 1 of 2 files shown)


package/config.js
```diff
@@ -20,6 +20,14 @@ if (isArray(app.extensions) && app.extensions.length) delete defaults.extensions
 
 const config = deepMerge(defaults, app)
 config.outputPath = resolve('public', config.public_output_path)
-config.publicPath = `/${config.public_output_path}/`.replace(/([^:]\/)\/+/g, '$1')
+config.publicPath = `/${config.public_output_path}/`
+
+// Add prefix to publicPath.
+if (process.env.RAILS_RELATIVE_URL_ROOT) {
+  config.publicPath = `/${process.env.RAILS_RELATIVE_URL_ROOT}${config.publicPath}`
```
gauravtiwari: Could you add some test for this, please?

jpickwell: I thought about that, but wasn't sure which test file to put them in? I was thinking of using the config test file, but wasn't sure because I didn't see anything in there for `publicPath`.

gauravtiwari: Ahh yes, but could you please add one :)

```js
/* global test expect, describe */

const { chdirTestApp, chdirCwd } = require('../utils/helpers')

chdirTestApp()

describe('Config', () => {
  beforeEach(() => jest.resetModules())
  afterAll(chdirCwd)

  test('public path', () => {
    delete process.env.RAILS_RELATIVE_URL_ROOT
    const config = require('../config')
    expect(config.publicPath).toEqual('/packs/')
  })

  test('public path with relative root', () => {
    process.env.RAILS_RELATIVE_URL_ROOT = '/foo'
    const config = require('../config')
    expect(config.publicPath).toEqual('/foo/packs/')
  })

  test('should return extensions as listed in app config', () => {
    const config = require('../config')
    expect(config.extensions).toEqual([
      '.js',
      '.sass',
      '.scss',
      '.css',
      '.module.sass',
      '.module.scss',
      '.module.css',
      '.png',
      '.svg',
      '.gif',
      '.jpeg',
      '.jpg'
    ])
  })
})
```

jpickwell: Will do. Thanks for the boilerplate.

gauravtiwari: I guess we need to set NODE_ENV in the test for consistent results: 

```js
 test('public path', () => {
    process.env.RAILS_ENV = 'development'
    delete process.env.RAILS_RELATIVE_URL_ROOT
    const config = require('../config')
    expect(config.publicPath).toEqual('/packs/')
  })

  test('public path with relative root', () => {
    process.env.RAILS_ENV = 'development'
    process.env.RAILS_RELATIVE_URL_ROOT = '/foo'
    const config = require('../config')
    expect(config.publicPath).toEqual('/foo/packs/')
  })
```

inkhrt:
I need your help

package/config.js
```diff
@@ -20,6 +20,14 @@ if (isArray(app.extensions) && app.extensions.length) delete defaults.extensions
...
+}
+
+// Remove extra slashes.
+config.publicPath = config.publicPath.replace(/(^\/|[^:]\/)\/+/g, '$1')
```
jpickwell: The `[^:]\/` part doesn't seem correct because if `public_output_path` contains a protocol, then `publicPath` would have a leading slash; e.g., (ignoring relative root) `public_output_path='https://cdn.com'` and `publicPath='/https://cdn.com/'`. The protocol would be preserved with the current RegExp, but that doesn't appear to be correct. Maybe I'm missing something.

package/config.js
jpickwell: I assume you meant `NODE_ENV` in your examples.

jpickwell: I assume you meant `NODE_ENV` in your examples.

jpickwell: Nope, you meant `RAILS_ENV`.

gauravtiwari: oh yes! I guess you figured it out 😉

package/config.js
gauravtiwari: Yeah, but at the moment Rails handles the CDN side of things so it's fine but you are right this needs more check if we expect the `public_output_path` to have a protocol as well (which we don't at the moment).

jpickwell: Sounds good.

PR #1440, opened 2018-04-21
add semicolons

coorasse:
Is pretty annoying to have linters complaining immediately after Webpacker setup :)
I'd suggest to terminate statements with semicolons.


gauravtiwari:
@coorasse Thanks for this but we have `no-semi` setting in eslint config for webpacker package.

Issue #1453, opened 2018-04-26
vue-loader v15 requires VueLoaderPlugin in webpack config


rlafranchi:
details: https://vue-loader.vuejs.org/migrating.html#migrating-from-v14

cc @ytbryan

ytbryan:
For anyone reading, the temporary solution is to downgrade to v14.2.2
 
Run: 

`yarn upgrade vue-loader@14.2.2`

@rlafranchi thanks for reporting!

bbugh:
A better temporary solution since `vue-loader` v15 fixes a bunch of bugs including the [css-out-of-order in production bug](https://github.com/vuejs/vue-loader/issues/808) is to add to environment.js

```javascript
// environment.js

// Required for vue-loader v15
const VueLoaderPlugin = require('vue-loader/lib/plugin')
environment.plugins.append(
  'VueLoaderPlugin',
  new VueLoaderPlugin()
)

module.exports = environment
```

gauravtiwari:
What's the permanent solution? 😄 Did we fix this in 4.x?

ytbryan:
Yup, it's fixed in 4.x but the user has  to choose the latest vue-loader at `rails webpacker:install:vue`. 
![screenshot 2018-10-06 11 54 41](https://user-images.githubusercontent.com/323891/46567120-719d2400-c95f-11e8-8f61-7b383598c687.png)

msencenb:
@bbugh solution seems to mostly work for me, although I also had to append a new rule into the webpack config that targets vue. But... that custom loader seems like it results in my vue components not respecting my .babelrc plugin of 'transform-class-properties', which causes my app to break on IE11 because of arrow functions. Not sure yet what to do about that. 

```
// Necesary configuration for vue-loader v15
const VueLoaderPlugin = require('vue-loader/lib/plugin');
environment.plugins.append(
    'VueLoaderPlugin',
    new VueLoaderPlugin()
);
environment.loaders.append('vue', {
    test: /\.vue$/,
    loader: 'vue-loader'
});
```

gauravtiwari:
Please try the latest RC, otherwise feel free to open a new issue with more details.

afuno:
Write this in the console:

```
yarn upgrade vue-loader@last
```

Next, choose version 14.2.3. Not 15 version, within the 14 version. In my case, this corrected the situation.

MelMacaluso:
[...]

PR #1459, opened 2018-04-26
yarn add vue-loader@14.2.2

ytbryan:
Prevent future issues by appending version to `yarn add vue-loader` until we get a permanent fix

Related to #1453


gauravtiwari:
@ytbryan Thank you ❤️ 

Is this issue only in 3.x? Or we have to update 4.x too?

ytbryan:
Affects both 3.x and 4.x of webpacker. 

For webpacker 4, we need to remove the appending of VueLoaderPlugin to environment.js

and apply `yarn upgrade vue-loader@14.2.2` to complete the fix.

PR #1462, opened 2018-04-28
Remove VueLoaderPlugin and yarn add vue-loader:14.2.2

ytbryan:
This is for webpacker 4. Related to https://github.com/rails/webpacker/issues/1453

Tested on localhost and it works. cc @gauravtiwari

### DIFF
--- a/lib/install/vue.rb
+++ b/lib/install/vue.rb
@@ -8,10 +8,6 @@ insert_into_file Rails.root.join("config/webpack/environment.js").to_s,
   "const { VueLoaderPlugin } = require('vue-loader')\n",
   after: "require('@rails/webpacker')\n"
 
-insert_into_file Rails.root.join("config/webpack/environment.js").to_s,
-  "environment.plugins.append('VueLoaderPlugin', new VueLoaderPlugin())\n",
-  before: "module.exports"
-
 say "Adding vue loader to config/webpack/environment.js"
 insert_into_file Rails.root.join("config/webpack/environment.js").to_s,
   "const vue =  require('./loaders/vue')\n",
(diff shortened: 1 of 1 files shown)


gauravtiwari:
@ytbryan Shouldn't we be using 15.0.0 instead? https://github.com/vuejs/vue-loader/releases & https://vue-loader.vuejs.org/migrating.html

ytbryan:
@gauravtiwari 

This PR is made as a temporary fix because vue-loader 15 and its new configuration are breaking the default vue pack even on webpacker 4. 

Ultimately, webpacker 4 should use vue-loader 15.

michaelize:
Any ETA on supporting vue-loader 15?

gauravtiwari:
@ytbryan Do you wanna explore how we can integrate v15 in Webpacker 4? I will take a look at this over the weekend.

ytbryan:
@gauravtiwari I tried. 

But even Vue-cli has not fully support webpack 4 and vue-loader 15 yet (just checked that vue-cli latest usess vue-loader 14). Nonetheless, if I figure out something, I will send a PR.

WoH:
I'd be very curious what the issues are you're running into, maybe I missed the description and you can point me towards it. I am using vue-loader 15.0.9, @rails/webpacker 4.0.0-pre.2 and the gem webpacker ~> 4.0.0.pre.pre.2 and I have not run into issues with the default pack as far as I can tell.
I checked the scoped styles and sass.
There should be changes for non-html templates (pug) and probably some non-standard styles + css modules, but a fresh installation of webpacker added the Plugin and the mini-css-extract rules needed.

gauravtiwari:
@WoH Thank you. Happy to make a PR? I guess the styles aren't working out of the box so we should fix that.

gauravtiwari:
Closing this until we have a proper solution for this.

PR #1466, opened 2018-04-29
Upgrade 3.x packages

gauravtiwari:


package.json
```diff
@@ -12,38 +12,38 @@
     "yarn": ">=0.25.2"
   },
   "dependencies": {
-    "babel-core": "^6.26.0",
-    "babel-loader": "^7.1.2",
```
swrobel: Why the hard version dependencies?

swrobel:
For some reason this is causing serious problems for me with yarn 1.6

https://github.com/yarnpkg/yarn/issues/5749#issuecomment-386335952

gauravtiwari:
@swrobel Just released 3.5.3 to fix this ^^

package.json
gauravtiwari: Sorry about this probably sneaked in while I was trying to update some packages between different branches.

PR #1467, opened 2018-04-30
Add PDF extension for file loader

dwightwatson:
This adds PDF to the test for the file loader, allowing PDF assets to be served through Webpacker. 

In my app everything goes through Webpacker - the only thing I have left in the old asset pipeline is PDF files. I think it makes sense for Webpacker to support serving these files out of the box as well, without needing to configure it on a per-app basis.

### DIFF
--- a/package/rules/file.js
+++ b/package/rules/file.js
@@ -2,7 +2,7 @@ const { join } = require('path')
 const { source_path: sourcePath } = require('../config')
 
 module.exports = {
-  test: /\.(jpg|jpeg|png|gif|tiff|ico|svg|eot|otf|ttf|woff|woff2)$/i,
+  test: /\.(jpg|jpeg|png|gif|tiff|ico|svg|eot|otf|ttf|woff|woff2|pdf)$/i,
   use: [
     {
       loader: 'file-loader',


webark:
@dwightwatson right now is it working for you to simply do..
```
const fileRule = environment.loaders.get('file');
fileRule.test = /\.(pdf|jpg|jpeg|png|gif|tiff|ico|svg|eot|otf|ttf|woff|woff2)$/i;
```
That's what I was going to do, and was wondering if you had ran into any issues.

dwightwatson:
Your solution is better than the one I came up with which was to add another file loader specifically for PDFs.

```js
const { source_path: sourcePath } = config
const { join } = require('path')

environment.loaders.append('pdf', {
    test: /\.(pdf)$/i,
    use: [
      {
        loader: 'file-loader',
        options: {
          name: '[path][name]-[hash].[ext]',
          context: join(sourcePath)
        }
      }
  ]
})
```

I'll switch over to your approach, but I still would like to see this ship out of the box.

gauravtiwari:
Will probably add an API for this so it's easier to add static file extensions through webpacker.yml

PR #1475, opened 2018-05-03
Honor BUNDLE_PATH if set in tasks

jfouchard:
If `$BUNDLE_PATH` is set the rake/rails binary will not be in the correct location and fail. The Ruby Docker images use this for instance.  I'm fairly new to Rails so there may be a better way to achieve this.

### DIFF
--- a/lib/tasks/installers.rake
+++ b/lib/tasks/installers.rake
@@ -17,11 +17,6 @@ namespace :webpacker do
   namespace :install do
     installers.each do |name, task_name|
       desc "Install everything needed for #{name}"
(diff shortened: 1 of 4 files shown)


gauravtiwari:
@jfouchard  Thanks, looks like this is removed in this commit: https://github.com/docker-library/ruby/commit/699a04311386ecc98ca242fc9bdee17fb4008863

jfouchard:
Thanks for the heads up about that change - I think the intention of the patch still stands `BUNDLE_BIN` is still supported according to the [docs](https://bundler.io/v1.16/bundle_config.html).

PR #1478, opened 2018-05-04
Match literal "." character instead of any char...

### LINKED ISSUE #1477: Example Stimulus application.js matches any single char before js extension, not "."
I believe in the regular expression contained here:

https://github.com/rails/webpacker/blob/89e4d602c29ca5cf89d7e634a7e901225d2e7a4f/lib/install/examples/stimulus/application.js#L5

we *mean* to match filenames ending in ".js" (as in, the literal "." character), rather than "js" preceded by any character.

Simple REPL example of the problem:

```javascript
> context = /.js$/
> context.test("hello_controller.js")  //=> true, as expected
> context.test("hello_controllerjs")  //=> also true, though not what new user likely expects

> context2 = /\.js$/
> context2.test("hello_controller.js")  //=> true
> context2.test("hello_controllerjs")  // => false
```

The latter form may lead to fewer unexpected issues for users.

bayprogrammer:
...in example application.js for Stimulus when specifying filename pattern in context. Fixes #1477.

### DIFF
--- a/lib/install/examples/stimulus/application.js
+++ b/lib/install/examples/stimulus/application.js
@@ -2,5 +2,5 @@ import { Application } from "stimulus"
 import { definitionsFromContext } from "stimulus/webpack-helpers"
 
 const application = Application.start()
-const context = require.context("controllers", true, /.js$/)
+const context = require.context("controllers", true, /\.js$/)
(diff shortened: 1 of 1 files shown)


gauravtiwari:
thanks 🍰

PR #1499, opened 2018-05-14
Default mode should be development for npm

euxn23:
In use from npm context, NODE_ENV is not set as default.
So default `mode` should be development on npm way, don't you think?
I appliciate if you consider.


gauravtiwari:
Thanks for the PR. Development mode is explicit mode and generally, we would want to compile in production mode.

Issue #1347, opened 2018-03-13
Slow Compile Times on Heroku


sunnyrjuneja:
This issue is an offshoot of https://github.com/rails/webpacker/issues/1319#issuecomment-372470188.

We're trying to debug why compile times are slow on Heroku. We are using https://github.com/sunnyrjuneja/wbspeedtest as a reproducible guide. 

Reproduce steps:
1. Git clone repo
2. `$ heroku create`
3. `$ heroku buildpacks:set heroku/ruby`
4. `$ heroku buildpacks:add --index 1 heroku/nodejs` This is necessary getting webpack working.

An application with no webpacker but both buildpacks added. Using commit: https://github.com/sunnyrjuneja/wbspeedtest/commit/db5fd721332fb3a80786ba13e9f6df170b2b1218

Result: Pretty fast.
```
λ →  git push heroku master
Counting objects: 6, done.
Delta compression using up to 4 threads.
Compressing objects: 100% (5/5), done.
Writing objects: 100% (6/6), 522 bytes | 522.00 KiB/s, done.
Total 6 (delta 3), reused 0 (delta 0)
remote: Compressing source files... done.
remote: Building source:
remote:
remote: -----> Node.js app detected
remote:
remote: -----> Creating runtime environment
remote:
remote:        NPM_CONFIG_LOGLEVEL=error
remote:        NODE_VERBOSE=false
remote:        NODE_ENV=production
remote:        NODE_MODULES_CACHE=true
remote:
remote: -----> Installing binaries
remote:        engines.node (package.json):  unspecified
remote:        engines.npm (package.json):   unspecified (use default)
remote:
remote:        Resolving node version 8.x...
remote:        Downloading and installing node 8.10.0...
remote:        Using default npm version: 5.6.0
remote:
remote: -----> Restoring cache
remote:        Skipping cache restore (not-found)
remote:
remote: -----> Building dependencies
remote:        Installing node modules (package.json)
remote:        up to date in 0.076s
remote:
remote: -----> Caching build
remote:        Clearing previous node cache
remote:        Saving 2 cacheDirectories (default):
remote:        - node_modules (nothing to cache)
remote:        - bower_components (nothing to cache)
remote:
remote: -----> Pruning devDependencies
remote:        Skipping because npm 5.6.0 sometimes fails when running 'npm prune' due to a known issue
remote:        https://github.com/npm/npm/issues/19356
remote:
remote:        You can silence this warning by updating to at least npm 5.7.1 in your package.json
[...]

PR #1508, opened 2018-05-19
POC: Parallel builds

kevincolten:
These tweaks allow you to enable parallel builds with https://github.com/trivago/parallel-webpack

You will have to manually separate your packs in your `config/webpack/environment.js`:
```js
const clone = require('clone');
const { environment } = require('@rails/webpacker');

module.exports = Object.keys(environment.entry).map(key => {
  const bundle = clone(environment);
  bundle.entry = {
    [key]: environment.entry[key]
  }
  return bundle;
})
```

and in your individual environment files (eg. `config/webpack/production.js`)
```js
const environment = require('./environment');
const { parse } = require('path');

module.exports = environment.map(env => {
  const path = `app/javascript/packs/${parse(Object.values(env.entry)[0]).base}`;
  return env.toWebpackConfig(path);
});
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -258,6 +258,7 @@ You can run following commands to upgrade Webpacker to the latest stable version
 
 ```bash
 bundle update webpacker
+rails webpacker:binstubs
 yarn upgrade @rails/webpacker --latest
 yarn add webpack-dev-server@^2.11.1
 


package/environments/base.js
```diff
@@ -111,7 +112,12 @@ module.exports = class Base {
     this.resolvedModules = getModulePaths()
   }
 
-  toWebpackConfig() {
+  toWebpackConfig(path) {
+    this.loaders = getLoaderList()
```
kevincolten: Had to reset these properties here. Something to do with the parallelization was stripping them of being ConfigObjects

guilleiguaran:
This is causing CI failures 😬

kevincolten:
Yeah, I’ll take a look. Wanted to gauge interest first

javan:
I'm hesitant to make changes just to support a 3rd-party webpack plugin. Could it be reframed as a more generalized feature like "Add support for compiling a single pack"?

drewB:
Bummed this didn't move forward.  I would hope would help with things like https://github.com/rails/webpacker/issues/1347

PR #1510, opened 2018-05-21
Add --show-config option to display computed webpack config

edslocomb:
Motivation:

Sometimes you just want to peek behind the curtain of Webpacker's config magic and see what it's telling webpack to do :).

This also dumps the computed Webpacker `environment`, because it can be quite helpful to know what that looks like, too.


edslocomb:
Any pointers on how to restart the CI build?   One config timed out after 10 minutes.

PR #1511, opened 2018-05-21
Reconcile RAILS_ENV and NODE_ENV

edslocomb:
### Motivation:
Webpacker picks a section from config/webpacker.yml based on RAILS_ENV.  But it determines which environment files from config/webpack/*.js will be used based on NODE_ENV.

Bad (or at least quite confusing) things will happen when NODE_ENV and RAILS_ENV are different.

It is especially easy to have them set to different things by setting one (but not both) to 'production' or 'test' on the command line, e.g.
```bash
$ RAILS_ENV=production ./bin/webpack
```
If a user doesn't otherwise have NODE_ENV set, this will use the `production:` section in webpacker.yml, but run the config/development.js setup file.

### Fix:

This PR makes sure RAILS_ENV and NODE_ENV are set to the same thing.  

If both are set, with different values, then RAILS_ENV wins*.  If neither is set, RACK_ENV is used, and if that's not set either, we default to 'development'.

\* js programmers coming to Rails are likely to learn about RAILS_ENV soon, while Rails programmers adopting webpacker are unlikely to learn about node.js or NODE_ENV from the front-end libraries/frameworks they've chosen.

### DIFF
--- a/lib/install/bin/webpack
+++ b/lib/install/bin/webpack
@@ -1,7 +1,9 @@
 #!/usr/bin/env ruby
 
-ENV["RAILS_ENV"] ||= ENV["RACK_ENV"] || "development"
-ENV["NODE_ENV"]  ||= "development"
+ENV["RAILS_ENV"] ||= ENV["RACK_ENV"]
+ENV["RAILS_ENV"] ||= ENV["NODE_ENV"]
+ENV["RAILS_ENV"] || "development"
+ENV["NODE_ENV"] = ENV["RAILS_ENV"]
 
 require "pathname"
 ENV["BUNDLE_GEMFILE"] ||= File.expand_path("../../Gemfile",
--- a/lib/install/bin/webpack-dev-server
+++ b/lib/install/bin/webpack-dev-server
@@ -1,7 +1,9 @@
 #!/usr/bin/env ruby
 
-ENV["RAILS_ENV"] ||= ENV["RACK_ENV"] || "development"
-ENV["NODE_ENV"]  ||= "development"
+ENV["RAILS_ENV"] ||= ENV["RACK_ENV"]
+ENV["RAILS_ENV"] ||= ENV["NODE_ENV"]
+ENV["RAILS_ENV"] || "development"
+ENV["NODE_ENV"] = ENV["RAILS_ENV"]
 
 require "pathname"
 ENV["BUNDLE_GEMFILE"] ||= File.expand_path("../../Gemfile",


gauravtiwari:
@edslocomb Thanks for this but the reason to keep them separate was to have more control over which environment one wants to load and execute from. After #1359, `RAILS_ENV` controls the settings from `webpacker.yml` and `NODE_ENV` controls, which Webpack config file to load. 

**For example: on staging**, you could have, `RAILS_ENV=staging and NODE_ENV=production`, which means it will load configuration settings from staging env but will run the production Webpack config to compile the packs. By default, one doesn't need to specifiy `NODE_ENV` at all. The compilation task is always run in production mode, whereas binstubs are run in development mode unless overridden elsewhere. 

The same applies for testing env as well so, `RAILS_ENV=test and NODE_ENV=development`

> I'm open to changing this such that if RAILS_ENV and NODE_ENV are both set in the user's environment, we leave them alone and assume they know what they're doing

Thank you. This seems like a good alternative but I am not sure if this might break custom environments like `staging` etc. where it's possible that `NODE_ENV` might not be set anywhere and it will try to use `staging` for loading webpack config. What do you think?

edslocomb:
I think it is preferable to have the program fail when RAILS_ENV is set to 'staging' and there is no userland staging.js file than to have the program set NODE_ENV to 'development' when the user sets RAILS_ENV to 'staging'.

(The rest of this is kind of speculative, feel free to ignore)

If I were designing this system, I would make NODE_ENV irrelevant to the end user, and use RAILS_ENV to select both the section from webpacker.yml and the supplementary js config file from config/webpack/.

If there were no section in webpacker.yml correstponding to the vaule of RAILS_ENV, I would use the 'default' section and print a warning to stderr.  If there were no file in webpack/ corresponding to the value of RAILS_ENV, I would use package/environments/base.js and print a warning to stderr.  I would not use lib/install/config/webpacker.yml to provide "default defaults".

I would not expose the internal configuration API of this gem/package to the user in config/webpack/*.js.  I would provide two entry points for those files; the default install would have no environment.js, just production.js and development.js, each of which would look like: 
```js
const webpackConfig = require('@rails/webpacker-development') // or -production
// dear user: webpackConfig is a valid webpack config object.  
// We have prepopulated it based on settings in webpacker.yml, plus some opinionated choices.  
// Modify it here, if you need more direct control.  See webpack docs.  
// Use --show-config to display the computed webpack config on the command line.
module.exports = webpackConfig
```
When adding a new environment, the user could then choose to require @rails/webpacker-development, @rails/webpacker-production, or (@rails/webpacker-base) when (if) setting up the new-environment.js file.  This would be the means of selecting between webpack's 'development' and 'production' --mode settings.  Realistically I'd probably also add a '-legacy' entrypoint for users who have invested in working with the exposed internal configuration API.

Issue #1502, opened 2018-05-16
Settings in config/webpack/test.js stopped working after of upgrade from 3.2 to 3.5


guilleiguaran:
I have expose loader set in `config/webpack/test.js` in a project with Webpacker 3.2:

```js
const environment = require('./environment')
const webpack = require('webpack')

// We must expose jQuery to the global object to be able to use `JQuery.active`
// in `wait_for_ajax` (spec/support/wait_for_ajax.rb)

environment.loaders.set('expose', {
  test: require.resolve('jquery'),
  use: {
    loader: 'expose-loader',
    options: 'jQuery'
  }
})

module.exports = environment.toWebpackConfig()
```

and I've modified the current configuration to match new conventions in 3.5 setting `NODE_ENV` and replacing `environment.loaders.set` with `environment.loaders.append`:

```js
process.env.NODE_ENV = process.env.NODE_ENV || 'development'

const environment = require('./environment')
const webpack = require('webpack')

// We must expose jQuery to the global object to be able to use `JQuery.active`
// in `wait_for_ajax` (spec/support/wait_for_ajax.rb)

environment.loaders.append('expose', {
  test: require.resolve('jquery'),
  use: {
    loader: 'expose-loader',
    options: 'jQuery'
  }
})

module.exports = environment.toWebpackConfig()
```

But after of upgrading, this loader stopped working.

Everything else in the project work as expected and compiling success without problems but I can't get this working, does someone has any pointers about this?

My `config/webpack/environment.js` is this:

```js
const { environment } = require('@rails/webpacker')
const webpack = require('webpack')

environment.plugins.append(
  'Provide',
  new webpack.ProvidePlugin({
    $: "jquery",
    jQuery: "jquery",
    "window.jQuery": "jquery"
  })
)

// resolve-url-loader must be used before sass-loader
environment.loaders.get('sass').use.splice(-1, 0, {
  loader: 'resolve-url-loader',
});

module.exports = environment
```

gauravtiwari:
@guilleiguaran Is ProvidePlugin not working either? Have made an example repo and everything seems to work fine: https://github.com/gauravtiwari/webpacker-provide-plugin

gauravtiwari:
expose loader issue might be related to this: https://github.com/webpack-contrib/expose-loader/issues/66

guilleiguaran:
@gauravtiwari I've cloned your app and added a `console.log` to every file under `config/webpack/` and looks like the `test.js` isn't loaded:

[...]

PR #1513, opened 2018-05-21
Include "test" in environment whitelist

schpet:
without this i don't think package/environments/test.js is ever loaded.

https://github.com/schpet/webpacker/blob/include-test-environment-in-whitelist/package/index.js#L13

#1502 possibly related?

### DIFF
--- a/package/__tests__/development.js
+++ b/package/__tests__/development.js
@@ -6,6 +6,10 @@ const { chdirTestApp, chdirCwd } = require('../utils/helpers')
 chdirTestApp()
 
 describe('Development environment', () => {
+  beforeEach(() => {
+    jest.resetModules()
+  })
+
   afterAll(chdirCwd)
 
   describe('toWebpackConfig', () => {
--- a/package/__tests__/production.js
+++ b/package/__tests__/production.js
@@ -6,6 +6,10 @@ const { chdirTestApp, chdirCwd } = require('../utils/helpers')
 chdirTestApp()
 
 describe('Production environment', () => {
+  beforeEach(() => {
+    jest.resetModules();
+  })
+
   afterAll(chdirCwd)
 
   describe('toWebpackConfig', () => {
@@ -13,6 +17,7 @@ describe('Production environment', () => {
 
     test('should use production config and environment', () => {
       process.env.RAILS_ENV = 'production'
+      process.env.NODE_ENV = 'production'
       const { environment } = require('../index')
 
       const config = environment.toWebpackConfig()
--- a/package/__tests__/staging.js
+++ b/package/__tests__/staging.js
@@ -6,6 +6,10 @@ const { chdirTestApp, chdirCwd } = require('../utils/helpers')
 chdirTestApp()
 
 describe('Custom environment', () => {
+  beforeEach(() => {
+    jest.resetModules()
+  })
+
   afterAll(chdirCwd)
 
   describe('toWebpackConfig', () => {
@@ -13,6 +17,7 @@ describe('Custom environment', () => {
 
     test('should use staging config and production environment', () => {
       process.env.RAILS_ENV = 'staging'
+      process.env.NODE_ENV = 'staging'
       const { environment } = require('../index')
 
       const config = environment.toWebpackConfig()
--- a/package/__tests__/test.js
+++ b/package/__tests__/test.js
@@ -6,6 +6,10 @@ const { chdirTestApp, chdirCwd } = require('../utils/helpers')
 chdirTestApp()
 
 describe('Test environment', () => {
+  beforeEach(() => {
+    jest.resetModules()
+  })
+
   afterAll(chdirCwd)
 
   describe('toWebpackConfig', () => {
@@ -13,6 +17,7 @@ describe('Test environment', () => {
 
     test('should use test config and production environment', () => {
       process.env.RAILS_ENV = 'test'
+      process.env.NODE_ENV = 'test'
       const { environment } = require('../index')
 
       const config = environment.toWebpackConfig()
--- a/package/env.js
+++ b/package/env.js
@@ -2,7 +2,7 @@ const { resolve } = require('path')
 const { safeLoad } = require('js-yaml')
 const { readFileSync } = require('fs')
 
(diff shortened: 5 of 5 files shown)


guilleiguaran:
@schpet I'm going to test this in my app and report if this fixes it 😊

edslocomb:
Ohhhh, so that's why things aren't working...

Maybe instead of hard-coding the environment list in 'NODE_ENVIRONMENTs' (which I suspect is/was meant to correspond to the webpack --mode setting, which has only those two legal values) this should read config/webpack/*.js and allow any of those except environment.js?

edslocomb:
Here's what I've got for package/env.js after refactoring to read config/webpack.js instead of relying on hard-coding:

```js
const { resolve } = require('path')                                                                         
const { safeLoad } = require('js-yaml')                                                                     
const { readdirSync, readFileSync } = require('fs')                                                         
                                                                                                            
const DEFAULT = 'production'                                                                                
const ymlConfigPath = resolve('config', 'webpacker.yml')                                                    
                                                                                                            
const railsEnv = process.env.RAILS_ENV                                                                      
const nodeEnv = process.env.NODE_ENV                                                                        
                                                                                                            
const ymlConfig = safeLoad(readFileSync(ymlConfigPath), 'utf8')                                             
const availableYmlEnvironments = Object.keys(ymlConfig)                                                     
                                       .filter((key) => key !== 'default')                                  
                                                                                                            
const jsConfigDir = resolve('config', 'webpack')                                                            
const jsConfigs = readdirSync(jsConfigDir)                                                                  
  .map((f) => f.replace(/\.[^\.]+$/, ''))                                                                   
  .filter((f) => f !== 'environment')                                                                       
                                                                                                            
module.exports = {                                                                                          
  railsEnv: railsEnv && availableYmlEnvironments.includes(railsEnv) ? railsEnv : DEFAULT,                   
  nodeEnv: nodeEnv && jsConfigs.includes(nodeEnv) ? nodeEnv : DEFAULT                                       
}                                                                                                           
```

gauravtiwari:
Thanks, @schpet. 

I forgot to remove the test.js file earlier but anyway, the idea behind this was since webpack and most of the JS libraries only understands two environments, we don't need a test environment. Ideally, we want to be testing or running CI against development or production bundle, i.e. the bundle produced uses the same webpack config as production bundle. Having totally different configs could potentially break the build on production although it might be passing on CI. 

What do you think?

schpet:
hey folks, thanks for checking this out! 

@gauravtiwari that makes sense, appreciate the context. 

the problem i was facing is that react throws a warning because my NODE_ENV !== 'production' but the JS was minified:

> Warning: It looks like you're using a minified copy of the development build of React. When deploying React apps to production, make sure to use the production build which skips development warnings and is faster. See https://fb.me/react-minification for more details.

i'm worried that webpacker's production defaults will result in other people seeing this console error. what do you think of setting `mode: "production"` only in production.js? and keeping test.js so that by default tests will not be minified. folks who want to test production JS can precompile assets or set NODE_ENV=production. @edslocomb's PR #1514 supports the idea that the prod setup isn't ideal for tests.

in my app, i've added the following workaround to `config/webpack/test.js` to make the react error go away:

```
const environment = require("./environment")

environment.config.merge({
  mode: "development"
})

```

interested in your thoughts on this 😄

schpet:
closing this in favor of #1524 based on @gauravtiwari's feedback.

i finally spun up a fresh app and realized that [by default, the development environment is used](https://github.com/rails/webpacker/blob/master/lib/install/config/webpack/test.js#L1), so I think these defaults work well 👍🏻

schpet:
@edslocomb you wrote:

> People who have seen and done this sort of thing before are going to expect to be able to pick a name for the new environment, drop a config file with that name into the appropriate folder, and have everything Just Work.

AFAIK this still works, e.g. 

```
$ cat config/webpack/foobar.js
throw new Error('foobar loaded')

$ NODE_ENV=foobar bin/webpack
/private/tmp/webpacker-may-25/config/webpack/foobar.js:1
(function (exports, require, module, __filename, __dirname) { throw new Error('foobar loaded')
                                                              ^

Error: foobar loaded
    at Object.<anonymous> (/private/tmp/webpacker-may-25/config/webpack/foobar.js:1:69)
    at Module._compile (internal/modules/cjs/loader.js:678:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:689:10)
    at Module.load (internal/modules/cjs/loader.js:589:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:528:12)
    at Function.Module._load (internal/modules/cjs/loader.js:520:3)
    at Module.require (internal/modules/cjs/loader.js:626:17)
    at require (internal/modules/cjs/helpers.js:20:18)
    at requireConfig (/private/tmp/webpacker-may-25/node_modules/webpack/bin/convert-argv.js:97:18)
    at /private/tmp/webpacker-may-25/node_modules/webpack/bin/convert-argv.js:104:17
```

am I missing something?

gauravtiwari:
That's correct @schpet 

Please see https://github.com/rails/webpacker/pull/1359 for more context in regards to using custom environments in Webpacker.

guilleiguaran:
@gauravtiwari @schpet can you help me to figure out what we should do to get #1502 working?

Does https://github.com/rails/webpacker/pull/1524 fix it?

gauravtiwari:
@guilleiguaran What do you have in your test.js file? Just curious to see your setup

> Should I move my configs from test.js to development.js?

yes, please

guilleiguaran:
> @guilleiguaran What do you have in your test.js file? Just curious to see your setup

```javascript
const webpack = require('webpack')

// We must expose jQuery to the global object to be able to use `JQuery.active`
// in `wait_for_ajax` (spec/support/wait_for_ajax.rb)

environment.loaders.append('expose', {
  test: require.resolve('jquery'),
  use: {
    loader: 'expose-loader',
    options: 'jQuery'
  }
})

```

gauravtiwari:
Thanks, @guilleiguaran  If you move this code to development.js, is everything good on CI and locally?

I am also thinking if there is a good case to bring back `test.js`, although I am a bit worried about unreliable tests if different configs are used - test vs production bundle.

PR #1515, opened 2018-05-22
Parse webpacker.yml with ERB to allow dynamic configurations

palkan:
This PR enhance Webpacker configuration management to allow having _dynamic_ values in `webpacker.yml`.

We have the following use case: some people in our team uses Docker for development (mostly backend devs), others use their host machines. We're running webpack-dev-server as a separate docker service, so we have to specify its hostname (within Docker network) in the `webpacker.yml`; but we cannot hard-code it since other devs use default "localhost".

With the provided patch our config looks like this:

```yml
development:
  <<: *default
  compile: true

  # Reference: https://webpack.js.org/configuration/dev-server/
  dev_server:
    https: false
    host: <%= ENV['WEBPACKER_HOST'] || 'localhost' %>
    port: 3035
```

### DIFF
--- a/lib/webpacker/dev_server_runner.rb
+++ b/lib/webpacker/dev_server_runner.rb
@@ -1,5 +1,6 @@
 require "shellwords"
 require "yaml"
+require "erb"
 require "socket"
 require "webpacker/runner"
 
@@ -14,7 +15,7 @@ module Webpacker
     private
       def load_config
         @config_file = File.join(@app_path, "config/webpacker.yml")
-        dev_server = YAML.load_file(@config_file)[ENV["RAILS_ENV"]]["dev_server"]
+        dev_server = YAML.load(ERB.new(File.read(@config_file)).result)[ENV["RAILS_ENV"]]["dev_server"]
 
         @hostname          = dev_server["host"]
         @port              = dev_server["port"]
--- a/lib/webpacker/configuration.rb
+++ b/lib/webpacker/configuration.rb
@@ -1,3 +1,6 @@
+require "yaml"
+require "erb"
+
 class Webpacker::Configuration
   delegate :root_path, :config_path, :env, to: :@webpacker
 
@@ -67,7 +70,7 @@ class Webpacker::Configuration
     end
 
     def load
-      YAML.load(config_path.read)[env].deep_symbolize_keys
+      YAML.load(ERB.new(config_path.read).result)[env].deep_symbolize_keys
 
     rescue Errno::ENOENT => e
       raise "Webpacker configuration file not found #{config_path}. " \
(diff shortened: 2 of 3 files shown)


lib/webpacker/dev_server_runner.rb
```diff
@@ -14,7 +15,7 @@ def run
     private
       def load_config
         @config_file = File.join(@app_path, "config/webpacker.yml")
-        dev_server = YAML.load_file(@config_file)[ENV["RAILS_ENV"]]["dev_server"]
+        dev_server = YAML.load(ERB.new(File.read(@config_file)).result)[ENV["RAILS_ENV"]]["dev_server"]
```
palkan: Not sure how to test this change (except from checking manually).

I think, it make sense to re-use `Configuration` here (but that would require its refactoring to not depend on `Webpacker` instance).

gauravtiwari:
@palkan Thanks for the PR but since Webpacker uses the same config file with both JS and Ruby, unfortunately, it's not possible to use ERB (JS code will fail). You could use env variable to change `HOST` or `PORT` at runtime - https://github.com/rails/webpacker#development. 

```bash
WEBPACKER_DEV_SERVER_HOST=example.com WEBPACKER_DEV_SERVER_INLINE=true WEBPACKER_DEV_SERVER_HOT=false ./bin/webpack-dev-server
```

Sounds good?

gauravtiwari:
It would be nice to have this though. Perhaps, we can pass parsed config from Rails to JS via JSON. Although this would require a few changes.

palkan:
> JS code will fail

Hm, for some reason my configuration with this patch doesn't fail. I'm using 3.x, maybe, that's the reason?

> Sounds good?

Not sure. `DevServerRunner` doesn't respect these env vars (https://github.com/rails/webpacker/blob/master/lib/webpacker/dev_server_runner.rb#L19).

What do you think about refactoring `Configuration` a little bit to re-use it in `DevServerRunner`?

palkan:
> Hm, for some reason my configuration with this patch doesn't fail

nvm, we have `WEBPACKER_DEV_SERVER_HOST` set for our `webpacker` service, that's why it works

palkan:
> What do you think about refactoring Configuration a little bit to re-use it in DevServerRunner

The change could be as simple as:

```ruby
class Webpacker::Configuration
  attr_reader :root_path, :config_path, :env

  def initialize(root_path:, config_path:, env: )
    @root_path = root_path
    @config_path = config_path
    @env = env
  end
  
  # ...
end
```

And then in `dev_server_runner.rb`:


```ruby
def load_config
  config = Configuration.new(
    config_path: "config/webpacker.yml",
    root_path: @app_path,
    env: ENV["RAILS_ENV"]
  )
  dev_server = config[:dev_server]
  # ...
end
```

gauravtiwari:
> What do you think about refactoring Configuration a little bit to re-use it in DevServerRunner

@palkan Sure sounds good to me 👍 I guess in the past the issue was that loading configuration in runner loaded the whole Rails stack, which wasn't ideal (if I am recalling correctly).

palkan:
Cool! I'll prepare another PR

PR #1516, opened 2018-05-22
Refactor configuration

palkan:
This PR refactors configuration classes to make it possible to re-use then in runners.

This way we can support env variable in `DevServerRunner` too (the same way as we do in Webpacker itself).

See also the discussion https://github.com/rails/webpacker/pull/1515#issuecomment-390898745

/cc @gauravtiwari

### DIFF
--- a/lib/webpacker/configuration.rb
+++ b/lib/webpacker/configuration.rb
@@ -1,3 +1,7 @@
+require "yaml"
+require "active_support/core_ext/hash/keys"
+require "active_support/core_ext/hash/indifferent_access"
+
 class Webpacker::Configuration
   attr_reader :root_path, :config_path, :env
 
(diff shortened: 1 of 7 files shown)


gauravtiwari:
thanks, @palkan looks great 👍

palkan:
@gauravtiwari 🙂

What about backporting this one to 3.x? I'm already using this patch with 3.x

gauravtiwari:
@palkan Sure will do this over the weekend and make a new release (or earlier if I have time) 👍

palkan:
@gauravtiwari I already have a patch for 3.x. Would you like me to create a PR?

gauravtiwari:
That would be great although I could cherry-pick it from master (might do this evening)

PR #1517, opened 2018-05-23
Add semicolon to default "Hello World"-message statement

neumanrq:
After upgrading `webpacker` for our Rails apps, our JS linter complained about the missing semicolon here. Do you agree on adding it?

### DIFF
--- a/lib/install/javascript/packs/application.js
+++ b/lib/install/javascript/packs/application.js
@@ -7,4 +7,4 @@
 // To reference this file, add <%= javascript_pack_tag 'application' %> to the appropriate
 // layout file, like app/views/layouts/application.html.erb
 
-console.log('Hello World from Webpacker')
+console.log('Hello World from Webpacker');


swrobel:
Just google "javascript semicolons" to see what a divisive issue this is. I don't think it's necessary to change this in webpacker.

guilleiguaran:
Some people have JS Linters that complain about non-required semicolons 😬

neumanrq:
> Just google "javascript semicolons" to see what a divisive issue this is. 

Sorry then. My (subjective) impression was that most JS developers around me tend to keep semicolons in to avoid ambiguities, so I thought it would be good to stick to this behavior here, too. 

I know it has been and will be a controversial debate. Being aware of this I explicitly formulated the PR as a question - just close it if you disagree on the change.

edslocomb:
Webpacker has its own [house style](https://github.com/rails/webpacker/blob/master/.eslintrc.js), and while it's in the minority in its semicolon preferences, I usually prefer the "when in Rome" approach to code formatting.

Anything that comes out of a generator, however, is fair game -- I'd say you can and should change those outputs to reflect your own Rome -- but the generator typically doesn't know what your house style is.

neumanrq:
Got it, nevermind. It was not my intention to criticize or nitpick on the overall style - I completely agree to the "When in Rome" rule of thumb. 

My mistake was that I believed that everybody in the Roman empire is used to end JS statements with semicolon - so I really thought (naively) that this would be an improvement. But I should have visited the local Forum Romanum beforehand to inform about the lint config information beforehand - I will do so next time ✌️ 

Thanks @edslocomb for the hint; lesson learnt.

PR #1524, opened 2018-05-25
Remove unused test environment file

schpet:
related: #1513

PR #1525, opened 2018-05-25
Remove outdated test files

schpet:
https://github.com/rails/webpacker/pull/1524#issuecomment-392164890


schpet:
still working on this, one min...

schpet:
@gauravtiwari the docs on testing reference test.js but i think it's still legit and should probably be kept. what do you think?

gauravtiwari:
Oh yes, let's leave it there. Thank you 💯 🍰

PR #1528, opened 2018-05-27
Replace postcss-cssnext with postcss-preset-env

ojab:
postcss-cssnext is deprecated (see https://github.com/MoOx/postcss-cssnext/issues/445) and postcss-preset-env should be used instead.

`yarn upgrade` is done in the first commit to minimize lockfile diff due to actual package change.


gauravtiwari:
@ojab Thanks for this but could you please remove the commit related to `yarn upgrade`. Is that necessary? or we can do yarn upgrade in a separate PR with package.json update.

ojab:
done

renchap:
`cssnext` deprecation blog post: https://moox.io/blog/deprecating-cssnext/

ojab:
`autoprefixer` is actually included into `postcss-present-env`, linked post was corrected. So no breaking changes expected.

gauravtiwari:
thanks @ojab

jaredbeck:
Can this be easily backported to v3?

PR #1538, opened 2018-05-31
Upgrade to babel@7

ojab:
Since we're at prerelease, why not use prerelase babel?

`babel@6` is basically frozen and new features (e. g. `@babel/plugin-syntax-nullish-coalescing-operator`, `@babel/plugin-syntax-optional-chaining`, you name it) is available only in 7.0.


ojab:
Also both `babel-preset-env@7` and `postcss-preset-env` supports `.browserslistrc`, [looks like it is the way forward](https://github.com/babel/babel/pull/7835) for declaring supported platforms.
Default browsers are `> 0.5%, last 2 versions, Firefox ESR, not dead`, so should we drop `browsers` from default `.babelrc` (`"browsers": "> 1%"`) and add documentation how to change it or we want to add `.browserslistrc` with current defaults?

package.json
```diff
@@ -12,13 +12,15 @@
...
-    "babel-polyfill": "^6.26.0",
-    "babel-preset-env": "^1.6.1",
+    "@babel/cli": "^7.0.0-beta.49",
+    "@babel/core": "^7.0.0-beta.49",
+    "@babel/plugin-proposal-class-properties": "^7.0.0-beta.49",
+    "@babel/plugin-proposal-object-rest-spread": "^7.0.0-beta.49",
+    "@babel/plugin-syntax-dynamic-import": "^7.0.0-beta.49",
+    "@babel/polyfill": "^7.0.0-beta.49",
+    "@babel/preset-env": "^7.0.0-beta.49",
+    "babel-core": "^7.0.0-0",
```
gauravtiwari: We need to remove the legacy babel-core no?

docs/es6.md
```diff
@@ -6,7 +6,7 @@
 Webpacker ships with [babel](https://babeljs.io/) - a JavaScript compiler so
 you can use next generation JavaScript, today. The Webpacker installer sets up a
 standard `.babelrc` file in your app root, which will work great in most cases
-because of [babel-env-preset](https://github.com/babel/babel-preset-env).
+because of [@babel/env-preset](https://github.com/babel/babel-preset-env).
```
gauravtiwari: should be `preset-env` right?

gauravtiwari:
Thanks @ojab 💯 

If you could please look into those comments and also add a `.browserslistrc` with current defaults, that would be great.

package.json
ojab: It's just a stub for packages that depend on `babel-core`, see https://www.npmjs.com/package/babel-core/v/7.0.0-bridge.0.
I'll drop it if you wish, but it probably will reduce potential breakage/incompatibility for the time being and it should be dropped later (at least after babel@7 release).

I could make it `^7.0.0-bridge` instead of `^7.0.0-0`, to be more clear.

docs/es6.md
ojab: whoops, fixed

ojab:
Done.

ojab:
(`"babel-core": "^7.0.0-0"` is replaced by `"babel-core": "^7.0.0-bridge"`, please tell if I should drop it completely)

package.json
gauravtiwari: Ahh, I see!  yes please, let's change to `^7.0.0-bridge` for clarity. Thank you

gauravtiwari:
@ojab We have to migrate .babelrc file as well and preset for react etc.

ojab:
ouch, will do shortly.

gauravtiwari:
No worries, made a PR #1543

PR #1543, opened 2018-06-03
Upgrade babelrc to use babel 7

gauravtiwari:
@ojab Could you please review this one?

### DIFF
--- a/lib/install/config/.babelrc
+++ b/lib/install/config/.babelrc
@@ -1,14 +1,24 @@
 {
   "presets": [
-    ["env", {
-      "modules": false,
-      "useBuiltIns": true
-    }]
+    [
+      "@babel/env",
+      {
+        "modules": false,
+        "targets": {
+          "forceAllTransforms": true
+        },
+        "useBuiltIns": "usage"
+      }
+    ]
   ],
-
   "plugins": [
-    "syntax-dynamic-import",
-    "transform-object-rest-spread",
-    ["transform-class-properties", { "spec": true }]
+    "@babel/plugin-syntax-dynamic-import",
+    "@babel/plugin-proposal-object-rest-spread",
+    [
+      "@babel/plugin-proposal-class-properties",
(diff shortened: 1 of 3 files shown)


lib/install/config/.babelrc
```diff
@@ -1,14 +1,24 @@
...
-    ["env", {
-      "modules": false,
-      "useBuiltIns": true
-    }]
+    [
+      "@babel/env",
+      {
+        "modules": false,
+        "targets": {
+          "forceAllTransforms": true
```
ojab: `forceAllTransforms` is unneeded because we're targeting uglify-es (it's needed for old uglify without es6 support).

lib/install/examples/react/.babelrc
```diff
@@ -1,6 +1,25 @@
 {
   "presets": [
-    ["env", { "modules": false } ],
-    "react"
+    [
+      "@babel/env",
+      {
+        "modules": false,
+        "targets": {
+          "forceAllTransforms": true
```
ojab: ditto

lib/install/config/.babelrc
gauravtiwari: Yeah I know, we sort of need it because the default target for Uglify is ES5 as well.  https://github.com/rails/webpacker/blob/master/package/environments/production.js#L29

lib/install/config/.babelrc
```diff
@@ -1,14 +1,24 @@
...
   "plugins": [
-    "syntax-dynamic-import",
-    "transform-object-rest-spread",
-    ["transform-class-properties", { "spec": true }]
+    "@babel/plugin-syntax-dynamic-import",
+    "@babel/plugin-proposal-object-rest-spread",
+    [
+      "@babel/plugin-proposal-class-properties",
+      {
+        "spec": true
```
pyromaniac: It is actually `loose` now

PR #1564, opened 2018-06-10
Babel 7 updates updates

gauravtiwari:
## Babel 7 related changes

- Support async, await and yield
- Migrate react installer to use babel 7 preset


gauravtiwari:
Perhaps, someone can help migrating `.babelrc` to `babel.config.js` - https://new.babeljs.io/docs/en/next/babelconfigjs.html

per comment: https://github.com/rails/webpacker/pull/1546#issuecomment-396067053

gauravtiwari:
Maybe we can make a `webpacker-preset` package that abstracts out all default plugins and presets we ship with `.babelrc`

swrobel:
@renchap maybe you can help w/ migrating to babel.config.js? https://github.com/rails/webpacker/pull/1564#issuecomment-396070868

PR #1582, opened 2018-06-21
Fix environment variable logic for dev_server options to allow falsey values

ElMassimo:
### Expected Behaviour ✅ 
A configuration in `config/webpacker.yml` such as: 

```yaml
  dev_server:
    hmr: true
```

and an environment variable with the value `WEBPACKER_DEV_SERVER_HOT=false`, should set `devServer.hot` to `false`.

### Current Behaviour 🐞 
Settings can not be set to `false`, `null`, `0`, or `''` using environment variables, only truthy values can override the configuration file settings.

### Description 📖 
This pull request changes the logic to only skip the override if the return value from `fetch` is `undefined`, as otherwise it means that a setting's environment variable is set, and the user intends to override the value.


gauravtiwari:
thanks @ElMassimo 🍰

Issue #1607, opened 2018-07-11
NoMethodError: undefined method `inquiry' for "production":String


cswilliams:
Hi,
I'm using webpacker 3.5.5, rails 5.2.0, and ruby 2.3.5 and when running `RAILS_ENV=staging bundle exec rake assets:precompile` I'm getting the following error:
```
RAILS_ENV=staging environment is not defined in config/webpacker.yml, falling back to production environment
NoMethodError: undefined method `inquiry' for "production":String
/var/www/.libs/customer/.bundle/2.3.5/1/ruby/2.3.0/gems/webpacker-3.5.5/lib/webpacker/env.rb:16:in `inquire'
/var/www/.libs/customer/.bundle/2.3.5/1/ruby/2.3.0/gems/webpacker-3.5.5/lib/webpacker/env.rb:7:in `inquire'
/var/www/.libs/customer/.bundle/2.3.5/1/ruby/2.3.0/gems/webpacker-3.5.5/lib/webpacker/instance.rb:11:in `env'
/var/www/.libs/customer/.bundle/2.3.5/1/ruby/2.3.0/gems/webpacker-3.5.5/lib/webpacker/instance.rb:18:in `config'
/var/www/.libs/customer/.bundle/2.3.5/1/ruby/2.3.0/gems/webpacker-3.5.5/lib/webpacker.rb:25:in `config'
/var/www/.libs/customer/.bundle/2.3.5/1/ruby/2.3.0/gems/webpacker-3.5.5/lib/tasks/webpacker/verify_install.rake:6:in `block (2 levels) in <top (required)>'
/var/www/.libs/customer/.bundle/2.3.5/1/ruby/2.3.0/gems/rake-12.3.1/exe/rake:27:in `<top (required)>'
/usr/local/rvm/gems/ruby-2.3.5/gems/bundler-1.16.1/lib/bundler/cli/exec.rb:75:in `load'
/usr/local/rvm/gems/ruby-2.3.5/gems/bundler-1.16.1/lib/bundler/cli/exec.rb:75:in `kernel_load'
/usr/local/rvm/gems/ruby-2.3.5/gems/bundler-1.16.1/lib/bundler/cli/exec.rb:28:in `run'
/usr/local/rvm/gems/ruby-2.3.5/gems/bundler-1.16.1/lib/bundler/cli.rb:424:in `exec'
[...]

PR #1608, opened 2018-07-11
Require active_support’s string inquiry

cswilliams:
It’s used in Webpacker::Env and depending on which rails frameworks are chosen at boot, it may not always be required before webpacker runs.

This fixes the issue detailed in #1607

### DIFF
--- a/lib/webpacker.rb
+++ b/lib/webpacker.rb
@@ -1,4 +1,5 @@
 require "active_support/core_ext/module/attribute_accessors"
+require "active_support/core_ext/string/inquiry"
 require "active_support/logger"
 require "active_support/tagged_logging"
 


gauravtiwari:
thanks @cswilliams 👍

maryammouse:
@gauravtiwari  Is this fixed in the latest version of webpacker? I'm not sure how to find the release it's a part of.

gauravtiwari:
@maryammouse Yes, both in 3.x and 4.x pre-release.

maryammouse:
Hmm, that's interesting - I am using webpacker 3.5.5 and I still got the issue in #1607 , which was fixed both by requiring the inquiry module and also by setting up an explicit uat config for webpack so no fallback was necessary. @gauravtiwari .

gauravtiwari:
hmm, that's strange. I see it's included in both

> https://github.com/rails/webpacker/blob/3-x-stable/lib/webpacker.rb#L2
> https://github.com/rails/webpacker/blob/master/lib/webpacker.rb#L2

Are you sure it's def 3.5.5? Please could you run `bundle update webpacker` and `yarn add @rails/webpacker`

corporealfunk:
@gauravtiwari this does appear in the the 3.x-stable branch, but it is not in the released 3.5.5 tag nor published gem:

https://github.com/rails/webpacker/blob/v3.5.5/lib/webpacker.rb

gauravtiwari:
Just published 3.6.0

PR #1617, opened 2018-07-18
Read postcss configs on a root project directory.

ducky-hong:
It is impossible to put logic in `.postcssrc.yml`. In order to make `postcss-loader` to read the `postcss.config.js` file, a `path` option should be deleted.

### DIFF
--- a/package/utils/get_style_rule.js
+++ b/package/utils/get_style_rule.js
@@ -3,7 +3,6 @@ const path = require('path')
 const devServer = require('../dev_server')
 const { nodeEnv } = require('../env')
 
-const postcssConfigPath = path.resolve(process.cwd(), '.postcssrc.yml')
 const isProduction = nodeEnv === 'production'
 const inDevServer = process.argv.find(v => v.includes('webpack-dev-server'))
 const isHMR = inDevServer && (devServer && devServer.hmr)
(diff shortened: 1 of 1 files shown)


tai2:
Right.

https://github.com/postcss/postcss-loader#path

>You can manually specify the path to search for your config (postcss.config.js) with the config.path option. This is needed if you store your config in a separate e.g ./config || ./.config folder.
>
>⚠️ Otherwise it is unnecessary to set this option and is not recommended
>
>⚠️ Note that you can't use a filename other than the supported config formats (e.g .postcssrc.js, postcss.config.js), this option only allows you to manually specify the directory where config lookup should start from

gauravtiwari:
Could you please rebase with master and update the PR? Think this is fixed in another commit but there is some dangling code that needs to be removed. 

Many thanks :)

JakeLaCombe:
The dangling code can be found here btw. https://github.com/rails/webpacker/blob/master/package/utils/get_style_rule.js#L14

ducky-hong:
@gauravtiwari My PR is the same as this commit: https://github.com/rails/webpacker/commit/31d9ec051e297a6dbe7f9238b69a451814289963
The dangling code that @JakeLaCombe mentioned is a different issue.
Close this PR.

PR #1625, opened 2018-07-26
Separate css extraction from build environment

tai2:
webpacker currently tie CSS extraction to build environment. It extracts stylesheets to files when production build or development with no dev-server.
I want to separate it from build environment because CSS extraction is merely a user preference.
Some frontend developers might desire to put stylesheets in style tag or to emit separate css files using [style-loader/url](https://github.com/webpack-contrib/style-loader#url) even in production.
These choices are reasonable in context of recent CSSinJS practices or HTTP/2 environment.

Another problem is that [webpacker checks whether command line arguments contains a keyword `webpack-dev-server`](https://github.com/rails/webpacker/blob/87c527dc33d1b693cfce7bf2bb08fa74f1bf0f85/package/utils/get_style_rule.js#L8) to switch css extraction. It prevents to use other development tools like storybook, which contains their own webpack-dev-server, with webpacker's configuration file.
I bereave this PR simplifies the condition checking logic and provides more flexible configuration.


gauravtiwari:
Thanks @tai2 Could you please rebase with master?

tai2:
@gauravtiwari rebased 🙏🏻

gauravtiwari:
Thanks @tai2 🙏

iczman:
After upgrading to 4.0.0.rc.2 with this change, my Rails app now doesn't emit css in dev env with `webpack-dev-server` running so my pages are missing all styling.

I changed the default `extract_css` to `true` in `webpacker.yml` and now it works, but I'm not quite sure if it's recommended to emit css in dev running `webpack-dev-server`?

```yaml
default: &default
  extract_css: true # default to false
```

aceofspades:
This is a breaking change, why default false?

ericdfields:
I get why it's now a config, but I did lose a bit of time wondering why my styles suddenly broke when a stray bundle install was done on a separate branch that got merged. Definitely strongly highlight this in the readme for upgraders!

aceofspades:
Every project needs to consider how much time a change, multiplied by the number of people it affects, can be extremely costly.

PR #1629, opened 2018-07-30
Let’s test against Rails 5.2

ttanimichi:
NOTE: https://github.com/rails/webpacker/commit/c5b63f843f25916a52d85da98691b3e37b04f811

> `~> 5.1` doesn't mean `5.1.x`. It means `5.x`. Now CIs aren't executed correctly.
> ref. https://bundler.io/v1.5/gemfile.html#gemfiles


gauravtiwari:
thanks 🍰

PR #1658, opened 2018-08-15
Fixes Webpacker::Compiler.env accessible to binstubs

### LINKED ISSUE #1641: Environment variables set with Webpacker::Compiler aren't available via process.env
#691 and [the docs](https://github.com/rails/webpacker/blob/cd36cd5/docs/env.md) suggest that `Webpacker::Compiler` would be able to pass environment variables over to the compiler, but I'm not finding this to be the case (at least based on the documentation example).

```ruby
# config/initializers/webpacker.rb
Webpacker::Compiler.env['FRONTEND_API_KEY'] = 'your_secret_key'
```

```javascript
// anywhere in environment.js (tried multiple places)
console.log(process.env['FRONTEND_API_KEY'])
=> undefined

console.log(process.env.FRONTEND_API_KEY)
=> undefined

console.log(process.env)
=> lots of environment variables
```

```bash
» FRONTEND_API_KEY='your_secret_key' ./bin/webpack-dev-server
your_secret_key
```

**Use case**

In our case, we have an existing Rails app that we're adding SPA to. We have a lot of legacy environment variable `if` checks and environment checks using Figaro. It would be quite complex to try to reproduce this in dotenv. By far the simplest solution for sharing app keys and variables with the SPA is simply passing them over from Rails at compile-time. 

Thank you for reading and for the really useful gem.

bbugh:
#691 added the ability to set environment variables that can be read by the compiler, but these are only set when using the compiler directly (e.g. via `assets:precompile`). Environment variables set via the compiler are not accessible to `./bin/webpack` or `./bin/webpack-dev-server`, so manual compilation and development access does not work.

This patch fixes #1641. Webpacker::Compiler.env will be accessible in the binstubs when manually compiling or in development mode.

This is a production deployment blocker for us.

### DIFF
--- a/lib/webpacker/dev_server_runner.rb
+++ b/lib/webpacker/dev_server_runner.rb
@@ -44,7 +44,7 @@ module Webpacker
       end
 
       def execute_cmd
-        env = { "NODE_PATH" => @node_modules_path.shellescape }
+        env = Webpacker::Compiler.env.merge("NODE_PATH" => @node_modules_path.shellescape)
         cmd = [
           "#{@node_modules_path}/.bin/webpack-dev-server",
           "--config", @webpack_config
--- a/lib/webpacker/webpack_runner.rb
+++ b/lib/webpacker/webpack_runner.rb
@@ -4,7 +4,7 @@ require "webpacker/runner"
 module Webpacker
   class WebpackRunner < Webpacker::Runner
     def run
-      env = { "NODE_PATH" => @node_modules_path.shellescape }
+      env = Webpacker::Compiler.env.merge("NODE_PATH" => @node_modules_path.shellescape)
       cmd = [ "#{@node_modules_path}/.bin/webpack", "--config", @webpack_config ] + @argv
 
       Dir.chdir(@app_path) do


gauravtiwari:
Thanks, @bbugh looks great. Could you do the same here as well, please?

gauravtiwari:
Oh never mind :)

bbugh:
❤️

siong1987:
@bbugh does this actually fix the issue? i have the same settings in `config/initializers/webpacker.rb` but `Webpacker::Compiler.env` is still empty.

```
# config/initializers/webpacker.rb
Webpacker::Compiler.env['FRONTEND_API_KEY'] = 'your_secret_key'
```

fikrikarim:
@siong1987 I also try this with `4.0.0.pre.3` but not work for me too. Not sure whether my configuration is different or what.

gauravtiwari:
That's strange. Please make sure both gem and npm package are up to date. If it doesn't work then, please file an issue and we can investigate further.

fikrikarim:
Okay I'll try thanks @gauravtiwari

chriscarpenter12:
Any plans to packport this to 3.5? I noticed just now that this only works when I compile and not when running the dev server.

ayrton:
> @bbugh does this actually fix the issue? i have the same settings in `config/initializers/webpacker.rb` but `Webpacker::Compiler.env` is still empty.
> 
> ```
> # config/initializers/webpacker.rb
> Webpacker::Compiler.env['FRONTEND_API_KEY'] = 'your_secret_key'
> ```

Stumbled upon the same problem. If you're not using foreman and instead use an initializer you can change the binstub to:

```rb
# ...

# Set environment variables from credentials
require_relative "../config/boot"
require_relative "../config/application"
require_relative "../config/initializers/webpacker"

APP_ROOT = File.expand_path("..", __dir__)
Dir.chdir(APP_ROOT) do
  Webpacker::DevServerRunner.run(ARGV)
end
```

PR #1669, opened 2018-08-27
move integrety check config to yaml

lsylvester:
Currently the `webpacker:install` task updates the `development.rb` and `production.rb` environment files. 

This is problematic in rails/rails#33079 as it causes some of the tests to hang while waiting for the user to confirm if they want to overwrite the files.

This PR moves this configuration into the `webpacker.yml` by default.

Setting it in environment files is still supported for backwards compatibility. I am not sure if this should be deprecated, or expanded to allow for anything to be set via the `Rails.application.config`.

In order to be able to merge the configs set in the environment files and those loaded in the yml file, I have moved the reloading of the config out of the bootstrapping, and added the webpacker.yml to the watch so that spring restarts the environment when it changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,16 +278,20 @@ yarn add @rails/webpacker@4.0.0-pre.2
 
 By default, in development, webpacker runs a yarn integrity check to ensure that all local JavaScript packages are up-to-date. This is similar to what bundler does currently in Rails, but for JavaScript packages. If your system is out of date, then Rails will not initialize. You will be asked to upgrade your local JavaScript packages by running `yarn install`.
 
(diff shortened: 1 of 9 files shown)


danielpowell4:
Why would you ever want to set the default of this to something other true? Seems like setting to false is promoting bad practices

gauravtiwari:
thanks @lsylvester 

@danielpowell4 On production this check doesn't really matter since it's not possible to alter the dependencies in any way (precompiled) compared to development.

danielpowell4:
I was thinking that failing precompilation, which uses a potentially outdated frozen lockfile, could also be a safe default.

What you all did also makes perfect sense to me though!

PR #1670, opened 2018-08-28
Upgrade babel to released 7.0.0 version

grk:


gauravtiwari:
thanks @grk

gauravtiwari:
We should update other babel family members too - https://github.com/babel/babel-loader/releases

Happy to make a PR? :)

PR #1672, opened 2018-08-28
allow dev runner to behave like the main runner

everplays:
We allowed the use of ARGS in `Webpacker::WebpackRunner` to pass
down any necessary arguments to webpack itself:

```ruby
cmd = [ "#{@node_modules_path}/.bin/webpack", "--config", @webpack_config ] + @argv
```

However, this was not the case for `Webpacker::DevServerRunner`. Hence,
it was not possible to run the dev runner as such:

```bash
./bin/webpack-dev-server --inline=false
```

This is depite the fact that ARGS are passed to DevServerRunner.

### DIFF
--- a/lib/webpacker/dev_server_runner.rb
+++ b/lib/webpacker/dev_server_runner.rb
@@ -50,6 +50,7 @@ module Webpacker
           "--config", @webpack_config
         ]
         cmd += ["--progress", "--color"] if @pretty
+        cmd += @argv
 
         Dir.chdir(@app_path) do
           exec env, *cmd


gauravtiwari:
Thanks, @everplays for the PR.

The reason for this is that we don't have support for overriding dev server options through ARGS. There are some middlewares that work with dev server and therefore, it's tricky to make things work with multiple configuration sources. 

You can pass options through webpacker.yml or an environment variable.    

```bash
WEBPACKER_DEV_SERVER_HOST=example.com WEBPACKER_DEV_SERVER_INLINE=true WEBPACKER_DEV_SERVER_HOT=false ./bin/webpack-dev-server
```

https://github.com/rails/webpacker/blob/master/lib/webpacker/dev_server_proxy.rb

everplays:
So it's not implemented to prevent the developers from shooting themselves in the foot.

In my case, changing the `webpacker.yml` is not desired because I am the only one in our team who wants to disable the `inline`. Anyway, even though dealing with arguments would be easier for something like `--inline=false`, the modification via env variable works for me.

Thanks for the reply @gauravtiwari and feel free to do as you please with this PR. :-)

Issue #1623, opened 2018-07-25
Make webpacker.yml and postcssrc.yml files optional.


JakeLaCombe:
Would it be possible to make the files optional for this Gem? My team is trying to create a node package that contains a webpacker configuration file they can import into their webpack scripts, and requiring the webpacker.yml and postcssrc.yml files make that a little difficult, as they have specific locations inside a rails application.

bjankord:
It would be really helpful to make this optional. Looking at [the code for this](https://github.com/rails/webpacker/blob/a2aa7399f4bc2b9ca2cd035bf456a78a054dba07/package/utils/get_style_rule.js#L6-L39), it could be done in a passive way.

You could check and see if the [postcssrc.yml file existed in the current location webpacker checks](https://github.com/rails/webpacker/blob/a2aa7399f4bc2b9ca2cd035bf456a78a054dba07/package/utils/get_style_rule.js#L6), and if it is not there, default to an empty postcssrc.yml file provided inside of wepacker. This would allow you keep the logic on the postcss-loader unchanged.

Another option would be to do the same type of check to see if the postcssrc.yml file existedi n the current location webpacker checks, and then based on that result, conditionally add the [postcss-loader](https://github.com/rails/webpacker/blob/a2aa7399f4bc2b9ca2cd035bf456a78a054dba07/package/utils/get_style_rule.js#L32-L37).

@gauravtiwari @dhh @tai2 would you be open to a PR for this?

dhh:
@bjankord Happy to go with a default inside-webpacker postcssrc.yml file when one isn't present where expected.

JakeLaCombe:
Also, I just noticed something. It looks like the postcssrc.yml file need was removed in this commit (https://github.com/rails/webpacker/commit/31d9ec051e297a6dbe7f9238b69a451814289963) making my last pull request extraneous.

PR #1676, opened 2018-08-29
Remove Dangling Code introduced from .postcssrc.yml change

JakeLaCombe:
Relates to https://github.com/rails/webpacker/issues/1623

In https://github.com/rails/webpacker/issues/1623, it was asked that we make the postcssrc.yml file optional, and that it would look for a file inside the lib file by default. However, there was another commit made before this was merged that removed the need for the postcssrc.yml file. This resulted in dangling code, which is removed in this pull request.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,8 +5,7 @@
   "main": "package/index.js",
   "files": [
     "package",
-    "lib/install/config/webpacker.yml",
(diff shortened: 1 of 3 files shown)


gauravtiwari:
Thanks @JakeLaCombe 🍰

PR #1691, opened 2018-09-06
Upgrade to Elm 0.19

### LINKED ISSUE #1668: Elm integration broken
Since the release of Elm 0.19 elm integration on webpacker is broken.

https://gist.github.com/ngw/0bc000fdf484a70dbc9693142790d2f0

```
 ngw@bluemonday ~ ruby -v
ruby 2.5.1p57 (2018-03-29 revision 63029) [x86_64-darwin17]

 ngw@bluemonday ~ elm --version
0.19.0

 ngw@bluemonday ~ rails -v
Rails 5.2.1 
```

Gemfile.lock: https://gist.github.com/ngw/4aa6e1ebb4de91c464554c91665f9625

klazuka:
Fixes #1668 

Notable changes that affect webpacker:

- `elm-package.json` renamed to `elm.json`
- the [old hot loader](https://github.com/fluxxu/elm-hot-loader) doesn't support Elm 0.19, so it has been replaced by [klazuka/elm-hot-webpack-loader](https://github.com/klazuka/elm-hot-webpack-loader)
- the Elm compiler no longer accepts the `--warn` flag
- elm-webpack-loader renamed the `pathToMake` option to `pathToElm`

### DIFF
--- a/lib/install/loaders/elm.js
+++ b/lib/install/loaders/elm.js
@@ -2,12 +2,11 @@ const { resolve } = require('path')
 
 const isProduction = process.env.NODE_ENV === 'production'
 const elmSource = resolve(process.cwd())
-const elmMake = `${elmSource}/node_modules/.bin/elm-make`
+const elmBinary = `${elmSource}/node_modules/.bin/elm`
 
-const elmDefaultOptions = { cwd: elmSource, pathToMake: elmMake }
+const elmDefaultOptions = { cwd: elmSource, pathToElm: elmBinary }
 const developmentOptions = Object.assign({}, elmDefaultOptions, {
   verbose: true,
-  warn: true,
   debug: true
 })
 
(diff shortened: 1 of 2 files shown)


gauravtiwari:
Thanks @klazuka

lib/install/loaders/elm.js
```diff
@@ -2,12 +2,11 @@ const { resolve } = require('path')
 
 const isProduction = process.env.NODE_ENV === 'production'
 const elmSource = resolve(process.cwd())
```
gauravtiwari: @klazuka One question, in the installer step we are using `src` as elm source location whereas here it's `process` root. Is this correct?

klazuka: This is correct. The local variable is poorly named. It's actually the path to the directory which contains the `elm.json` project manifest file.

gauravtiwari: thanks I see, maybe we can make that clear :)

Issue #1411, opened 2018-04-05
Specifying Custom localIdentName for css Modules as part of config.


bytheway875:
I noticed there was a recent change to the webpack config to add scssModule and cssModule support. However, it currently uses the default localIdentName, which is [hash:base64], and I still have to do this in order to get my more-descriptive localIndentName applied:

```
const cssModulesOptions = {
  // This must match the value of generateScopedName
  // in the .babelrc settings of react-css-modules.
  localIdentName: "[path]--[name]--[local]--[hash:base64:5]"
}

const CSSLoader = environment.loaders.get('moduleSass').use.find(el => el.loader === 'css-loader')

CSSLoader.options = merge(CSSLoader.options, cssModulesOptions)
```

I'm not sure how css-modules and react-css-modules work, but with the lighter-weight babel-plugin-react-css-modules, I need to make sure that my localIdentName in the css-loader config matches the generated class names. Since there is a huge performance advantage to using the babel-plugin vs the main package, and the default value of localIndentName is the not-so-descriptive [hash:base64] seems like this will be something people will need to do fairly often.

Is there any precendent for adding a config variable to webpacker.yml or elsewhere that we can then use in the getStyleRule file to feed it the localIdentName? 

If this seems like something that is feasible and useful, I can work on it.

gauravtiwari:
@bytheway875 Thanks for filling the issue. I have been thinking about this too. Would make sense to provide a default minimal `localIdentName` that remains unique. We could have an options object on environment instance that can be responsible for providing some of these settings. 

Please feel free to make a PR to add this option.

bytheway875:
Cool. I've forked the project, but I'm not sure how to get it running locally against a sample applications. Is it enough to link the gem locally?

gauravtiwari:
@bytheway875 Thank you :)

In your gemfile: 

```
gem "foo", path: "/path/to/local_webpacker_path"
```

then to link webpacker npm package, run this inside webpacker directory: 

```
yarn link
```

this will create a local copy of the package, and to use in your sample application. Run below command inside sample application folder: 

```
yarn link @rails/webpacker
```

Note: Run above whenever you run `yarn install` inside sample application since that will override the local link.

gauravtiwari:
Please try the latest RC, otherwise feel free to open a new issue with more details.

PR #1692, opened 2018-09-07
Specify class name in CSS modules in development

timhwang21:
Addresses #1411.

I took the `localIdentName` from the docs, if there's a more sensible default please let me know.

If this isn't something the team thinks is necessary, I can close this.

Thanks!

### DIFF
--- a/package/utils/get_style_rule.js
+++ b/package/utils/get_style_rule.js
@@ -22,6 +22,7 @@ const getStyleRule = (test, modules = false, preprocessors = []) => {
       options: {
         sourceMap: true,
         importLoaders: 2,
+        localIdentName: '[name]__[local]___[hash:base64:5]',
         modules
       }
     },


package/utils/get_style_rule.js
```diff
@@ -22,6 +22,7 @@ const getStyleRule = (test, modules = false, preprocessors = []) => {
       options: {
         sourceMap: true,
         importLoaders: 2,
+        localIdentName: isProduction ? undefined : '[name]__[local]___[hash:base64:5]',
```
gauravtiwari: I guess we can make this a default behavior, even in production and perhaps provide an easier setting to change it, but this is good enough for this PR.

gauravtiwari:
Thanks @timhwang21, looks good just one comment

timhwang21:
@gauravtiwari I've made this the default behavior in both dev and prod now. Thanks for the review!

gauravtiwari:
thanks @timhwang21

Issue #1124, opened 2017-12-19
Unable to run webpacker:install in a rails engine.


etewiah:
When I install the webpacker gem in a rails engine the "webpacker:install" command is not available.  There is however a "app:webpacker:install" command available.

When I run it:

`bundle exec rails app:webpacker:install`

I get the following error:

```
rails aborted!
Don't know how to build task 'app:template' (see --tasks)
./bin/rails:13:in `require'
./bin/rails:13:in `<main>'
(See full trace by running task with --trace)
```

The only way to get the command to run is to cd into "spec/dummy" and run "webpacker:install".  That installs webpacker into the "spec/dummy" directory though which is not ideal.

Just so I can learn more about webpacker I went ahead and installed it in the dummy app in my project here and it actually works but does not seem like the correct way to do things:

https://github.com/RealEstateWebTools/property_web_scraper

gauravtiwari:
@etewiah Thanks for filling the issue. Webpacker doesn't support engines yet, please see this thread for more details: https://github.com/rails/webpacker/issues/348

gauravtiwari:
Lets track this in #1694

zedtux:
I'm facing this exact error, @etewiah how did you fixed it?

PR #1694, opened 2018-09-08
Resolve errors blocking usage in a mountable Rails engine

kevindew:
This addresses https://github.com/rails/webpacker/issues/1124 and https://github.com/rails/webpacker/issues/348 and is related to this PR: https://github.com/rails/rails/pull/33819

This intends to improve support for usage with a mountable Rails engine by fixing the various errors a user encounters running the rake tasks. 

A mountable Rails engine tends to include a full Rails app inside a `test/dummy` directory which can be used to test mounting the engine. This is a somewhat unusual situation where you have a Rails app nested within directories from the Rakefile and all the Rails tasks are prefixed with `app`. Thus, these changes mostly solves those issues by running binstubs and yarn commands from the context of the `Rails.root`, rather than the root directory of the engine and checking whether a prefix is needed to run tasks.

## An example of usage in a mountable Rails engine

Given we create an engine with `rails plugin new rails-plugin --mountable` we then have a directory `test/dummy` where a new application has been created to allow us to test our engine.

If we run `rake app:webpacker:install` this will create binstubs in `test/dummy/bin`, config in `test/dummy/config`, update the `test/dummy/package.json`, etc - and then run yarn from within the context of `test/dummy`.

If we run `rake assets:precompile` it will look for packs in `test/dummy/javascript/packs` (or wherever else you configure it too) and output them to `test/dummy/public/packs`. Which is consistent with how sprockets would output assets to `test/dummy/public/assets`.

## How this affects engine usage and distribution

The fixes here aren't themselves opinionated - they merely maintain a pattern introduced by other Rails tasks - but it does offer a chance to think about engine usage:

- Engines can publish their javascript to NPM in the same way as Ruby gems such as Activestorage. Other applications that use the engine can install the module and import it into their code.
- A mounting app for an engine can have it's own `package.json` file which pulls in `@rails/webpacker` and can link to the root module via a relative import (as developers shouldn't push `@rails/webpacker` out as a dependency when it's not needed - and they shouldn't mark it as a dev dependency if they want to test `assets:precompile`)
- A pack file within the mounting app can be used to import the module into the mounting app

However these changes don't enforce working in this way and a second package.json isn't necessary as yarn will just fall back to the root one.

Hope this all makes sense - there's more info in the commit messages.


gauravtiwari:
Thanks @kevindew. Could you please rebase with master?

kevindew:
@gauravtiwari Sure, just done so.

gauravtiwari:
Thanks Kevin, will take a look at this over the weekend :)

gauravtiwari:
Please can you rebase again 🙏

mathewbaltes:
Anything we can do to help this one along?

kevindew:
I've just rebased this now - sorry @gauravtiwari I missed your previous notification.

@mathewbaltes if you have some time it'd be useful for you to just try this out and check it all still plays nice. It has been quite a while since I first raised this, but all still seemed to be working good for me giving it a run just now.

lib/install/angular.rb
```diff
@@ -6,8 +6,10 @@
 say "Copying hello_angular app to #{Webpacker.config.source_path}"
 directory "#{__dir__}/examples/angular/hello_angular", "#{Webpacker.config.source_path}/hello_angular"
 
-say "Installing all angular dependencies"
-run "yarn add core-js zone.js rxjs @angular/core @angular/common @angular/compiler @angular/platform-browser @angular/platform-browser-dynamic"
+Dir.chdir(Rails.root) do
+  say "Installing all angular dependencies"
+  run "yarn add core-js zone.js rxjs @angular/core @angular/common @angular/compiler @angular/platform-browser @angular/platform-browser-dynamic"
+end
```
gauravtiwari: @kevindew Do we need to do this? Is it not possible to use root package.json and yarn as a source of truth? But I see, what you mentioned in the description, in case someone wants to publish an engine package to npm.

kevindew: I'd suggest it's preferable to do this as it keeps the installation consistent to the Rails root (this would be a piece in a different place) and also if you have just a single package.json for both the Rails engine itself and the dummy app then they need to share dependencies when this may not makes sense (e.g. you wouldn't want to have a dependency on webpacker for your distributed engine JS files).

A nice thing though is that if you don't have a package.json in your Rails root then this will just use the global one for your application. So this dir change only prefers a package.json in your mounted app and will fallback to the root one if it doesn't exist.

kevindew:
@gauravtiwari Any chance we could look at this again soon? I notice Rails 6 is coming rather soon.

m4tm4t:
I'm looking for creating an engine using Vue.js and looking for advices on the setup
I'm done with the following steps : 

- In dummy/root app : 
    - adding engine to package.json dependencies
    - `require('engine/app/javascript/packs/engine.js')` in `rootapp/app/javascript/packs/engine.js`

- Then in my engine application.html.erb I can call `<%= javascript_pack_tag 'engine' %>`

It work but it's the right way to do that ? 
BTW @kevindew branch works to run webpacker tasks :+1:

mhoad:
Just following up on a comment from @kevindew to @gauravtiwari with the release of Rails 6 and making Webpacker the default answer for JS does this mean that without this getting merged that Webpacker wouldn't be able to work with engines correctly in Rails 6?

junket:
It would be great to see this merged. Thanks @kevindew @gauravtiwari 👍

gauravtiwari:
Hello @kevindew Thanks for working on this and apologies for long silence 🙏 very busy at work. 

Could you please rebase and ping me here for review?

kevindew:
Hey @gauravtiwari this is now rebased. Let me know when you've had a chance to do a review. 

I've done a pretty simple rebase here without going through exactly how the gem has changed so there may be some issues to resolve. So once you let me know you're happy with the approach and we're close to merging I'll carve out some time to do a sweep.

ThiagoCasao:
Guys, any update on this?

ranand:
@gauravtiwari - any chance this can be reviewed and merged soon? thanks!

Issue #1135, opened 2017-12-20
don't run `yarn check --integrity` on every rails command


doits:
`yarn check --integrity` is always run. Even if I do a `rails g migration ...`, which obviously has nothing to do with yarn.

I benchmarked it and noticed it adds about 400ms delay to such commands in my case, which is huge imo. It should be changed that it only runs when it makes sense to check for it, like when starting up the web server or the webpack-dev-server.

Since I am not familiar with the rails/railtie internals, I don't know if this is possible or how to implement such checks at the right place.

Is this possible? What do you think?

kriansa:
You can disable that in `development.rb`. 

Just set `config.webpacker.check_yarn_integrity = false`

doits:
@kriansa yes, I know this, but this disables it completely. It often saved me already when running `bin/webpacker` or `rails server` with outdated deps, so in general it makes sense to have it enabled on development. But not when I run `rails g migration`.

doits:
Why was this closed? Will/can not be fixed?

ansonhoyt:
This check also breaks adding webpacker to an existing project if `yarn.lock` exists. Running `rails webpacker:install` loads rails, so this initializer tries to fetch :check_yarn_integrity, which loads `config/webpacker.yml`. However, this doesn't exist until after we run this rake task.

I had a chuckle at the error message suggesting I run `rails webpacker:install`:
```
$ bundle exec rails webpacker:install
RAILS_ENV=development environment is not defined in config/webpacker.yml, falling back to production environment
rails aborted!
Webpacker configuration file not found /Users/ahoyt/code/tiu/pd-tracker/config/webpacker.yml. Please run rails webpacker:install Error: No such file or directory @ rb_sysopen - /Users/ahoyt/code/tiu/pd-tracker/config/webpacker.yml
```

We could update the [installation instructions](https://github.com/rails/webpacker#installation) to:
```
rm yarn.lock
bundle
bundle exec rails webpacker:install
```

...but that seems hacky. I like the idea of this initializer being smarter. I haven't learned enough about it to suggest how.

mcmire:
I would love to see this revisited. If anything, I feel like `yarn.lock` should be checked on running `rails assets:precompile` (and, as mentioned, on running `bin/webpack` or `bin/webpack-dev-server`). Otherwise, it doesn't really matter, since Webpack is run in a separate process.

aliraxa-max:
How I can resolve this problem,
=> Booting Puma
[...]

PR #1712, opened 2018-09-25
check if `webpacker.yml` exists before check_yarn_integrity?

key-sn:
If you already have `yarn.lock` and you want to add webpacker later, that is, if you add `webpacker` to your Gemfile, `rails new webpack:install` fails. 

https://github.com/rails/webpacker/blob/48d9fd52c3e9637dbb21e551b48c84c0f2dbb2ed/lib/webpacker/railtie.rb#L32
`Webpacker.config.check_yarn_integrity?` happens to load` config / webpacker.yml`, the following error occurs.
https://github.com/rails/webpacker/blob/48d9fd52c3e9637dbb21e551b48c84c0f2dbb2ed/lib/webpacker/configuration.rb#L82
```
      raise "Webpacker configuration file not found #{config_path}. " \
            "Please run rails webpacker:install " \
            "Error: #{e.message}"
```

So, I change to check if `webpacker.yml` exists before `Webpacker.config.check_yarn_integrity?`

### DIFF
--- a/lib/webpacker/railtie.rb
+++ b/lib/webpacker/railtie.rb
@@ -29,7 +29,7 @@ class Webpacker::Engine < ::Rails::Engine
   #     - edit config/environments/production.rb
   #     - add `config.webpacker.check_yarn_integrity = true`
   initializer "webpacker.yarn_check" do |app|
-    if File.exist?("yarn.lock") && Webpacker.config.check_yarn_integrity?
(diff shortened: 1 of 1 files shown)


gauravtiwari:
thanks :)

ansonhoyt:
Nice, clean fix! I ran into this problem a couple weeks ago: https://github.com/rails/webpacker/issues/1135#issuecomment-419453065

PR #1716, opened 2018-09-29
Update hello_erb.js.erb

Kastrcn:


ytbryan:
<%% is literal tag delimiters I don't think you should remove it.

https://puppet.com/docs/puppet/5.3/lang_template_erb.html#literal-tag-delimiters

seuros:
@Kastrcn  `<%%` is used there because it part of a comment.  It should not be converted .

gauravtiwari:
closing, thanks @ytbryan @seuros

PR #1729, opened 2018-10-03
Set NODE_ENV to test instead of development.

seuros:


lib/install/config/webpack/test.js
```diff
@@ -1,4 +1,4 @@
-process.env.NODE_ENV = process.env.NODE_ENV || 'development'
+process.env.NODE_ENV = process.env.NODE_ENV || 'test'
```
danielpowell4: having the default fallback be `test`, which is neither `production` (package.json dependencies only) or `development` (package.json dependencies + devDependencies), doesn't make sense.

If you want this for your projects I'd recommend setting it as `NODE_ENV=test` from the command line or via a ENV var

Issue #1622, opened 2018-07-24
Provide SplitChunksPlugin documentation


tinderfields:
The configuration docs at https://github.com/rails/webpacker/blob/master/docs/webpack.md currently refer to CommonsChunkPlugin as the way to deduplicate your js packs.

Since Webpack 4 "the CommonsChunkPlugin was removed in favor of optimization.splitChunks": https://webpack.js.org/plugins/split-chunks-plugin/

The documentation link above should be updated to reflect that change as webpacker now supports webpack 4.

tinderfields:
This will involve changes to the `javascript_pack_tag` as it doesn't seem to support `entrypoints` entries in the manifest. 

The may not be the most 'railsy' way but here is how I got it working:

```javascript
// config/webpack/split_chunks.js
module.exports = {
  optimization: {
    splitChunks: {
      chunks: 'all'
    }
  }
}

// config/webpack/environment.js
const splitChunks = require('./split_chunks')
const WebpackAssetsManifest = require('webpack-assets-manifest')

environment.config.merge(splitChunks)

// Should override the existing manifest plugin
environment.plugins.insert(
  'Manifest',
  new WebpackAssetsManifest({
    entrypoints: true, // default in rails is false
    writeToDisk: true, // rails defaults copied from webpacker 
    publicPath: true // rails defaults copied from webpacker 
  })
)
```

```ruby
# app/helpers/application_helper.rb
module ApplicationHelper
  def javascript_entrypoint_pack_tag(*names, **options)
    javascript_include_tag(*sources_from_pack_manifest_entrypoints(names, type: :javascript), **options)
  end

private
  def sources_from_pack_manifest_entrypoints(names, type:)
    names.map do |name|
      Webpacker.manifest.lookup!('entrypoints')[name][manifest_type(type)]
    end.flatten
  end

  def manifest_type(type)
    case type
    when :javascript then 'js'
    when :stylesheet then 'css'
    else type.to_s
    end
  end
end

# view file
= javascript_entrypoint_pack_tag 'pack_name'
```

tinderfields:
I also found the need to define:
```ruby
  def stylesheet_entrypoint_pack_tag(*names, **options)
    unless Webpacker.dev_server.running? && Webpacker.dev_server.hot_module_replacing?
      stylesheet_link_tag(*sources_from_pack_manifest_entrypoints(names, type: :stylesheet), **options)
    end
  end
```

stefanvermaas:
[...]

PR #1735, opened 2018-10-05
Have the *_pack_tag support entrypoints

stefanvermaas:
This addresses #1622 by adding support and documentation for the `splitChunks` functionality of `Webpack 4.0`. I'm not entirely sure whether this is the best way to add support for it, so feedback is welcome and appreciated. 

The downside of this approach is that you need to add `split_chunk: true` to your `*_pack_tag` options list to make it work. If it's possible to do this in a more neat way; please tell me. It would be wonderful if the next version of `Webpacker` is entirely ready for this kind of functionality. 

Hope this helps @gauravtiwari releasing the final release of Webpacker, that is compatible with `Webpack 4.0`!

### DIFF
--- a/docs/webpack.md
+++ b/docs/webpack.md
@@ -256,6 +256,47 @@ const { environment } = require('@rails/webpacker')
 environment.resolvedModules.append('vendor', 'vendor')
 ```
 
+### Add SplitChunks (Webpack V4)
+Originally, chunks (and modules imported inside them) were connected by a parent-child relationship in the internal webpack graph. The CommonsChunkPlugin was used to avoid duplicated dependencies across them, but further optimizations were not possible
+
+Since webpack v4, the CommonsChunkPlugin was removed in favor of optimization.splitChunks.
+
+For the full configuration options of SplitChunks, see the [Webpack documentation](https://webpack.js.org/plugins/split-chunks-plugin/).
+
+```js
+// config/webpack/environment.js
+const WebpackAssetsManifest = require('webpack-assets-manifest');
+
+const splitChunks = {
+  optimization: {
+    splitChunks: {
+      chunks: 'all'
+    },
+  },
+};
+
+environment.config.merge(splitChunks);
+
+// Should override the existing manifest plugin
+environment.plugins.insert(
+  'Manifest',
+  new WebpackAssetsManifest({
+    entrypoints: true, // default in rails is false
+    writeToDisk: true, // rails defaults copied from webpacker
+    publicPath: true // rails defaults copied from webpacker
+  })
+)
+```
+
+The `javascript_pack_tag` and `stylesheet_pack_tag` doesn't know (by default) whether they should use the entrypoints or that it should use the normal way of adding a pack. To make sure the `*_pack_tag` works with `SplitChunks`, you should use the following syntax;
+
+```erb
+javascript_pack_tag "your-entrypoint-javascript-file.js", split_chunks: true
+stylesheet_pack_tag "your-entrypoint-stylesheet-file.css", split_chunks: true
+```
+
+For the old configuration with the CommonsChunkPlugin see below
+
 ### Add common chunks
 
 The CommonsChunkPlugin is an opt-in feature that creates a separate file (known as a chunk), consisting of common modules shared between multiple entry points. By separating common modules from bundles, the resulting chunked file can be loaded once initially, and stored in the cache for later use. This results in page speed optimizations as the browser can quickly serve the shared code from the cache, rather than being forced to load a larger bundle whenever a new page is visited.
--- a/docs/webpack.md
+++ b/docs/webpack.md
@@ -288,11 +288,11 @@ environment.plugins.insert(
 )
 ```
 
-The `javascript_pack_tag` and `stylesheet_pack_tag` doesn't know (by default) whether they should use the entrypoints or that it should use the normal way of adding a pack. To make sure the `*_pack_tag` works with `SplitChunks`, you should use the following syntax;
+To use the `javascript_pack_tag` or the `stylesheet_pack_tag` with `SplitChunks` or `RuntimeChunks` you can refer to the packs as usual.
 
 ```erb
-javascript_pack_tag "your-entrypoint-javascript-file.js", split_chunks: true
-stylesheet_pack_tag "your-entrypoint-stylesheet-file.css", split_chunks: true
+javascript_pack_tag "your-entrypoint-javascript-file"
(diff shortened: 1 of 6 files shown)


gauravtiwari:
thanks @stefanvermaas ❤️  will take a look over the weekend.

jakeNiemiec:
Note that [splitChunks fully replaced CommonsChunkPlugin in webpack 4.0](https://webpack.js.org/plugins/commons-chunk-plugin/). Additionally, it could be useful to be familiar with what it means to set a [runtimeChunk](https://webpack.js.org/configuration/optimization/#optimization-runtimechunk).

stefanvermaas:
@gauravtiwari awesome 👏 

@jakeNiemiec thanks! I already added the replacement notice to the new docs. 

Would you mind being more specific about the runtimeChunk part? I haven’t played around with the runtimeChunk so far.

jakeNiemiec:
>Would you mind being more specific about the runtimeChunk part? I haven’t played around with the runtimeChunk so far.

As far as I can tell, it is a [formal version of the 'manifest' chunk for the CommonsChunkPlugin](https://github.com/rails/webpacker/blame/9d7363bda220899ec7321802541257dbd0fb62ed/docs/webpack.md#L323). 


It's everything webpack needs to be webpack. Personally, we use it to load a 'critical' entrypoint that contains the minimal JS for every page (error handling etc).
image

gauravtiwari:
Sorry it took me a while. 

I am wondering, if it would make sense to enable this by default without using `split_chunks` option. We can check if "entrypoints" hash exist and includes chunk for a pack? If it does, we can send that along with the main pack.

gauravtiwari:
Perhaps then, we can also do same for runtime chunk too, if enabled.

stefanvermaas:
Hi @gauravtiwari,

I guess it could make sense to enable it by default and checking for the runtime and entrypoints in the hash would definitely be an option. I'll provide a new version of the PR tomorrow.

Thanks for looking into this.

docs/webpack.md
```diff
@@ -256,6 +256,47 @@ const { environment } = require('@rails/webpacker')
...
+```
+
+The `javascript_pack_tag` and `stylesheet_pack_tag` doesn't know (by default) whether they should use the entrypoints or that it should use the normal way of adding a pack. To make sure the `*_pack_tag` works with `SplitChunks`, you should use the following syntax;
+
+```erb
+javascript_pack_tag "your-entrypoint-javascript-file.js", split_chunks: true
+stylesheet_pack_tag "your-entrypoint-stylesheet-file.css", split_chunks: true
+```
+
+For the old configuration with the CommonsChunkPlugin see below
```
jakeNiemiec: You might want to remove this since you can't use CommonsChunkPlugin with webpack 4+

stefanvermaas: Wouldn’t be a depreciation notice be a better option here? 

The publicly available version of webpacker is still the version that is not compatible with webpack 4. 

I would suggest to add a deprecation notice until the webpack 4 version of webpacker is out of beta. As Github is the only source for documentation. 

But; I’m not the expert on maintaining large OS projects like these, so I’m fine with both options.

stefanvermaas:
Hi @gauravtiwari and @jakeNiemiec, I've updatet the PR. 

You don't have to add the `split_chunks: true` anymore to the `*_pack_tag` to make it work with `SplitChunks` or `RuntimeChunks` (as requested). Both the `SplitChunks` and the `RuntimeChunks` end up in the `entrypoints` object of the manifest and `Webpacker` can now handle both cases.

Using the `begin/rescue` isn't the most elegant solution, but using several `if/else` statements wasn't elegant either. When there is no `entrypoints` key in the `manifest` it will raise a `NoMethodError`, when the pack isn't defined in the `entrypoints` object it will raise a `NoMethodError` and thus it seems reasonable to use this approach.

I'm looking forward to your feedback.

stefanvermaas:
I see I didn't specify any tests for the `RuntimeChunks` (tested it locally in a new Rails app), do you want me to add a test for the `RuntimeChunks` too?

stefanvermaas:
Hi @gauravtiwari and @jakeNiemiec,

Any chance one of you could take a look at this? It would be great to have these changes merged into webpack.

Cheers (and enjoy your weekend!).

ratenbuuren:
Any updates on this PR? Would love to see this feature integrated 🥇

mathewbaltes:
Agreed, this would be very useful to many people.

gauravtiwari:
Thanks, everyone 🙏 will take a look today and get this merged.

lib/webpacker/manifest.rb
```diff
@@ -18,13 +18,29 @@ def refresh
...
+    # an extra object called "entrypoints". When the entrypoints key is not
+    # present in the manifest, or the name is not found in the entrypoints hash,
+    # it will raise a NoMethodError. If this happens, we should try to lookup
+    # a single instance of the pack based on the given name.
+    begin
+      manifest_pack_type = manifest_type(pack_type[:type])
+      manifest_pack_name = manifest_name(name, manifest_pack_type)
+
+      # Lookup the pack in the entrypoints of the manifest
+      find("entrypoints")[manifest_pack_name][manifest_pack_type]
```
gauravtiwari: Perhaps, we can use `dig` here then? https://ruby-doc.org/core-2.3.0_preview1/Hash.html#method-i-dig

gauravtiwari:
Thanks, @stefanvermaas for the wait and PR 🙏 

I might tweak this a bit in another PR I am working relating to split chunks.

stefanvermaas:
Thanks for reviewing and merging this @gauravtiwari!! If you need an extra pair of hands for the tweaks, just let me know. More than happy to help!

gauravtiwari:
Thanks, @stefanvermaas Great work, everything working well. Perhaps, let's enable entrypoint by default. 

I am thinking to unify CSS as well so we don't have to write extra `stylesheet_pack_tag`. The `javascript_pack_tag` should include all compiled resources required for it to render.

gauravtiwari:
But I guess that might not fit everyone's use case, perhaps we enable above by default and allow a way to turn it off. `<%= javascript_pack_tag 'hello_vue', include_css: false %>`

zernie:
@gauravtiwari considering that stylesheets usually go to `head`, and scripts to the bottom of `body` tag, I don't think it's a good default option.

gauravtiwari:
Yep that makes sense, was thinking in terms of efficiency (if there are a lot of packs) but I guess that might complicate things as you have pointed out.

Issue #1723, opened 2018-10-01
PostCSS config file should be provided in loader


renchap:
`postcss-loader`'s `options.config.path` has been removed since beta 2.

This breaks compilation when you import some CSS files from `node_modules`. For example I am getting:
```
ERROR in /home/xxx/www/xxx/shared/node_modules/react-calendar/dist/Calendar.css
Module build failed (from /home/xxx/www/xxx/shared/node_modules/mini-css-extract-plugin/dist/loader.js):
ModuleBuildError: Module build failed (from /home/xxx/www/xxx/shared/node_modules/postcss-loader/src/index.js):
Error: No PostCSS Config found in: /home/xxx/www/xxx/shared/node_modules/react-calendar/dist
```

This is fixed by providing the `config.path` option to the loaded.

There have been a few PR around this:
- https://github.com/rails/webpacker/pull/1676 tries to use the default `postcss.yml` when not present
- https://github.com/rails/webpacker/pull/1660 removed the config option

Ping @JakeLaCombe as you worked on this.

aganov:
The problem is that the `node_modules` directory is often **symlinked** with capistrano based deployment and `postcss` is seeking its config file for every import you've specified in your bundles. You can either copy `.postcssrc.yml` to `/var/www/app/shared/` during deployment, but this is a temporary **DIRTY** fix.

```ruby
execute :cp, File.join(release_path, ".postcssrc.yml"), shared_path
```

More cleaner solution would be to fetch postcss loader and set `config.path` option to it.

In your `config/webpack/environment.js`

```js
const path = require("path")
Object.assign(environment.loaders.get("css").use.find(el => el.loader === "postcss-loader").options, {
  config: {
    path: path.resolve(__dirname, "../..", ".postcssrc.yml")
  }
})
```

JakeLaCombe:
In my opinion, at that point I think you are better to use a straight up JSON object to store the postcss configuration as opposed to using a .postcssrc.yml file.

```
const postCssOptions = {
  ident: 'postcss',
  plugins() {
    return[
      Autoprefixer({ browsers:[
        'iOS >= 10',
        'last 2 and_chr versions',
        'last 2 android versions',
        'last 2 chrome versions',
        'last 2 edge versions',
        'last 2 firefox versions',
        'ie 11',
        'last 2 safari versions',
      ]})
    ];
  }
};

environment.loaders.forEach((loader) => {
   loader.value.use.forEach((use) => {
[...]

PR #1738, opened 2018-10-06
Add PostCSS config path in postcss-loader options.

jpickwell:
Solves an issue with symlinked node_modules directory.


lakim:
Solves https://github.com/rails/webpacker/issues/1723

gauravtiwari:
thanks :)

Issue #1611, opened 2018-07-12
Is modification time the best default to calculate the compilation digest?


oriolbcn:
https://github.com/rails/webpacker/pull/1522 introduced a neat change, to be able to calculate the compilation digest based on the contents of the file instead of the modification time.

Now I wonder: Wouldn't that calculation (based on contents) be a better default than the modification time? In the end, the contents of the file is what matters and what will always be 100% accurate. The modification time may or may not reflect a change in the file.

Specifically, this doesn't work well with Capistrano, which is probably one of the most (the most?) popular deployment tool for Rails apps. With Capistrano, every time you make a new deploy, it copies over all the app files to a new folder, which makes the files have a different modification time, which defeats the purpose of the digest. Copying files is one of the scenarios when the modification time does not work as an indication of file modification, there may be others.

I understand that currently it is done this way for performance reasons, but is it worth compromising the usefulness of the digest due to some (not a lot, I guess) speed? What I propose is to make the calculation based on the contents by default and maybe ditch the other one completely or make it configurable with a different environment variable or option (like "FASTER_DIGESTS" or something).

jpickwell:
relates to #1439

gauravtiwari:
@oriolbcn Thanks, that makes sense. 

`mtime` works pretty well in development, in fact, it's not needed if one is using dev server or watcher. How about if we keep the `mtime` for on-demand compilation in development but use file digest for precompilation in other environments?

oriolbcn:
Yep sounds good I like the approach it would be clearer than now.

grk:
After https://github.com/rails/webpacker/commit/95c00e23b98397706bece763b3f45421821efcb7 this can proably be closed.

PR #1743, opened 2018-10-10
Always check files recompilation by content

### LINKED ISSUE #1439: Recompile only modified assets in production
Working in dev is great. Webpack recompiles and dev-server send to browser only changed files.

But in production it occurs to recompile ALL files. And it WAY TOO LONG.

Is there any method to recompile ONLY CHANGED files in production?

mdesantis:
Fix #1439

### DIFF
--- a/lib/webpacker/compiler.rb
+++ b/lib/webpacker/compiler.rb
@@ -45,11 +45,7 @@ class Webpacker::Compiler
 
     def watched_files_digest
       files = Dir[*default_watched_paths, *watched_paths].reject { |f| File.directory?(f) }
-      file_ids = if ENV["CI"]
-        files.sort.map { |f| "#{File.basename(f)}/#{Digest::SHA1.file(f).hexdigest}" }
-      else
-        files.map { |f| "#{File.basename(f)}/#{File.mtime(f).utc.to_i}" }
-      end
+      file_ids = files.sort.map { |f| "#{File.basename(f)}/#{Digest::SHA1.file(f).hexdigest}" }
       Digest::SHA1.hexdigest(file_ids.join("/"))
     end
 


mdesantis:
@gauravtiwari I thought about that too but I'm worried about CI: it's unlikely that you run tests under production environment. What about doing `CI env var check || env.production?`?

Also, an alternative fix could be just documenting the CI env var in the readme (and maybe renaming that; I don't think it's descriptive enough); developers might just be confused by the fact that this behaviour isn't written loud and clear

latortuga:
As someone who is not super familiar with webpacker internals, forgive me if this is a dumb question. Is this codepath even used in development? My impression is that this feature is for when assets are precompiled, which doesn't happen in development mode.

mdesantis:
@latortuga (not a Webpacker expert neither) but you can precompile assets even in development if for some reason you don't want to use `webpack-dev-server`, although it isn't very common; and anyway you precompile them in test mode, if we are considering `development` as `non-produciton`

swrobel:
> I thought about that too but I'm worried about CI: it's unlikely that you run tests under production environment. What about doing `CI env var check || env.production?`?

Please see @gauravtiwari's [comment here](https://github.com/rails/webpacker/issues/1611#issuecomment-417745121) for an example of what he means. Production & Test (or any env other than Development) would use the file hash method.

jpickwell:
Is there a reason why the results of `watched_files_digest` isn't memoized? The function gets called once by `stale?` and again by `record_compilation_digest` (if `stale?` returns `true`) when calling `compile`.

swrobel:
@jpickwell I raised this with my [earlier PR](https://github.com/rails/webpacker/pull/1522#issuecomment-394158222) and got the 👍 but didn't get around to implementing it.

gauravtiwari:
Thanks everyone 🙏 

Lets use `mtime` in development and file hash for other environments, where most likely we would be precompiling assets.

**Why in development?**

Mainly because, Webpacker ships with this "naive" on-demand compiler, which works well for simple cases and we don't want to slow it down, even if it's tiny by using file hash. If I am not missing anything, I think `mtime` should work fine there? 

----
But as I wrote the above paragraph, I realised for simple cases using file hash wouldn't be slow either. So, perhaps just use file hash for everything to keep things simple?

gauravtiwari:
Thanks @mdesantis :)

kjleitz:
Any chance this will be put into 3.x? This fix would _really_ help our deployment process out. Deployment went from 3-4 minutes to ~15 minutes for us. It's a big pain.

latortuga:
@kjleitz looks like this got released on March 6, 2019: https://github.com/rails/webpacker/commit/8c25c092dede858d8c5ba2aa708e0134216dc8b0

:tada:

Issue #1410, opened 2018-04-05
Add equivalent task to "rake assets:clean"


swrobel:
I don't see any straightforward way to remove all of the outdated packs from `public/packs[-test]` ... am I missing something?

swrobel:
Thanks, I'm aware of that one, but that's like `assets:clobber`, not `assets:clean`.

swrobel:
For context, Heroku's ruby buildpack depends on assets:clean to make sure slug sizes don't get too big. https://github.com/heroku/heroku-buildpack-ruby/blob/v177/lib/language_pack/rails4.rb#L92

cc @schneems

investment-radar:
And for me another thing I want just like assets:clean, by default it will keep 3 versions of the assets files.

swrobel:
possible solution: https://github.com/heroku/heroku-buildpack-ruby/pull/803#issuecomment-428772751

ericboehs:
I've started on a solution here: https://github.com/rails/webpacker/compare/master...ericboehs:eb-4-clean-task

ytbryan:
looking good! Send a PR!

ericboehs:
I don't think it's ready. I believe it needs to keep recent versions of the assets or active user sessions will break after a deploy.

ericboehs:
[It's ready](
https://github.com/rails/webpacker/pull/1744). Or as close to ready as I'm going to get it without some input. cc: @ytbryan @swrobel

NielsKSchjoedt:
+1

henrypoydar:
+1

NielsKSchjoedt:
Looks like there as a cleanup plugin introduced for Webpacker itself, I added this, and it works great! It's a more integrated way of cleaning up those assets instead of relying on rails doing it. https://github.com/johnagan/clean-webpack-plugin

PR #1744, opened 2018-10-11
Webpack Clean Task

ericboehs:
Resolves https://github.com/rails/webpacker/issues/1410

 ## Goal
Add a task to clean old packs so that our packs directory doesn't get bloated with hundreds of old, unused assets.

 ## Approach
Create a new `webpacker:clean` task (which enhances the `assets:clean` task) to delete unused assets by detecting old assets (using the Webpack-generated manifest.json to find alternative versions and by using `File.mtime` to determine file age).

 ## Context
Users' browsers of a Webpacker-enabled app will request packs from the public packs directory. During a deploy the packs are replaced with a new version (at a new location) that the browser may not yet know about which results in 404s or other errors to the end user. Best practices recommend to keep a version or two of your assets around to avoid this scenario. Currently, if the packs directory is persisted across deploys, **all compiled packs are kept forever**. Sprockets has an `assets:clean` task which removes unused assets. This pull request intends to add the equivalent task to Webpacker.

 ## Notes
I'm not 100% confident with how I find the older versions but it's the best I can come up with. I'm open to suggestions.

 ## Related
* https://github.com/heroku/heroku-buildpack-ruby/pull/803

### DIFF
--- a/lib/tasks/webpacker.rake
+++ b/lib/tasks/webpacker.rake
@@ -2,6 +2,7 @@ tasks = {
   "webpacker:info"                    => "Provides information on Webpacker's environment",
   "webpacker:install"                 => "Installs and setup webpack with Yarn",
   "webpacker:compile"                 => "Compiles webpack bundles based on environment",
+  "webpacker:clean"                   => "Remove old compiled webpacks",
   "webpacker:clobber"                 => "Removes the webpack compiled output directory",
   "webpacker:check_node"              => "Verifies if Node.js is installed",
   "webpacker:check_yarn"              => "Verifies if Yarn is installed",
--- a/lib/tasks/webpacker/clean.rake
+++ b/lib/tasks/webpacker/clean.rake
@@ -2,8 +2,8 @@ require "webpacker/configuration"
 
 namespace :webpacker do
   desc "Remove old compiled webpacks"
-  task clean: ["webpacker:verify_install", :environment] do
-    Webpacker.clean
+  task :clean, [:keep] => ["webpacker:verify_install", :environment] do |_, args|
+    Webpacker.clean(Integer(args.keep || 2))
   end
 end
 
--- a/lib/tasks/webpacker/clean.rake
+++ b/lib/tasks/webpacker/clean.rake
@@ -1,0 +1,15 @@
+require "webpacker/configuration"
+
+namespace :webpacker do
+  desc "Remove old compiled webpacks"
+  task clean: ["webpacker:verify_install", :environment] do
+    Webpacker.clean
+  end
+end
+
+# Run clean if the assets:clean is run
+if Rake::Task.task_defined?("assets:clean")
+  Rake::Task["assets:clean"].enhance do
+    Rake::Task["webpacker:clean"].invoke
+  end
+end
--- a/lib/webpacker/commands.rb
+++ b/lib/webpacker/commands.rb
@@ -10,8 +10,18 @@ class Webpacker::Commands
 
     files_on_filesystem = Dir.glob("#{config.public_output_path}/**/*").select { |f| File.file? f }
     files_in_manifest = manifest.refresh.values.map { |f| File.join config.root_path, "public", f }
+
+    count_to_keep = 2
+    versions_to_keep = files_in_manifest.flat_map do |file_in_manifest|
+      file_prefix, file_ext = file_in_manifest.scan(/(.*)[0-9a-f]{20}(.*)/).first
+      versions_of_file = Dir.glob("#{file_prefix}*#{file_ext}").grep(/#{file_prefix}[0-9a-f]{20}#{file_ext}/)
+      versions_of_file.map do |version_of_file|
+        [version_of_file, File.mtime(version_of_file).utc.to_i]
+      end.sort_by(&:last).reverse.first(count_to_keep).map(&:first)
+    end
+
     manifest_file = File.join config.public_output_path, "manifest.json"
-    files_to_be_removed = files_on_filesystem - files_in_manifest - [manifest_file]
+    files_to_be_removed = files_on_filesystem - files_in_manifest - versions_to_keep - [manifest_file]
 
     files_to_be_removed.each { |f| File.delete f }
   end
--- a/lib/webpacker/commands.rb
+++ b/lib/webpacker/commands.rb
@@ -5,6 +5,17 @@ class Webpacker::Commands
     @webpacker = webpacker
   end
 
+  def clean
+    return if !config.public_output_path.exist? || !config.public_manifest_path.exist?
+
+    files_on_filesystem = Dir.glob("#{config.public_output_path}/**/*").select { |f| File.file? f }
+    files_in_manifest = manifest.refresh.values.map { |f| File.join config.root_path, "public", f }
+    manifest_file = File.join config.public_output_path, "manifest.json"
+    files_to_be_removed = files_on_filesystem - files_in_manifest - [manifest_file]
+
+    files_to_be_removed.each { |f| File.delete f }
+  end
+
   def clobber
     config.public_output_path.rmtree if config.public_output_path.exist?
     config.cache_path.rmtree if config.cache_path.exist?
--- a/lib/webpacker/commands.rb
+++ b/lib/webpacker/commands.rb
@@ -6,20 +6,20 @@ class Webpacker::Commands
   end
 
   def clean(count_to_keep = 2)
-    return if !config.public_output_path.exist? || !config.public_manifest_path.exist?
+    if config.public_output_path.exist? && config.public_manifest_path.exist?
+      files_in_manifest = manifest.refresh.values.map { |f| File.join config.root_path, "public", f }
+      files_to_be_removed = files_in_manifest.flat_map do |file_in_manifest|
+        file_prefix, file_ext = file_in_manifest.scan(/(.*)[0-9a-f]{20}(.*)/).first
+        versions_of_file = Dir.glob("#{file_prefix}*#{file_ext}").grep(/#{file_prefix}[0-9a-f]{20}#{file_ext}/)
+        versions_of_file.map do |version_of_file|
+          next if version_of_file == file_in_manifest
 
-    files_in_manifest = manifest.refresh.values.map { |f| File.join config.root_path, "public", f }
-    files_to_be_removed = files_in_manifest.flat_map do |file_in_manifest|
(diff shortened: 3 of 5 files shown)


lib/webpacker/commands.rb
```diff
@@ -5,6 +5,27 @@ def initialize(webpacker)
...
+  def clean
+    return if !config.public_output_path.exist? || !config.public_manifest_path.exist?
+
+    files_on_filesystem = Dir.glob("#{config.public_output_path}/**/*").select { |f| File.file? f }
+    files_in_manifest = manifest.refresh.values.map {|f| File.join config.root_path, 'public', f }
+
+    count_to_keep = 2
+    versions_to_keep = files_in_manifest.flat_map do |file_in_manifest|
+      file_prefix, file_ext = file_in_manifest.scan(/(.*)[0-9a-f]{20}(.*)/).first
+      versions_of_file = Dir.glob("#{file_prefix}*#{file_ext}").grep(/#{file_prefix}[0-9a-f]{20}#{file_ext}/)
```
ericboehs: I've added the `grep` here to help filter out mismatches that `Dir.glob` might match. I'm thinking of something like `packs/users-abc123.js` and `packs/users-profile-def456.js`.

I'm trying to think of something that could still match. Maybe `users-baddadbabebaddadbabe-abc123.js` (i.e. the dev named the file something that matches the digest character space and length).

ericboehs:
In a previous commit, I removed all files in the directory that we didn't explicitly know about. I think it's better to build up a list of files that we know are ours and remove those. I left the previous commit intact for historical purposes.

gauravtiwari:
Thanks @ericboehs Will review this tomorrow :)

ericboehs:
I think I also need to delete the`.js.map` and `.js.gz` files. Doing more testing now.

ericboehs:
Looks like the current code also deletes the `.js.map` and `.js.gz` files.

lib/tasks/webpacker.rake
```diff
@@ -2,6 +2,7 @@ tasks = {
   "webpacker:info"                    => "Provides information on Webpacker's environment",
   "webpacker:install"                 => "Installs and setup webpack with Yarn",
   "webpacker:compile"                 => "Compiles webpack bundles based on environment",
+  "webpacker:clean"                   => "Remove old compiled webpacks",
```
swrobel: Should probably be "Removes" to be consistent with other task descriptions

swrobel: Also would be great if this took a `keep` arg like the `assets:clean` rake task does : https://github.com/rails/sprockets/blob/master/lib/sprockets/manifest.rb#L245

lib/webpacker/commands.rb
```diff
@@ -5,6 +5,24 @@ def initialize(webpacker)
     @webpacker = webpacker
   end
 
+  def clean
+    return if !config.public_output_path.exist? || !config.public_manifest_path.exist?
+
+    count_to_keep = 2
+    files_in_manifest = manifest.refresh.values.map { |f| File.join config.root_path, "public", f }
+    files_to_be_removed = files_in_manifest.flat_map do |file_in_manifest|
```
swrobel: Might be worth comparing to the logic that Sprockets uses: https://github.com/rails/sprockets/blob/master/lib/sprockets/manifest.rb#L245

lib/tasks/webpacker.rake
ericboehs: I'm mimicking `rails assets:clean` for my wording.

I've added the `keep` arg to the task. Good idea!

lib/webpacker/commands.rb
ericboehs: I've compared the logic and I believe it's fairly identical. They're going about getting the files differently though since they keep old version names in the manifest whereas Webpack does not.

lib/webpacker/commands.rb
```diff
@@ -5,6 +5,23 @@ def initialize(webpacker)
     @webpacker = webpacker
   end
 
+  def clean(count_to_keep = 2)
+    return if !config.public_output_path.exist? || !config.public_manifest_path.exist?
```
gauravtiwari: @ericboehs I am wondering if we do this other way around for readability. 

```rb
if config.public_output_path.exist? && config.public_manifest_path.exist?
  # do the cleaning
end
```

ericboehs: RuboCop usually barks at wrapping a whole method in a conditional and recommends using a guard clause such as this. I thought the way I wrote it was clear enough (I avoided using `unless`). Would it be easier to understand if I expanded it to two guard clauses?

```ruby
return unless config.public_output_path.exist?
return unless config.public_manifest_path.exist?

# ...
```

swrobel: Why not `return unless config.public_output_path.exist? && config.public_manifest_path.exist?`. Reads clearly to me.

ericboehs: I usually stay away from compound `unless` statements but if I'm the oddball on thinking those are confusing I'd be glad to change it to your suggestion @swrobel.

gauravtiwari: Thanks @ericboehs and @swrobel 

The reason mainly is that we don't use/prefer early returns inside Webpacker or Rails codebase. Rubocop shouldn't complain if it's multi-line conditional right?

ericboehs: Okay. I updated it. I didn't realize the Rails style guide differed from the Ruby style guide here.

ericboehs:
@gauravtiwari Would you have time to review this soon?

Thanks!

lib/tasks/webpacker.rake
ytbryan: I think the verbs of webpacker's tasks should follow singular verbs of `rails assets`. Send a separate PR?

gauravtiwari:
Thanks, @ericboehs for this, looks great. I am wondering if you would be happy to add some test for this? Something along this line: https://github.com/rails/sprockets/blob/master/test/test_manifest.rb#L415

lib/tasks/webpacker/clean.rake
```diff
@@ -0,0 +1,15 @@
...
+  desc "Remove old compiled webpacks"
+  task :clean, [:keep] => ["webpacker:verify_install", :environment] do |_, args|
+    Webpacker.clean(Integer(args.keep || 2))
+  end
+end
+
+# Run clean if the assets:clean is run
+if Rake::Task.task_defined?("assets:clean")
+  Rake::Task["assets:clean"].enhance do
+    Rake::Task["webpacker:clean"].invoke
```
lazyatom: It would be a very nice touch if we could pass on the same arguments given to `assets:clean`; I'm not sure Rake provides a particularly clean way of doing this though; my best approach so far has been this:

```ruby
Rake::Task["assets:clean"].enhance do |_, args|
  Rake::Task["webpacker:clean"].invoke(*args.to_a)
end
```

lazyatom:
Another potential issue here is that compressed files (generated in production) aren't included in the manifest, so if we want them to be cleaned, we'll also need to check explicitly for them.

phlegx:
This is a part of my manifest file:

```json
{
  "entrypoints": {
    "frontend": {
      "css": [
        "/packs/css/application-34645e16.css"
      ],
      "js": [
        "/packs/js/application-c7b3dd68a6ac50abc8ca.js"
      ],
      "css.map": [
        "/packs/css/application-34645e16.css.map"
      ],
      "js.map": [
        "/packs/js/application-c7b3dd68a6ac50abc8ca.js.map"
      ]
    }
  },
  "frontend.css": "/packs/css/application-34645e16.css",
  "frontend.css.map": "/packs/css/application-34645e16.css.map",
  "frontend.js": "/packs/js/application-c7b3dd68a6ac50abc8ca.js",
  "frontend.js.map": "/packs/js/application-c7b3dd68a6ac50abc8ca.js.map",
  ...
```

And I see that only JS files are cleaned and all other files not, because the JS files has a 20 hash length and all other files (css, images, fonts, etc.) only a 8 hash length. Rake task `assets:clean` does not clean files in `/public/packs` directory, because the files are generated by webpacker.

Here my clean code:

```ruby
    def clean(count_to_keep = 1)
      if config.public_output_path.exist? && config.public_manifest_path.exist?
        files_in_manifest = manifest.refresh.values.map { |f| File.join config.root_path, "public", f if f.is_a?(String) }.compact
        files_to_be_removed = files_in_manifest.flat_map do |file_in_manifest|
          file_prefix, file_ext = file_in_manifest.scan(/(.*)(?:[0-9a-f]{8}|[0-9a-f]{20})(.*)/).first
          versions_of_file = Dir.glob("#{file_prefix}*#{file_ext}").grep(/#{file_prefix}(?:[0-9a-f]{8}|[0-9a-f]{20})#{file_ext}/)
          versions_of_file.map do |version_of_file|
            next if version_of_file == file_in_manifest

            [version_of_file, File.mtime(version_of_file).utc.to_i]
          end.compact.sort_by(&:last).reverse.drop(count_to_keep).map(&:first)
        end

        files_to_be_removed.each { |f| File.delete f }
      end
    end
```

The regular expression is the important part: `(?:[0-9a-f]{8}|[0-9a-f]{20})`

drewB:
I am curious if there has been anymore movement on this.

beauraF:
We are very interested in that, here too.

jakeNiemiec:
Webpack has changed since the creation of this PR.

@drewB @beauraF If you want this functionality, you can use https://github.com/johnagan/clean-webpack-plugin#about

That plugin would be passed into webpacker according to https://github.com/rails/webpacker/blob/0feb1474a7e06ac2a6bff745475bee192643a29f/docs/webpack.md#plugins

🤔 Is there any reason _not_ to include `clean-webpack-plugin` by default?

HarrisonB:
@jakeNiemiec I would vote against inclusion by default. The behavior of `clean-webpack-plugin` is closer to `webpacker:clobber`. One of the nice things about `assets:clean` (and what this PR is suggesting) is it maintains a limited number of (recent) old versions of bundled assets. This is great when your web servers/workers have non-trivial cutover time and users could possibly fetch the previous deploy's assets.

gauravtiwari:
Yes, makes sense. Merging....

Thanks @ericboehs 🙏  sorry it took ages.

viniciusgama:
@ericboehs I was just using this but I noticed it does not remove files that are not referenced in the manifest file. I saw your comment above mentioning you remove this option but can you explain why is that? Looks like a desired behaviour for me.

Thanks,

PR #1747, opened 2018-10-11
Fix undefined method `watch' for Spring:Module

zernie:
This PR fixes `undefined method watch for Spring:Module` error.


danielpowell4:
@zernie are you interested in getting the build to pass by appeasing the linter?

Would be appreciated!

zernie:
@danielpowell4 yeah, sure. I've pushed the fix.

danielpowell4:
@gauravtiwari though small this is a fairly decent sized issue with the `.pre.3`, at least for my app in development

Otherwise, I need to add the following patch inside somewhere like development.rb

```
require 'spring/watcher' if defined?(::Spring)
```

gauravtiwari:
merged, thanks :)

PR #1749, opened 2018-10-11
Backport SHA digest file checking to 3-x-stable

ericboehs:
See https://github.com/rails/webpacker/pull/1743

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -117,6 +117,3 @@ environment.plugins.append('ContextReplacement',
   )
 )
 ```
-
-## Compilation triggered even when files haven't changed
-Webpacker compares the modified times (mtimes) of files to see if they have changed & determine whether to recompile on-demand. For CI environments that clone the repository before running tests, the modified times will be set to when the repository was cloned, not when the file was last modified. To prevent unnecessary recompilation on CI, Webpacker watches for the `CI` environment variable, and instead compares the SHA hashes of all files to determine if they've actually changed. You can enable this behavior outside of CI environemnts, or in those that don't set the `CI` env var, by running `CI=true bin/webpack` (or `webpack-dev-server`).
--- a/lib/webpacker/compiler.rb
+++ b/lib/webpacker/compiler.rb
@@ -44,11 +44,7 @@ class Webpacker::Compiler
 
     def watched_files_digest
       files = Dir[*default_watched_paths, *watched_paths].reject { |f| File.directory?(f) }
-      file_ids = if ENV["CI"]
-        files.sort.map { |f| "#{File.basename(f)}/#{Digest::SHA1.file(f).hexdigest}" }
-      else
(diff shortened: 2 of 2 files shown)


gauravtiwari:
thanks @ericboehs ❤️

mdesantis:
Thank you @ericboehs!!!

kjleitz:
Thanks @ericboehs!

Do you guys know when you might release a new version with the changes in this PR? Over here we run multiple app servers under a load balancer (with a Capistrano-style deployment strategy) so I've been pulling my hair out trying to solve the problem we were having with the fact that our `public/packs/*` fingerprints were _different_ between the servers, whereas our `public/assets/*` fingerprints were all _the same_ between the servers. Lots of inconsistent 404s for webpacker-compiled files as a result (which is never an issue we had with the traditional asset pipeline).

I'd rather lock to a gem version than a git branch, but I guess that's the option I'll have to go with in the meantime. Is there somewhere I can find a release schedule @gauravtiwari? 😃

ericboehs:
> Lots of inconsistent 404s for webpacker-compiled files as a result
Interesting. Thinking about this, it would seem this would be a much wider spread problem. With sprockets the previous 2 versions of an asset were kept around for active users of the site during a deploy. With Webpacker, currently no packs are caching across deploys if the mtime isn't working correctly (all Heroku builds and it sounds like possibly capistrano as well).

Once you update your app to use this fix, you'll have another problem. All versions of an asset is kept forever. This is the existing behavior for users where mtime is working correctly. I have [a pull request](https://github.com/rails/webpacker/pull/1744) open to fix this as well.

> Do you guys know when you might release a new version with the changes in this PR?
It would make sense to wait to deploy both of these fixes together but I'm not a maintainer here so I'm not sure what the plan is.

mdesantis:
@kjleitz why don't you use `CI` env var till the next Webpacker release?

kjleitz:
@mdesantis I'm having trouble finding documentation for the `CI` environment variable you mentioned, or for what it's supposed to help with. Mind pointing me in the right direction?

ericboehs:
@kjleitz See the last paragraph: https://github.com/rails/webpacker/blob/v3.5.5/docs/troubleshooting.md#compilation-triggered-even-when-files-havent-changed

> You can enable this behavior outside of CI environemnts, or in those that don't set the CI env var, by running CI=true bin/webpack

kjleitz:
Ah, thanks @ericboehs. A little bit concerned that the environment variable is named `CI`—potential clashes there in a big app. But thank you for the tip! I'll check for collisions and try it out if there aren't any.

Side note (and I'm sorry, I know this isn't exactly the venue for this, but): it's pretty rough navigating the docs for Webpacker—any reason they aren't all just in `README.md` so their contents could be searchable and visible? Or am I missing something? Because IMHO the current structure really obscures a lot of the caveats and configuration options, if you're coming to the GitHub page for documentation. Just my two cents 😄

PR #1754, opened 2018-10-15
Master install uses github default

justin808:
Newer versions of yarn like

yarn add rails/webpacker

to get the github based package

Tested with Yarn 1.9.4.


justin808:
@gauravtiwari what do you need to get this merged?

ytbryan:
`yarn add https://github.com/rails/webpacker` is working on my end. 

is there a reason to use a shorter version?

justin808:
@ytbryan might be the version of yarn. Can you try mine? Per the docs, I'm pretty sure you can't argue with mine.

gauravtiwari:
Let's keep it as is because this format is supported throughout 1.x versions, don't know if the shorter format is supported in earlier versions of yarn (just in case if someone is still using 1.3 or 1.5)?

PR #1757, opened 2018-10-16
add block context to CSP-related initializer snippets

kjleitz:
These now match the block at README.md:202

The documentation currently assumes the reader will place it within a `#content_security_policy` block. But, without context, a user may mistakenly plant the snippet in `config/initializers/content_security_policy.rb` bare, since in a new Rails app there is no existing block (it's commented out in that file).


gauravtiwari:
thanks :)

PR #1762, opened 2018-10-19
Invoke webpacker:yarn_install before assets:precompile in old rails

printercu:
Make it work similar to new rails, and it's now possible to use node_modules in sprockets.


gauravtiwari:
thanks @printercu

Welziewagers2015:
Run

PR #1764, opened 2018-10-20
Only record compilation digest if webpack runs successfully.

frodsan:
This fixes an issue where current request gets an stale pack,
because previous request wrote the compilation digest before
webpack ends up running.


frodsan:
Here is a gif explaining the issue:

before

And here is another gif showing how it works after these changes:

after

gauravtiwari:
thanks :)

PR #1771, opened 2018-10-23
Add possible solution for `Webpacker::Manifest::MissingEntryError`

strzibny:


gauravtiwari:
We ship webpack-cli with webpacker so this shouldn't be a problem? Could you please explain what problem you were experiencing?

PR #1788, opened 2018-11-02
Allow commands to run via `yarn`

y-yagi:
Currently, commands(`webpack` and `webpacker-dev-server`) need to exist under the `node_modules/.bin` directory. 

But when Yarn Plug'n'Play using, `node_modules/.bin` is not created.
Ref: https://github.com/yarnpkg/rfcs/blob/master/accepted/0000-plug-an-play.md 

In this case, I think it is better to run the commands via `yarn`.  

For this reason, I added new config key, and if it is specified, run  commands via `yarn`. 

I'm not sure if it is appropriate to set this behavior in the configuration file. 
If env is appropriate, I will update it. Thanks!

### DIFF
--- a/lib/webpacker/dev_server_runner.rb
+++ b/lib/webpacker/dev_server_runner.rb
@@ -45,15 +45,22 @@ module Webpacker
 
       def execute_cmd
         env = Webpacker::Compiler.env
-        cmd = [
-          "#{@node_modules_bin_path}/webpack-dev-server",
-          "--config", @webpack_config
-        ]
+
+        cmd = if node_modules_bin_exist?
+          ["#{@node_modules_bin_path}/webpack-dev-server"]
+        else
+          ["yarn", "webpack-dev-server"]
+        end
+        cmd += ["--config", @webpack_config]
         cmd += ["--progress", "--color"] if @pretty
 
         Dir.chdir(@app_path) do
-          exec env, *cmd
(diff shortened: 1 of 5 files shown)


gauravtiwari:
@y-yagi Thanks for this. Looks great, I am not sure about config option though therefore, I am wondering would it make sense to check file presence and fallback to yarn? (do this implicitely)

```rb
if File.exist?('node_modules/.bin/webpack')
else 
  # try yarn
end
```

y-yagi:
Thanks for your feedback! 
I added config to think that it is confusing when using Yarn Plug'n'Play to break by whether `node_modules` exists or not. 
(This is because it predicts that `node_modules` will remain in many cases when migrating to Yarn Plug'n'Play) 

But this is probably the only problem in the beginning. Otherwise, it makes more sense to check if the file exists. 
I updated to fix that check `node_modules` exists or not.

y-yagi:
Because of the effect that `BigDecimal.new` removed, the build of 4.2 + ruby-head is an fails. This fixes with #1825.

gauravtiwari:
Merged #1825 thanks for the update

gauravtiwari:
If you could rebase with the master it should pass 👍

gauravtiwari:
Great, merged 👍 🍰

y-yagi:
Thanks!

Issue #1615, opened 2018-07-16
Allow environment settings in webpacker.yml with ERB


tamouse:
Follow the conventions of Rails `config_for` that processes ERB settings before running the YAML loader.

ckahle33:
Any update on this?

jasonkarns:
Related to https://github.com/rails/webpacker/issues/956

I also expected ERB parsing in webpacker.yml to work. We need to get the dev server port from `ENV`

SophiaAP:
Similar problem here. Some of our team does not want the javascript compiling in development but since we can't use environment variables in the YAML this is not an option. 

If anyone has advice on a workaround for this, we'd really appreciate it! 

Thanks for all your hard work on this gem.

23tux:
Same issue here: We are running parallel specs and need to suffix the public path with a test env number to avoid clashes between parallel specs on the file system. Any updates on this?

latortuga:
@gauravtiwari since you closed #1801, are you opposed to supporting ERB in the config completely? Is there some other way to support this?

My use-case for ERB in the config is dynamic setting of the dev server port. I run multiple apps in development on my machine and I have to do the dev server dance by shutting down all of them except current each time I want to switch work between apps. It's my impression that this file is intended to be checked into source control so, if so, it should allow configuring from the environment in some way.

schuylr:
My use-case to have ERB is to be able to control the `dev_server: public` endpoint so I can duplicate my Rails app across multiple development domains to behave differently.

gingerlime:
@schuylr same here :( did you find a solution/workaround?

latortuga:
I gave up on this feature and ended up just changing my webpacker-dev-server port in each app I work on. It works fine. I still think this feature might be nice.

gingerlime:
In case it helps anyone, or now [this workaround](https://github.com/rails/webpacker/issues/956#issuecomment-416265099) helped us ... (and there's another [potential workaround](https://github.com/rails/webpacker/issues/956#issuecomment-341484600) but I haven't tried it).

Our use case was to override `dev_server: public` by running webpack dev server with `WEBPACKER_DEV_SERVER_PUBLIC=$SOME_ENV bin/webpack-dev-server`... So for a couple of overrides, I guess that's ok, but feels a bit dirty.

rpbaptist:
It took me some iterations to realize ERB was not supported by this one Rails `.yml` config file. Unexpected indeed.

[...]

PR #1801, opened 2018-11-14
fixes issue #1615 by adding ERB processing to the yml loader

DaKaZ:
This PR allows for ERB inside the `config/webpacker.yml` file.  Previous attempts failed to address the case of using `bin/webpack`, but this PR fixes that by processing the ERB in the yml file and outputing a hidden yml that the node process will look for by default.

### DIFF
--- a/lib/webpacker/configuration.rb
+++ b/lib/webpacker/configuration.rb
@@ -1,3 +1,4 @@
+require "erb"
 require "yaml"
 require "active_support/core_ext/hash/keys"
 require "active_support/core_ext/hash/indifferent_access"
@@ -81,7 +82,7 @@ class Webpacker::Configuration
     end
 
     def load
-      YAML.load(config_path.read)[env].deep_symbolize_keys
+      YAML.load(ERB.new(config_path.read).result)[env].deep_symbolize_keys
 
     rescue Errno::ENOENT => e
       raise "Webpacker configuration file not found #{config_path}. " \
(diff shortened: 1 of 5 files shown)


DaKaZ:
This also addresses issue #956

DaKaZ:
Note, the CI jobs that are failing here are due to a known issue in ruby 2.6.0-preview2 and rubocop 0.60: https://github.com/rubocop-hq/rubocop/issues/6412

gauravtiwari:
We do support env variables for some runtime options. Could you please add more info about your usecase?

DaKaZ:
I think there are hundreds of use cases. In our case we need to set certain settings if we’re on our Jenkins platform and we typically do that by checking for an ENV var.  As seen by the multitude of comments, the rails convention is to support ERB processing in yaml files. This PR makes that possible in a fairly unobtrusive way.  Do you see downside to having this PR merged?

gauravtiwari:
@DaKaZ Thanks for this but I still find this awkward especially, where someone can go run webpack directly without having to run ruby binstubs and in that case the JS side would fail due to embedded erb. 

```
yarn run webpack --config config/webpack/development.js
```

I am mainly concerned that this is a partial implementation and it may surely work for an individual usecase but certainly not for the community. I hope you understand.

DaKaZ:
That is unfortunate and disappointing

mikeyhew:
@gauravtiwari it looks like you're on vacation - hope you're having a great time.

For when you get back: config/webpacker.yml is parsed on the JS side in package/env.js. After the file is read, and before it is parsed, it would be possible to check if the yaml contains ERB tags, and report an error in that case, e.g. "webpacker.yml contains ERB tags, so you need to use the Ruby wrappers around webpack and webpack-dev-server located in the bin directory". It's a compromise, but at least there would be an appropriate error message, and if people don't use ERB in webpacker.yml then it would be a non-issue.

jasonperrone:
Ok, so if you guys won't support ERB in the Webpacker.yml, tell me how to do this. I want to set check_yarn_integrity to false only if my rails server is being started by RubyMine, as opposed to rails s.

PR #1808, opened 2018-11-20
Refactor singleton usage

palkan:
## Context

I'm working on the application consisting of multiple engines. Some of these engines provide user interface.

We wanted to make our engines as isolated as possible (though keeping in the same monorepo); hence we needed our webpack(-er) to be isolated too: separate `package.json`, dev server, compilation process.

It turned out that it could be done with Webpacker if we replace the _global_ `Webpacker.instance` usage with the _local_ instance–i.e. if we "fix" dependency injection.

**NOTE:** this PR should considered as an experimental (though the functionality has been extracted from the production app); I'd like to discuss this approach first of all.

**NOTE 2:** this PR doesn't introduce any backward compatibility changes and only deals with the internals.
 
## Related tickets

https://github.com/rails/webpacker/issues/348

## What's inside?

- `Compiler` class refactored to use `@webpacker` instance instead of `Webpacker` (
https://github.com/rails/webpacker/commit/7a0a58c108e8dbd338702fea0e42dd594a0ba5eb)
- Added `webpacker` option to `DevServerProxy` to make it possible to use with non-default instance (
https://github.com/rails/webpacker/commit/53c8a777df30eb8a8dd6088a2179d766a4e2f617)
- Added explicit `chdir` for scripts (`webpack`, `webpack-dev-server`) to make it possible to run not only from the app root (
https://github.com/rails/webpacker/commit/e2bc2363cccaadd3a9b71092cafcbcd610bc527b)
- Added `current_webpacker` to `Helper` to make it possible to override the default Webpacker (
https://github.com/rails/webpacker/commit/aa2aabd71835261587e7afbcb5dfbb728433a585)
- Added `env_prefix` config param to `dev_server` section of `webpacker.yml` to make it possible to use different env variable for different webpackers (https://github.com/rails/webpacker/pull/1808/commits/32fd72eeb6dcf61cfd09c7465301565712299de5)

## Usage example

Suppose that we have a Rails engine `MyEngine`.

To add an isolated Webpacker configuration to that gem we need:
- Add `config/webpacker.yml` and `config/webpack/*.js` files
- Add `bin/webpack` and `bin/webpack-dev-server` files
- Configure engine's instance:

```ruby
module MyEngine
  ROOT_PATH = Pathname.new(File.join(__dir__, ".."))

  class << self
    def webpacker
      @webpacker ||= ::Webpacker::Instance.new(
        root_path: ROOT_PATH,
        config_path: ROOT_PATH.join("config/webpacker.yml")
      )
    end
  end
end
```

- Configure dev server proxy

```ruby
module MyEngine
  class Engine < ::Rails::Engine
        insert_middleware = begin
                            MyEngine.webpacker.config.dev_server.present?
                          rescue
                            nil
                          end
        next unless insert_middleware

        app.middleware.insert_before(
          0, "Webpacker::DevServerProxy",
          ssl_verify_none: true,
          webpacker: MyEngine.webpacker
        )
      end
  end
end
```

- (optional) If you have multiple webpacker, you would probably want to run multiple dev servers at a time, and hence be able to configure their setting though env vars (we do use env vars within a docker-compose dev env):

```yml
# webpacker.yml
# ...
development:
  # ...
  dev_server:
    env_prefix: "MY_ENGINE_WEBPACKER_DEV_SERVER"
    # ...
```

- Configure helper:

```ruby
require "webpacker/helper"

module MyEngine
  module ApplicationHelper
    include ::Webpacker::Helper

    def current_webpacker
      MyEngine.webpacker
    end
  end
end
```

Now you can use `stylesheet_pack_tag` and `javascript_pack_tag` from within your engine.

- Add Rake task to compile assets in production (`rake my_engine:webpacker:compile`)

```ruby
namespace :my_engine do
  namespace :webpacker do
    desc "Install deps with yarn"
    task :yarn_install do
      Dir.chdir(File.join(__dir__, "../..")) do
        system "yarn install --no-progress --production"
      end
    end

    desc "Compile JavaScript packs using webpack for production with digests"
    task compile: [:yarn_install", :environment] do
      Webpacker.with_node_env("production") do
          if MyEngine.webpacker.commands.compile
            # Successful compilation!
          else
            # Failed compilation
            exit!
          end
      end
    end
  end
end
```

I think, most of this work could be automated via generators in the future, but that's not the scope of this PR: here I want to slightly change the Webpacker design to make it more flexible.

### DIFF
--- a/lib/webpacker/helper.rb
+++ b/lib/webpacker/helper.rb
@@ -1,4 +1,11 @@
 module Webpacker::Helper
+  # Returns current Webpacker instance.
+  # Could be overriden to use multiple Webpacker
+  # configurations within the same app (e.g. with engines)
+  def current_webpacker
+    Webpacker.instance
+  end
+
   # Computes the relative path for a given Webpacker asset.
   # Return relative path using manifest.json and passes it to asset_path helper
   # This will use asset_path internally, so most of their behaviors will be the same.
@@ -11,8 +18,8 @@ module Webpacker::Helper
   #   # In production mode:
   #   <%= asset_pack_path 'calendar.css' %> # => "/packs/calendar-1016838bab065ae1e122.css"
   def asset_pack_path(name, **options)
-    unless stylesheet?(name) && Webpacker.dev_server.running? && Webpacker.dev_server.hot_module_replacing?
-      asset_path(Webpacker.manifest.lookup!(name), **options)
+    unless stylesheet?(name) && current_webpacker.dev_server.running? && current_webpacker.dev_server.hot_module_replacing?
+      asset_path(current_webpacker.manifest.lookup!(name), **options)
     end
   end
 
@@ -28,8 +35,8 @@ module Webpacker::Helper
   #   # In production mode:
   #   <%= asset_pack_url 'calendar.css' %> # => "http://example.com/packs/calendar-1016838bab065ae1e122.css"
   def asset_pack_url(name, **options)
-    unless Webpacker.dev_server.running? && Webpacker.dev_server.hot_module_replacing?
-      asset_url(Webpacker.manifest.lookup!(name), **options)
+    unless current_webpacker.dev_server.running? && current_webpacker.dev_server.hot_module_replacing?
+      asset_url(current_webpacker.manifest.lookup!(name), **options)
     end
   end
 
@@ -40,7 +47,7 @@ module Webpacker::Helper
   #  <%= image_pack_tag 'application.png', size: '16x10', alt: 'Edit Entry' %>
   #  <img alt='Edit Entry' src='/packs/application-k344a6d59eef8632c9d1.png' width='16' height='10' />
   def image_pack_tag(name, **options)
-    image_tag(asset_path(Webpacker.manifest.lookup!(name)), **options)
+    image_tag(asset_path(current_webpacker.manifest.lookup!(name)), **options)
   end
 
   # Creates a script tag that references the named pack file, as compiled by webpack per the entries list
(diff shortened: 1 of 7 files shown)


gauravtiwari:
Thanks, @palkan Is this still a WIP? 

I would very much like the engines support to be in.

palkan:
Hi @gauravtiwari,

It's not WIP from the functionality point of view (we're already using this brunch). Btw, added one more commit to make env prefix for dev server configurable (
https://github.com/rails/webpacker/commit/32fd72eeb6dcf61cfd09c7465301565712299de5).

What's left then?

I think, we need a way to somehow automate the process of configuring Webpacker for an engine (all the steps I described above). Something like `rails g webpacker:install` but for the engine (`rails g webpacker:engine:install`?).

Or maybe it's better to add Webpacker support to `rails plugin new`?

Another thing not covered here: how to deploy the application?

We use the following trick:

```ruby
# main app Rakefile
Rake::Task["assets:precompile"].enhance do
  Rake::Task["my_engine:webpacker:compile"].invoke
end
```

So, we don't need separate tasks to be invoked during the deployment (works perfectly on Heroku).

To sum up, from the code perspective this PR looks complete to me (that means that we're happily use this branch without additional monkey-patches).

lib/webpacker/helper.rb
```diff
@@ -1,4 +1,11 @@
 module Webpacker::Helper
+  # Returns current Webpacker instance.
+  # Could be overriden to use multiple Webpacker
+  # configurations within the same app (e.g. with engines)
+  def current_webpacker
```
gauravtiwari: I would just call `Webpacker.instance` directly instead of putting into a method. What do you think?

gauravtiwari:
Thanks, @palkan This is great 👍 

Perhaps, make a changelog entry with experimental tag and link to this PR. We can work on another PR to add installation support either via plugin or rake task. Add relevant documentation.

gauravtiwari:
Could you please rebase with the master branch?

lib/webpacker/helper.rb
palkan: The goal is to make it possible to override the Webpacker instance, so in the engine's helper we do:

```ruby
require "webpacker/helper"

module MyEngine
  module ApplicationHelper
    include ::Webpacker::Helper
    
    def current_webpacker
       MyEngine.webpacker # or whatever
    end
  end
end
```

`Webpacker.instance` would point to _global_ Webpacker (i.e. `::Webpacker.instance`), so that won't work

palkan:
> Could you please rebase with the master branch?

This branch is based on `3-x-stable`. Would you like me to add a new PR for v4?

mathewbaltes:
Probably would be best to have this for v4.

palkan:
> Probably would be best to have this for v4.

Absolutely.

Right now I'm migrating our app to v4) I'll open a new PR as soon as I test this feature for v4 in prod/dev.

palkan:
Closed in favour of #1836

PR #1819, opened 2018-11-30
Fix test matrix exclusion

tricknotes:
This test matrix was excluded by https://github.com/rails/webpacker/pull/1539 due to always failing.
However, https://github.com/rails/webpacker/pull/1777 upgraded Ruby 2.3 version without updating this exclusion.


tricknotes:
In favor of https://github.com/rails/webpacker/pull/1821

gauravtiwari:
Thanks forgot to leave to a comment ❤️

PR #1823, opened 2018-12-02
Stable webpacker 4.0 release

gauravtiwari:
This PR is final PR to finish up some remaining bits and release Webpacker 4.x. 

Apologies it took us a while but looks like Webpacker is now in a great shape. Thanks, everyone for chipping in and making things better. 

Things I would like to finish and ship: 

- [ ] Integrity hashes for assets - https://github.com/rails/webpacker/issues/323 (currently working)
- [x] New split chunks API
- [x] New file loader option for adding static extensions
- [x] Basic node modules compilation apart from app code
- [ ] CDN support - https://github.com/rails/webpacker/issues/1084 Perhaps someone can help with this one?
- [ ] Engines support - there are two pending PRs https://github.com/rails/webpacker/pull/1808 and https://github.com/rails/webpacker/pull/1694

PR #1825, opened 2018-12-02
Exclude Rails 4.2 + ruby-head build

y-yagi:
Rails 4.2 is using `BigDecimal.new`. But `BigDecimal.new` was removed.
Ref: https://github.com/ruby/bigdecimal/commit/d73eca27de842928dd7f4480d7d6e435f50c9bc2

This causes the build of Rails 4.2 + ruby-head to fail.
https://travis-ci.org/rails/webpacker/jobs/462579174#L1292

Let's exclude this in order to keep the build state clean.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,3 +36,5 @@ matrix:
   exclude:
     - rvm: 2.3.8
       gemfile: gemfiles/Gemfile-rails-edge
+    - rvm: ruby-head
+      gemfile: gemfiles/Gemfile-rails.4.2.x


gauravtiwari:
Yep, lets done that already in #1823 but let's merge this one 🙏

mrkn:
I decided to revert removing BigDecimal.new for the next version of bigdecimal.
Anyway, I think you can stick the version of bigdecimal to `1.3.5`.

PR #1827, opened 2018-12-04
Use vue-loader stable version

holidayworking:
vue-loader v15 has already been released.

https://github.com/vuejs/vue-loader/blob/v15.4.0/CHANGELOG.md


gauravtiwari:
Thanks, fixed in #1823 but lets merge this one

PR #1830, opened 2018-12-11
Use RAILS_ENV in runner webpack config.

phlegx:
This allows to have webpack configs for other environments then development, test and production. For example using environment staging.js.


phlegx:
@gauravtiwari you write: _You can supply env variables to change settings during builds. For example, a different API URL for staging vs production._

Can you show me please an example how to do that?

gauravtiwari:
@phlegx Sure, here is some documentation: https://github.com/rails/webpacker/blob/master/docs/env.md

But basically, you can pass env variable like so: 
```
API_URL=foo bundle exec rails s
```

Issue #1824, opened 2018-12-02
Compiling in production fails silently and isn't verbose enough


ameft:
I have a Rails 5.2 app with Webpacker 4.0.0.pre.3 with runs smoothly in development and instead couldn't compile in production: `[Webpacker] Compiling…` just hanged there for hours (locally and on Heroku, where @schneems helped me a lot), in a similar way as described in #1534, #1217 and #1003.
After many hours of testing and research I found out that the problem was that two dependencies in `package.json` (@fortawesome/fontawesome-free and expose-loader) were in `devDependencies` instead of `dependencies`, that made compiling failing silently.
Running this command put me on the right path to find the solution, giving me some hits of errors `NODE_ENV=production ./bin/webpack --progress --config config/webpack/production.js`.
Problem is that instad running `RAILS_ENV=production NODE_ENV=production bundle exec rails assets:compile` didn't give me any indication of errors, and I don't know why but shouldn't be the case.
What do you think?

jakeNiemiec:
[FontAwesome did a writeup on this known issue](https://fontawesome.com/how-to-use/with-the-api/other/tree-shaking#issues). Issue https://github.com/terser-js/terser/issues/50

There is also a webpacker PR to fix this https://github.com/rails/webpacker/issues/1719

jakeNiemiec:
It's not that Webpacker is erroring out, uglify is incorrectly trying to minify all of the FA glyphs & exports. In other words, it's trying to do an impossible amount of work (memory & cpu wise). This is an issue that is just as hard to diagnose with vanilla webpack.

ameft:
@jakeNiemiec are you sure that that is what it's happening? Why moving the libraries from `devDependencies` to `dependencies` should fix that? Compiling with `./bin/webpack` was giving me errors, but Webpacker didn't.

jakeNiemiec:
@ameft Sorry, my mistake, I conflated some of the issues you pointed to with what you were saying. Now my understanding is: FA was in `devDeps` when it needed to be in `deps` and webpacker did not alert you to this error.

It looks like webpacker is configured correctly to display errors:
https://webpack.js.org/configuration/stats/#stats
https://github.com/rails/webpacker/blob/8d496a0b871085648d6ee21c0e1e469d11cfe593/package/environments/production.js#L34
https://github.com/rails/webpacker/blob/227661b546701e107213aa9b6f83451bc89cf7d5/package/environments/development.js#L43

[...]

PR #1834, opened 2018-12-14
Reduce the output of the webpack-dev-server

Zyphrax:
The webpack-dev-server normally outputs a lot of information about how the packages are resolved. During development you are often only interested in the packages that webpack creates after building and the names and sizes of those packages.

After this pull request has been merged, you can always increase logging again by adding the following lines to your `environment.js`:
```javascript

// Reduce the output of the Webpack Dev Server
environment.config.merge({
  devServer: {
    stats: {
      entrypoints: true,
      errorDetails: true,
      modules: true,
      moduleTrace: true
    }
  }
});

module.exports = environment
```

Or run webpack manually instead of the webpack-dev-server. But in my opinion the stats settings in this pull request are a better default.

### DIFF
--- a/package/environments/development.js
+++ b/package/environments/development.js
@@ -40,7 +40,10 @@ module.exports = class extends Base {
         headers: devServer.headers,
         overlay: devServer.overlay,
         stats: {
-          errorDetails: true
+          entrypoints: false,
+          errorDetails: false,
+          modules: false,
+          moduleTrace: false
         },
         watchOptions: devServer.watch_options
       }


gauravtiwari:
thanks 👍

jakeNiemiec:
As a developer, I wouldn't expect `webpacker` to suppress error details and stack traces by default (especially if that's not the default for vanilla webpack). 

There even some issues asking for `webpacker` to be more verbose or complaining about a build hanging with no error message. https://github.com/rails/webpacker/issues/1824

Zyphrax:
@jakeNiemiec Perhaps it depends on the loaders that are used. In our project we are using TypeScript and if I keep the webpack-dev-server at the defaults it will output 132 lines.

A lot of it is things like:
```
Child mini-css-extract-plugin node_modules/css-loader/index.js??ref--6-1!node_modules/postcss-loader/src/index.js??ref--6-2!node_modules/resolve-url-loader/index.js!node_modules/sass-loader/lib/loader.js??ref--6-4!app/frontend/packs/platform/visitor.scss:
    Entrypoint mini-css-extract-plugin = *
    [./app/frontend/images/background.jpg] 100 bytes {mini-css-extract-plugin} [built]
    [./node_modules/css-loader/index.js?!./node_modules/postcss-loader/src/index.js?!./node_modules/resolve-url-loader/index.js!./node_modules/sass-loader/lib/loader.js?!./app/frontend/packs/platform/visitor.scss] ./node_modules/css-loader??ref--6-1!./node_modules/postcss-loader/src??ref--6-2!./node_modules/resolve-url-loader!./node_modules/sass-loader/lib/loader.js??ref--6-4!./app/frontend/packs/platform/visitor.scss 7.52 KiB {mini-css-extract-plugin} [built]
    [./node_modules/css-loader/lib/css-base.js] 2.21 KiB {mini-css-extract-plugin} [built]
    [./node_modules/css-loader/lib/url/escape.js] 448 bytes {mini-css-extract-plugin} [built]
```

That doesn't provide any useful information for me for normal development. If I have a specific problem with a loader it could be useful, but otherwise it is just a lot of text.

The webpack-dev-server settings are injected in https://github.com/rails/webpacker/blob/master/package/environments/development.js. It would be great if we could merge in `stats` settings from the app's `config/webpacker.yml`.

This would probably work, but I'm not sure if it is the right way:

**package/environments/development.js**

```javascript
    overlay: devServer.overlay,
    stats: {
      entrypoints: true,
      errorDetails: true,
      modules: true,
      moduleTrace: true,
      ...devServer.stats
    },
    watchOptions: devServer.watch_options
```

ch-tactica:
@yvbeek I'm sorry, but I don't think this was a very good change. If you decide to increase the loaders in your own project and the default output becomes to verbose, that's something that is on you to deal with yourself.

Having this off by default hides very important compilation errors that are a pain for people who aren't as familiar with Webpacker/Webpack to track down.

yvbeek:
I’m fine with reverting this change. Especially if it is causing issues for people who don’t use TypeScript in their projects.

Issue #1795, opened 2018-11-07
Error supporting Elm 0.19 webpacker somehow injecting 0.18 even though 0.19 is installed.


koctya:
WEbpacker is throwing errors for 0.19, even though it is installed;
I'm still getting this error in ver 3.5.5 even though the comments in 
https://github.com/rails/webpacker/issues/1668 stated the 3.5.x stable was patched for 0.19.
```
ERROR in ./app/javascript/packs/Api/Union.elm
Module build failed: Error: Compiled JS from the Elm compiler is not valid. You must use the Elm 0.19 compiler.
    at inject (/Users/kurtlandrus/code/mcs/cost_model/node_modules/elm-hot/src/inject.js:35:15)
    at Object.module.exports (/Users/kurtlandrus/code/mcs/cost_model/node_modules/elm-hot-webpack-loader/src/index.js:4:12)
```
I only have version 0.19 installed;
```✗ elm --version
0.19.0
✗ bundle list webpacker
/Users/kurtlandrus/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/webpacker-3.5.5
```
my env;
```
✗ rails about
About your application's environment
Rails version             5.2.1
Ruby version              2.3.1-p112 (x86_64-darwin15)
RubyGems version          2.5.1
Rack version              2.0.5
JavaScript Runtime        Node.js (V8)
Middleware                Webpacker::DevServerProxy, Rack::Sendfile, ActionDispatch::Static, ActionDispatch::Executor, ActiveSupport::Cache::Strategy::LocalCache::Middleware, Rack::Runtime, Rack::MethodOverride, ActionDispatch::RequestId, RequestStore::Middleware, ActionDispatch::RemoteIp, Sprockets::Rails::QuietAssets, Rails::Rack::Logger, ActionDispatch::ShowExceptions, WebConsole::Middleware, ActionDispatch::DebugExceptions, BetterErrors::Middleware, ActionDispatch::Reloader, ActionDispatch::Callbacks, ActiveRecord::Migration::CheckPending, ActionDispatch::Cookies, ActionDispatch::Session::CookieStore, ActionDispatch::Flash, ActionDispatch::ContentSecurityPolicy::Middleware, Rack::Head, Rack::ConditionalGet, Rack::ETag, Rack::TempfileReaper, Warden::Manager
Environment               development
Database adapter          mysql2
```

koctya:
```
✗ rails webpacker:info
[...]

PR #1843, opened 2018-12-20
Adding support for Reason and ReasonReact applications

fuentesjr:
I would like to contribute support for Reason applications to this project. The following changes setup the necessary changes to compile and run Reason applications. Furthermore, it includes a small ReasonReact application to get started right away. Comments and feedback are welcomed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -403,6 +403,23 @@ The Elm library and its core packages will be added via Yarn and Elm.
 An example `Main.elm` app will also be added to your project in `app/javascript`
 so that you can experiment with Elm right away.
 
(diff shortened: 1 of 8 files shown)


gauravtiwari:
Thanks, @fuentesjr for this

We are trying to keep the default installers list small and limited to popular choices only since it requires effort to keep them in sync and maintain. We once let someone add elm, which isn't  properly maintained/supported now: https://github.com/rails/webpacker/issues/1795 

I suggest creating a package called - `webpacker-reason`, which has similar installers: `webpacker:reason:install`. What do you think?

Issue #1856, opened 2018-12-27
Circular JSON error when using Angular AoT Compiler


getabetterpic:
**Steps to reproduce:**
1. Set up new Rails app with webpacker
1. Upgrade Webpacker to v4.0.0.rc2
1. `yarn add @ngtools/webpack`
1. Add the following to `config/production.js` to use AngularCompilerPlugin:
```javascript
const AngularCompilerPlugin = require('@ngtools/webpack').AngularCompilerPlugin

environment.plugins.append(
  'AngularCompilerPlugin',
  new AngularCompilerPlugin({
    tsConfigPath: 'tsconfig.json',
    entryModule: 'app/javascript/hello_angular/app/app.module#AppModule'
  })
)

environment.loaders.append('typescript', {
  test: /\.ts$/,
  use: '@ngtools/webpack'
})
```
5. Run `NODE_ENV=production rake assets:precompile`

**Results:**
Compile errors with the following message:
```
Compiling…
Compilation failed:
/Users/dan/projects/circular_json/node_modules/webpack-cli/bin/cli.js:244
                                throw err;
                                ^

TypeError: Converting circular structure to JSON
    at JSON.stringify (<anonymous>)
    at isEqual (/Users/dan/projects/circular_json/node_modules/@rails/webpacker/package/utils/helpers.js:20:69)
    at deepMerge (/Users/dan/projects/circular_json/node_modules/@rails/webpacker/package/utils/deep_merge.js:8:7)
    at ConfigObject.merge (/Users/dan/projects/circular_json/node_modules/@rails/webpacker/package/config_types/config_object.js:50:25)
    at module.exports.toWebpackConfig (/Users/dan/projects/circular_json/node_modules/@rails/webpacker/package/environments/base.js:151:24)
    at Object.<anonymous> (/Users/dan/projects/circular_json/config/webpack/production.js:19:30)
    at Module._compile (/Users/dan/projects/circular_json/node_modules/v8-compile-cache/v8-compile-cache.js:178:30)
    at Object.Module._extensions..js (module.js:663:10)
    at Module.load (module.js:565:32)
    at tryModuleLoad (module.js:505:12)
    at Function.Module._load (module.js:497:3)
    at Module.require (module.js:596:17)
    at require (/Users/dan/projects/circular_json/node_modules/v8-compile-cache/v8-compile-cache.js:159:20)
[...]

PR #1859, opened 2018-12-30
Use flatted to compare object equality

getabetterpic:
Issue #1856 where an error occurs if there is a circular JSON reference somewhere in the config. This PR introduces the [`flatted`](https://www.npmjs.com/package/flatted) package which can stringify those circular references.


gauravtiwari:
thanks @getabetterpic

PR #1868, opened 2019-01-04
Make css-loader an optional dependency

jaredbeck:
I think that many people do not use css-loader, and those
people should not be forced to install it.

Disclaimer: I'm primarily a ruby programmer, so if there are
related changes outside of package.json that I am missing, sorry,
and thanks for your help.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,6 @@
     "babel-plugin-macros": "^2.4.2",
     "case-sensitive-paths-webpack-plugin": "^2.1.2",
     "compression-webpack-plugin": "^2.0.0",
-    "css-loader": "^2.0.1",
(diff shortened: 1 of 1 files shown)


gauravtiwari:
Hi @jaredbeck I will look into making loaders more automatic in the future Webpacker release.

jaredbeck:
> I think that many people do not use css-loader, and those people should not be forced to install it.

I still think this, but ironically I now use `css-loader` in my rails project. 😆 "Life comes at you fast"

> Hi @jaredbeck I will look into making loaders more automatic in the future Webpacker release.

Thanks! I am curious to find out what "automatic" means! 😄

PR #1860, opened 2019-01-02
Exclude Rails Edge (6.0) + Ruby 2.4 or lower build

koic:
Rails Edge (6.0) requires Ruby 2.5 or higher.
https://github.com/rails/rails/commit/1b7c3222e86b98ce886110ae0860ba62e35e2299

This PR excludes the following error matrix from Travis CI.

```console
$ bundle install
Fetching https://github.com/rails/rails.git
Fetching https://github.com/rails/arel.git
Fetching https://github.com/rubocop-hq/rubocop.git
Fetching gem metadata from https://rubygems.org/.............
Fetching gem metadata from https://rubygems.org/..........
Fetching gem metadata from https://rubygems.org/..........
Resolving dependencies...
activesupport-6.0.0.alpha requires ruby version >= 2.5.0, which is incompatible
with the current version, ruby 2.4.5p335
The command "bundle install" failed and exited with 5 during .
```

https://travis-ci.org/rails/webpacker/jobs/472092576#L887-L897

PR #1872, opened 2019-01-06
Restore Rails Edge (6.0) + Ruby 2.5 build matrix

koic:
Follow up of #1860.

Ruby 2.5 needs to be built with Rails Edge (6.0).
That is my mistake and I restored it 💦


guilleiguaran:
Looks like this was already done in master!

PR #1882, opened 2019-01-10
Add installer for jest

coorasse:
This PR adds support for the command `webpacker:install:jest`.
The command will install jest and add a couple of files as an example, saving you from the pain of configuring it manually.

Personally, I would propose to have it installed by default when installing webpacker. What is nice about Rails (and puts it one step in front of other frameworks) is to have a development environment that **just works**. 

In such an environment, I expect to have a testing framework for my Javascript ready to use. As Rails provides minitest out-of-the-box, it should also provide jest, to start writing tests immediately.

Things to consider: 

* I put the tests under the `test/javascripts` folder. Should this go under webpacker configuration file?

* Shall I write tests? I didn't see tests for the other installers. I have an idea on how to run the tests for these rake commands. you can see it here: https://github.com/coorasse/webpacker/commit/5195aef0a113b62fe9ba4275235ae9fd6c954476

### DIFF
--- a/lib/install/jest.rb
+++ b/lib/install/jest.rb
@@ -1,0 +1,39 @@
+require "webpacker/configuration"
+
+say "Copying the example test files"
+tests_folder = "#{defined?(RSpec) ? 'spec' : 'test'}/javascripts"
+
+copy_file "#{__dir__}/examples/jest/hello_jest.js",
+          "#{Webpacker.config.source_entry_path}/hello_jest.js"
+copy_file "#{__dir__}/examples/jest/hello_jest.spec.js",
+          "#{Webpacker.config.root_path.join(tests_folder)}/hello_jest.spec.js"
+
+say "Updating package.json"
+package_json = JSON.parse(File.read(Rails.root.join("package.json")))
+relative_source_path = Webpacker.config.source_path.relative_path_from(Webpacker.config.root_path)
+
+jest_config = {
+  "scripts" => {
+    "test" => "jest",
+    "test:watch" => "jest --watch"
+  },
+  "jest" => {
+    "roots" => [relative_source_path.to_s, tests_folder],
+    "moduleDirectories" => ["node_modules", relative_source_path.to_s],
+    "transform" => {
+      "^.+\\.js?$" => "babel-jest"
+    }
+  }
+}
+
+order_list = %w(name private jest dependencies devDependencies)
+new_package_json = jest_config.deep_merge(package_json).sort_by do |key, value|
+  order_list.index(key) || Float::INFINITY
+end.to_h
+
+File.write(Rails.root.join("package.json"), JSON.pretty_generate(new_package_json))
+
+say "Installing all Jest dependencies"
+run "yarn add --dev jest babel-jest"
+
+say "Webpacker now supports Jest 🎉. Execute 'yarn test' to run your tests.", :green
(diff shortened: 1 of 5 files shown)


coorasse:
CI is failing because of an incompatibility between bundler 2.0 and Rails 4.2. I see also other PR fail for the same reason.

joecochran:
+1 on this idea. if jest comes installed out of the box, it might be useful to yarn install `@vue/test-utils` in `webpacker:install:vue` and `jest-enzyme enzyme enzyme-adapter-react-16` with a `webpacker:install:react` as well. I might take a crack at that if this makes it in.

ytbryan:
A shameless plug on my open-source project - Vue on Rails. It works well with webpacker 3/4 and  supports Jest and @vue/test-utils too.

1. To generate a new webpacker 3 project with jest, run `rails new app -m https://vueonrails.com/vue` 
> For webpacker 4, please run `rails new app -m https://vueonrails.com/vue4`
2. To scaffold a vue component  with jest test, run `rails g vue something --test`
3. To setup jest test on your rails project, run `rails vue:test`

--

IMHO, It is better to leave test support to a 3rd party webpacker library.

zernie:
Great job @coorasse, but specs break down if you have any `*.js.erb` files in your code. I assume a custom jest transformer is necessary :(

coorasse:
I have no `.js.erb` files. I rebased and the CI is now ok. 🎉

lib/install/jest.rb
```diff
@@ -0,0 +1,39 @@
+require "webpacker/configuration"
+
+say "Copying the example test files"
+copy_file "#{__dir__}/examples/jest/hello_jest.js",
+          "#{Webpacker.config.source_entry_path}/hello_jest.js"
+copy_file "#{__dir__}/examples/jest/hello_jest.spec.js",
+          "#{Webpacker.config.root_path.join(tests_folder)}/hello_jest.spec.js"
```
1st8: It looks like `test_folder` is not defined yet, that happens in line 13, or am I missing something?

coorasse: you are not. I updated the jest.rb. I guess a set of tests wouldn't hurt here, right? It would have been nice to have some examples but there are not. If you reckon, I can write the first one here.

coorasse:
Hello @1st8, 
is there something else I can do to push forward this feature? Please give me a feedback if there is something I can improve or if is not a feature we want in general. Thank you 🙇

1st8:
Hi @coorasse ,

I am not affiliated with this project so I can't tell you if this is something that belongs to the webpacker scope, sorry for the misunderstanding.
Your code helped me with configuring jest in our environment (that's when I saw the error), but the final configuration needed a few changes that I am not sure will be mandatory on a default webpacker setup.

If you wanted, I could do a code review, but it would be my personal opinion only. 
Same applies for our final configuration, if you wanted details on that.

Maybe this topic could find easier adoption if written as documentation?

Cheers!

coorasse:
😅  If you have any consideration, @1st8, is more than welcome!

gauravtiwari:
Closing per https://github.com/rails/webpacker/pull/1882#issuecomment-461850212

coorasse:
I respect your point of view, although I still think having a testing library pre-configured would give webpacker a boost in terms of productivity. 
Having it "ready to use" is exactly one of these things that, as a developer, make me say "That's nice! 😄" 
Thanks for your time, guys. 🙇

PR #1889, opened 2019-01-14
Update v3.x to use babel v7

dev-cprice:
Update babel dependencies on v3.x stable branch to use babel v7


maschwenk:
This won't just be a drop in replacement because all the references to the package name in config will have to change. I imagine @gauravtiwari has no plans to backport Babel v7 to Webpacker 3.x

gauravtiwari:
Thanks @maschwenk

PR #1898, opened 2019-01-17
Add a new javascript_packs_with_chunks_tag tag helper

### LINKED ISSUE #1835: [4.0.0-rc.1] Javascript_asset_tag is broken
I've upgraded from `4.0.0-pre.3` to `4.0.0-rc.1` and now I get a really strange error in my application.slim:

> Sprockets::Rails::Helper::AssetNotFound at /
> The asset "{\"src\"=>\"/packs/vendor-4699109b154b54e1ef4e.js\", 
> \"integrity\"=>\"sha256-hkCqb1IPUS4z7o4PyMRG7t3+J0U2MgbmVUty9/hviU0= sha384-
> En5fK1fCL66RKx2Ovx8/n6AEwom4R6OgFxxNvKdrL6nQYow56QN66K0CE0N1c2z/ sha512-
> nTC02x/4ytkiw9of3MH55Xm4KkZaaGnbEkqBJ+OeToZmQH+fYrasEosB95s+bvNKQ6tjtcEIe
> zHJE8aHCyDqSw==\"}.js" is not present in the asset pipeline.

This errors seems to be caused by this call in my project:

_app/views/layout/application.slim_
```slim
= javascript_pack_tag 'vendor'
```

It seems that `javascript_pack_tag` generates a hash and is incorrectly converted to a Rails include tag. It seems to think that the hash is the `src`.

----
**Edit: It might have something to do with my splitChunks configuration**
This was working fine in `4.0.0-pre.3`, but now something goes wrong in [manifest.rb - lookup](https://github.com/rails/webpacker/blob/3aa4a5dc517106935b3d6f45aa82f4707a5de217/lib/webpacker/manifest.rb#L21).

```javascript
environment.config.merge({
  optimization: {
    splitChunks: {
      cacheGroups: {
        commons: {
          test: /\/node_modules\//,
          name: 'vendor',
          chunks: 'all'
        }
      }
    }
  }
});
```

Configurations with `cacheGroups` no longer work?

gauravtiwari:
This PR adds a new chunks helper that addresses issues outlined here: https://github.com/rails/webpacker/issues/1835

Fixes #1835 
----

```erb
<%= javascript_packs_with_chunks_tag 'hello_react', 'application', 'hello_vue', 'data-turbolinks-track': 'reload' %>
```

which outputs following html, with depedencies de-deduped 

```html
<script src="/packs/runtime~hello_react-26febc8eea663e32aadb.js" data-turbolinks-track="reload"></script>
<script src="/packs/0-d471c80da3d3a8a51c72.chunk.js" data-turbolinks-track="reload"></script>
<script src="/packs/1-8ef1f3ec1d6838a685eb.chunk.js" data-turbolinks-track="reload"></script>
<script src="/packs/hello_react-d26893900192b71057ec.chunk.js" data-turbolinks-track="reload"></script>
<script src="/packs/runtime~application-2a426143d8d4ad5ac92e.js" data-turbolinks-track="reload"></script>
<script src="/packs/application-c45c132bb54ceeafd2e5.chunk.js" data-turbolinks-track="reload"></script>
<script src="/packs/runtime~hello_vue-b068488867759681a7da.js" data-turbolinks-track="reload"></script>
<script src="/packs/2-4bd0417a6b1fb648dbf1.chunk.js" data-turbolinks-track="reload"></script>
<script src="/packs/hello_vue-7e0743cde157093b96bb.chunk.js" data-turbolinks-track="reload"></script>
```

`javascript_packs_with_chunks_tag` is introduced to handle new webpack split chunks API: https://webpack.js.org/plugins/split-chunks-plugin/

```json
 "application": {
      "js": [
        "/packs/runtime~application-2a426143d8d4ad5ac92e.js",
        "/packs/0-d471c80da3d3a8a51c72.chunk.js",
        "/packs/application-c45c132bb54ceeafd2e5.chunk.js"
      ],
      "js.map": [
        "/packs/runtime~application-2a426143d8d4ad5ac92e.js.map",
        "/packs/0-d471c80da3d3a8a51c72.chunk.js.map",
        "/packs/application-c45c132bb54ceeafd2e5.chunk.js.map"
      
``` 

`javascript_pack_tag`, will keep working the same way as before, so no breaking changes there. 

```erb
<%# This will include all deps + pack %>
<%= javascript_packs_with_chunks_tag "stimulus", "hello_stimulus" %>

<%# this will include transpiled packs only, no deps (if using split chunks) %>
<%= javascript_pack_tag "stimulus" %>
<%= javascript_pack_tag "hello_stimulus" %>
```

```erb
<%# DO %>
<%= javascript_packs_with_chunks_tag 'calendar', 'map' %>

<%# DON'T %>
<%= javascript_packs_with_chunks_tag 'calendar' %>
<%= javascript_packs_with_chunks_tag 'map' %>
```

Thanks @javan, could you please review?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,37 @@
 **Please note that Webpacker 3.1.0 and 3.1.1 have some serious bugs so please consider using either 3.0.2 or 3.2.0**
 
+## [4.0.0] - Unreleased
+
+### Fixed
+ - Issue with javascript_pack_tag asset duplication [#1898](https://github.com/rails/webpacker/pull/1898)
+
+
+### Added
+ - `javascript_packs_with_chunks_tag` helper, which creates html tags
+  for a pack and all the dependent chunks, when using splitchunks.
+
+```erb
+<%= javascript_packs_with_chunks_tag 'calendar', 'map', 'data-turbolinks-track': 'reload' %>
+
+<script src="/packs/vendor-16838bab065ae1e314.js" data-turbolinks-track="reload"></script>
+<script src="/packs/calendar~runtime-16838bab065ae1e314.js" data-turbolinks-track="reload"></script>
+<script src="/packs/calendar-1016838bab065ae1e314.js" data-turbolinks-track="reload"></script>
+<script src="/packs/map~runtime-16838bab065ae1e314.js" data-turbolinks-track="reload"></script>
+<script src="/packs/map-16838bab065ae1e314.js" data-turbolinks-track="reload"></script>
+```
+
+**Important:** Pass all your pack names when using `javascript_packs_with_chunks_tag`
+helper otherwise you will get duplicated chunks on the page.
+
+```erb
+<%# DO %>
+<%= javascript_packs_with_chunks_tag 'calendar', 'map' %>
+
+<%# DON'T %>
+<%= javascript_packs_with_chunks_tag 'calendar' %>
+<%= javascript_packs_with_chunks_tag 'map' %>
+```
+
 ## [4.0.0.rc.2] - 2018-12-15
 
 ### Fixed
--- a/README.md
+++ b/README.md
@@ -142,6 +142,31 @@ can use the `asset_pack_path` helper:
 <img src="<%= asset_pack_path 'images/logo.svg' %>" />
 ```
 
+If you are using new webpack 4 split chunks API, then consider using `javascript_packs_with_chunks_tag` helper, which creates html
+tags for a pack and all the dependent chunks.
+
+```erb
+<%= javascript_packs_with_chunks_tag 'calendar', 'map', 'data-turbolinks-track': 'reload' %>
+
+<script src="/packs/vendor-16838bab065ae1e314.js" data-turbolinks-track="reload"></script>
+<script src="/packs/calendar~runtime-16838bab065ae1e314.js" data-turbolinks-track="reload"></script>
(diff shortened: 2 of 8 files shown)


aceofspades:
Is there an approach to using `defer: true`? Presently I get different issues popping up when true and false.

aceofspades:
Also, there's a need to be able to generate a list of all the paths to packs, not just tags, i.e. when building a service worker that caches files. 

@gauravtiwari

jakeNiemiec:
I'd advise against building a service worker with webpack. You will end up with multiple workers registered under differently hashed names within the same domain. (plus your sw file needs to be in the `/public` dir to work on some browsers)

aceofspades:
@jakeNiemiec Not sure what you mean. Are you suggesting that if an app requires service workers then to not use webpack at all? That seems like a huge shortcoming

jakeNiemiec:
https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerContainer/register#Examples
> There is frequent confusion surrounding the meaning and use of scope. Since **a service worker can't have a scope broader than its own location**, only use the scope option when you need a scope that is narrower than the default.

If your service worker is in `www.example.com/packs/SW.js`, it can only control pages under `www.example.com/packs` which does not make much sense. 

This is why service workers are generally not "built". Some SW libraries may copy the file to the correct place, it all depends on the implementation.

PR #1755, opened 2018-10-15
Create upgrade-v4.md

justin808:

PR #1905, opened 2019-01-22
Add a Webpacker v4 Upgrade Guide to docs

connorshea:
This is based on my experience [upgrading my Rails app](https://github.com/connorshea/ContinueFromCheckpoint/pull/77) to v4.0.0-rc.5, plus some content taken from #1755.

Credit to @justin808 for the original v4 upgrade guide this is based on.

### DIFF
--- a/docs/v4-upgrade.md
+++ b/docs/v4-upgrade.md
@@ -1,0 +1,18 @@
+To update a Webpacker v3.5 app to v4, follow these steps:
+
+1. Update the `webpacker` gem and the `@rails/webpacker` package to v4. This will upgrade webpack itself from 3.x to 4.x, make sure you're aware of [any deprecations which might effect you](https://webpack.js.org/migrate/4/). Also make sure any other packages you depend on support webpack 4 and don't require any changes, e.g. if you explicitly include `webpack` you need to upgrade it to 4.x, and if you use `webpack-dev-server` you need to upgrade it to 3.x.
+1. Browser support definitions have been moved to [`.browserslistrc`](../lib/install/config/.browserslistrc) to `/`.
+1. Merge any differences between [`config/webpacker.yml`](../lib/install/config/webpacker.yml) and your `config/webpacker.yml`.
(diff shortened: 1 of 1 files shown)


Zyphrax:
@connorshea Great work! I think it would be good to mention `javascript_packs_with_chunks_tag` and `stylesheet_packs_with_chunks_tag` for people with a splitChunks configuration.

connorshea:
@Zyphrax I don't fully understand that change unfortunately (I haven't really used the splitChunks config before), could you make a quick summary for me to include in the guide?

Zyphrax:
@connorshea SplitChunks is a Webpack feature that avoids duplicate code by combining shared dependencies and including them only once. If you have script A that uses (e.g.) React and you have script B that uses React, then normally Webpack will merge React into script A and again into script B.

With SplitChunks you can move the dependency (React) into a separate "chunk" and include it only once. A common configuration is to move all the node_modules dependencies into a vendors chunk. This will result in smaller js files and makes it easier for your browser to cache the relatively static vendor packages.

Maybe it is a good idea to include these lines from the [Usage instructions](https://github.com/rails/webpacker#usage):

> If you are using the new Webpack 4 split chunks API, then consider using the javascript_packs_with_chunks_tag and stylesheet_packs_with_chunks_tag helpers, which creates html tags for a pack and all the dependent chunks.

(I've added stylesheet_packs_with_chunks_tag to clarify what to use for stylesheets)

connorshea:
@Zyphrax I've added a section on the change, does it look accurate to you? I've also added the config from https://github.com/rails/webpacker/pull/1755#issuecomment-449521155

justin808:
Cursory look...LGTM.

gauravtiwari:
@connorshea For split chunks, https://github.com/rails/webpacker/blob/c4cf2e99c2ae4dffdd2942c1d2282936b8aaf36e/docs/webpack.md#add-splitchunks-webpack-v4

We have an option in the environment to setup splitchunks (officially suggested - https://twitter.com/wSokra/status/969633336732905474) but obviously, it's customisable.

arkhamRejek:
as an alternative to .browserlistrc file, I added it to the package that json like so. 

image

can we add this in the read me as well for those of us who don't want more files in our project? or is that file absolutely necessary @gauravtiwari

connorshea:
@gauravtiwari should I just copy-paste that section into this upgrade guide?

gauravtiwari:
@arkhamRejek Sure, that's another option. 

@connorshea Yes please, just mention that we support the basic implementation out of the box however it's customisable if need to be. 

Thanks :)

connorshea:
Done :)

connorshea:
@Zyphrax you can feel free to handle it

Zyphrax:
See pull request: https://github.com/rails/webpacker/pull/1929

PR #1906, opened 2019-01-22
Remove an extra space in the vue.rb installer

connorshea:
I was wondering why my `environment.js` file had that weird space in it :)


Deterio:
Such a meaningful commit.

Issue #895, opened 2017-10-04
Overcomplicated @rails/webpacker package?


bytewalls:
I love that this just works out of the box, but configuring it after that leaves a lot to be desired.

The basis of the problem being that generally webpack will need some configuration and it is currently 90% extracted out into a node module, and I could start from scratch but it already _mostly_ works. 

Maybe I'm missing something, but it seems like you could have a much simpler easy to extend, well documented by webpack configuration with a `webpack.base.config.js` that is simply extended in `dev`, `test`, `prod` file. Instead it is using a yarn package with and environment class that builds an object and requires me to read the code of the yarn package to change. (The current webpacker documentation here says to look at webpack for configuration information, but that isn't particularly useful IMO when you have built helper libraries to build the config object.)

renchap:
You can easily add your own Webpack plugins / loaders, see here: https://github.com/rails/webpacker/blob/master/docs/webpack.md
This does not solve the problem to see what the default Webpack config is, but IMO adding your own bits is quite easy.

bytewalls:
I looked at this, but it did not allow me to do what I needed to (specifically I need to set the order that loaders are ordered in). I ended removing the package entirely and building my own webpack config.

gauravtiwari:
Hey, working on a PR that adds same feature you are looking for - #884

Perhaps, we can also have an eject feature that copies over config files from npm module to your app for customisation.

gauravtiwari:
@bytewalls Merged #884 and here is the new API:

```js
// environment.js

environment.loaders.append('json', jsonLoader)
environment.loaders.prepend('json', jsonLoader)
environment.loaders.insert('json', jsonLoader, { after: 'style' } )
environment.loaders.insert('json', jsonLoader, { before: 'babel' } )
// Same API for plugins
environment.plugins.insert('Manifest', /* manifestPlugin instance */, { before: 'ExtractText' } )

// Same for resolved modules
environment.resolvedModules.prepend('vendor', 'vendor')

module.exports = environment
```

`environment.config` now returns webpack config object: 

```js
// environment.js

const environment = require('@rails/webpacker')

environment.config = // {} webpack config object

environment.config.set('resolve.extensions', ['.foo', '.bar'])
environment.config.set('output.filename', '[name].js')
environment.config.delete('output.chunkFilename')
environment.config.get('resolve')

// Just merge as you would do with webpack-merge
environment.config.merge({ output: {
    filename: '[name].js'
  }
})

module.exports = environment
```

PR #1916, opened 2019-01-26
[POC] Eject feature

banyan:
Hi! I want to pursue a possibility that if we can have eject feature in webpacker.

I understand webpacker can configure all of webpack settings using `ConfigList` and `ConfigObject`. If a use case is just simple, then we can go with webpacker without having many custom settings, but if we want to customize, eventually we have to understand internal webpacker codes. In that case, I think `webpacker:eject` would be useful especially who knows webpack well and wants to configure directly.

I believe `webpacker:eject` would be an escape route of a tall mountain. Even it might not be used, it will make people feel easy and it will strengthen webpacker. That being said, this PR is just POC and has a problem how to maintain the initial `webpack.config.js`. IMO, `webpack.config.js` is not need to be completely compatible with webpacker's one, but just core logic is enough. 

eject will do following things:

* Generete webpacker compatible `webpack.config.js` while referring to `config/webpacker.yml`'s settings.
  * `config/webpack/loaders/*` will be used from generated `webpack.config.js`. so for example, if you run `bundle exec rake webpacker:install:elm` before, you can still use `elm`.
* webpacker gem will have remained for asset helpers like `javascript_pack_tag`.
* npm-scripts `yarn start` and `yarn build` will be added.

This PR works well with fresh `rails new` projects. 

I'm happy if I can get any feedbacks. 
@gauravtiwari I saw [this](https://github.com/rails/webpacker/issues/895#issuecomment-335948228). do you have any idea regarding implementation?

### DIFF
--- a/lib/eject/template.rb
+++ b/lib/eject/template.rb
@@ -1,0 +1,50 @@
+# Eject Webpacker
+yes = yes? 'Are you sure you want to eject? This action is permanent. (y/N)'
+return unless yes
+
+loaders = Dir.glob("#{Rails.root.join('config/webpack/loaders')}/*")
+  .map { |loader|
+    basename = File.basename(loader, File.extname(loader))
+    "require('./loaders/#{basename}')"
+  }
+webpacker_config = YAML.load_file(Rails.root.join('config/webpacker.yml'))
+
+say 'Copying webpack.config.js'
+
+File.write(Rails.root.join('config/webpack/webpack.config.js'), ERB.new(File.read("#{__dir__}/webpack.config.js.erb"), nil, '-').result(binding))
+
+say "Installing all dependencies"
+
+dependencies = %w[
+  webpack
+  webpack-cli
+  mini-css-extract-plugin
+  webpack-assets-manifest
+  case-sensitive-paths-webpack-plugin
+  pnp-webpack-plugin
+  babel-loader
+]
+
+run "yarn add #{dependencies.join(' ')}"
+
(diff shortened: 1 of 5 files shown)


jakeNiemiec:
You beat me to it 😆, I had been working out an `eject` function for a while. I stopped because I felt like there needs to be a completly seperate gem for resolving the `javascript_pack_tag` entry files from a `manifest.json` file. 

Looks good!

nikushi:
Very interesting work. FYI, I made [WebpackManifest](https://github.com/nikushi/webpack_manifest) gem a while ago, which provides helpers to resolve path and chunk.

h0jeZvgoxFepBQ2C:
Anyone knows if this PR will get merged?

derekblank:
@banyan Updating lines 2, 10, 12, 14, and 33 of `template.rb` to use double-quoted strings will resolve the [build failure](https://travis-ci.org/rails/webpacker/jobs/484713109#L701-L717).

trcarden:
Does this have support broadly from the rest of the rails team? Seems like an excellent idea and I would even say that if webpacker was just reconstructed into a set of generators it may be even better. We have to learn parts of webpack anyway and in order to know why our CSS is showing up inline instead of a file in dev you have to dive into webpacker to determine the config anyway.
See: https://github.com/rails/webpacker/issues/2161

kimihito:
@banyan has extracted this PR as [webpacker_ejetor](https://github.com/banyan/webpacker_ejector) gem.

PR #1918, opened 2019-01-28
Re-enable Source maps in production, reverts #770

guilleiguaran:
per https://github.com/rails/webpacker/issues/769#issuecomment-458216151


connorshea:
This should have a changelog entry (and possibly should be mentioned in the v4 upgrade doc as well).

Make sure to mention that it makes debugging in production easier, will be compressed by default, and shouldn’t increase the asset size by very much, as source maps aren’t loaded unless the dev tools are opened. :)

gauravtiwari:
@wildjcrt Sure, please see: https://github.com/rails/webpacker/blob/master/docs/webpack.md#configuration

```js
// development.js or production.js

environment.config.merge({ devtool: 'none' })

```

PR #1941, opened 2019-02-13
Enable static asset extensions to be overwritten from app config

dreyks:
Currently you can only add more custom static asset extensions to app's webpacker config. If you remove some of them (e.g. for them to be loaded by url-loader) the default webpacker config merges them back in

So let's treat them the same way we do the usual extensions i.e. delete the default list if there's a user-provided one


gauravtiwari:
Thanks :)

PR #1952, opened 2019-02-19
Remove deprecated option of resolve-url-loader and make code correct to be executed before "css" instead of "sass"

pedrofurtado:
Hey everyone! :+1: 

The current code example does not work with the latest version of webpacker gem (3.x) and the latest version of resolve-url-loader (3.x).

@gauravtiwari

### DIFF
--- a/docs/css.md
+++ b/docs/css.md
@@ -147,10 +147,10 @@ yarn add resolve-url-loader
 const { environment } = require('@rails/webpacker')
 
 // resolve-url-loader must be used before sass-loader
-environment.loaders.get('sass').use.splice(-1, 0, {
+environment.loaders.get('css').use.splice(-1, 0, {
   loader: 'resolve-url-loader',
   options: {
-    attempts: 1
+    // the resolve-url-loader options
   }
 });
 ```


pedrofurtado:
@jakeNiemiec Is something wrong? Let me know, then I can fix that in PR :handshake:

jakeNiemiec:
Mismanagement of the `select2` repo aside, like the doc says: `resolve-url-loader` must be used before sass-loader.

image

The word "Before" is kind of a misnomer since the loaders are executed bottom-to-top. 

Now that I am digging into the code https://github.com/rails/webpacker/blob/master/docs/webpack.md#loaders I see that webpackERs loaders are what webpack calls rules. By appending to the .css loader, I don't think you would be parsing any .scss files:

```js
rules: [
    {
      test: /\.css$/,
      use: [
        {
          loader: 'css-loader',
          options: {
            'import': true,
          },
        },
        {
          loader: 'resolve-url-loader',
          options: {
            // the resolve-url-loader options
          }
        }
      ]
    }
  ]
```

If your intent is to modify .css files, you can do that via css-loader options: https://github.com/webpack-contrib/css-loader#function

pedrofurtado:
@jakeNiemiec Thanks for explanation! :+1: It's more clear now. I fixed it, I made the wrong configuration here.

gauravtiwari:
Thanks, I guess the previous example still stands correct?

jakeNiemiec:
> I guess the previous example still stands correct?

@gauravtiwari Yes. There also may be an oppertunity to mirror `resolve-url-loader` function with expanded `css-loader` [functions](https://github.com/webpack-contrib/css-loader#function). I planned to dig into this a bit after the dust settles a bit from 4.0, but I am not yet convinced that this would work for every use-case.

PR #1929, opened 2019-02-06
Update v4 guide, add section about new namespaces

Zyphrax:
- A few small layout changes to keep it in style with the other docs
- Added a section about the new namespaces
- Mention of the v4 upgrade guide in the main README

PR #1958, opened 2019-02-22
Upgrade the upgrade (instructions)

jaredbeck:
After a major-version upgrade (eg. webpacker 3 to 4), I found it necessary to run the `webpacker:install` task, to update (replace) files such as `babel.config.js`

I barely know what I'm doing, so if there's a better way, please correct me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -298,6 +298,10 @@ yarn add webpack-dev-server@^3.1.14
 
 # Or to install a latest release (including pre-releases)
 yarn add @rails/webpacker@next
+
+# After a major-version upgrade (eg. webpacker 3 to 4)
+rm .babelrc
+bundle exec rails webpacker:install # or webpacker:install:react
 ```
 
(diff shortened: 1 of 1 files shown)


danielpowell4:
@jaredbeck here is a bit more thorough guide:

https://github.com/rails/webpacker/blob/master/docs/v4-upgrade.md

My guess is that when stable 4 comes out there will be a decent how-to release post of some sort. With Rails 6 just around the corner, I assume it's pretty close!

jaredbeck:
> @jaredbeck here is a bit more thorough guide:
> 
> https://github.com/rails/webpacker/blob/master/docs/v4-upgrade.md

Cool. I didn't realize that existed, thanks. Should we link to it from the readme?

gauravtiwari:
Thanks, @jaredbeck will merge #1929 which includes this change.

PR #1968, opened 2019-02-28
Updated example of json loader usage

nirname:
Corrected wrong example of usage
Blind copying and pasting leads to `ReferenceError`

### DIFF
--- a/docs/webpack.md
+++ b/docs/webpack.md
@@ -69,12 +69,13 @@ yarn add json-loader
 // config/webpack/environment.js
 const { environment } = require('@rails/webpacker')
 
-environment.loaders.append('json', {
+const jsonLoader = {
   test: /\.json$/,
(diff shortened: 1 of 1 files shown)


jakeNiemiec:
I would consider defining it at the top like: 
```js

const jsonLoader =  {
  test: /\.json$/,
  use: 'json-loader'
}

// Insert json loader at the end of list
environment.loaders.append('json', jsonLoader)

// Insert json loader at the top of list
environment.loaders.prepend('json', jsonLoader)

// Insert json loader after/before a given loader
environment.loaders.insert('json', jsonLoader, { after: 'style'} )
environment.loaders.insert('json', jsonLoader, { before: 'babel'} )

module.exports = environment
```

nirname:
Agree, since this commands are equal, you suggestion seems to be clearer and straightforward

Issue #1938, opened 2019-02-12
Media assets ignore folder structure


gaffneyc:
We just upgraded an app to RC7 and ran into the change to `asset_pack_path` where all assets are now collapsed into a single `media` directory. We use nested directories to namespace our assets and I find the new behavior [surprising](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) and potentially dangerous.

For example, let's say we have two icons that each represent a component or module within our system.
```
/app/javascripts/images/projects/icon.svg
/app/javascripts/images/users/icon.svg
```

Webpacker will copy both of these files to `media/icon.svg` with one overwriting the other. It was previously suggested to [rename the images](
https://github.com/rails/webpacker/issues/1915#issuecomment-457608282) but that violates the conventions we have in our application. Additionally it would create either a junk drawer (every random file in the same directory) situation or a lot of name stuttering (`projects/projects-icon.svg`).

I believe the default should be to maintain the nesting that folks choose for their applications, allowing us to reference images in a way that matches that nesting and namespacing.

joelmoss:
I too found this really quite disturbing. Both for the reasons stated above and also because it's such a change with little or no benefit.

gaffneyc:
#1887 - Original issue that caused [the change](https://github.com/rails/webpacker/pull/1914)

dwightwatson:
I commented on that other issue but glad to see it's not just me. The silent overwriting of assets with the same name is dangerous, and it feels like a less organised way to deal with assets.

joelmoss:
Now I know it's the "Rails Way" with convention over configuration and all that, which I would usually subscribe to, but the default Webpack config that Webpacker produces is now ways away from what I need in my projects. And this means me having to make more than just changes to the webpacker config.

Yes I know I can edit the config by adding/removing/merging config options, etc. but this is so messy and unwieldy.

I also find that a YAML config AND the actual Webpack config is a conflicting construct.

I've now got to the point where I simply want to create my own Webpack config from scratch, without any external forces making further (unwanted) changes.

But I appreciate the conveniance of the server side support that Webpacker provides out of the box. Things like Rails integration, compilation, and deployment are hugely conveniant.

So right now I'm exploring my own solution that will solve the above issues.

theJoeBiz:
[...]

Issue #1947, opened 2019-02-15
postcss config cascading isn't working


timonbandit:
Hey!
I'm trying to use several postcss.config.js for different folders (different templates) but postcss-loaders loads only one config from the root of a project.
I'm sure this is the problem on this line  get_style_rule.js:31
` config: { path: resolve() },`
I think this is a not correct way of using `path` in postcss-loader because post-css-loader documentation tells us - `Config lookup starts from path.dirname(file) and walks the file tree upwards until a config file is found.`

when I removed this line it helped me.

timonbandit:
small workaround for this issue is changing the rule with postcss-loader
```
environment.loaders.get('sass').use.splice(2, 1, {
  loader: 'resolve-url-loader',
  options: {}
});
```

in `config/webpack.environment.js`

guillaumebriday:
Can this issue be closed ?

Issue #1922, opened 2019-01-29
CDN host for staging environments


kpelelis:
**Webpacker version**: 4.0.0.rc7
**Rails Version**: 5.1.6

Greetings! We have integrated webpacker into our asset pipeline and now we are in the process of updating it to v4 in order to be in sync with the latest updates. Both testing and development environments are working just fine but we are facing some issues our staging environment which,is almost a mirror of the production setup.

Our assets are built using [our build server](https://github.com/skroutz/mistry/) with `RAILS_ENV=production` and are then moved into the application servers. A key concept in this build server is that it only generates production builds, as we want to keep staging and production environments as similar as possible and also leverage incremental building.

One of the few differences that the staging environment has, is that we do not use CDNs to distribute our assets. With the introduction of #1084 the URLs generated by the Manifest Plugin are prefixed with the production asset server URL (which in our case is a CDN). Currently there is no way to opt out of this feature without generating staging specific builds. Is there a chance to make this feature opt-in? I can get onto the task if it a valid change.

For the shake of completeness, the manifest file now looks something like:

```json
{
  "bundle1": "//my.cdn.com/assets/webpack/bundle1-abcdef0123456789.js"
}
```

Whereas before the update it looked like:

```json
{
  "bundle1": "/assets/webpack/bundle1-abcdef0123456789.js"
}
```

iggant:
this is duplicate https://github.com/rails/webpacker/issues/1845

guilleiguaran:
Closing in favor of #1845

Issue #1915, opened 2019-01-25
Webpack v4 rc7 images wrong pack reference


JanStevens:
Hello,

We are following every RC update so today I upgraded from rc5 to rc7 and I noticed that all my images now need to be references differently. To get started some configuration files:

```yml
default: &default
  source_path: app/javascript
  source_entry_path: packs
  public_root_path: public
  public_output_path: packs
  cache_path: tmp/cache/webpacker
  check_yarn_integrity: false
  webpack_compile_output: false

  # Additional paths webpack should lookup modules
  # ['app/assets', 'engine/foo/app/assets']
  resolved_paths: ['app/javascript/src']

  # Reload manifest.json on all requests so we reload latest compiled packs
  cache_manifest: false

  # Extract CSS files
  extract_css: true
  
  static_assets_extensions:
    - .jpg
    - .jpeg
    - .png
    - .gif
    - .tiff
    - .ico
    - .svg
    - .eot
    - .otf
    - .ttf
    - .woff
    - .woff2

  extensions:
    - .mjs
    - .jsx
    - .js
    - .sass
    - .scss
    - .css
    - .module.sass
    - .module.scss
    - .module.css
    - .png
    - .svg
    - .gif
    - .jpeg
    - .jpg
    - .graphql

development:
  <<: *default
  compile: true

  check_yarn_integrity: true
```

In our main `application.js` file we require the images (else they would never be included in the pack)

```js
import 'bootstrap'
// Start Rails UJS
import Rails from 'rails-ujs'
Rails.start()
// Require images
require.context('../images/', true, /\.(gif|jpg|png|svg|ico)$/i)
```

Our `environment.js` didn't change, we do add the sass `resolve-url-loader`

```js
environment.loaders.get('sass').use.splice(-1, 0, {
  loader: 'resolve-url-loader'
})
```

For example the favicon in RC5 could be referenced as:

```erb
  <link rel="shortcut icon" href="<%= asset_pack_path('images/favicon.ico') %>" data-environment="<%= Rails.env %>">
```

But in RC7 this needs to be

```erb
<link rel="shortcut icon" href="<%= asset_pack_path('media/favicon.ico') %>" data-environment="<%= Rails.env %>">
```

Even more problematic are subdirectories in the images (for example `app/javascript/images/logo/brand-color.png`), in RC5

```erb
<%= image_pack_tag('images/logo/brand-color.png', class: 'd-inline-block align-top') %>
```

[...]

PR #1973, opened 2019-03-03
Correct file loader path for media outside of context

gauravtiwari:
Updated file loader to have filename based on the path. This change
keeps the old behaviour intact i.e. let people use namespaces for media
inside `app/javascript` and also include media outside of `app/javascript`
with simpler paths, for example from `node_modules` or `app/assets`

```bash
# Files inside app/javascript (i.e. packs source path)
# media/[full_path_relative_to_app_javascript]/name_of_the_asset_with_digest

media/images/google-97e897b3851e415bec4fd30c265eb3ce.jpg
media/images/rails-45b116b1f66cc5e6f9724e8f9a2db73d.png
media/images/some_namespace/google-97e897b3851e415bec4fd30c265eb3ce.jpg

# Files outside app/javascript (i.e. packs source path)
# media/[containing_folder_name]/name_of_the_asset_with_digest

media/some_assets/rails_assets-f0f7bbb5.png
media/webfonts/fa-brands-400-4b115e11.woff2
```

This change is done so we don't end up paths like `media/_/assets/images/rails_assets-f0f7bbb5ef00110a0dcef7c2cb7d34a6.png` or `media/_/_/node_modules/foo-f0f7bbb5ef00110a0dcef7c2cb7d34a6.png` for media outside of
`app/javascript`

Properly fixes:  #1938, #1915, #1947, and #1922

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,20 @@ module.exports = environment.toWebpackConfig()
 
 - Reintroduced `context` to the file loader. Reverting the simpler paths change
 
+- Updated file loader to have correct paths based on location of assets.
+
+```bash
+# Files inside app/javascript (packs source)
+# media/[full_path_relative_to_app_javascript]/name_of_the_asset_with_digest
+media/images/google-97e897b3851e415bec4fd30c265eb3ce.jpg
+media/images/rails-45b116b1f66cc5e6f9724e8f9a2db73d.png
+media/images/some_namespace/google-97e897b3851e415bec4fd30c265eb3ce.jpg
+
+# Files outside app/javascript (packs source)
+# media/[containing_folder_name]/name_of_the_asset_with_digest
+media/some_assets/rails_assets-f0f7bbb5.png
+media/webfonts/fa-brands-400-4b115e11.woff2
+```
 
 ## [4.0.0.rc.7] - 2019-01-25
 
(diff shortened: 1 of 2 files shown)


gauravtiwari:
Does this change work for everybody? @gaffneyc @JanStevens @dwightwatson 

Apologies for the trouble :) The main reason for introducing flat structure was that it's easy to work with - including in the views and has cleaner paths (outside of pack source) but I see since we have been using this approach for a while it's quite a big change, hence this PR.

gauravtiwari:
Perhaps, we can move to use just the `[folder]` name instead of `[path]` but again it would be a big change for people who are heavily using a nested structure and including them in views.

JanStevens:
This indeed seems to fix it, thanks! 👍

gaffneyc:
Does the alternate version of this (with just the folder) have the same problem of files clobbering each other? Are the URLs you're trying to simplify / prettify human generated or system generated?

gauravtiwari:
> Does the alternate version of this (with just the folder) have the same problem of files clobbering each other? 

It will preserve the containing folder but not the whole path and no, the assets won't be clobbered together. 

> Are the URLs you're trying to simplify / prettify human generated or system generated?

This comes from `[path]` option supplied in the name, `[path]` includes the full relative path of the included asset, which works well but not for assets outside of the source path.

PR #1980, opened 2019-03-05
Update installation instructions

seuros:


connorshea:
The readme probably doesn’t need the section for the legacy version, I’d say you can get rid of that.

gauravtiwari:
Oops, didn't see this PR. Fixed on master already

PR #1989, opened 2019-03-08
Don't show 'Compilation failed' unless the status exit is non-zero

### LINKED ISSUE #1988: Deprecation warnings trigger "Compilation failed" warning message in webpacker compile
If there are any DeprecationWarnings from webpack, like everyone's favorite:

```
(node:71336) DeprecationWarning: Tapable.plugin is deprecated. Use new API on `.hooks` instead
```

Webpacker will print out "Compilation failed..." even though the status code from webpack was 0.

The reason for this is `util.deprecate` prints out to STDERR: https://nodejs.org/api/util.html#util_util_deprecate_fn_msg_code

And Webpacker reports a compilation error even though compilation succeeded:

https://github.com/rails/webpacker/blob/f88bbad3ae3aa3e6172e1664e5eca9faa7fbb256/lib/webpacker/compiler.rb#L67-L72

I think the right thing to do is to stop checking for `stderr == ""` and just rely on status exit code, as it's wildly unpredictable what libs will print to stderr or not.

Thoughts?

dbalatero:
Fixes #1988

### DIFF
--- a/lib/webpacker/compiler.rb
+++ b/lib/webpacker/compiler.rb
@@ -58,17 +58,21 @@ class Webpacker::Compiler
     def run_webpack
       logger.info "Compiling…"
 
-      stdout, sterr , status = Open3.capture3(
+      stdout, stderr, status = Open3.capture3(
         webpack_env,
         "#{RbConfig.ruby} ./bin/webpack",
         chdir: File.expand_path(config.root_path)
       )
 
-      if sterr == "" && status.success?
+      if status.success?
         logger.info "Compiled all packs in #{config.public_output_path}"
-        logger.info stdout if config.webpack_compile_output?
+        logger.error "#{stderr}\n" unless stderr.empty?
       else
-        logger.error "Compilation failed:\n#{sterr}\n#{stdout}"
+        logger.error "Compilation failed:\n#{stderr}"
+      end
+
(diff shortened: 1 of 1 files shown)


gauravtiwari:
thanks

lib/webpacker/compiler.rb
```diff
@@ -58,17 +58,21 @@ def record_compilation_digest
...
       )
 
-      if sterr == "" && status.success?
+      if status.success?
         logger.info "Compiled all packs in #{config.public_output_path}"
-        logger.info stdout if config.webpack_compile_output?
+        logger.error "#{stderr}\n" unless stderr.empty?
       else
-        logger.error "Compilation failed:\n#{sterr}\n#{stdout}"
+        logger.error "Compilation failed:\n#{stderr}"
```
sharang-d: Why does the `else` case have no `\n` after `#{stderr}`?

dbalatero: Hmm I think I've been used to loggers appending their own newlines. Is that not the case here?

sharang-d: I only asked because there's a discrepancy. We append a new line in the `if` case but not in the `else` case. Both of the places use `logger.error()`.

Issue #2000, opened 2019-03-12
./bin/webpack-dev-server command line arguments are ignored


marisveide:
I am using the `rails/webpacker` `v4.0.2`.

When I run the `./bin/webpack-dev-server` with command line arguments, those actually don't get passed to the actual `webpack-dev-server`. As a result - I cannot pass the needed arguments.

## Steps to replicate the issue:
1. Set the breakpoint here:
https://github.com/rails/webpacker/blob/7a241e75bd5376524348466aa2daa11661c49f41/lib/webpacker/dev_server_runner.rb#L58

2. Run the command with the debuger:
```bash
./bin/webpack-dev-server --https --cert ~/.ssl/private.pem --key ~/.ssl/private.key
```

3. See that the arguments were ignored.
The command being built here basically is not passing any of the original `@args` to the `webpack-dev-server` command:
```ruby
      def execute_cmd
        env = Webpacker::Compiler.env

        cmd = if node_modules_bin_exist?
          ["#{@node_modules_bin_path}/webpack-dev-server"]
        else
          ["yarn", "webpack-dev-server"]
        end
        cmd += ["--config", @webpack_config]
        cmd += ["--progress", "--color"] if @pretty

        Dir.chdir(@app_path) do
          Kernel.exec env, *cmd
        end
      end
```

## Related issues
* https://github.com/rails/webpacker/issues/612#issuecomment-319930567
* https://github.com/webpack/webpack-dev-server/issues/1695

## Suggested solution

Add the line to pass the arguments from `@argv` to the real `webpack-dev-server` command.

Changed code:
```ruby
      def execute_cmd
        env = Webpacker::Compiler.env

        cmd = if node_modules_bin_exist?
          ["#{@node_modules_bin_path}/webpack-dev-server"]
        else
          ["yarn", "webpack-dev-server"]
        end
        cmd += ["--config", @webpack_config]
        cmd += ["--progress", "--color"] if @pretty
        cmd += @argv  # <===== Add this line here

        Dir.chdir(@app_path) do
          Kernel.exec env, *cmd
        end
      end
```

marisveide:
I think, this will also solve the #1758.

jipiboily:
This is super useful, and in fact required for me.

Use case:

I integrate with a bunch of third party platforms via OAuth. Some of them doesn't work with localhost (Mailchimp, for example), so I moved to 127.0.0.1 but others don't support 127.0.0.1 (FB Ads). So I decided to use a custom domain.

[...]

PR #2002, opened 2019-03-12
Pass the command line arguments to the webpack-dev-server module (#2000)

marisveide:
Fixes the issue: https://github.com/rails/webpacker/issues/2000

### DIFF
--- a/lib/webpacker/dev_server_runner.rb
+++ b/lib/webpacker/dev_server_runner.rb
@@ -53,6 +53,7 @@ module Webpacker
         end
         cmd += ["--config", @webpack_config]
         cmd += ["--progress", "--color"] if @pretty
+        cmd += @argv
 
         Dir.chdir(@app_path) do
           Kernel.exec env, *cmd
--- a/test/dev_server_runner_test.rb
+++ b/test/dev_server_runner_test.rb
@@ -18,6 +18,13 @@ class DevServerRunnerTest < Webpacker::Test
     verify_command(cmd, use_node_modules: true)
   end
 
+  def test_run_cmd_via_node_modules_with_arguments
+    argv = ["--key", "~/.ssl/private.key"]
+    cmd = ["#{test_app_path}/node_modules/.bin/webpack-dev-server", "--config", "#{test_app_path}/config/webpack/development.js"] + argv
+
+    verify_command(cmd, use_node_modules: true, argv: argv)
+  end
+
   def test_run_cmd_via_yarn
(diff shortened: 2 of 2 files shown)


gauravtiwari:
Thanks @marisveide 

The reason for not allowing config options in dev server runner is because of sync problem in options like `host` and `port` etc since they are used on both Ruby and JS side and it's hard to keep them in sync if we have too many sources to set them up (plus it might be confusing for some people). 

At the moment, Webpacker supports `.yml` file and `environment variables`

Perhaps, we can allow all other options except certain like `host` and `port`. This may need some testing.

marisveide:
@gauravtiwari - so are you saying this Pull Request will not get rejected?

This change really helped us set up our development environments properly. And looks like it helped others too already who need to pass the file names from the command line...

nicklozon:
@gauravtiwari using ```webpacker.yml``` and environment variables is okay, but it does make the configuration less flexible because @marisveide will need to be copying his key and cert into the yaml file whenever it changes.

Another issue I ran into is the webpack-dev-server ```https``` configuration isn't a boolean, but webpacker's reverse proxy won't use https unless that configuration (or the environment variable) is explicitly set to true, and if it is set to true then the key and cert files will never make it to the webpack-dev-server configuration.

Here is my comment about it: https://github.com/rails/webpacker/issues/2000#issuecomment-485312873

dj0nes:
Ran into this issue in development.

@gauravtiwari - If we filter out `host` and `port` from the argument list, would we be closer to getting this feature merged?

marisveide:
Yeah, it would be great to merge this... otherwise we are sitting on our fork, and would be glad to not have to maintain the forked version just because we need to pass the SSL certificate params.

marisveide:
Hi!

What do I have to do to get this PR merged?

Thanks!

Maris

gauravtiwari:
@marisveide and everyone, apologies for long silence 🙏 

One solution: 

If you can filter out the options supported in webpacker.yml inside dev server then I am happy to review. Perhaps, if one supplies the options already supported in `webpacker.yml`, we can display a friendly warning i.e. "Please set the option in `webpacker.yml` instead"

The reason to do this is because, we use some options on both node and ruby side (host, port etc.) and if there are multiple ways to declare, it will cause confusion and issues. 

---

I know this still doesn't solve the whole problem, for example colleagues working together want to change these options on runtime without committing. Maybe let's try to figure out a better solution, together. 

Simply, allowing options to dev server would cause other problems.

PR #2008, opened 2019-03-13
Added Custom Capistrano Hook for Yarn

majksner:
In case of `node_modules` being linked to `shared` folder we need to run `yarn install` so changed `package.json` gets processed.

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -83,3 +83,19 @@ Make sure you have `public/packs` and `node_modules` in `:linked_dirs`
 append :linked_dirs, "log", "tmp/pids", "tmp/cache", "tmp/sockets", "public/packs", ".bundle", "node_modules"
 ```
 
+If you have `node_modules` added to `:linked_dirs` you'll need to run yarn install before `deploy:assets:precompile`, so you can add this code snippet at the bottom deploy.rb
+
+```
+before "deploy:assets:precompile", "deploy:yarn_install"
+namespace :deploy do
+  desc "Run rake yarn install"
+  task :yarn_install do
+    on roles(:web) do
+      within release_path do
+        execute("cd #{release_path} && yarn install --silent --no-progress --no-audit --no-optional")
+      end
+    end
+  end
+end
+```
+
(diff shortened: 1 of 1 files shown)


docs/deployment.md
```diff
@@ -83,3 +83,19 @@ Make sure you have `public/packs` and `node_modules` in `:linked_dirs`
 append :linked_dirs, "log", "tmp/pids", "tmp/cache", "tmp/sockets", "public/packs", ".bundle", "node_modules"
 ```
 
+If you have `node_modules` added to `:linked_dirs` you'll need to run yarn install before `deploy:assets:precompile`, so you can add this code snippet at the bottom deploy.rb
+
+```
```
jakeNiemiec: Adding this will give you highlighting:
```
```ruby 
```

b-nik:
Maybe I'm reading this wrong
https://github.com/rails/webpacker/blob/master/lib/tasks/webpacker/compile.rake#L18-L24
https://github.com/rails/webpacker/blob/master/lib/tasks/webpacker/compile.rake#L49

But it seems that webpacker runs "yarn install" anyway during assets:precompile. 

Am I missing something?

majksner:
@b-nik Please read it carefully "If you have node_modules added to :linked_dirs, ...."

b-nik:
@majksner Yes I know, what I'm saying is that webpacker runs "yarn install"  anyway during assets:precompile, so I can't see why we'd have to add the extra capistrano task.

This is what I see during deploy. I do have node_modules in the linked_dirs. I __do not__ have the extra yarn install task.

```
00:32 deploy:assets:precompile
      01 $HOME/.rbenv/bin/rbenv exec bundle exec rake assets:precompile
      01 yarn install v1.17.3
      01 [1/4] Resolving packages...
      01 success Already up-to-date.
      01 Done in 0.88s.
      01 Compiling…
      01 Compiled all packs in ......
```

My understanding is that adding the task will make capistrano run "yarn install" once, and then when it runs assets:precompile it will run yarn install again (from webpacker)

majksner:
@b-nik Could it by that it was fixed by https://github.com/rails/webpacker/commit/c5065217284e4540c07fefc35fe27d33fc531082

Issue #2003, opened 2019-03-12
Inconsistent file name / hash with extract css


agrieser:
Rails 5.2.2
webpacker 4.0.2
`extract_css` is set to `true`

The app is deployed in heroku, and on every release the filename/hash of the css changes, even if there are no changes to the css. I verified this by comparing the md5 of the css between releases and it is identical, however the filename/hash of the generated css file changes.

environment.js
```
const { environment } = require('@rails/webpacker')

const vue =  require('./loaders/vue')
const vueResolver = require("./resolve/vue")

environment.loaders.append('vue', vue);
environment.config.merge(vueResolver);

module.exports = environment
```

postcss.config.js
```
var tailwindcss = require("tailwindcss");

module.exports = {
  plugins: [
    tailwindcss("./app/javascript/src/tailwind.js"),
    require('postcss-import'),
    require('postcss-flexbugs-fixes'),
    require('postcss-preset-env')({
      autoprefixer: {
        flexbox: 'no-2009'
      },
      stage: 3
    })
  ]
}
```

Edit: running `RAILS_ENV=production rails assets:precompile` on my local machine produces the same asset hash/filename every time (clearing out all packs and tmp/cache in between), but on heroku the asset hash/filename is different on each deploy, but downloading the css from heroku and comparing via md5 indicates they are identical, even though the filenames are different. The javascript assets have the same filenames between releases.

agrieser:
For anyone else who ends up here: the issue was with intermediate sourcemaps that contained build path information being included as part of the hash, and thus filename. The reason this didn't show up on my local machine is that the build paths were the same every time, but in heroku the buildpaths include a randomly named tmp directory. Which means the hash of the content will be different for each deploy. In the final step, when the css was written into it's own file, the sourcemap info was removed so the files were identical, however the hashes still included that information and thus the filenames were different.

I tried using the option `environment.config.merge({ devtool: 'none' })` to disable sourcemaps, but this only disables sourcemaps at the final step. Intermediate sourcemaps are still built (and looking at the code they are hard coded: https://github.com/rails/webpacker/blob/1e9e2d04de100a3d5beb085ccaad061440265dc7/package/utils/get_style_rule.js#L13)

The workaround was to iterate through all the style loaders and merge into their configs:

`config/webpacker/environment.js`:
```
const merge = require('webpack-merge')

const sassLoader = environment.loaders.get('sass')
const cssLoader = environment.loaders.get('css')

sassLoader.use.map(function(loader) {
[...]

PR #2009, opened 2019-03-13
Fix inconsistent hash generation in environments with randomized build paths

agrieser:
Fixes https://github.com/rails/webpacker/issues/2003

Use the `devtool` option to conditionally disable sourcemap generation in css-loader. 

The `devtool` setting was used because it is the recommended way to disable sourcemaps from the v4 upgrade docs: https://github.com/rails/webpacker/blob/master/docs/v4-upgrade.md#source-maps-enabled-by-default

### DIFF
--- a/package/utils/get_style_rule.js
+++ b/package/utils/get_style_rule.js
@@ -15,11 +15,14 @@ const styleLoader = {
 }
 
 const getStyleRule = (test, modules = false, preprocessors = []) => {
+  const options = modules ? { include: /\.module\.[a-z]+$/ } : { exclude: /\.module\.[a-z]+$/ }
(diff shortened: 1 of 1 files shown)


package/utils/get_style_rule.js
```diff
@@ -15,11 +15,14 @@ const styleLoader = {
 }
 
 const getStyleRule = (test, modules = false, preprocessors = []) => {
+  const options = modules ? { include: /\.module\.[a-z]+$/ } : { exclude: /\.module\.[a-z]+$/ }
+  const sourceMap = options.devtool === 'none'
```
jakeNiemiec: options.devtool will always be `undefined`

Issue #2011, opened 2019-03-16
Changelog should include links to diff


justin808:
@gauravtiwari It would be great if somebody could update the changelog to be like the one on [react_on_rails/blob/master/CHANGELOG.md](https://github.com/shakacode/react_on_rails/blob/master/CHANGELOG.md) so that your square braces work to show the diffs.

justin808:
Was done a long time ago, since

https://github.com/rails/webpacker/blob/master/CHANGELOG.md#403---2019-05-28

PR #2012, opened 2019-03-16
Update CHANGELOG.md with diff link

justin808:
Reported in #2011.

Just did the first one as an example.

I recommend merging and leave a help-wanted tag for the others.


gauravtiwari:
Closing, since we do this from 4.x onwards...

PR #2015, opened 2019-03-17
Remove ^ from webpack-dev-server@^3.1.14

Diogomartf:
When I run `yarn add webpack-dev-server@^3.1.14`, I've got no matches found: webpack-dev-server@^3.1.14.

It works great by removing the ^


gauravtiwari:
Thanks @Diogomartf Just updated the section, we don't need it for the stable version anymore.

PR #2020, opened 2019-03-19
Allow NODE_ENV to be overridden in webpacker:compile task

### LINKED ISSUE #1558: Compile is always using production env?
It seems like with the changes introduced in d8dd4460144e63a9120290631c492aba3cbea332, the webpack gem is always going to build in the production environment, regardless of the environment that's set via the environment variable.

Was this change intentional?  Or am I missing something in terms of how to build non-production with this gem?

dbalatero:
Closes #1558

Since users might have a default `NODE_ENV=development` set in their `.envrc` or `.env` file, I added a warning message letting them know that they were choosing a different `NODE_ENV` than the default `production` value. I think this will help avoid opaque compilation bugs.

### DIFF
--- a/lib/tasks/webpacker/compile.rake
+++ b/lib/tasks/webpacker/compile.rake
@@ -26,7 +26,18 @@ end
 namespace :webpacker do
   desc "Compile JavaScript packs using webpack for production with digests"
   task compile: ["webpacker:verify_install", :environment] do
-    Webpacker.with_node_env("production") do
+    node_env = "production"
+
+    if ENV.has_key?("NODE_ENV")
+      node_env = ENV["NODE_ENV"]
+
+      if node_env != "production"
+        $stdout.puts "Using NODE_ENV=#{ENV["NODE_ENV"]} from local "\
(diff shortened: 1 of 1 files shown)


lib/tasks/webpacker/compile.rake
```diff
@@ -26,7 +26,18 @@ end
 namespace :webpacker do
   desc "Compile JavaScript packs using webpack for production with digests"
   task compile: ["webpacker:verify_install", :environment] do
-    Webpacker.with_node_env("production") do
+    node_env = "production"
```
awcodify:
```
node_env = ENV["NODE_ENV"] || "production"

if node_env != "production"
  $stdout.puts "Using NODE_ENV=#{ENV["NODE_ENV"]} from local "\
     "environment instead of default NODE_ENV=production"
end
```

gauravtiwari:
Fixed here: https://github.com/rails/webpacker/commit/4c79d225a7da4178f3c9e1169548ff7199d3b821

Issue #2058, opened 2019-04-17
Webpacker default react config


JohnClema:
I currently am configuring a `babel.config.js` for a long lasting rails/webpacker/react project. 

Why does the provided install config use a preset/plugin format of `require('package').default` instead of `'package'`. 

For example [here](https://github.com/rails/webpacker/blob/404751f36f883db19221eeab257a5dbcdbc21540/lib/install/examples/react/babel.config.js) under line 21 `require('@babel/preset-env').default` is used as a preset, why not use `@babel/preset-env`?

jakeNiemiec:
Agree, we should stick closely with what the babel docs say to do. 

If you created a PR, it would probably be accepted.

JohnClema:
Thanks for the quick response 🙏 will do

JohnClema:
Just opened it here: #2060

guillaumebriday:
Can this issue be closed ?

PR #2060, opened 2019-04-18
Update webpacker config to not include require('babel-plugin').default syntax

JohnClema:
## Issue ( #2058 ) 
Currently the example react babel.config.js example uses verbose syntax in some areas. Eg. 
`require('@babel/preset-env').default`

## Solution
Quite sure the `require('plugin-or-preset').default` syntax is not required for the babel config and we can just use 'plugin-or-preset' eg. '@babel/preset-env'

### DIFF
--- a/lib/install/examples/react/babel.config.js
+++ b/lib/install/examples/react/babel.config.js
@@ -18,15 +18,17 @@ module.exports = function(api) {
   return {
     presets: [
       isTestEnv && [
-        require('@babel/preset-env').default,
+        '@babel/preset-env',
         {
           targets: {
             node: 'current'
-          }
-        }
+          },
+          modules: 'commonjs'
+        },
+        '@babel/preset-react'
       ],
       (isProductionEnv || isDevelopmentEnv) && [
-        require('@babel/preset-env').default,
+        '@babel/preset-env',
         {
           forceAllTransforms: true,
           useBuiltIns: 'entry',
@@ -35,7 +37,7 @@ module.exports = function(api) {
         }
       ],
       [
-        require('@babel/preset-react').default,
+        '@babel/preset-react',
         {
           development: isDevelopmentEnv || isTestEnv,
           useBuiltIns: true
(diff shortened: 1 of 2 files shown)


JohnClema:
@jakeNiemiec is there any further process I need to go through to merge this?

jakeNiemiec:
It's been a month since any PRs were merged. Every so often gauravtiwari will go through the PRs (I'm not an Owner).

If you get impatient, you can request a review from them, but I don't think it would speed things up (everyone spams pings).

jakeNiemiec:
It's been a month since any PRs were merged. Every so often ＠gauravtiwari will go through the PRs (I'm not an Owner). **You might link to the babel docs to show that your changes match what babel recommends.**

If you get impatient, you can request a review from them, but I don't think it would speed things up (everyone spams pings).

JohnClema:
Cheers for the heads up, will link to the docs, much appreciated

gauravtiwari:
Apologies for the delay. The reason for using this syntax is, it would fail beforehand if the dependencies aren't satisfied. 

What issues have you encountered due to this?

jakeNiemiec:
>Apologies for the delay.

No apologies necessary, FOSS works best when slow & steady.

> The reason for using this syntax is, it would fail beforehand if the dependencies aren't satisfied.

The dependencies _should_ be satisfied via `yarn`, plus one of the plugins there doesn't contain an `export default`. 

If you just pass a string as your [EntryTarget](https://babeljs.io/docs/en/options#entrytarget), babel will do all of the heavy lifting for you with [`name-normalization`](https://babeljs.io/docs/en/options#name-normalization) and [`merging`](https://babeljs.io/docs/en/options#merging).

Additionally, all of the babel docs use this convention.

maschwenk:
I think I ran into issues with AVA not understanding my babel config in this format. Also, as @jakeNiemiec mentioned, if you do mess up here, your compile process will fail-fast.

JohnClema:
@gauravtiwari if I am honest we found the syntax cumbersome, and encountered similar blockers to @maschwenk with tools not recognizing the syntax. It seems pretty standardized syntax with babel v7 as @jakeNiemiec mentioned. 

In addition to all of the docs on the [babel website](https://babeljs.io/docs/en/usage). It seems like the [babel repository's config has this syntax too](https://github.com/babel/babel/blob/master/babel.config.js). 

I too am definitely for the slow FOSS approach and discussing the downsides of adopting this. How does the `require('plugin-or-preset').default` cause faster failures?

gauravtiwari:
Could you please update babel config for react as well to use this syntax?

JohnClema:
@gauravtiwari of course, I will investigate this weekend

JohnClema:
Thanks for the review 🙏 @gauravtiwari. I just updated the main babel config in `lib/install/config` with this syntax. Just confirming this is what you meant? Or were their other changes you wanted?

gauravtiwari:
Thanks @JohnClema for the patience. Looks great 👍

lib/install/examples/react/babel.config.js
```diff
@@ -18,15 +18,17 @@ module.exports = function(api) {
...
+        '@babel/preset-env',
         {
           targets: {
             node: 'current'
-          }
-        }
+          },
+          modules: 'commonjs'
+        },
+        '@babel/preset-react'
```
roumanite, line 28: hi may I know what is this for? why is the '@babel/preset-react' mentioned 2 times and the first one is in 1 array with @babel/preset-env ? thanks

PR #2075, opened 2019-04-30
'node_modules' compilation should be configurable through webpacker.yml

maschwenk:
1. The `nodeModules` rule should be configurable, not buried in a v4 upgrade doc
1. I think the default of `true` is overly defensive/paranoid, and will likely cause compilation speeds to go _way_ down.
    - in future versions I think we should reverse this default

### DIFF
--- a/package/environments/base.js
+++ b/package/environments/base.js
@@ -22,7 +22,9 @@ const config = require('../config')
 
 const getLoaderList = () => {
   const result = new ConfigList()
-  Object.keys(rules).forEach(key => result.append(key, rules[key]))
+  Object.keys(rules)
+    .filter(rule => rule !== 'nodeModules' || (rule === 'nodeModules' && config.compile_dependencies))
+    .forEach(key => result.append(key, rules[key]))
   return result
 }
 
--- a/package/environments/base.js
+++ b/package/environments/base.js
@@ -22,9 +22,10 @@ const config = require('../config')
 
 const getLoaderList = () => {
   const result = new ConfigList()
-  Object.keys(rules)
-    .filter(rule => rule !== 'nodeModules' || (rule === 'nodeModules' && config.compile_dependencies))
-    .forEach(key => result.append(key, rules[key]))
+  const ruleKeys = config.compile_dependencies
+    ? Object.keys(rules)
+    : Object.keys(rules).filter(rule => rule !== 'nodeModules')
+  ruleKeys.forEach(key => result.append(key, rules[key]))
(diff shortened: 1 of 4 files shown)


package/environments/base.js
```diff
@@ -22,7 +22,9 @@ const config = require('../config')
 
 const getLoaderList = () => {
   const result = new ConfigList()
-  Object.keys(rules).forEach(key => result.append(key, rules[key]))
+  Object.keys(rules)
+    .filter(rule => rule !== 'nodeModules' || (rule === 'nodeModules' && config.compile_dependencies))
+    .forEach(key => result.append(key, rules[key]))
```
jakeNiemiec: You would have to test for yourself, but this might be a bit easier to read:

```js
if (config.compile_dependencies == false) {
  delete rules.nodeModules;
}

result = { ...result, ...rules };
// const result = { ...new ConfigList(), ...rules }; // I don't think this would work
```

maschwenk: Made it less terrible here: 26e0891

maschwenk:
@gauravtiwari Is this something we'd consider? Otherwise I will close.

0xdevalias:
Throwing my /2c on here. My understanding is still that within the ecosystem, most packages transpile themselves still. While I expect that may eventually change once proper modules are supported everywhere, etc, that really isn't the reality of things as it is now.

So given most things transpile themselves, making the default here to transpile things again seems at best inefficient, but likely to just cause a lot of headaches for people. The fact that it's apparently not easily configurable just compounds this even more.

Perhaps exposing a way to say ‘please transpile these specific things’ would make more sense? IMO it would definitely end up being a shorter whitelist list than trying to blacklist every dependency that is already transpiled.

maschwenk:
There have already been two recent bugs caused directly by this feature:

https://github.com/rails/webpacker/pull/2111/files#diff-456094c8451b5774db50028dfecf4aa8L10

https://github.com/rails/webpacker/pull/2110/files

package/environments/base.js
jakeNiemiec: I would still go with a simple `if`, there's no reason to iterate if we know exactly which value we don't want. The `keys`/`forEach`/`filter`/`append` stuff adds enough to the complexity without additional ternaries.

```js
if (config.compile_dependencies == false) {
  delete rules.nodeModules;
}
```

gauravtiwari:
Thanks @maschwenk Closing in favour of #2624

PR #2093, opened 2019-05-15
Allows setting of parallelization options

masato-hi:
If you set the parallel option of TersorPlugin to true, resource usage issues may occur.
This problem is noticeable when used on containers, and will fail to fork when the number of CPU cores on hardware is high and the memory usage allowed for the container is low.

This is due to the fact that the number of CPU cores count - 1 process is generated when the parallel option of TersorPlugin is true.

This patch makes it possible to limit the number of parallelizations in the following way.

How to write to `config/initializers/webpacker.rb`
```
Webpacker::Compiler.env['WEBPACKER_PARALLEL'] = '2'
```

or how to write it before doing `require('@rails/webpacker')` in `config/webpack/production.js`
```
process.env.WEBPACKER_PARALLEL = '2'

```

I'm sorry if you find something strange because I use Google Translate.

### DIFF
--- a/package/environments/production.js
+++ b/package/environments/production.js
@@ -36,7 +36,7 @@ module.exports = class extends Base {
       optimization: {
         minimizer: [
           new TerserPlugin({
-            parallel: true,
+            parallel: Number.parseInt(process.env.WEBPACKER_PARALLEL) || true,
             cache: true,
             sourceMap: true,
             terserOptions: {
--- a/package/environments/production.js
+++ b/package/environments/production.js
@@ -36,7 +36,7 @@ module.exports = class extends Base {
       optimization: {
         minimizer: [
           new TerserPlugin({
-            parallel: Number.parseInt(process.env.WEBPACKER_PARALLEL) || true,
+            parallel: Number.parseInt(process.env.WEBPACKER_PARALLEL, 10) || true,
             cache: true,
             sourceMap: true,
             terserOptions: {


masato-hi:
Thank you for the review.
But this is not a problem in the WSL environment.

TersorPlugin gets cpu count in this line.
https://github.com/webpack-contrib/terser-webpack-plugin/blob/master/src/TaskRunner.js#L22

When `isWSL = false && parallel = true`, `cpus.length -1` is assigned to maxConcurrentWorkers.
https://github.com/webpack-contrib/terser-webpack-plugin/blob/master/src/TaskRunner.js#L25

And the process is forked and executed by the number of maxConcurrentWorkers.
https://github.com/webpack-contrib/terser-webpack-plugin/blob/master/src/TaskRunner.js#L47

As a result, if the usable memory is 1GB and the CPU count is 4 cores and consumes 500MB per process, 1.5GB of memory is required in total and the build fails because of insufficient memory.

jakeNiemiec:
@masato-hi Could you please add documentation for this change?

YurySolovyov:
It seems that if you want to disable parallel completely there is no way to do that currently:
```js
> Number.parseInt('false', 10) || true
true
> Number.parseInt('0', 10) || true
true
```
Am I missing something?

masato-hi:
@YurySolovyov Please specify '1'.
`Number.parseInt ('false')` and `Number.parseInt ('0')` are NaN and 0, respectively, both of which evaluate to false.

PR #2106, opened 2019-05-28
Upgrade deps

gauravtiwari:
- [x] Node packages
- [x] Gems


aried3r:
Hey I just saw this PR, is there a reason to limit bundler to `< 2`? CI only installs <2, but is there a reason to limit it during development?

gauravtiwari:
Oops, my bad - forgot to remove it after fixing CI.

gauravtiwari:
@aried3r Fixed, thanks for pointing out.

PR #2031, opened 2019-03-26
deprecates babel/polyfill in favor of core-js@3

somebody32:
[babel/polyfill is deprecated](https://github.com/zloirock/core-js/blob/master/docs/2019-03-19-core-js-3-babel-and-a-look-into-the-future.md) and using core-js directly is the way now.

This pr adds the required modifications to support core-js@3. 
~I didn't remove `babel/polyfill` from the dependencies to not make this change breaking.~

Update: removed `babel/polyfill` because updating `core-js` would break old usage anyway, so updating after that PR would require changes to the user's codebase

PR #2107, opened 2019-05-28
Add changelog entry with @babel/polyfill upgrade instructions

jakeNiemiec:
Taking @connorshea's suggestion from https://github.com/rails/webpacker/pull/2031#issuecomment-483065730 to warn users who were using `@babel/polyfill`. With https://github.com/rails/webpacker/pull/2031 merged, users who upgrade will need to remove `import '@babel/polyfill'` from their pack files and add the replacements to those files and their `package.json`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,26 @@
 
 Please see the diff
 
+### Breaking changes
+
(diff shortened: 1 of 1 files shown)


jaredbeck:
> Change this ..
> import "@babel/polyfill";

What if we don't have such an import? Should we still add the new imports?

PikachuEXE:
If you use polyfill differently you can skip it
My project use feature detection and only load polyfill chunk if any feature missing

jakeNiemiec:
I would add the new imports, the docs say that the new imports can handle both core-js 2 & 3 https://babeljs.io/blog/2019/03/19/7.4.0#migration-from-core-js-2. 

It's hard to make a general recommendation since all of this depends on how you use `babel.config.js`. I expect issues to pop up, babel has always been that tool that you forget about until it stops working.

Issue #2109, opened 2019-05-29
Uncaught TypeError: __webpack_require__(...) is not a function - error on fresh app


ankane:
## Repro Steps

Create fresh rails app

```sh
rails _6.0.0.rc1_ new hello
```

Add a controller and a view, then load it.

## Expected Result

JavaScript loads correctly.

## Actual Result

```
Uncaught TypeError: __webpack_require__(...) is not a function
    at Module../node_modules/webpack/buildin/harmony-module.js (harmony-module.js:33)
    at __webpack_require__ (bootstrap:19)
    at Module../node_modules/@rails/ujs/lib/assets/compiled/rails-ujs.js (rails-ujs.js:822)
    at __webpack_require__ (bootstrap:19)
    at Object../app/javascript/packs/application.js (application.js:6)
    at __webpack_require__ (bootstrap:19)
    at bootstrap:83
    at bootstrap:83
```

May be related to recent release / #2031

ankane:
For others experiencing this issue, you can edit your Gemfile to:

```ruby
gem 'webpacker', '4.0.2'
```

And then run:

```sh
bundle
bundle exec rails webpacker:install
```

jakeNiemiec:
Can you post your `package.json`?

d-simon:
I have the same error on a fresh app. In addition to your solution I also had to change the corejs version in the `babel.config.js` to `2` (two occurences) to get this to work.

ankane:
package.json

```json
{
  "name": "hello",
  "private": true,
  "dependencies": {
    "@rails/actioncable": "^6.0.0-alpha",
    "@rails/activestorage": "^6.0.0-alpha",
    "@rails/ujs": "^6.0.0-alpha",
    "@rails/webpacker": "^4.0.4",
    "turbolinks": "^5.2.0"
  },
  "version": "0.1.0",
  "devDependencies": {
    "webpack-dev-server": "^3.4.1"
  }
}
```

babel.config.js

```js
module.exports = function(api) {
  var validEnv = ['development', 'test', 'production']
  var currentEnv = api.env()
  var isDevelopmentEnv = api.env('development')
  var isProductionEnv = api.env('production')
  var isTestEnv = api.env('test')

  if (!validEnv.includes(currentEnv)) {
    throw new Error(
      'Please specify a valid `NODE_ENV` or ' +
        '`BABEL_ENV` environment variables. Valid values are "development", ' +
        '"test", and "production". Instead, received: ' +
[...]

PR #2111, opened 2019-05-30
fix: webpack should not be transpiled

jakeNiemiec:
fixes: #2109


jakeNiemiec:
@gauravtiwari This fixes additional errors in #2109. 

**Note**: Tell me if the 4.0.6 commit is helpful, otherwise feel free to remove.

PR #2117, opened 2019-06-03
Record the compilation digest even on webpack error

### LINKED ISSUE #2113: With ts-loader, files aren't correctly cached if there is a compilation error. This leads to too many or too few calls to compile.
Webpacker 4.0.2 with ts-loader 6.0.1.

Summary: The cache file: `tmp/cache/webpacker/last-compilation-digest-development` is only updated if webpack compilation succeeds. However, even though ts-loader fails the webpack build if there is a Typescript error, webpack still emits an updated manifest.json and an output .js file. So, if you introduce an error in your source Typescript file and the webpack compiler runs, the cache file will still indicate that the output is based on the contents of the file BEFORE the error was introduced. So, if you then undo your change to the source file and reload the page, webpack is NOT invoked as it should be.

Furthermore, this means that as long as there is an error in any Typescript file, webpack will ignore the cache and will ALWAYS invoke compilation, causing repeated compilations of the exact same files with the exact same source.

Details:

Not using webpack dev server, but compiling on demand, meaning that it should only compile if there is a change.

I have a file in app/javascript/packs/authentications.ts that compiles correctly.
I load my login page and see that Webpack is compiling, as I expect:
```
web_1       | 18:07:28 web.1               | [Webpacker] Compiling…
web_1       | 18:07:39 web.1               | [Webpacker] Compiled all packs in /app/ynab_api/public/assets/packs
web_1       | 18:07:39 web.1               |   Rendered authentications/new.html.erb within layouts/application (10567.7ms)
web_1       | 18:07:39 web.1               |   Rendered shared/_favicon.erb (20.3ms)
web_1       | 18:07:39 web.1               | Completed 200 OK in 10919ms (Views: 10904.9ms | ActiveRecord: 0.0ms)
````
I look at my cache file and see:
```
$ cat tmp/cache/webpacker/last-compilation-digest-development
66a64d57cdc91eabf692e2bc6318c555fa6e10b2%
```

So far, so good. Now, introduce a compile error and reload the page.
Again, it compiles, but this time it displays a compilation error. So far, so good:
```
web_1       | 18:09:25 web.1               | [Webpacker] Compiling…
web_1       | 18:09:35 web.1               | [Webpacker] Compilation failed:
web_1       | 18:09:35 web.1               |
web_1       | 18:09:35 web.1               | Hash: 63e8d239697434693810
web_1       | 18:09:35 web.1               | Version: webpack 4.29.6
web_1       | 18:09:35 web.1               | Time: 8590ms
web_1       | 18:09:35 web.1               | Built at: 2019-05-31 18:09:35
web_1       | 18:09:35 web.1               |                                      Asset       Size           Chunks             Chunk Names
web_1       | 18:09:35 web.1               | js/authentications-f513c39af604d8951a62.js    377 KiB  authentications  [emitted]  authentications
web_1       | 18:09:35 web.1               |            js/otps-822bdb1661898df52846.js   5.48 KiB             otps  [emitted]  otps
web_1       | 18:09:35 web.1               |                              manifest.json  391 bytes                   [emitted]
web_1       | 18:09:35 web.1               | Entrypoint authentications = js/authentications-f513c39af604d8951a62.js
web_1       | 18:09:35 web.1               | Entrypoint otps = js/otps-822bdb1661898df52846.js
web_1       | 18:09:35 web.1               | [./app/javascript/packs/authentications.ts] 39 KiB {authentications} [built] [1 error]
web_1       | 18:09:35 web.1               | [./app/javascript/packs/otps.coffee] 1.65 KiB {otps} [built]
web_1       | 18:09:35 web.1               | [./node_modules/webpack/buildin/global.js] (webpack)/buildin/global.js 878 bytes {authentications} [built]
web_1       | 18:09:35 web.1               |     + 5 hidden modules
web_1       | 18:09:35 web.1               |
web_1       | 18:09:35 web.1               | ERROR in /app/ynab_api/app/javascript/packs/authentications.ts
web_1       | 18:09:35 web.1               | ./app/javascript/packs/authentications.ts
web_1       | 18:09:35 web.1               | [tsl] ERROR in /app/ynab_api/app/javascript/packs/authentications.ts(6,1)
web_1       | 18:09:35 web.1               |       TS2304: Cannot find name 'asdfsfsd'.
```

But now the output `.js` file and `manifest.json` HAS been updated, but the cache file has NOT been updated!
```
$ cat tmp/cache/webpacker/last-compilation-digest-development
66a64d57cdc91eabf692e2bc6318c555fa6e10b2%
```

So now, if I undo my change to the source file, restoring the source authentications.ts file to the state it was in before the error and reload the page, I do NOT see Webpack compile my file. 

And of course, the manifest.json and output .js files remain the same, meaning that the compiled file loaded is the one compiled based on the error-containing source file.

Taytay:
Fixes #2113 

Previously, in #1764, the recording of the compilation digest was ensured to happen only after webpacker ran, but during that fix, it was also made that the digest was only recorded if Webpack succeeded. However, whether Webpack succeeds is not a measure of whether or not webpack emits output, meaning that the digest wouldn't be recorded in the case of a Typescript compilation error, even though new output file(s) were emitted. This divergence could cause either too many or too few compilations, as noted in #2113. This fix still moves the writing of the digest after Webpack runs, but does so _regardless_ of whether Webpack succeeds.

### DIFF
--- a/lib/webpacker/compiler.rb
+++ b/lib/webpacker/compiler.rb
@@ -19,7 +19,12 @@ class Webpacker::Compiler
   def compile
     if stale?
       run_webpack.tap do |success|
-        record_compilation_digest if success
+        # We used to only record the digest on success
+        # However, the output file is still written on error, (at least with ts-loader), meaning that the
+        # digest should still be updated. If it's not, you can end up in a situation where a recompile doesn't
+        # take place when it should.
+        # See https://github.com/rails/webpacker/issues/2113
+        record_compilation_digest
       end
     else
       true
--- a/test/compiler_test.rb
+++ b/test/compiler_test.rb
@@ -49,14 +49,13 @@ class CompilerTest < Minitest::Test
     end
   end
 
-  def test_staleness_on_compile_fail
+  def test_freshness_on_compile_fail
     status = OpenStruct.new(success?: false)
 
     assert Webpacker.compiler.stale?
     Open3.stub :capture3, [:sterr, :stdout, status] do
-
       Webpacker.compiler.compile
-      assert Webpacker.compiler.stale?
+      assert Webpacker.compiler.fresh?
     end
   end
 


jakeNiemiec:
@frodsan It has been awhile, but can you check to see if this PR works in tangent with the test you describe here https://github.com/rails/webpacker/pull/1764#issuecomment-431624443

If I don't hear back in a few days, I can make a rough approximation.

jakeNiemiec:
We also need to address the test for failing builds: https://github.com/rails/webpacker/blob/master/test/compiler_test.rb#L52

jakeNiemiec:
We also need to address the test for failing builds: https://github.com/rails/webpacker/blob/master/test/compiler_test.rb#L52

@Taytay If you want me to take care of this, you can allow me to edit with:
image

Taytay:
Doh! I totally missed that test. My apologies! I've got "allow edits" checked, so I would indeed appreciate it if you could add the fix!

justin808:
@gauravtiwari @jakeNiemiec Hi! I think this should be changed to check that at least the manifest.json was written. 

If there's some error on the webpack build, the cache digest is still saved. Then you go to update your webpack config and it's confusing that there is no recompile.

If somebody wants TS suggestions, that can be done by editor configuration or other things separate from the webpack build process.

I can't think of a valid case where the digest should get written if the output of webpack is non-zero.

Issue #1874, opened 2019-01-07
Add preload_link_tag helper


dmix:
Rails recently released support for preloading <link> content:

https://github.com/rails/rails/pull/31251

Documented here: https://api.rubyonrails.org/classes/ActionView/Helpers/AssetTagHelper.html#method-i-preload_link_tag

Which is recommended by Google and Mozilla to speed up pageloading performance:

https://developers.google.com/web/fundamentals/performance/resource-prioritization

https://developer.mozilla.org/en-US/docs/Web/HTML/Preloading_content

Any chance we could get similar support for "stylesheet_pack_tag" et al?

guillaumebriday:
I did it, I will create a PR in the afternoon 👍

jakeNiemiec:
I left feedback about this over here: https://github.com/rails/webpacker/pull/2124#pullrequestreview-248282772

Apart from that, I feel like there is some de-sync happening between all of us. Let me try and straighten it out:

<hr>

> Rails recently released support for preloading content: rails/rails#31251

This is useful for preloading fonts, videos, and web workers, but preloading all of your JS scripts is antithetical to progressive web enhancement and chunk lazy loading. **Plus, webpack gives you this as a feature**.

> is recommended by Google and Mozilla to speed up pageloading performance

In the google link:

>Note that <link rel="preload"> is a compulsory instruction to the browser; unlike the other resource hints we’ll be talking about, it’s something the browser must do, rather than merely an optional hint. **This makes it particularly important to test carefully, to insure that you’re not accidentally causing anything to fetch twice by using it, or fetching something that’s not needed.**

Mozilla does not currently support this at all:

![image](https://user-images.githubusercontent.com/16782063/59290265-cbb45e80-8c3d-11e9-879b-4a05af204b6e.png)

> Any chance we could get similar support for "stylesheet_pack_tag" et al?

This is my main source of confusion. **You don't want to preload the css, you want to preload the fonts and images inside the css so that fonts, css, and images can all be downloaded in parallel instead of waiting for the browser to parse the css**.

> My own versions of helpers for this...

The initial question was about stylesheet assets. While your approach is completely valid, I am not sure it makes for a good general recommendation to preload all chunks like this.

I would instead let webpack inject the `<link rel="preload">` _for you_ with `import(/* webpackPreload: true */ "./my/file.js")`. You can read more about this here (article linked from docs): https://medium.com/webpack/link-rel-prefetch-preload-in-webpack-51a52358f84c

[...]

PR #2124, opened 2019-06-11
Add a preload_pack_asset helper  

guillaumebriday:
Cf: https://github.com/rails/webpacker/issues/2125 and https://github.com/rails/webpacker/issues/2125#issuecomment-501022712

### DIFF
--- a/docs/webpack.md
+++ b/docs/webpack.md
@@ -277,7 +277,7 @@ environment.splitChunks()
 environment.splitChunks((config) => Object.assign({}, config, { optimization: { splitChunks: false }}))
 ```
 
-Then use, `javascript_packs_with_chunks_tag` helper to include all the transpiled
+Then use the `javascript_packs_with_chunks_tag` and `stylesheet_packs_with_chunks_tag` helpers to include all the transpiled
 packs with the chunks in your view, which creates html tags for all the chunks.
 
 ```erb
@@ -303,6 +303,20 @@ get duplicated chunks on the page.
 ```
 
 For the old configuration with the CommonsChunkPlugin see below. **Note** that this functionality is deprecated in Webpack V4.
+
+#### Preloading
+
+Before preload or prefetch your assets, please read [https://developer.mozilla.org/en-US/docs/Web/HTML/Preloading_content](https://developer.mozilla.org/en-US/docs/Web/HTML/Preloading_content).
+
+Webpack also provide it's own methods for preload or prefetch [https://medium.com/webpack/link-rel-prefetch-preload-in-webpack-51a52358f84c](https://medium.com/webpack/link-rel-prefetch-preload-in-webpack-51a52358f84c).
+
+You can preload your assets with the `preload_pack_asset` helper.
+
+```erb
+<%= preload_pack_asset 'fonts/fa-regular-400.woff2' %>
+```
+
+**Warning:** You don't want to preload the css, you want to preload the fonts and images inside the css so that fonts, css, and images can all be downloaded in parallel instead of waiting for the browser to parse the css.
 
 ### Add common chunks (deprecated in Webpack V4)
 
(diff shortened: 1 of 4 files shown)


guillaumebriday:
Thank you for the feedback !

I was not aware that rails provides the `preload_link_tag`, I changed my PR to use this instead but it's only available in rails 5.2.

In the other hand, I don't know we can preload only images and fonts with Webpacker, if you have any ideas ?

I read your comment in the related issue and I got your point. Maybe we could add a helper just for fonts and images, and let Webpacker handle only JS ?

jakeNiemiec:
> but it's only available in rails 5.2.

I'll let @gauravtiwari chime in on this. I am not sure about that, I don't want to break things for some users.

> Maybe we could add a helper just for fonts and images, and let Webpacker handle only JS ?

Agreed. I think the best course of action is to create a `preload_pack_asset` helper. It will need to resolve the file assets outside of the `entrypoints` object in `manifest.json`. Example from one of my projects:

```erb
  <%= preload_pack_asset 'fa-regular-400.woff2' %>
  <%# turns into something like %>
  <link href="/packs/fa-regular-400.d8O6LxGN.woff2" rel="preload" as="font" type="font/woff2" crossorigin>
```

What do you think?

jakeNiemiec:
@guillaumebriday We may be able to get away with something like this: https://github.com/rails/webpacker/issues/2125#issuecomment-501022712

guillaumebriday:
@jakeNiemiec I ended up doing a very basic wrapper for `preload_link_tag`. It's still only available on Rails >= 5.2 :/

I think we should insist in the document on what you said in this comment https://github.com/rails/webpacker/issues/1874#issuecomment-500936752

What do you think ?

guillaumebriday:
Hey,

I added a condition to check if Rails version is greater than 5.2.x.

The tests are now green, what do you think ?

Thank you !

cc @jakeNiemiec @gauravtiwari

jakeNiemiec:
I would need to defer to @gauravtiwari, I don't have too much experience with Rails versions inter-op. I would raise an error/warning when that if statement is false though.

gauravtiwari:
Thanks @guillaumebriday @jakeNiemiec 

How about we define a link tag if one isn't available? So pre-5.2 we define a link tag by hand and set the preload option otherwise use the one provided by 5.2+

Something like this:

```rb
if self.class.method_defined?(:preload_link_tag)
  preload_link_tag(current_webpacker_instance.manifest.lookup!(name), options)
else 
  # define_tag
  # https://github.com/rails/rails/blob/b9ca94caea2ca6a6cc09abaffaad67b447134079/actionview/lib/action_view/helpers/asset_tag_helper.rb#L268
end
```

guillaumebriday:
Hey @gauravtiwari @jakeNiemiec  !

I think this way is a good compromise until the end of support of rails 5.2.

I updated the PR, waiting for your feedbacks

Thank you

guillaumebriday:
Thank you @gauravtiwari for the idea !

`self.class.method_defined?(:preload_link_tag)` works very well !

However, it's not that easy if it's not defined and I'm facing multiple issues with < 5.2 versions.

For example:
`Template::Types` is not available in Webpacker.
So I tried with `Mime::Type.lookup_by_extension(extname).try(:to_s)` but it does not work with, at least, `woff2`, I don't know why.

The method `resolve_link_as` is a private method in Rails and I'm not sure it's a good idea to duplicate it in Webpacker.

At this point I think the best solution is to made it available only for Rails > 5.2 with a clear warning in the documentation.

Why do you think @gauravtiwari @jakeNiemiec ?

Thank you

gauravtiwari:
Thanks @guillaumebriday for your efforts, really appreciate it. What do you think of failing in the  else clause so a user knows that tag isn't supported. Something like,  "you need `rails > 5.2` to use this tag". 

Will also clarify this in changelog.

guillaumebriday:
@gauravtiwari Yes I think it's a pretty reasonable compromise !

I pushed it if you want to check again.

Thanks

gauravtiwari:
Thanks @guillaumebriday 🍰

guillaumebriday:
Thanks @gauravtiwari @jakeNiemiec for the feedbacks !

PR #2130, opened 2019-06-13
update yarn packages & update get_style_rule to handle css-loader@3.0.0

mmaloon:
After upgrading css-loader to version 3, we get a build error from the CssLoader that there are invalid options. Turns out the `localIdentName` option needs to be nested within the `modules` option.


jakeNiemiec:
Thanks for the submission, but this is a dupe of https://github.com/rails/webpacker/pull/2169 without tests. 

Regardless of how @smondal responds in #2169, feel free to submit a PR with a narrower focus on just updating packages without breaking changes.

PR #2136, opened 2019-06-19
change sourcemap quality at production

Sn005:
Current sourcemp state is `sourcemap` at production.
And TerserPlugin's sourceMap option is true.
But soucemap is not necessary at production.
More to say, an Unspecified number of people see development code by soucemap at  `sourcemap`.
So, If the Environment is production, soucemap is good to be hidden.


sharang-d:
@Sn005 Having sourcemaps in production is a decision made by DHH.
Read here: https://github.com/rails/webpacker/issues/769#issuecomment-458216151 (and the whole thread).

jakeNiemiec:
As @sharang-d said. Sorry about that.

PR #2139, opened 2019-06-20
configure engine asset delivery path from within engine

lazylester:
b/c it's tidier to keep engine-related configuration within the engine itself!

### DIFF
--- a/docs/engines.md
+++ b/docs/engines.md
@@ -166,7 +166,7 @@ or if you prefer to keep your engine-related configuration within the engine its
 module MyEngine
   class Engine < ::Rails:Engine
     config.app_middleware.use(
-      "Rack::Statuc",
+      "Rack::Static",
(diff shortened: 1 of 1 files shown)


docs/engines.md
```diff
@@ -159,5 +159,18 @@ config.middleware.use(
...
 )
 ```
+or if you prefer to keep your engine-related configuration within the engine itself
+
+```ruby
+# my-engine-root/lib/my-engine/engine.rb
+module MyEngine
+  class Engine < ::Rails:Engine
+    config.app_middleware.use(
+      "Rack::Statuc",
```
gauravtiwari: Typo here ^^ should be `Rack::Static` no?

lazylester: indeed... I've fixed it, thanks for catching it

Issue #400, opened 2017-05-19
Angular CLI compatibility


berga:
It would be great to use the CLI to generate angular components inside a rails app as happen with rails generators.

Since now I've been able to (starting from a fresh Rails 5.1.1 project):

- [X] integrate a generated cli project pointing src root to app/javascript[/packs] 
- [X] adjust the webpack config to use external html and css files generated from CLI
- [X] compile frontend in development
- [X] generate components using CLI 
- [ ] test production env
- [ ] add AOT to webpack
- [ ] wrap ng CLI to a rails generator (optional but cool) 

I'm wondering if it would be possible to enhance _webpacker:install_ 
to generate an Angular CLI compatible structure ready to work, 
an hello_word example a little bit more complex (with separated html and css) or take it out to a another gem, something like angular_rails_cli wrapping/using webpacker?

gauravtiwari:
@berga Sounds like a good idea to start a companion webpacker gem with some installers - `webpacker-angular` that provided these advanced features plus automatic mounting using helpers - `<%= angular_component('Home') %>` inside views.

returnlytom:
@berga Started implementing this on my own, then found your PR. Is there anything you would like help with?

berga:
hi @returnlytom, I'me a bit late on it, I could push a bare gem structure, can you help me building it?

returnlytom:
@berga Go ahead and push what you got, we can work on it collaboratively

gauravtiwari:
Closing this - Feel free to make a PR for README once this is ready 👍

jakehockey10:
Did this ever happen?

TylerRick:
Has anyone added this feature yet?

TylerRick:
To be able to use `ng generate` to generate new components, etc., and `ng test` to run tests for Angular components...

This would be huge for those using Angular.

berga:
> @berga Go ahead and push what you got, we can work on it collaboratively

I'm sorry @returnlytom, I've abandoned too early the "monolytical" approach 
in favour of 2 levels separated app: rails api + angular front,
btw I could try to resume some dusty code

PR #2141, opened 2019-06-21
Add docs for Angular AOT compilation

### LINKED ISSUE #1737: Rails 5.2, Webpacker 3.5.5 + Angular CSP issue when deploying to prod (Heroku)
Hi,
firstly thank you for all your hard work, you're all awesome!

I created a mostly "fresh" app using Rails 5.2, Webpacker 3.5.5 and Angular. 
I've follow the angular instructions in the README:

> The installer will add the TypeScript and Angular core libraries using Yarn alongside a few changes to the configuration files. An example component written in TypeScript will also be added to your project in app/javascript so that you can experiment with Angular right away.
> 
> By default, Angular uses a JIT compiler for development environment. This compiler is not compatible with restrictive CSP (Content Security Policy) environments like Rails 5.2+. You can use Angular AOT compiler in development with the @ngtools/webpack plugin.
> 
> Alternatively if you're using Rails 5.2+ you can enable unsafe-eval rule for your development environment. This can be done in the config/initializers/content_security_policy.rb with the following code:

```rb
  if Rails.env.development?
    policy.script_src :self, :https, :unsafe_eval
  else
    policy.script_src :self, :https
  end
```

and added some code that runs well in development, but when I deploy it to production, namely Heroku, I get the following error in the console:

`Uncaught EvalError: Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "script-src 'self' https:`

and the angular component does not load.

I commented out all of the CSP stuff in the initializer and deployed again and things started working again. 

Then I replaced this part:

```rb
  if Rails.env.development?
    policy.script_src :self, :https, :unsafe_eval
  else
    policy.script_src :self, :https
  end
```

with 

```rb
  policy.script_src :self, :https, :unsafe_eval
```

basically removing the check for the development environment and that worked as well.
I don't know a lot about CSP but I'm assuming that adding :unsafe_eval` would defeat the purpose of having any policy, so probably not the best solution for the problem.

Any recommendations on how I can get around this issue without the need to completely circumvent CSP?

TylerRick:
Resolves #1737

The current docs only talk about enabling AOT for development:

> By default, Angular uses a JIT compiler for **development environment**. … You can use Angular AOT compiler **in development** with the @ngtools/webpack plugin.
> Alternatively … you can enable unsafe-eval rule for your **development environment**. 

While it's fine to enable AOT in development, _too_, security is a lot more important in **production** so it seems a lot more important to have AOT enabled in *production*, where you *should* have a restrictive CSP that doesn't simply allow `unsafe-eval`.

The way it is currently worded seems to imply that _no additional configuration changes from the default are needed for it to work in production_.

While that may be true for [Vue](https://vuejs.org/v2/guide/installation.html#CSP-environments):

> On the other hand, the runtime-only build is fully CSP-compliant. When using the runtime-only build with Webpack + vue-loader or Browserify + vueify, your templates will be precompiled into render functions which work perfectly in CSP environments.

, that is simply not true for Angular with the way webpacker configures things by default! The default config breaks CSP in production. (If I am wrong, please prove me wrong.)

https://github.com/TylerRick/webpacker_angular_csp/tree/aot_broken_by_default is proof of that, and #1737 contains reports of 2 people who ran into this problem in production builds.

I spent most of a day figuring out how to enable AOT. Hopefully these doc improvements can help save others the pain I went through.

Aside from being safer, using AOT also results in 30-60% less code being served to users due to the JIT Compiler module not needing to be included in the bundle (see https://slides.com/wassimchegham/demystifying-ahead-of-time-compilation-in-angular-2-aot-jit#/32), which results in faster load times. Wins all around!

I wonder if we should even just enable AOT compilation by default when you `webpacker:install:angular` so that it just works out of the box? (#2142)

### DIFF
--- a/README.md
+++ b/README.md
@@ -358,25 +358,79 @@ a few changes to the configuration files. An example component written in
 TypeScript will also be added to your project in `app/javascript` so that
 you can experiment with Angular right away.
 
-By default, Angular uses a JIT compiler for development environment. This
-compiler is not compatible with restrictive CSP (Content Security
-Policy) environments like Rails 5.2+. You can use Angular AOT compiler
-in development with the [@ngtools/webpack](https://www.npmjs.com/package/@ngtools/webpack#usage) plugin.
+#### AOT compiler and Content Security Policy
 
-Alternatively if you're using Rails 5.2+ you can enable `unsafe-eval` rule for your
-development environment. This can be done in the `config/initializers/content_security_policy.rb`
-with the following code:
+By default, Angular uses a Just-In-Time (JIT) compiler which uses `eval` and therefore results in
+errors if you are using a Content Security Policy that doesn't include `unsafe-eval` as an allowed
+`script_src`.
+
+To prevent this, you can use the Angular Ahead-Of-Time (AOT) compiler, provided by the
+[@ngtools/webpack](https://www.npmjs.com/package/@ngtools/webpack#usage)
+plugin. To enable AOT:
+
+1. Add/update the following dependencies in `package.json`:
+    ```javascript
+      "dependencies": {
+        …
+        "@angular/compiler-cli": "^8.0.2",
+        "@ngtools/webpack": "^8.0.3",
+        "typescript": "~3.4.5",
+        …
+      },
+    ```
+
+2. Add/update the following config files:
+    ```javascript
+    // config/webpack/loaders/angular.js
+    const { AngularCompilerPlugin } = require('@ngtools/webpack')
+
+    module.exports = {
+      test: /(?:\.ngfactory\.js|\.ngstyle\.js|\.ts)$/,
+      use: [
+        {
+          loader: '@ngtools/webpack'
+        }
+      ]
+    }
+
+    // config/webpack/environment.js
+    const { environment }           = require('@rails/webpacker')
+    const { AngularCompilerPlugin } = require('@ngtools/webpack')
+    const angular                   = require('./loaders/angular')
+
+    environment.loaders.prepend('angular', angular)
+    environment.plugins.prepend('angular',
+      new AngularCompilerPlugin({
+        tsConfigPath: './tsconfig.json',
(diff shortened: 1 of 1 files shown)


ericrav:
Is it possible to use JIT for development with this configuration?

TylerRick:
> Is it possible to use JIT for development with this configuration?

Not with this configuration, no. I'm sure it's possible (because I believe Angular CLI does it that way), but it seems too difficult based on what I know currently:

1. Instead of changing `config/webpack/environment.js`, I guess you'd have to change `config/webpack/development.js` to add the regular `ts-loader` and `config/webpack/production.js` to `environment.loaders.prepend('angular' …`

2. In every file where you have bootstrap code, I guess you'd have to somehow make this conditional so it only did this for production:

```
    import { AppModuleNgFactory } from './app/app.module.ngfactory';
    platformBrowser().bootstrapModuleFactory(AppModuleNgFactory);
```

(Maybe there's a technique we could use like this that I saw in an `angular.json` file that could replace your normal .ts files with `.prod.ts` ones (but what a pain to maintain both)?:
 ```
              "fileReplacements": [
                {
                  "replace": "src/environments/environment.ts",
                  "with": "src/environments/environment.prod.ts"
                }
              ],
```
)

Better yet, figure out how Angular CLI is able to keep the regular `platformBrowserDynamic` stuff and work for both JIT and AOT. Even better yet, make it so you can just use Angular CLI directly with webpacker (#400) so wouldn't have to duplicate anything that it already provides.

Someone with more Angular/webpacker experience than me would have to be the one to take that on...

eric-simonton-sama:
This would make the docs easier to follow for sure. Thank you.

Question: Angular's CLI still uses JIT for its jasmine tests. Our project has many jasmine tests that use angular components. Do you know if this change would affect our ability to run them?

PR #2149, opened 2019-06-25
Remove unnecessary `require "rubygems"`

tatsuyafw:
In Ruby 1.9 and later, `require "rubygems"` happens automatically.
Therefore, this PR removes those `require` s.


gauravtiwari:
thanks @tatsuyafw

Issue #2140, opened 2019-06-20
publicPath should not include RAILS_RELATIVE_URL_ROOT


sspread:
It's my understanding that all paths defined in rails server are relative to the `RAILS_RELATIVE_URL_ROOT`, and webpack publicPath should behave the same way. For example:

`RAILS_RELATIVE_URL_ROOT = /app`
`public_output_path = packs` (webpacker.yml)

Working result would be
`javascript_pack_tag` -> `<script src= '/app/packs/...'>`
`manifest.json` -> `/packs/...`
`publicPath` -> `/packs`

However, under the current implementation,
`javascript_pack_tag` -> `<script src= '/app/packs/...'>`
`manifest.json` -> `/packs/...`
`publicPath` -> `/app/packs`
which will not be found because rails assumes routes relative to `RAILS_RELATIVE_URL_ROOT`.

See: https://github.com/rails/webpacker/issues/1357 and comments. Essentially I'm proposing that the merged code that addresses that issue should be reverted. Maybe something changed since then or maybe I'm missing something.

jakeNiemiec:
What is the difference between [`public_output_path` and `RAILS_RELATIVE_URL_ROOT`](https://github.com/rails/webpacker/pull/1428/files#diff-ebfa15a9cc1df2aca52913e46e89cc48R27)? Do we need both?

sspread:
I believe `public_output_path` is _supposed_ to be synonymous with [webpack's](https://webpack.js.org/guides/public-path/) `publicPath`.  Under the current Webpacker config, `publicPath` now becomes `RAILS_RELATIVE_URL_ROOT + public_output_path`. 

`RAILS_RELATIVE_URL_ROOT` is used to [configure rails](https://guides.rubyonrails.org/configuring.html#deploy-to-a-subdirectory-relative-url-root) to mount on a sub-directory. If configured, rails already assumes the relative url root as the base, and it need not be a concern for webpack/webpacker.

rdavid369:
> If configured, rails already assumes the relative url root as the base, and it need not be a concern for webpack/webpacker.

Not in all cases.  If I require a css file inside my module and said css file includes an image from it's package, that image doesn't properly load. It doesn't know anything about the relative url root.  It becomes a broken asset.  It works fine if the app isn't deployed in a sub directory.  But for apps in sub directories, which is what we use relative url root for in the first place, it breaks.

How is everyone else telling webpack that the app is deployed in a sub directory, so assets properly load?

[...]

PR #2151, opened 2019-06-26
Remove RAILS_RELATIVE_URL_ROOT from publicPath

sspread:
Fixes https://github.com/rails/webpacker/issues/2140

### DIFF
--- a/package/config.js
+++ b/package/config.js
@@ -28,14 +28,7 @@ config.outputPath = resolve(config.public_root_path, config.public_output_path)
 // (code-splitting chunks and static assets) load from the CDN instead of a relative path.
 const getPublicPath = () => {
   const rootUrl = process.env.WEBPACKER_ASSET_HOST || '/'
-  let packPath = `${config.public_output_path}/`
-  // Add relative root prefix to pack path.
-  if (process.env.RAILS_RELATIVE_URL_ROOT) {
-    let relativeRoot = process.env.RAILS_RELATIVE_URL_ROOT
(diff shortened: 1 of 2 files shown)


package/config.js
```diff
@@ -28,14 +28,7 @@ config.outputPath = resolve(config.public_root_path, config.public_output_path)
...
-  let packPath = `${config.public_output_path}/`
-  // Add relative root prefix to pack path.
-  if (process.env.RAILS_RELATIVE_URL_ROOT) {
-    let relativeRoot = process.env.RAILS_RELATIVE_URL_ROOT
-    relativeRoot = relativeRoot.startsWith('/') ? relativeRoot.substr(1) : relativeRoot
-    packPath = `${ensureTrailingSlash(relativeRoot)}${packPath}`
-  }
-
+  const packPath = `${config.public_output_path}/`
   return ensureTrailingSlash(rootUrl) + packPath
```
jakeNiemiec: Nitpick: you might move `ensureTrailingSlash` like:
```js
const rootUrl = ensureTrailingSlash(process.env.WEBPACKER_ASSET_HOST || '/')

return `${rootUrl}${config.public_output_path}/`
```
I could even see this, but it might not read as well:
```js
return `${ensureTrailingSlash(process.env.WEBPACKER_ASSET_HOST || '/')}${config.public_output_path}/`
```

sspread: Your first suggestion is more readable. I wasn't trying to touch more than was necessary here but I can add that change.

Issue #2162, opened 2019-07-03
Webpacker + Vue compiled failed


Jaroost:
I tried to create a new Rails app using Webpacker and Vue. 

Info:
- Rails 5.2.3
- Ruby 2.5.5p157 x64
- OS: Windows Server 2016

Here what I have done:

`rails new myapp --webpack=vue -d mysql`

While installing package with yarn I have peer dependency warnings:
```
warning " > vue-loader@15.7.0" has unmet peer dependency "css-loader@*".
warning " > vue-loader@15.7.0" has unmet peer dependency "webpack@^4.1.0 || ^5.0.0-0".
```

After that my `package.json` file looks like this:
```
{
  "name": "myapp",
  "private": true,
  "dependencies": {
    "@rails/webpacker": "^4.0.7",
    "vue": "^2.6.10",
    "vue-loader": "^15.7.0",
    "vue-template-compiler": "^2.6.10"
  }
}
```

When I use command `rails webpacker:compile` it works perfectly and compile :
```
Compiling…
Compiled all packs in D:/myapp/public/packs
```
So far so good but when I try to add peer dependency in package.json things go wrong:
`yarn add webpack@^4.35.2`
`yarn add css-loader`

Here, peer dependencies warnings are gone.

Now package.json looks like this:

```
{
  "name": "myapp",
  "private": true,
  "dependencies": {
    "@rails/webpacker": "^4.0.7",
    "css-loader": "^3.0.0",
    "vue": "^2.6.10",
    "vue-loader": "^15.7.0",
    "vue-template-compiler": "^2.6.10",
    "webpack": "4.35.2"
  }
}
```

And when I try to run the `rails webpacker:compile`  it leave me with this (blank) error :

```
Compiling…
Compilation failed:


```

if I use the command `ruby .\bin\webpack-dev-server` I got the following error (It's normal I don't have webpack-dev-server in dev dependency of package.json. Is it normal that this package is not added by default?)

```
yarn run v1.16.0
error Command "webpack-dev-server" not found.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

So I add webpack-dev-server to package.json:
`yarn add webpack-dev-server --dev`

So now I can run the `ruby .\bin\webpack-dev-server` and I got this following error:

```
[...]

PR #2169, opened 2019-07-07
Upgrade Node Js  package 

smondal:
Upgrade Node Js packages with latest version


jakeNiemiec:
For example, `css-loader` is one of the largest problems on that list since the acceptable config structure changed.

smondal:
Thanks @jakeNiemiec for your reply. I will be working on dependencies upgrade. I have updated the webpacker version. But is their any way I can find all of the error which we need to fix after dependencies upgrade?  I found few tase cases is written under utils. Is this sufficient to find all the error after dependencies upgrade?

jakeNiemiec:
Here is one example for the `css-loader`:  https://github.com/rails/webpacker/issues/2162#issuecomment-508827256. You can see that `getLocalIdent` was moved to `module.getLocalIdent` as a breaking change.

In this case, you would need to modify the structure that webpacker expects according to the breaking changes from `css-loader`: https://github.com/webpack-contrib/css-loader/blob/master/CHANGELOG.md#breaking-changes

Here is the webpacker file that you would need to change:

https://github.com/rails/webpacker/blob/41d79c96187154b5485289e8c3c42428dd819bfc/package/utils/get_style_rule.js#L18-L27

I don't think the tests currently cover this.

smondal:
Thanks @jakeNiemiec, Still now we found, its issue with css-loader. After fixing  this issue "getLocalIdent was moved to module. getLocalIdent as a breaking change", compile was working fine for Vue JS.
But I will review other package's CHANGELOG. If there are any relevant changes I will update this.

smondal:
Hi @jakeNiemiec  I see there is "babel-preset" is internally using '@babel/plugin-transform-destructuring', Still we are include this plugin inside "babel.config.js". do we really need this to include.

https://github.com/rails/webpacker/blob/master/lib/install/config/babel.config.js#L43

https://github.com/babel/babel/blob/eb54adb75589ef82041ad23cdfcb3168a52317a7/packages/babel-preset-env/package.json#L29

Can you please help me on this?

jakeNiemiec:
You can read about how the plugins work here: https://babeljs.io/docs/en/options#plugins

That file dictates which plugins (_already listed in `package.json`_) are applied to our webpack build during the babel step. You can read more about that here: https://babeljs.io/docs/en/configuration

smondal:
@jakeNiemiec My point was that we should removed '@babel/plugin-transform-destructuring' from "babel.config.js". 
https://github.com/rails/webpacker/blob/master/lib/install/config/babel.config.js#L43

PR #2177, opened 2019-07-12
Only include configured extensions when generating the watched files digest

calmyournerves:
Hey there 👋 

We're using a "component" based approach where we have our template files inside `app/javascript/` next to the corresponding JS and CSS files. We noticed the following issue when running our test suite locally: Whenever we changed a template file, it would trigger a Webpack compile in the next test run. 

It seems this is caused by the logic around watched files and the digest. We hoped to be able to configure this for the test env via the `extensions:` list in `webpacker.yml`. But in `Compiler#watched_files_digest`, all files are considered for the digest, regardless of the extension.

This PR changes that behaviour and only includes the configured extensions for the watched files digest.

I'm not sure if there could be any side effects with this or if this is desirable at all for you. Let me know what you think 🙂.

### DIFF
--- a/lib/webpacker/compiler.rb
+++ b/lib/webpacker/compiler.rb
@@ -86,7 +86,7 @@ class Webpacker::Compiler
     def default_watched_paths
       [
         *config.resolved_paths_globbed,
-        "#{config.source_path.relative_path_from(config.root_path)}/**/*",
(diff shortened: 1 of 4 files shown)


gauravtiwari:
thanks @calmyournerves

PR #2179, opened 2019-07-14
Update docs to reflect Docker best practices for apt-get

juliantrueflynn:
Improve docs for to represent Docker best practices.

Reference to Docker documentation for best practices:
https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#apt-get

> Always combine RUN apt-get update with apt-get install in the same RUN statement. (Snippet.) Using apt-get update alone in a RUN statement causes caching issues and subsequent apt-get install instructions fail.

### DIFF
--- a/docs/docker.md
+++ b/docs/docker.md
@@ -23,10 +23,23 @@ add nodejs and yarn as dependencies in Dockerfile,
 ```dockerfile
 FROM ruby:2.4.1
 
-RUN apt-get update -qq
-RUN apt-get install -y build-essential nodejs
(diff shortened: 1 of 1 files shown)


juliantrueflynn:
I had to re-update this PR. It turns out when I tried the example given in docs that was updated by @ctrlaltdylan, it didn't work when using `assets:precompile` in the Dockerfile. Using `assets:precompile` in the Dockerfile is common practice and what we do at my workplace.

I went ahead and added that to the documentation as well and left @ctrlaltdylan's example (but still refactored it to use Docker best practices).

PR #2183, opened 2019-07-15
Change installers to be style agnostic

### LINKED ISSUE #2170: Adding semicolons to environment.js causes installers to silently fail
Hello,

I ran into an issue where some of the lines from the ERB and Vue installers were not added to my `environment.js`, though this could apply to any of the installers.

I figured out after looking through the source that it was because I added semicolons to my `environment.js`

https://github.com/rails/webpacker/blob/5d5b3b867d35b376f3fa2af7c9cee3bfdf6f0f7c/lib/install/vue.rb#L6-L9

I can see a couple solutions for how to resolve this:
- Change the behavior so that it doesn't look exclusively for a line that doesn't end in a semi-colon (ideal, and I believe I could create a PR for this)
- Emit a warning in the console when it can't find the string it is looking for, and add documentation detailing why this error was thrown (I'm not actually sure if this is possible, I can't find any functionality like this in thor)

SkipKayhil:
Fixes #2170

### DIFF
--- a/lib/install/coffee.rb
+++ b/lib/install/coffee.rb
@@ -6,7 +6,7 @@ copy_file "#{__dir__}/loaders/coffee.js", Rails.root.join("config/webpack/loader
 say "Adding coffee loader to config/webpack/environment.js"
 insert_into_file Rails.root.join("config/webpack/environment.js").to_s,
   "const coffee =  require('./loaders/coffee')\n",
-  after: "require('@rails/webpacker')\n"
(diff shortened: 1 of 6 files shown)


SkipKayhil:
I learned some more about regex, and updated the PR to match quotes types instead of potentially matching `'@rails/webpacker"`

gauravtiwari:
thanks @SkipKayhil

PR #2188, opened 2019-07-17
Add test for extract_css?

smondal:


gauravtiwari:
Thanks @smondal Could you please use this helper here instead of reinitialising? https://github.com/rails/webpacker/blob/35c638cc4761e15abf3f54cd5dc1ae022e387aaf/test/test_helper.rb#L24

smondal:
Thanks @gauravtiwari  for your comment. I have applied this helper with_rails_env(env) and modify the test cases. This is working properly

gauravtiwari:
thanks @smondal

PR #2196, opened 2019-07-26
Fix proxy for chunked streaming templates

### LINKED ISSUE #2195: Webpacker appends and prepends junk to pages when streaming
There is a bug caused by an interaction between Webpacker and ActionController Streaming.

If you include the Webpacker gem and turn streaming on you get weird extra output prepended and appended to your content (and the page doesn't stream). What's prepended appears to be the content length in base64, then appended is a zero. If you change the content the prepended content length changes but the zero does not. This happens with templates too but here I've provided the most minimal reproduction:

```ruby
class StreamingsController < ActionController::Base
  def show
    render plain: 'Hello, World!', stream: true, layout: false
  end
end
```

Weird output:

```
d
Hello, World!
0
```

With streaming off there's no issue:

```ruby
class StreamingsController < ActionController::Base
  def show
    render plain: 'Hello, World!', layout: false
  end
end
```

Normal output:

```
Hello, World!
```

## Reproduction

https://github.com/sfcgeorge/webpacker-streaming-bug

This happens with the latest gem and master. I encountered this in a real app we're trying to add streaming to.

Only happens when `dev_server` is enabled and thus the [middleware is inserted](https://github.com/rails/webpacker/blob/cc98e59652f22c723fac0308199cff5a42ad49ce/lib/webpacker/railtie.rb#L54).

Disabling `dev_server` in `webpack.yml` or removing the webpacker gem works around this.

sfcgeorge:
Fixes #2195 

Webpacker is currently incompatible with Rails template streaming (Transfer-Encoding: chunked) in development. Firstly it prevents streaming; the whole layout blocks. Secondly it inserts junk into the page that breaks rendering in browsers.

This patch fixes it. I'm unclear on why we even need to proxy everything, or why the proxy is messing with headers. Nonetheless, [this line](https://github.com/rails/webpacker/blob/cc98e59652f22c723fac0308199cff5a42ad49ce/lib/webpacker/dev_server_proxy.rb#L13) deletes the header that is needed for streaming to work, which this patch amends. It also fixes the junk in the page, I assume because deleting that header when the server is streaming leads to an invalid format.

### DIFF
--- a/lib/webpacker/dev_server_proxy.rb
+++ b/lib/webpacker/dev_server_proxy.rb
@@ -8,13 +8,6 @@ class Webpacker::DevServerProxy < Rack::Proxy
     super
   end
 
-  def rewrite_response(response)
-    _status, headers, _body = response
-    headers.delete "transfer-encoding" unless headers["transfer-encoding"] == "chunked"
-    headers.delete "content-length" if dev_server.running? && dev_server.https?
-    response
-  end
-
   def perform_request(env)
     if env["PATH_INFO"].start_with?("/#{public_output_uri_path}") && dev_server.running?
       env["HTTP_HOST"] = env["HTTP_X_FORWARDED_HOST"] = env["HTTP_X_FORWARDED_SERVER"] = dev_server.host_with_port
--- a/lib/webpacker/dev_server_proxy.rb
+++ b/lib/webpacker/dev_server_proxy.rb
@@ -10,7 +10,7 @@ class Webpacker::DevServerProxy < Rack::Proxy
 
   def rewrite_response(response)
     _status, headers, _body = response
-    headers.delete "transfer-encoding"
+    headers.delete "transfer-encoding" unless headers["transfer-encoding"] == "chunked"
     headers.delete "content-length" if dev_server.running? && dev_server.https?
     response
   end


lib/webpacker/dev_server_proxy.rb
```diff
@@ -10,7 +10,7 @@ def initialize(app = nil, opts = {})
 
   def rewrite_response(response)
     _status, headers, _body = response
-    headers.delete "transfer-encoding"
+    headers.delete "transfer-encoding" unless headers["transfer-encoding"] == "chunked"
```
jakeNiemiec: This does not account for: `Transfer-Encoding: gzip, chunked`

I would just remove that line, it is from: https://github.com/rails/webpacker/pull/637

sfcgeorge: Thanks for the review. Ooh good point, so 3 options:

1. Delete the line (I'm hesitant as it must've been added for a reason so I've asked https://github.com/rails/webpacker/pull/637#discussion_r308964409).
2. Add the `gzip, chunked` case as well.
3. Leave it; [chunking is done after gzipping](https://en.wikipedia.org/wiki/Chunked_transfer_encoding#Use_with_compression) which implies this case won't happen with Rails because Rails streaming happens first so you can't gzip afterwards. I can confirm the the Rack::Deflate middleware breaks streaming. That said, someone could potentially have _another_ middleware that does chunking later for whatever reason.

I'm also in favour of 1 for simplicity if the line turns out to not be needed so I'll see if I get a response then go from there :)

jakeNiemiec:
>I'm unclear on why we even need to proxy everything, or why the proxy is messing with headers. 

Webpack dev server needs it. It's an express server that hosts assets and creates a websocket that your page connects to for hot reloading.

lib/webpacker/dev_server_proxy.rb
jakeNiemiec: I say delete because:

1. 2 years ago, webpacker was a different tool with different goals
2. This prevents any file size compression
3. `transfer-encoding` does not seem to even be present to delete. I tested the major webpacker flows and this seems to be nonessential.

sfcgeorge: Thanks for the testing! Awesome I'll go with 3 then. 

In my testing the header wasn't deleted but the page was broken. I assume something else was adding it back in again, possibly [this middleware](https://www.rubydoc.info/github/rack/rack/Rack/Chunked) that adds `chunked` if there's no `content-length` set.

sfcgeorge:
I see from the PR you referenced that the proxy is so assets are served from the same host to prevent CORS issues, makes sense :)

lib/webpacker/dev_server_proxy.rb
sfcgeorge: What do you think about the line below deleting the `content-length` header? One or the other is required by HTTP 1.1, so I thought maybe this middleware was relying on the Rack::Chunked middleware to add back `Transfer-Encoding: chunked`. But running `bin/rails middleware` I don't see Rack::Chunked in the stack, yet I always seem to have that `Transfer-Encoding: chunked` header as you said.

jakeNiemiec: > I assume something else was adding it back in again

Weirdly enough, chromes devtool adds it to the request for me. `curl http://localhost:3000/streaming -I` shows it missing.

> What do you think about the line below deleting the content-length header? 

I honestly have no idea; I am not that much of an expert on Rack/middleware stuff.

gauravtiwari: Thanks, both and apologies for the delay. Let's get rid of that line if it works now. Earlier, there was some weird error from dev server (like `invalid chunk`) or something. Basically, the packs arrived broken in browser. 

I can't exactly remember but if it works now let's remove it.

sfcgeorge: Thank you @gauravtiwari 😄 

Okay I've removed `rewrite_response` now. I was nervous about simply removing but: 

* Without knowing what the purpose was and without a test I agree it shouldn't blindly stay. 
* This fixes the chunked encoding bug.
* This middleware isn't inserted in production (usually) so previously there would've been a difference in headers between development and production envs that could cause other hard to debug issues.
* The old behaviour modified the headers of _all_ requests, not just asset requests which is overkill. 

That said, if after merging someone does have an issue due to an unusual WDS config then we can add a test case and put the lines back; but wrapped in a check so only assets have their headers modified:

```ruby
if env["PATH_INFO"].start_with?("/#{public_output_uri_path}") && dev_server.running?
  headers.delete "transfer-encoding"
  headers.delete "content-length" if dev_server.running? && dev_server.https?
end
```

gauravtiwari:
Thanks @sfcgeorge @jakeNiemiec 

Let give this a try and will cut a new release if no issues arises.

PR #2207, opened 2019-08-02
Possible syntax error fix.

angarc:
I got a syntax error, but what fixed it was changing "Rack::Static" to just Rack::Static. I don't think it's supposed to be a string. This worked for me.


gauravtiwari:
thanks @angarc

Issue #2059, opened 2019-04-17
stylesheet_pack_tag not working in webpacker 4??


AmirolAhmad:
based on the usage mention [here](https://github.com/rails/webpacker#usage), _you can link them by using stylesheet_pack_tag_ which mean you add `<%= stylesheet_pack_tag 'application' %>` under `views>layout>application.html.erb`

but then when I view-source the web, there is no css loaded inside the `<head>`.

then to fixed this issue, i need to import `import '../src/application.css'` into `packs/application.js` to make the css file to work and load.

my question, is this how it should work? if so, what is the purpose to have `<%= stylesheet_pack_tag 'application' %>`? if you commented out this line, it is still work.

I have create the [sample app](https://github.com/AmirolAhmad/myspookytech_backend) for better picture about this

or maybe i miss some point somewhere else?

hishammalik:
having the same issue.

bbugh:
This sounds to me like misunderstanding how webpack works (which is reasonable, because it's incredibly confusing).

Webpack is used for building many things, and the way we tell webpack what to load are `import` or `require` statements. This includes css files, images, and everything else. When you do `import '../src/application.css'`, you're telling webpack *include application.css in the build*. This does not mean it's going to be compiled into your javascript, only that webpack now knows that you want to compile this file. How that file compilation is handled is depending on how your loaders (`css-loader`, `sass-loader`, `file-loader`, etc.. ) are configured.

When you do `<%= stylesheet_pack_tag 'application' %>`, that's a *run-time* inclusion from Rails, that has nothing to do with whether webpack has built the stylesheets or not. All that is doing is saying "if you have a pack named `application-*.css`, include it here. If webpack didn't build a separate pack of css, then this statement will have nothing to load because no stylesheets were compiled.

You didn't mention what framework you're using, and how this works is dependent on that. As far as I remember, `stylesheet_pack_tag` is only used for Vue because `vue-loader` automatically collects all of the stylesheets into one file. I am not sure if any other frameworks require this (I don't think React does, for example).

It is perfectly normal and expected to `import` or `require` your stylesheets like you're doing in your entry point file.

hishammalik:
I saw that its covered in the document as

> If you have hmr turned to true, then the stylesheet_pack_tag generates no output, as you will want to configure your styles to be inlined in your JavaScript for hot reloading. During production and testing, the stylesheet_pack_tag will create the appropriate HTML tags.

[...]

PR #2214, opened 2019-08-06
Clarify steps needed to load styles

vfonic:
I was having some troubles loading scss styles in my app and I figured I'd try to contribute to the docs a bit to make it easier for the next person to understand how it works.

I'm not sure what I wrote is 100% correct, though. Please check and let me know if I misunderstood something.

I've used part of the explanation from @bbugh's comment in this issue:
https://github.com/rails/webpacker/issues/2059#issuecomment-486491924

Thanks!

### DIFF
--- a/docs/css.md
+++ b/docs/css.md
@@ -3,6 +3,18 @@
 
 Webpacker supports importing CSS, Sass and SCSS files directly into your JavaScript files.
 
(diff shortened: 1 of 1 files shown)


gauravtiwari:
Thanks :)

PR #2224, opened 2019-08-09
Print STDOUT on failed compilation (fixed #2190)

### LINKED ISSUE #2190: Webpacker::Compiler swallows stdout in the event of failed compilation
In the event of a failed compilation `Webpacker::Compiler#run_webpack` will swallow the output of stdout and only show the output of stderr.

If webpack is incorrectly outputting on StdOut, as seems to be the case with a missing node module, then the user just see's a blank error with no further information

```
Compilation failed:
```

Happy to submit a patch to resolve this - not sure on the best approach? Options I can think of  are

1. Extend line L71 to something like

```ruby
logger.error "Compilation failed:\n#{stdout}\n#{stderr}"
```

2. Log the stdout to a different level in the logger

```ruby
logger.error "Compilation failed:\n#{stderr}"
logger.debug "#{stdout}"
```

Thanks

botanicus:
This fixes #2190.

The issue is that some errors that webpack prints out go to `STDOUT` rather than `STDERR`, but `STDOUT` gets swallowed.

One such instance was that declaring `const` twice. Before this fix no error would be printed, now it is printed out correctly.

_@jebw was the first one proposing such solution, I don't want to take the credit, but at the same time this needs to get fixed, so I'm submitting this PR._

### DIFF
--- a/lib/webpacker/compiler.rb
+++ b/lib/webpacker/compiler.rb
@@ -75,6 +75,7 @@ class Webpacker::Compiler
         logger.error "#{stderr}" unless stderr.empty?
       else
         logger.error "Compilation failed:\n#{stderr}"
+        logger.error "#{stdout}" unless stdout.empty?
       end
 
       if config.webpack_compile_output?
--- a/lib/webpacker/compiler.rb
+++ b/lib/webpacker/compiler.rb
@@ -74,8 +74,8 @@ class Webpacker::Compiler
         logger.info "Compiled all packs in #{config.public_output_path}"
         logger.error "#{stderr}" unless stderr.empty?
       else
-        logger.error "Compilation failed:\n#{stderr}"
-        logger.error "#{stdout}" unless stdout.empty?
+        non_empty_streams = [stdout, stderr].delete_if(&:empty?)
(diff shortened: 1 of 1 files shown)


jakeNiemiec:
I couldn't comment fast enough 😅

jakeNiemiec:
Actually, do you think 77 and 78 would be better flipped?

botanicus:
@jakeNiemiec I was thinking about that as well. Just committed it, let me test it.

botanicus:
Confirming that this works as intended.

jebw:
@botanicus no worries about credit - just keen to see this get fixed, great to see the PR

botanicus:
@jakeNiemiec oh and about using `logger.info`: this gets triggered only if there's an error and `STDOUT` is likely to contain that error. It's not conclusive, perhaps, the original issue was about the error being swallowed, and I think it's better to log some debugging info to `logger.error` than to log an error to `logger.info` and have it potentially ignored is log level is set to error only for instance.

gauravtiwari:
Thanks @botanicus

PR #2228, opened 2019-08-12
Nodejs upgrade

smondal:
Here is the change log from Node 6 to 8. No other changes required except version.
As Node Js still support 8, So I am changing from 6 to 8.
https://github.com/nodejs/wiki-archive/blob/master/Breaking-changes-between-v6-LTS-and-v8-LTS.md


gauravtiwari:
thanks @smondal

frenkel:
Would it be possible to set 8.10 as a minimum? Ubuntu LTS doesn't provide 8.16 and I think quite a number of servers run plain Ubuntu LTS.

mvz:
Also, Travis uses Xenial which comes with node 8.12.0.

PR #2231, opened 2019-08-13
set configured path to railties config

andrewkress:
This change goes along with rails pull request rails/rails#36803 and issue rails/rails#36799

On initialization, we set the configured Webpacker path so that rails will exclude that path from autoloading.

### DIFF
--- a/lib/webpacker/railtie.rb
+++ b/lib/webpacker/railtie.rb
@@ -91,6 +91,6 @@ class Webpacker::Engine < ::Rails::Engine
   end
 
   initializer "webpacker.set_source" do |app|
-    config.javascript_path = Webpacker.config.source_path.relative_path_from(Rails.root.join("app")).to_s
+    app.config.javascript_path = Webpacker.config.source_path.relative_path_from(Rails.root.join("app")).to_s
   end
 end
--- a/lib/webpacker/railtie.rb
+++ b/lib/webpacker/railtie.rb
@@ -89,4 +89,8 @@ class Webpacker::Engine < ::Rails::Engine
       end
     end
   end
+
+  initializer "webpacker.set_source" do
+    config.javascript_path = Webpacker.config.source_path.relative_path_from(Rails.root.join("app")).to_s
+  end
 end
--- a/lib/webpacker/railtie.rb
+++ b/lib/webpacker/railtie.rb
@@ -90,7 +90,7 @@ class Webpacker::Engine < ::Rails::Engine
     end
   end
 
-  initializer "webpacker.set_source" do
+  initializer "webpacker.set_source" do |app|
     config.javascript_path = Webpacker.config.source_path.relative_path_from(Rails.root.join("app")).to_s
   end
 end


lib/webpacker/railtie.rb
```diff
@@ -89,4 +89,12 @@ class Webpacker::Engine < ::Rails::Engine
       end
     end
   end
+
+  initializer "webpacker.set_source" do
+    if defined?(Rails::Server) || defined?(Rails::Console)
```
rafaelfranca: Why we need to care about the server or the console?

andrewkress: I thought it would only be necessary for those scenarios.  I guess there would be no need to limit.

lib/webpacker/railtie.rb
```diff
@@ -89,4 +89,8 @@ class Webpacker::Engine < ::Rails::Engine
       end
     end
   end
+
+  initializer "webpacker.set_source" do
```
rafaelfranca:
```suggestion
  initializer "webpacker.set_source" do |app|
```

lib/webpacker/railtie.rb
```diff
@@ -89,4 +89,8 @@ class Webpacker::Engine < ::Rails::Engine
...
+    config.javascript_path = Webpacker.config.source_path.relative_path_from(Rails.root.join("app")).to_s
```
rafaelfranca:
```suggestion
    app.config.javascript_path = Webpacker.config.source_path.relative_path_from(Rails.root.join("app")).to_s
```

agrobbin:
@rafaelfranca @andrewkress I think there might still be a subtle bug in this initialization logic. We're running into some issues with this in Webpacker 4.2.0, and I'm not sure how best to go about fixing it.

When we set a custom `source_path` in `config/webpacker.yml` (in our case, `app/frontend`), Zeitwerk still attempts to load that folder when eager-loading the application due to this initializer block not executing early enough in the initialization chain.

Based on what I can see, [`Rails::Engine::Configuration#paths`](https://github.com/rails/rails/blob/v6.0.1/railties/lib/rails/engine/configuration.rb#L38-L72) is invoked before `javascript_path` has been set, which causes it to use the default of `javascript`, and eventually causes the [`eager_load!`](https://github.com/rails/rails/blob/v6.0.1/railties/lib/rails/application/finisher.rb#L116-L125) initializer block to fail due to Zeitwerk errors.

I'd love to fix this myself, but I'm struggling to follow how initializers are loaded, and how that plays with things like engine path customization (i.e. [Paper Trail](https://github.com/paper-trail-gem/paper_trail/blob/v10.3.1/lib/paper_trail/frameworks/rails/engine.rb#L26)).

If someone can look into this, and maybe point me in the right direction for how best to fix this, that'd be awesome, thanks!!

schinery:
@agrobbin do you know if what you asked was resolved re: using a custom directory? Am currently upgrading an app from Rails 5.2 to 6 with a custom directory of `app/webpacker` and have the same issue, although oddly only when I deploy to Heroku.

agrobbin:
@schinery unfortunately I don't think it has been fixed. We've worked around the issue by doing this in our `config/application.rb`:

```ruby
config.javascript_path = Webpacker.config.source_path.relative_path_from(Rails.root.join('app')).to_s
```

This way, `javascript_path` is set early enough in the initialization chain.

schinery:
Thanks @agrobbin for the workaround, I'll give that a try. Hopefully, it'll get resolved properly at some point.

PR #2232, opened 2019-08-14
Remove yarn note from deployment doc

hjiangwen:
Because webpacker runs `yarn install` anyway during `assets:precompile`, we should not run this task again.

related discussion: https://github.com/rails/webpacker/pull/2008


jakeNiemiec:
It should probably still be documented right?

hjiangwen:
> It should probably still be documented right?

Capistrano note should still be there, but `yarn install` should be remove

gauravtiwari:
Please update the pr and we will merge. Thank you

PR #2234, opened 2019-08-14
Enable production optimizations when compiling in production

botandrose:
Smaller assets and better performance. More info here: https://elm-lang.org/0.19.0/optimize


gauravtiwari:
Thanks @botandrose just made some small changes (hope that's okay) :)

botandrose:
@gauravtiwari Much nicer. Thank you!

PR #2250, opened 2019-08-21
We should used deepmerge plugin for merging object

smondal:
There is a well defined plugin regarding deep merging and used by 500242 people and lots of test cases covered.

Instead of our own writing we can used this plugin.
https://github.com/TehShrike/deepmerge


shakacode-tools:
@jakeNiemiec @gauravtiwari this one should be closed. I've tried deepmerge and it had issues with webpack config objects.

justin808:
@jakeNiemiec @gauravtiwari this one should be closed. I've tried deepmerge and it had issues with webpack config objects.

PR #2251, opened 2019-08-23
coerce https yaml config to boolean

hwhelchel:
This is a proposed solution to this discussion: https://github.com/rails/webpacker/issues/2000

Allows passing https configuration to the webpacker dev server within `webpacker.yml`

Before this PR if you configured webpacker.yml like:

```yml
  dev_server:
    https:
      key: './config/ssl/key.pem'
      cert: './config/ssl/cert.pem'
```

The https configuration would be passed to webpacker but the webpacker DevServerProxy would not recognize `https` being enabled. Therefore it would set `env["HTTPS"] = env["HTTP_X_FORWARDED_SSL"] = "off"` causing errors.

### DIFF
--- a/lib/webpacker/dev_server.rb
+++ b/lib/webpacker/dev_server.rb
@@ -31,7 +31,7 @@ class Webpacker::DevServer
   end
 
   def https?
-    case fetch(:https)
+    case !!fetch(:https)
     when true, "true"
       true
     else
--- a/lib/webpacker/dev_server.rb
+++ b/lib/webpacker/dev_server.rb
@@ -31,8 +31,8 @@ class Webpacker::DevServer
   end
 
   def https?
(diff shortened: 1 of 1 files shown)


frankjmattia:
Forgive me for jumping in with my $0.02, but I was fighting with this issue this PR helped me immensely.

It got me thinking: would it be better to instead change the first case to read:

`when true, "true", Hash`

My reasoning was that coercing https to boolean means the string "true" is never actually reached, so instead of removing it/hiding the intention, just make it explicit by adding `Hash` to the list of truthy values.

hwhelchel:
@frankjmattia your suggestion makes sense to me. I've just updated to check if a Hash is passed instead.

gauravtiwari:
thanks @hwhelchel

Issue #1163, opened 2018-01-09
Loading scss.erb


timonbandit:
Hi!
i have a problem, i can't load my variables from .scss.erb files.
```
ERROR in ./app/themes/theme1/assets/stylesheets/theme1/variables.scss.erb
Module parse failed: Unexpected token (3:5)
You may need an appropriate loader to handle this file type.
| 
| 
| body {
|   background: blue!important;
| }
```

The source file contains:

```
<% color = 'blue' %>

body {
  background: <%=color%>!important;
}
```
So .erb works, but webpack cannot load it.

js.erb works well after ```bundle exec rails webpacker:install:erb```

How can i fix this problem with scss.erb?

gauravtiwari:
@timonbandit Ahh seems like you need to add `erb-loader` in front of sass loader. From the log it looks like erb loader is compiling it correctly but then it doesn't know which loader to hand off afterwards?

```
You may need an appropriate loader to handle this file type.
| 
| 
| body {
|   background: blue!important;
| }
```
You could use sass variables thought no?

renatodeleao:
This is an old thread, but i have to deal with this problem recently, and it's still number one google result for `webpacker scss.erb` query. 

After struggling for a while, based on @gauravtiwari lead:
```js
// default bundle exec rails webpacker:install:erb

// loaders/erb.js
module.exports = {
  test: /\.erb$/,
  enforce: 'pre',
  exclude: /node_modules/,
  use: [{
    loader: 'rails-erb-loader',
    options: {
      runner: (/^win/.test(process.platform) ? 'ruby ' : '') + 'bin/rails runner'
    }
  }]
}

// environment.js
environment.loaders.prepend('erb', erb)
```
As stated this works out of the box for `.js.erb`, not for (scss|sass).erb


#### My attempt to ship it
```js
// loaders/sass-erb.js
const { environment } = require('@rails/webpacker');
const erbLoader  = require('./erb');
const sassLoaders = environment.loaders.get('sass').use

module.exports = {
  test: /\.(scss|sass)(\.erb)$/i,
  enforce: 'pre',
  exclude: /node_modules/,
  use: [
    ...sassLoaders,
    erbLoader.use[0]
  ]
}

```

```diff
// environment.js
environment.loaders.prepend('erb', erb)
+ environment.loaders.insert('sass-erb', sassErb, { before: 'erb'} )
```

This seems to be working, although not extensively tested. 
[...]

PR #2259, opened 2019-08-28
Support .(sass|scss).erb

swrobel:
@gauravtiwari maybe I'm misunderstanding [your comment](https://github.com/rails/webpacker/issues/1163#issuecomment-357438020), but it seems that the erb loader forces itself to the front of the loader chain (I checked my loaders and it is first) but the sass loader still needs to be ok with the fact that the file might end with `.erb`

### DIFF
--- a/package/rules/sass.js
+++ b/package/rules/sass.js
@@ -1,6 +1,6 @@
 const getStyleRule = require('../utils/get_style_rule')
 
-module.exports = getStyleRule(/\.(scss|sass)$/i, false, [
+module.exports = getStyleRule(/\.(scss|sass)(\.erb)?$/i, false, [
   {
     loader: 'sass-loader',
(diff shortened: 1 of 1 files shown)


package/rules/sass.js
```diff
@@ -1,6 +1,6 @@
 const getStyleRule = require('../utils/get_style_rule')
 
-module.exports = getStyleRule(/\.(scss|sass)$/i, false, [
+module.exports = getStyleRule(/\.(scss|sass)(\.erb)?$/i, false, [
```
gauravtiwari: Looks great, did you test the regex? `/\.(scss|sass)?(\.erb)?$/` (optional .erb ext) I am no expert but feel like this is correct.

gauravtiwari:
Just tested, this one looks fine too :) Thanks

PR #2263, opened 2019-09-01
Fix typo in hello_erb.js.erb 

neko314:
I found a typo so fixed it.


petitkriket:
this is actually not a typo and here on purpose, it prevent the code to be executed as erb fragment

neko314:
@petitkriket 
Oh, I found this PR but I missed the conversations. Thank you so much for your telling me👍

Issue #2089, opened 2019-05-08
Adding support for compression with Brotli


guillaumebriday:
Hey ! 👋 

As far as I know, Webpacker is using https://github.com/webpack-contrib/compression-webpack-plugin to generate the gzip compressed files.

This plugin added the support for Brotli https://github.com/webpack-contrib/compression-webpack-plugin/pull/119 and I think it's could be great to add it [in Webpacker](https://github.com/rails/webpacker/blob/9e671a3ebe368cfdc624309a9b4a55e998f87186/package/environments/production.js#L13)

Adding he proper documentation for nginx headers and an opt-out configuration in the webpacker.yml could be great too.

What do you think ? I could open a PR if you're ok with this feature !

Thank you in advance

aried3r:
Hey @guillaumebriday, I really like your idea. What do you think about submitting a PR were this can be discussed further? I think once there's a PR it becomes more real and others can maybe already use it in their configs.

As this feature is only available after node.js 11.7.0 and later this should only be the default for compatible setups, since webpacker (so far) works with node.js 6 and above, while 10 is the currently active LTS release and 11 is EOL. https://github.com/nodejs/Release#release-schedule

What do you think? :)

guillaumebriday:
Ok I'll will try to open a PR soon,

and yes, this should be the default only for compatible systems !

Thank you

julianrubisch:
I think this issue can be closed?

guillaumebriday:
Sure ! Thanks @clupprich

PR #2273, opened 2019-09-06
Add support for Brotli compression

clupprich:
Adds support for Brotli compression via the `compression-webpack-plugin` package/plugin when running Node versions >=11.7.0, as described in https://github.com/rails/webpacker/issues/2089.

I could use some help in writing tests for this - any good starting points?
~I'm also not really happy about the added dependency on the `semver` package - maybe some small utility function would do the trick here?~ Fixed that in 716ac1d by checking the presence of `process.versions.brotli`.

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -1,7 +1,7 @@
 # Deployment
 
 
-Webpacker hooks up a new `webpacker:compile` task to `assets:precompile`, which gets run whenever you run `assets:precompile`. 
+Webpacker hooks up a new `webpacker:compile` task to `assets:precompile`, which gets run whenever you run `assets:precompile`.
 If you are not using Sprockets `webpacker:compile` is automatically aliased to `assets:precompile`. Remember to set NODE_ENV environment variable to production during deployment or when running the rake task.
 
 The `javascript_pack_tag` and `stylesheet_pack_tag` helper method will automatically insert the correct HTML tag for compiled pack. Just like the asset pipeline does it.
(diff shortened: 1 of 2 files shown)


clupprich:
Ah, Brotli has its own entry in `process.versions`, I can probably use that. Thanks to @julianrubisch for that hint!

guillaumebriday:
Hey!

That is so cool, thank you @clupprich 👍 

I think we should add configuration example and installation details in the https://github.com/rails/webpacker/blob/master/docs/deployment.md?

What do you think?

Thanks

clupprich:
@guillaumebriday Definitely! Do you think we should also add a config flag to enable/disable this? Because Brotli isn't widely supported yet by e.g. hosted solutions (Heroku, afaik), this can be overhead that people don't need.

guillaumebriday:
I think we should not add a specific config option only for Brotli even if it's not supported by some hosted solutions. These solutions will simply ignore the Brotli files.

I think, we can add sentence like that but for Brotli:
> Some servers support sending precompressed versions of files with the .gz extension when they're available. For example, nginx offers a gzip_static directive.

And add a link (or steps) to install it with a warning on the minimal node version required.

What do you think?

clupprich:
Added the instructions in f70f7ef. Turns out the hard part is not the code, but the documentation ;). Let me know if that section makes sense.

guillaumebriday:
It makes sens ! The google's instruction are not very clear imo.

I think [this line]( https://github.com/rails/webpacker/pull/2273/files#diff-00e64a8af4dd5116b95476a6691792dcR45) should be removed.

gauravtiwari:
Thanks @clupprich

clupprich:
🎉 Thanks @guillaumebriday and @gauravtiwari

guillaumebriday:
🎉

PR #2280, opened 2019-09-10
Update nginx config example

guillaumebriday:
Just a try to improve the nginx config example.

Inspired by https://laravel.com/docs/5.8/deployment#nginx.


gauravtiwari:
thanks @guillaumebriday

PR #2281, opened 2019-09-10
Added favicon_pack_tag to generate favicon links

kpumuk:
There is no way at the moment to generate a favicon link that uses a webpacker managed image. This PR adds `favicon_pack_tag` that behaves in a similar way to `image_pack_tag`, but instead of an image tag, it produces a favicon link.


gauravtiwari:
thanks @kpumuk

PR #2287, opened 2019-09-17
Directly call webpack from compiler

gauravtiwari:
Changes compiler to directly invoke webpack instead of calling another binstub, that invokes webpack. The actual problem is not clear since the compiler hangs indefinitely, when using the old setup. 

Here are some compilation results, with new changes: 

### Before change:

Compiler hangs indefinitely on Heroku and locally takes about 10 minutes (Macbook Pro). 

After change
----

### Without cache:

Screenshot 2019-09-17 at 12 28 28

### With cache: 

Screenshot 2019-09-17 at 12 25 37

### With opts: 

Screenshot 2019-09-17 at 12 33 54

### DIFF
--- a/lib/webpacker/compiler.rb
+++ b/lib/webpacker/compiler.rb
@@ -10,6 +10,10 @@ class Webpacker::Compiler
   # Webpacker::Compiler.env['FRONTEND_API_KEY'] = 'your_secret_key'
   cattr_accessor(:env) { {} }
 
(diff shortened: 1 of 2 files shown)


gauravtiwari:
Closing, since this seems to be a deeper problem with webpack. It doesn't seem to send non-zero exit status if there are any errors that's why the `compiler` hangs.

PR #2303, opened 2019-09-26
Enforced nokogiri >= 1.10.4 due to CVE-2019-5477

kamilc:
We've found that the latest `webpacker` is pulling `nokogiri 1.10.3` which is known to be vulnerable to the command injection via the `Kernel.open` method ([CVE-2019-5477](https://nvd.nist.gov/vuln/detail/CVE-2019-5477)).

This pull request locks `nokogiri` at `1.10.4` which is considered safe.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,6 +5,7 @@ gemspec
 gem "rails"
 gem "rake", ">= 11.1"
 gem "rack-proxy", require: false
+gem "nokogiri", ">= 1.10.4"
 
 group :test do
   gem "minitest", "~> 5.0"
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -5,6 +5,7 @@ gemspec
 gem "rails"
 gem "rake", ">= 11.1"
 gem "rack-proxy", require: false
+gem "nokogiri", ">= 1.10.4"
```
rafaelfranca: This line is not needed. Can you remove it?

gauravtiwari:
Updated deps on master so this can be closed.

PR #2306, opened 2019-09-29
update deprecated option

mathieujobin:
rubocop is telling me this option has been renamed
since I bundle install with `--path vendor` 
this is now breaking me CI

thank you


mathieujobin:
Not like it would have hurt anything to merge... But whatever. Thank you for your time

PR #2316, opened 2019-10-07
Enable webpack_compile_output and errorDetails by default

### LINKED ISSUE #2312: Why isn't webpack_compile_output enabled by default?
How you doing guys.

I was wondering why isn't `webpack_compile_output` enabled by default. If webpack fails to compile your bundles you won't see anything unless you enable it. Compiling issues are so commonplace when working on code or deploying that it doesn't make sense to me to have all webpack output stripped out by default.

Would you guys consider having `webpack_compile_output` enabled by default? I think most people would benefit from that. Of course some devs may not want to see all the webpack output every time they do a change and refresh the browser, but they can always disable it.

LuanGB:
Fixes #2312

### DIFF
--- a/lib/install/config/webpacker.yml
+++ b/lib/install/config/webpacker.yml
@@ -7,7 +7,7 @@ default: &default
   public_output_path: packs
   cache_path: tmp/cache/webpacker
   check_yarn_integrity: false
-  webpack_compile_output: false
+  webpack_compile_output: true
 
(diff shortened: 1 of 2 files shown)


jakeNiemiec:
Closes https://github.com/rails/webpacker/issues/2312 
//cc @gauravtiwari

cmitz:
Could we release this, like, yesterday? 😜

PR #2333, opened 2019-10-29
NodeModules Babel exclude regex

ncodech:
We had a problem using the module "webpacker-svelte" and IE11 compatibility.
We found out that the module is not processed by babel because of the exclude pattern used for the nodeModules rule.

I changed the regex according to the info here: https://regex101.com/r/CglKdg/5/

Hope this helps!


jakeNiemiec:
Great catch! Thank you.

ncodech:
Thank you! :)

Issue #1374, opened 2018-03-23
Railtie yarn check --integrity fails on development


doits:
just updated to 3.4.0 and I get this:

```
========================================
  Your Yarn packages are out of date!
  Please run `yarn install` to update.
========================================


To disable this check, please add `config.webpacker.check_yarn_integrity = false`
to your Rails development config file (config/environments/development.rb).


yarn check v1.5.1
warning package.json: No license field
warning No license field
warning Integrity check: Flags don't match
error Integrity check failed
error An unexpected error occurred: "Found 1 errors.".
info If you think this is a bug, please open a bug report with the information provided in "/Users/markus/svn/caseman-events/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/check for documentation about this command.
```

If I run it manually, everything is fine:

```
→   yarn check --integrity                                                                                                                                                                       [demo/develop]
yarn check v1.5.1
warning package.json: No license field
warning No license field
success Folder in sync.
   Done in 0.10s.
```

If I use `NODE_ENV=production` I can reproduce it:

```
→   NODE_ENV=production yarn check --integrity                                                                                                                                                   [demo/develop]
yarn check v1.5.1
warning package.json: No license field
warning No license field
warning Integrity check: Flags don't match
error Integrity check failed
error An unexpected error occurred: "Found 1 errors.".
info If you think this is a bug, please open a bug report with the information provided in "/Users/markus/svn/caseman-events/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/check for documentation about this command.
```

If I install with production ...

```
NODE_ENV=production yarn install
```

... I can use rails again and the rails integrity check works. But of course I want to use my development yarn packages, so this is no solution.

The problems seems to be that `yarn check --integrity` is run with `NODE_ENV=producion`, even locally when developing.

gauravtiwari:
@doits When is this happening -running binstubs or rake tasks?

doits:
It happened to me when trying to upgrade the binstubs

```
bundle exec rails webpacker:binstubs
```

kurtnovack:
This is happening to me on a fresh install:

created a new app:
```
rails new install_test --webpack
```

created databases:
```
rails db:create
```

tried starting server
```
rails server
```

output:
```
=> Booting Puma
=> Rails 5.1.5 application starting in development
=> Run `rails server -h` for more startup options


========================================
  Your Yarn packages are out of date!
  Please run `yarn install` to update.
========================================


[...]

PR #2341, opened 2019-10-30
Revert of production env enforcement

### LINKED ISSUE #2340: Webpacker not respecting configuration
Since #1395 got closed, continuing the discussion here because this problem needs to be fixed.

With the change introduced in https://github.com/rails/webpacker/commit/d8dd4460144e63a9120290631c492aba3cbea332 , running `rake assets:precompile` breaks our CI since it always loads the `config/webpack/production.js` instead of the `test.js` file, causing the CI to error out since the code in that file does not have the required ENV variables available.

Explicitly setting `NODE_ENV=test` or `RAILS_ENV=test` has no bearing whatsoever on webpacker.
The only way we can currently get our assets compiled is by running `RAILS_ENV=test NODE_ENV=test ./bin/webpack`

coding-bunny:
# Description
The change made in https://github.com/rails/webpacker/commit/d8dd4460144e63a9120290631c492aba3cbea332 forces the `production` environment on every compile, rake task and usage of the webpacker gem, ignoring the environment variables used by the system and/or developer.

1. The tool does not get to decide how it's running, it needs to respect the ENV.
2. This breaks CI systems that have specific configurations
3. It steps away from convention over configuration.

By retaining the original code, and defaulting to the `production` environment, it's possible for people to compile their assets and dependencies under their requirements, ensuring no breaking changes in the systems. Proper documentation can/should be added to explain the usage of the `NODE_ENV` and how it affects packages for people not familiar with Node and NPM, but this change has too much breaking potential.

Resolves #2340

### DIFF
--- a/lib/tasks/webpacker/compile.rake
+++ b/lib/tasks/webpacker/compile.rake
@@ -29,12 +29,12 @@ namespace :webpacker do
   desc "Compile JavaScript packs using webpack for production with digests"
   task compile: ["webpacker:verify_install", :environment] do
     ensure_log_goes_to_stdout do
-	  if Webpacker.compile
-	    # Successful compilation!
-	  else
-	    # Failed compilation
-	    exit!
-	  end
+      if Webpacker.compile
+        # Successful compilation!
+      else
+        # Failed compilation
+        exit!
+      end
     end
   end
 end
--- a/lib/tasks/webpacker/compile.rake
+++ b/lib/tasks/webpacker/compile.rake
@@ -1,5 +1,3 @@
-ENV["NODE_ENV"] ||= "production"
-
 $stdout.sync = true
 
 def ensure_log_goes_to_stdout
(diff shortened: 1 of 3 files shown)


rossta:
To add another take here, what about instead, something like the following?

```ruby
Webpacker.with_node_env(ENV.fetch("NODE_ENV", "production")) do
# ...
```

I believe this would maintain [the spirit of the change in question](https://github.com/rails/webpacker/pull/1377), which was to remove global env pollution, i.e. `ENV["NODE_ENV"] ||= 'production'` to fix https://github.com/rails/webpacker/issues/1374, while also satisfying the desire stated here: to allow the override of the default behavior.

coding-bunny:
Implemented your change @rossta 
This makes the README actually correct again, and allows overwriting/using different environments.
Default is production, which allows the suggestions from #1374 to work as well by not setting the NODE_ENV

PR #2353, opened 2019-11-05
Add detailed instructions for HMR

vfonic:
I just enabled HMR in my rails/webpacker project. I couldn't find a single place with all the documented changes needed to be done for it to work fully.*

*work fully => without full page reload after every change and with support for editing project files directly from the browser developer tools

Preview the new markdown here:
https://github.com/rails/webpacker/blob/df59eb47f2eba2863587287d10ab2af6ade6c2bf/docs/webpack-dev-server.md

### DIFF
--- a/docs/webpack-dev-server.md
+++ b/docs/webpack-dev-server.md
@@ -19,6 +19,74 @@ Now if you refresh your Rails view everything should work as expected.
 Webpacker out-of-the-box supports HMR with `webpack-dev-server` and
 you can toggle it by setting `dev_server/hmr` option inside `webpacker.yml`.
 
+Before turning HMR on, consider upgrading to latest stable gems and packages:
+https://github.com/rails/webpacker#upgrading
+
+Here's the full set of changes you need to do to get the most out of HMR (this is subject to change, for the latest solution, check the webpack/HMR official documentation links below):
+
+```diff
+# Procfile.dev
+web: bundle exec puma -C config/puma.rb
+- webpacker: ./bin/webpack-dev-server
++ webpacker: ./bin/webpack-dev-server --hot
+```
+
+```diff
+// app/javascript/app.jsx
+
+import React from 'react';
++ import { hot } from 'react-hot-loader';
+
+const App = () => <SomeComponent(s) />
+
+- export default App;
++ export default hot(module)(App);
+```
+
+```diff
+development:
+  # ...
+  dev_server:
+-    hmr: false
++    hmr: true
+    # Inline should be set to true if using HMR
+-    inline: false
++    inline: true
+```
+
+```diff
+// config/webpack/development.js
+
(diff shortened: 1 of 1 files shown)


jakeNiemiec:
I'm not sure if that's broad enough or how much @hot-loader is doing for you. There seems to be missing content e.g. https://webpack.js.org/api/hot-module-replacement/#accept (at least from the HMR stuff I have done)

Does this just work for react? Perhaps this should be in a react-specific doc?

vfonic:
> Does this just work for react?

I've tested this only with react. I'm not familiar with other frameworks nor whether it would work with them.

> Perhaps this should be in a react-specific doc?

You're right. I guess this would be best moved to react doc, at least until we figure out whether it works for other frameworks.

> I'm not sure if that's broad enough or how much @hot-loader is doing for you. There seems to be missing content e.g. webpack.js.org/api/hot-module-replacement/#accept (at least from the HMR stuff I have done)

Indeed, I had this code snippet before:

```js
if (module.hot) {
  module.hot.accept('./library.js', function() {
  // ...
}
```

But it seems to be working without it as well. I think the "magic" comes from webpacker and new version of "react-hot-loader" (as far as I know it was recently updated with a lot of changes to hot reloading). Here's the "Getting started" section:
https://github.com/gaearon/react-hot-loader#getting-started

After having another look, I realized that `hot(module)(App)` is the old API. I'll update the PR now to use the new API and also add the babel plugin. I'm not sure if it's required or not, but it seems to be working with it.

vfonic:
Updated

docs/webpack-dev-server.md
```diff
@@ -19,6 +19,72 @@ Now if you refresh your Rails view everything should work as expected.
...
+Before turning HMR on, consider upgrading to latest stable gems and packages:
+https://github.com/rails/webpacker#upgrading
+
+Here's the full set of changes you need to do to get the most out of HMR (this is subject to change, for the latest solution, check the webpack/HMR official documentation links below):
+
+```diff
+# Procfile.dev (only if you need/use it)
+web: bundle exec puma -C config/puma.rb
+- webpacker: ./bin/webpack-dev-server
++ webpacker: ./bin/webpack-dev-server --hot
```
lewispb: Do you need this _and_ hmr: true?

vfonic:
It would be great if someone could take over this PR. I currently unfortunately don't have the time to wrap it up properly. :/

justin808:
@jakeNiemiec I'm happy to take this one over. Let me know as I don't want to duplicate the work.

vfonic:
Hey @justin808! Thanks for taking over! I'll close this PR in favor of yours.

docs/webpack-dev-server.md
anikolaiev: yes, at least in my case with Vue.js

PR #2360, opened 2019-11-11
Update dependencies

gauravtiwari:
Updates gem and npm dependencies


package/utils/helpers.js
```diff
@@ -1,29 +1,29 @@
...
   && (value.length === undefined || value.length === null)
 
-const isNotObject = value => !isObject(value)
+const isNotObject = (value) => !isObject(value)
 
-const isBoolean = str => /^true/.test(str) || /^false/.test(str)
+const isBoolean = (str) => /^true/.test(str) || /^false/.test(str)
 
-const isEmpty = value => value === null || value === undefined
+const isEmpty = (value) => value === null || value === undefined
```
justin808: @javan @jakeNiemiec @gauravtiwari  What do you guys think about separating linting  updates (eslint and rubocop) from other updates?

The "blame" of files makes more sense if this is done.

PR #2381, opened 2019-11-29
Fix config file path in error message

arashm:
Fix printing blank instead of config path when the port is already in use.


gauravtiwari:
thanks :)

PR #2391, opened 2019-12-09
Fix assets clean step

tachyons:
File.exists? return true for directories and it is causing issue in assets clean step

```
rake aborted!
Errno::EISDIR: Is a directory @ apply2files - /home/circleci/project/public/packs-test/js
/home/circleci/project/vendor/bundle/ruby/2.6.0/gems/webpacker-4.2.1/lib/webpacker/commands.rb:11:in `delete'
```


gauravtiwari:
Made a new release - 4.2.2

gauravtiwari:
thanks for testing 🙏

Issue #2375, opened 2019-11-26
Compilation fails due to 'Maximum call stack size exceeded'


Volosh1n:
Hi!

Recently project moved from Sprockets to Webpacker. All styles were moved from `app/assets/stylesheets` to `app/webpack/stylesheets` (webpack folder name was configured in `webpacker.yml`).

In `webpacker.yml` I only added following configs:
```yml
source_path: app/webpack
resolved_paths: ['app/assets'] # because we still using images from app/assets/images
extract_css: true
```
And Bootstrap/JQuery/Popper added to webpacker too.

Current file structure:
```
app
├── # ...
└── webpack
    ├── packs
    │   ├── application.js
    │   ├── landings.scss
    │   └── services.scss
    └── stylesheets
```

Files like `landings.scss` just imports many other styles from `app/webpacker/stylesheets/` folder with many `@import`s. But the following issue occurred many times:
![webpacker_screenshot](https://user-images.githubusercontent.com/16169679/69623921-5eaa4f00-104c-11ea-91f9-2ace4d685c44.png)

For the first time, we just cropped large CSS bundles (that's why there is no application.scss), which was included by controller name (for example, `services_controller` requires only `services.css`, etc.) and it temporary fixed problem. But with style files growing, this problem returned. Splitting CSS no longer helps, large CSS bundle crashes compilation.

Has anyone faced this issue too? Can stack size be increased in config? I found flag `--stack-size=1000`, but it didn't help with any value.
Also, I checked all styles files for recursive `@import`s, but everything is fine.

jakeNiemiec:
In order to help debug, could you please post as much as you can of:
- `services.scss`
- `./config/webpack/environment.js`
- `./config/webpacker.yml`
- `./app/javascript/packs/application.js` 
- the full error message (even the parts you wouldn’t think of as relevant)

It looks like it's having a problem with [acornjs](https://github.com/acornjs/acorn), how are you using this?

Volosh1n:
@jakeNiemiec , thanks for reply!
Actually, we don't use acorn explicitly anywhere, just Webpacker.
About error: screenshot covers it almost fully - there are some lines about acorn.js after that screenshot and then default webpack output about built packs, nothing more about error, unfortunately.

Here is some files you requested:

* `./config/webpacker.yml`:
```yml
default: &default
  source_path: app/webpack
  source_entry_path: packs
  public_root_path: public
  public_output_path: packs
  cache_path: tmp/cache/webpacker
[...]

PR #2409, opened 2019-12-20
Add ability to pass node params

za4emnik:
Hi guys! We recently faced a problem #2375. I think it would be nice to pass parameters to the node.


gauravtiwari:
@za4emnik thanks, would you be able to look at this at some point?

za4emnik:
@gauravtiwari yes, I will do it shortly.

PR #2421, opened 2020-01-02
Remove defaults based on webpack mode development

justin808:
Default of cache is true:
https://webpack.js.org/configuration/other-options/#cache

Default of pathinfo is true:
https://webpack.js.org/configuration/output/#outputpathinfo


justin808:
@gauravtiwari Any reason to include settings that are already at the default for "development" mode?

gauravtiwari:
thanks @justin808 and @jakeNiemiec

PR #2422, opened 2020-01-02
Update debugging instructions

justin808:
Since webpack supports a --debug option and since we pass
all additional args to webpack, we don't want to overload
the --debug option.

### DIFF
--- a/lib/webpacker/dev_server_runner.rb
+++ b/lib/webpacker/dev_server_runner.rb
@@ -52,7 +52,7 @@ module Webpacker
           ["yarn", "webpack-dev-server"]
         end
 
-        if ARGV.include?("--debug-webpacker")
+        if @argv.include?("--debug-webpacker")
           cmd = [ "node", "--inspect-brk"] + cmd
         end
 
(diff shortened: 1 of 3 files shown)


justin808:
@gauravtiwari this is a small, useful adjustment.

justin808:
Thanks @jakeNiemiec! How do you handle merges?

lib/webpacker/webpack_runner.rb
```diff
@@ -12,9 +12,8 @@ def run
         ["yarn", "webpack"]
       end
 
-      if ARGV.include?("--debug")
+      if @argv.include?("--debug-webpacker")
```
gauravtiwari: Does webpack complains, if an unsupported option is present? We might need to delete the custom ones then?

lib/webpacker/dev_server_runner.rb
```diff
@@ -52,9 +52,8 @@ def execute_cmd
           ["yarn", "webpack-dev-server"]
         end
 
-        if ARGV.include?("--debug")
+        if ARGV.include?("--debug-webpacker")
```
gauravtiwari: We can use instance variable here too? `@argv`

PR #2426, opened 2020-01-07
Ensure missing entry error message includes asset file type

rossta:
Webpack view helpers that indicate pack type in the method name, e.g.
`javascript_pack_tag`, `stylesheet_pack_tag`, support manifest lookup both
with or without the extension in the given file name. For example, given
an existing pack for application.js, all of the following manifest
lookups will succeed:

```
javascript_pack_tag "application"    # or
javascript_pack_tag "application.js"

stylesheet_pack_tag "application"    # or
stylesheet_pack_tag "application.css"
```

When the lookup does not succeed, a helpful error message is displayed.
However, if the ".js" extension is omitted from the helper method
invocation, the error message also omits the pack type. For example:

```
javascript_pack_tag "not-a-pack"
stylesheet_pack_tag "not-a-pack"
```
results in the following error message:

```
Webpacker can't find not-a-pack in
/path/to/app/public/packs/manifest.json
```

The ambiguity is confusing when using both view helpers: does this
indicate the server was looking for the javascript bundle or the
stylesheet?

This changeset resolves the ambiguity. We can leverage the existing support
in the manifest for determining the qualified pack name when printing the 
missing entry info.

Now, for a missing stylesheet such as:

```
stylesheet_pack_tag "not-a-pack"
```

the following error message is generated:

```
Webpacker can't find not-a-pack.css in
/path/to/app/public/packs/manifest.json
```

Also, adds some assertions and extracts a couple shared methods in
manifest_test.rb

### DIFF
--- a/lib/webpacker/manifest.rb
+++ b/lib/webpacker/manifest.rb
@@ -29,7 +29,7 @@ class Webpacker::Manifest
   end
 
   def lookup_pack_with_chunks!(name, pack_type = {})
-    lookup_pack_with_chunks(name, pack_type) || handle_missing_entry(name)
+    lookup_pack_with_chunks(name, pack_type) || handle_missing_entry(name, pack_type)
   end
 
   # Computes the relative path for a given Webpacker asset using manifest.json.
@@ -46,7 +46,7 @@ class Webpacker::Manifest
 
   # Like lookup, except that if no asset is found, raises a Webpacker::Manifest::MissingEntryError.
   def lookup!(name, pack_type = {})
-    lookup(name, pack_type) || handle_missing_entry(name)
+    lookup(name, pack_type) || handle_missing_entry(name, pack_type)
   end
 
   private
@@ -75,8 +75,8 @@ class Webpacker::Manifest
       "#{name}.#{manifest_type(pack_type)}"
     end
 
-    def handle_missing_entry(name)
-      raise Webpacker::Manifest::MissingEntryError, missing_file_from_manifest_error(name)
+    def handle_missing_entry(name, pack_type)
+      raise Webpacker::Manifest::MissingEntryError, missing_file_from_manifest_error(full_pack_name(name, pack_type[:type]))
     end
 
     def load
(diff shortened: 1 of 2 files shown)


jakeNiemiec:
It doesn't have to hold up this PR, but I had a few thoughts on this message since it seems to be the one that users run into the most.

It may be more "_educational_" if we refer to `not-a-pack` [like a webpack entry point](https://webpack.js.org/concepts/#entry)(or a hash), instead of a file name. e.g.:
- Webpacker can't find `not-a-pack` under `entrypoints` in `/public/packs/manifest.json` (the entry point is missing)
- Webpacker can't find `css` under `entrypoints['not-a-pack']` in `/public/packs/manifest.json` (the entry point exists, but no styles were extracted)
- `entrypoints['not-a-pack']['css']` does not exist in `/public/packs/manifest.json` (shorter?)

<hr>

> Webpacker can't find #{bundle_name} in #{config.public_manifest_path}. Possible causes:
> 1. You want to set **webpacker.yml value of compile to true for your environment**
   unless you are using the `webpack -w` or the webpack-dev-server.
> 2. webpack has **not yet re-run to reflect updates**.
> 3. You have misconfigured Webpacker's config/webpacker.yml file.
> 4. Your webpack configuration is **not creating a manifest**.
> Your manifest contains:
> #{JSON.pretty_generate(@data)}

What do you think about [making that message](https://github.com/rails/webpacker/pull/2426/files#diff-2ba273647eedaf91b83407694dcb4bedR106-R117) more dynamic/helpful?  We could check for:
- dev_server.running?
- if compile is `true` for the current environment
- if webpack has re-run to reflect updates. (`compiling?`)
- existence or validity of `config/webpacker.yml` & `manifest.json`

Like I said, not directly related to this PR. I have just seen lots of users take issue with webpacker's error messages.

rossta:
@jakeNiemiec Thanks for your thoughts on improving the missing entry error message. I'm glad the minor change I suggested here is leading to a discussion around more useful improvements; I've been feeling some of the same pain points you've alluded to.

I like your suggestions. To summarize my understanding, since it's common for developers to run into this error, there's an opportunity to improve both the "what happened" and the "what to do about it" portions of the error message. We may be able to do a better job of educating based on what we know from the config. manifest, dev_server, etc.

I'll see if I can make time to put together a separate PR which we can discuss independently of this one.

PR #2428, opened 2020-01-07
Update js dependencies

pustovalov:

https://github.com/webpack-contrib/file-loader/blob/master/CHANGELOG.md
```
minimum required nodejs version is 10.13.0
```


pustovalov:
I'd rather switch from `8.16.0` to `12.14.1`. This is current LTS version

Active LTS Start - 	2019-10-21



WDYT?

jakeNiemiec:
Just a few questions. Why is terser under resolutions? `"terser-webpack-plugin": "^2.3.1"`

Should we avoid wrapping breaking changes in a commit like "Update js dependencies". Remember that users can fine-tune the loaders in `environment.js`. This is unlikely, but it has happened with other loaders in the past couple months.

pustovalov:
>Just a few questions. Why is terser under resolutions? "terser-webpack-plugin": "^2.3.1"

Webpacker use `"terser-webpack-plugin": "^2.3.1"`
https://github.com/rails/webpacker/blob/master/package.json#L48

And webpack has this dependency:
`terser-webpack-plugin "^1.4.3"`
https://github.com/rails/webpacker/blob/master/yarn.lock#L8101

a separate dependency was introduced in this commit

https://github.com/rails/webpacker/commit/0d36af9929ac344b8d1d06ad314829f5128d5858#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L42

and then updated in this commit

https://github.com/rails/webpacker/commit/13756f1ea6b3393117878bb339b9dd507b801eb2#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L48

as a result, we have two terser-webpack-plugin:
1.4.3
https://github.com/rails/webpacker/blob/master/yarn.lock#L7615

2.3.1
https://github.com/rails/webpacker/blob/master/yarn.lock#L7630

I think terser-webpack-plugin will not be updated for webpack 4, because:

https://github.com/webpack-contrib/terser-webpack-plugin/blob/master/CHANGELOG.md#-breaking-changes
```
2.0.0 (2019-09-05)
⚠ BREAKING CHANGES
minimum require Node.js version is 8.9.0
```

Webpack 4 supporting `node >= 6.11`
https://github.com/webpack/webpack/tree/webpack-4

I use resolutions for `terser-webpack-plugin` with webpacker in my project, and all works fine

pustovalov:
>Remember that users can fine-tune the loaders in environment.js
yea, but webpacker in default configuration does not use those parameters that were changed. 

I think with the version update for Node.js, because
```
The Node.js 8.x Maintenance LTS cycle will expire on December 31, 2019 - which means that Node 8 won’t get any more updates, bug fixes or security patches.
```
https://blog.risingstack.com/update-nodejs-8-end-of-life-no-support/

We can also update and loaders

package.json
```diff
@@ -43,22 +43,25 @@
...
-    "eslint-plugin-import": "^2.18.2",
+    "eslint-plugin-import": "^2.19.1",
     "eslint-plugin-jsx-a11y": "^6.2.3",
-    "eslint-plugin-react": "^7.16.0",
+    "eslint-plugin-react": "^7.17.0",
     "jest": "^24.9.0"
   },
+  "resolutions": {
+    "terser-webpack-plugin": "^2.3.1"
+  },
```
jakeNiemiec: > as a result, we have two terser-webpack-plugin ... I think terser-webpack-plugin will not be updated for webpack 4

It is ok that we have 2 versions of terser. Webpack needs to use the version of terser that it was built to support. There is no need to micromanage the lock file 👍.

package.json
```diff
@@ -43,22 +43,25 @@
     "postcss-preset-env": "^6.7.0",
     "postcss-safe-parser": "^4.0.1",
     "regenerator-runtime": "^0.13.3",
-    "sass-loader": "7.3.1",
-    "style-loader": "^1.0.0",
+    "sass-loader": "^8.0.0",
```
jakeNiemiec: > yea, but webpacker in **default configuration does not use those parameters** that were changed...We can also update and loaders

This will break things for users that **chose to use those parameters** since we can only update loaders in our own codebase. We could have changelog entry that informs users how to migrate their `environment.js`, but this would need to be a breaking change.

package.json
```diff
@@ -8,31 +8,31 @@
     "lib/install/config/webpacker.yml"
   ],
   "engines": {
-    "node": ">=8.16.0",
+    "node": ">=10.13.0",
```
jakeNiemiec: > I think with the version update for Node.js because...

👍 I agree with this part

package.json
```diff
@@ -8,31 +8,31 @@
...
-    "@babel/runtime": "^7.7.2",
+    "@babel/core": "^7.7.7",
+    "@babel/plugin-proposal-class-properties": "^7.7.4",
+    "@babel/plugin-proposal-object-rest-spread": "^7.7.7",
+    "@babel/plugin-syntax-dynamic-import": "^7.7.4",
+    "@babel/plugin-transform-destructuring": "^7.7.4",
+    "@babel/plugin-transform-regenerator": "^7.7.5",
+    "@babel/plugin-transform-runtime": "^7.7.6",
+    "@babel/preset-env": "^7.7.7",
+    "@babel/runtime": "^7.7.7",
```
kimyu92: may want to consider pushing es2020 as 7.8.6 is out https://github.com/babel/babel/releases/tag/v7.8.6

package.json
```diff
@@ -43,20 +43,20 @@
     "postcss-preset-env": "^6.7.0",
     "postcss-safe-parser": "^4.0.1",
     "regenerator-runtime": "^0.13.3",
-    "sass-loader": "7.3.1",
-    "style-loader": "^1.0.0",
+    "sass-loader": "^7.3.1",
+    "style-loader": "^1.1.2",
     "terser-webpack-plugin": "^2.3.4",
-    "webpack": "^4.41.3",
+    "webpack": "^4.41.5",
```
kimyu92: https://github.com/webpack/webpack/releases/tag/v4.42.0

pustovalov: Hey @gauravtiwari, does it make sense to support this PR? Maybe it would be better to break it into smaller ones?

gauravtiwari: Thanks @pustovalov Yep it does. I was just concerned if there is going to be any breaking changes? I guess everything looks good except file loader node version requirements? https://github.com/webpack-contrib/file-loader/releases/tag/v5.0.0

Thanks for working on this and sorry about the delay.

gauravtiwari:
If you could double check this and update. I will merge and make a release later today 👍 

Thank you all 🙏

package.json
pustovalov: you are right, looks like breaking change only in `file-loader`

gauravtiwari:
thanks @pustovalov

amoshydra:
This PR bumped `file-loader` version from 4 to 5 which
> switch to ES modules by default (the option esModule is true by default)
> -- https://github.com/webpack-contrib/file-loader/blob/2a18cba98f1e4bd64734791a657e8acff4b11fac/CHANGELOG.md

This broke asset loading mentioned in https://github.com/rails/webpacker/issues/2613

PR #2432, opened 2020-01-12
basic yarn autoupdate functionality

ivar:
Implements a new option for the yarn integrity check setting that allows users to configure webpacker to update and/or install any outdated or missing packages. The change is backwards compatible will the existing configuration options.

This PR is inspired by the fact that webpacker's message upon integrity check failure tells the user what command they should run to rectify the issue. Why not allow the user to set webpacker to perform the action automatically? 

Note: I'll be happy to update the documentation or tidy this up if there is interest in this feature.


gauravtiwari:
Thanks @ivar Could you please rebase the pr with master?

PR #2433, opened 2020-01-16
Update terser-webpack-plugin to 2.3.2

njradford:
Updates to 2.3.2 containing this change https://github.com/webpack-contrib/terser-webpack-plugin/pull/203

related to this open bug https://github.com/webpack-contrib/terser-webpack-plugin/issues/143#issuecomment-569087575


njradford:
a

PR #2434, opened 2020-01-16
Update testing docs to reflect change in options configuration.

mrsweaters:
It appears that `environment.plugins.get('Manifest').opts` has changed to `environment.plugins.get('Manifest').options`. This updates the testing docs to reflect this change.


gauravtiwari:
Thanks, this doc page needs to be updated though since we are no longer using manifest plugin but https://github.com/webdeveric/webpack-assets-manifest

Could you please update?

PR #2438, opened 2020-01-20
Update the HMR docs for React

justin808:
Taking over from https://github.com/rails/webpacker/pull/2353

### DIFF
--- a/docs/react.md
+++ b/docs/react.md
@@ -1,0 +1,134 @@
+# React
+
+If you need more advanced React-integration, like server-side rendering, SSR with react-router, SSR with code splitting, then you should consider these gems:
+
+| Gem | Props Hydration | Server-Side-Rendering (SSR) | SSR with HMR | SSR with React-Router | SSR with Code Splitting |
+| --- | --------------- | --- | --------------------- | ----------------------| ------------------------|
+| [shakacode/react_on_rails](https://github.com/shakacode/react_on_rails) | ✅ | ✅ | ✅ | ✅ | ✅ |
+| [react-rails](https://github.com/reactjs/react-rails)  | ✅ | ✅ |  | | | |
+| [webpacker-react](https://github.com/renchap/webpacker-react) | ✅ | | | | | |
+
+## HMR and React Hot Reloading
+
+Before turning HMR on, consider upgrading to latest stable gems and packages:
+https://github.com/rails/webpacker#upgrading
+
+First, check that the `hmr` option is `true` in your `config/webpacker.yml` file.
+
+```diff
+development:
+  # ...
+  dev_server:
+-    hmr: false
++    hmr: true
+# ...
+```
+
+The basic setup will have HMR working with the default webpacker setup. However, the basic will cause a full page refresh each time you save a file.
+
+Webpack's HMR allows replacement of modules in-place without reloading the browser.
+
+To do this next part, you have two options:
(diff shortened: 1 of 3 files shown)


justin808:
@gauravtiwari @jakeNiemiec any feedback?

clearyandzap:
@justin808 You're missing one step: 

```
{
    "plugins": [
        [
            "react-hot-loader/babel",
            {
            "safetyNet": false
            }
        ]
    ]
}
```

Without this, it doesn't work.

justin808:
@jakeNiemiec and @clearyandzap I'll get to your comments by this weekend.

@clearyandzap do you have any doc link for safetyNet item?

clearyandzap:
@justin808 Yup, https://github.com/gaearon/react-hot-loader#getting-started

justin808:
@clearyandzap docs there say:

> You **may** disable it to get more control on the module execution order.

Please clarify on why this is necessary with rails/webpacker.

clearyandzap:
@justin808 Oh, I'm sorry. The safetyNet , here https://github.com/gaearon/react-hot-loader#hotcomponent-options

csalvato:
I just used these instructions to set up my React app with HMR.

Why not just merge this as is? It can only help people.

justin808:
@jakeNiemiec @gauravtiwari This PR is ready per requested changes.

justin808:
@jakeNiemiec and @gauravtiwari I believe I've addressed all your comments.

justin808:
@jakeNiemiec @gauravtiwari Any reason that this cannot be merged? It's just a doc improvement. I took great care to get this ready.

docs/react.md
```diff
@@ -0,0 +1,184 @@
...
+    inline: true
+```
+
+The basic setup will have HMR working with the default webpacker setup. However, the basic will cause a full page refresh each time you save a file.
+
+Webpack's HMR allows replacement of modules in-place without reloading the browser.
+
+To do this next part, you have two options:
+
+1. Use the
```
rchiba, line 80: Is this line missing text?

justin808: Yes -- fixing now. 😄

rchiba:
For anyone who's using these docs / this thread to add HMR to their app, I had to do two things to get my Sass to be included as part of the HMR process:

1. In webpacker.yml, set `extract_css: false` for the development environment, which will cause styles to be compiled to a JavaScript module that webpack will insert as CSS into the page dynamically. [Additional documentation on extract_css.](https://rossta.net/blog/how-to-use-webpacker-yml.html#extract_css)
2. Doing step 1 caused my `stylesheet_pack_tag "application"` to throw an error in development as the styles are no longer being compiled into a separate stylesheet file, so I had to add a condition to prevent the error `stylesheet_pack_tag "application" if ENV["RAILS_ENV"] != "development"`

docs/react.md
rchiba: Thank you @justin808 for putting this documentation together, it was very helpful in getting HMR working for my app. 🙏

justin808:
@rchiba you should not need step 2.

https://github.com/rails/webpacker/blob/d456b1b3783313bf0b3927bed6f434d1de5854c4/lib/webpacker/helper.rb#L127-L147

justin808:
@rchiba see my last commit.

PR #2439, opened 2020-01-20
Move hot to the dev server runner

justin808:
This simplifies the need to add the plugin to the webpack config.

Anything we can do to simplify the webpack config is a bonus.

Docs:
https://webpack.js.org/configuration/dev-server/#devserverhot


justin808:
@gauravtiwari @jakeNiemiec I'm on the fence on this one. I'd say this one is optional. I'm not sure if there's any advantage in starting the webpack-dev-server with the `--hot` arg.

justin808:
Simplifies the package/environments/development.js. However, optional for sure.

However, let's close this if you don't want to do this.

justin808:
@gauravtiwari FWIW, 

https://webpack.js.org/plugins/mini-css-extract-plugin/#hot-module-reloading-hmr

```
You should not use HotModuleReplacementPlugin plugin if you are using a webpack-dev-server. webpack-dev-server enables / disables HMR using hot option.
```

Issue #2399, opened 2019-12-16
Outputting .br files in production even when not using Brotli


JoshRobertson:
after the support was added for Brotli in 4.1.0, I noticed our prod build started emitting both `.gz` and `.br` without anything changing in our setup or configs except updating Webpacker.

Rolling back to 4.0.7 predictably solves the issue.

Is there some way to explicitly disable this output?

njradford:
A config to disable .br output would be great.

njradford:
Also wondering if there's a way to disable this

arturopuente:
@JoshRobertson @njradford this works, but I'd love to see a more accesible config option exposed to disable brotli.

```javascript
// in production.js
const config = environment.toWebpackConfig();
// Disables brotli (.br) compression
config.plugins = config.plugins.filter(conf => {
  if (conf.options) {
    return conf.options.filename !== "[path].br[query]";
  } else {
    return conf;
  }
});
```

dreyks:
you can do it before `toWebpackConfig`
```js
environment.plugins.delete('Compression Brotli')
```

but I'd really wanna see plugins section be more configurable through webpacker.yml maybe

woodhull:
Yes, wish we could specify `compress: false` somewhere to skip both gzip and brotli compression, since we do this with our CDN.

guillaumebriday:
You can remove the plugin in your configuration. But why would you do that? 🤔

Cervenka:
If you want to skip both gzip and br compression:
```
config.plugins = config.plugins.filter(plugin => {
  return (plugin.constructor.name != 'CompressionPlugin')
})
```

PR #2442, opened 2020-01-24
Use zopfli for gzip compression

joankaradimov:
Zopfli is a compression by Google. It produces files compatible with gzip, but the compression time is much higher. That makes it ideal for asset pre-compilation. The idea (and as far as I know - the implementation too) is similar to Brotli, but with backwards compatibility to gzip and slightly lower compression ratios.

The implementation I suggest drops node's gzip and replaces it with [@gfx/zopfli](https://www.npmjs.com/package/@gfx/zopfli). This implementation of zopfli uses web assembly and is portable (so no native components and compilation). It runs on Node 8+. I could provide a gzip fallback for older node versions, but that doesn't seem necessary as [the requirements of webpacker cover that](https://github.com/rails/webpacker/blob/c7292e9/package.json#L11).

Here are size comparisons for files from a real-life project I work on:

<table>
  <tr>
    <th>Uncompressed</td>
    <th>GZip</td>
    <th>Zopfli</td>
    <th>Brotli</td>
  </tr>
  <tr>
    <td>890 KiB</td>
    <td>224 KiB</td>
    <td>216 KiB</td>
    <td>183 KiB</td>
  </tr>
  <tr>
    <td>1.74 MiB</td>
    <td>431 KiB</td>
    <td>416 KiB</td>
    <td>343 KiB</td>
  </tr>
  <tr>
    <td>716 KiB</td>
    <td>132 KiB</td>
    <td>126 KiB</td>
    <td>102 KiB</td>
  </tr>
</table>

Each line is a separate file. I've hidden the filenames as they are irrelevant. All files are **minified JavaScript**. I cannot provide size comparisons for CSS (or other formats) at the moment. But from previous experience I can confirm they are very similar.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "@babel/plugin-transform-runtime": "^7.6.2",
     "@babel/preset-env": "^7.7.1",
     "@babel/runtime": "^7.7.2",
+    "@gfx/zopfli": "1.0.14",
     "babel-loader": "^8.0.6",
     "babel-plugin-dynamic-import-node": "^2.3.0",
     "babel-plugin-macros": "^2.6.1",
--- a/package/environments/production.js
+++ b/package/environments/production.js
@@ -2,6 +2,7 @@ const TerserPlugin = require('terser-webpack-plugin')
 const CompressionPlugin = require('compression-webpack-plugin')
 const OptimizeCSSAssetsPlugin = require('optimize-css-assets-webpack-plugin')
 const safePostCssParser = require('postcss-safe-parser')
+const zopfli = require('@gfx/zopfli')
 const Base = require('./base')
 
 module.exports = class extends Base {
@@ -12,8 +13,9 @@ module.exports = class extends Base {
       'Compression',
       new CompressionPlugin({
         filename: '[path].gz[query]',
-        algorithm: 'gzip',
+        algorithm: zopfli.gzip,
         cache: true,
+        compressionOptions: { numiterations: 15 },
         test: /\.(js|css|html|json|ico|svg|eot|otf|ttf|map)$/
       })
     )
--- a/yarn.lock
+++ b/yarn.lock
@@ -700,6 +700,13 @@
   resolved "https://registry.yarnpkg.com/@csstools/convert-colors/-/convert-colors-1.4.0.tgz#ad495dc41b12e75d588c6db8b9834f08fa131eb7"
   integrity sha512-5a6wqoJV/xEdbRNKVo6I4hO3VjyDq//8q2f9I6PBAvMesJHFauXDorcNCsr9RzvsZnaWi5NYCcfyqP1QeFHFbw==
 
+"@gfx/zopfli@1.0.14":
+  version "1.0.14"
+  resolved "https://registry.yarnpkg.com/@gfx/zopfli/-/zopfli-1.0.14.tgz#52319e9bce0482f03ff02845e13fd95bb0ec7152"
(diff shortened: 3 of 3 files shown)


joankaradimov:
> with the sentiment dhh expresses here

These comments are irrelevant here. What ddh writes about is code executed in the browser. This is just a compile-time step. No webassemly will be bundled in the release file.

> I don't think it is wise to have web assembly as a default

Maybe my wording was wrong here. The packaging is webassembly, the implementation itself is in C (native code like any production-ready compression I am aware of). A better comparison would be the current gzip/brotli implementations in webpacker. They use native node modules (machine code) which is way more obfuscated.

> What do you think about documenting how a user would go about doing this?

I'd love to document this, but I'm not sure what you mean. A user wouldn't have to do anything. This PR makes zopfli the default gzip compression. From the users point of view, a updated webpacker will just produce smaller gzip files. Maybe I can drop a line in the release notes?

jakeNiemiec:
> These comments are irrelevant here...No webassemly will be bundled in the webpack output.

You are correct, I misunderstood was you were saying.

>  A better comparison would be the current gzip/brotli implementations that webpacker uses. They are native node modules (machine code) which is way more obfuscated.

The difference is that gzip compression libs have [10,000,000+](https://www.npmjs.com/package/pako) **weekly** downloads while [@gfx/zopfli](https://www.npmjs.com/package/@gfx/zopfli) has only 14k. This is not to say that `@gfx/zopfli`, should not be used; but, please understand the increased scrutiny.

<hr>

> This PR makes zopfli the default gzip compression. 

That package has a different way of enabling zopfli: https://github.com/webpack-contrib/compression-webpack-plugin/blob/2fa1ef6ff8deb9416ebe6be097f64ccd90dfafb3/README.md#using-zopfli

Also here with additional caveats: https://github.com/webpack-contrib/compression-webpack-plugin/commit/f6f8c6cf370460c2ad685d093343b8ee3404cc01#diff-1fdf421c05c1140f6d71444ea2b27638R17 

> From the users point of view, a updated webpacker will just produce smaller gzip files.

I was saying that it might be better to document how to enable stronger compression (that also takes longer), rather than forcing it on everyone.

joankaradimov:
> This is not to say that @gfx/zopfli, should not be used; but, please understand the increased scrutiny.

No, I understand. That's absolutely fair.

> That package has a different way of enabling zopfli: https://github.com/webpack-contrib/compression-webpack-plugin/blob/2fa1ef6ff8deb9416ebe6be097f64ccd90dfafb3/README.md#using-zopfli

Well, it is not identical, but it's pretty much the same code.

> Also here with additional caveats: webpack-contrib/compression-webpack-plugin@f6f8c6c#diff-1fdf421c05c1140f6d71444ea2b27638R17

That's specific code for using `node-zopfli` (a different binding to the zopfli library) and doesn't appear to be the recommended way of doing it - as can be seen in the link to the README you provided.

> I was saying that it might be better to document how to enable stronger compression (that also takes longer), rather than forcing it on everyone.

I assumed the opposite to be the preferred convention, as that is what happens in master. Webpacker uses Brotli, which is slow (maybe slower that zopfli) and is forced on everyone by default. Without a trivial way to disable.

There's appears to be [a bug](https://github.com/rails/webpacker/issues/2399) that requests a way to disable Brotli. Exposing a configuration for switching between gzip/zopfli would be pretty much equal to fixing the aforementioned bug.

Maybe I should fix the Brotli bug and only then we can continue discussing this?

Issue #2441, opened 2020-01-23
`webpacker:clean` is deleting files when they are outputted more than 2 seconds apart


sbeam:
In 4.2.2, the `webpacker:clean` task is grouping all files that are not mentioned in manifest.json by their last modified time, down to a one-second resolution. This is meant to find very old versions of assets files that can be safely deleted.

After upgrading and deploying to Heroku yesterday, noticed many lines in the build log where assets were being removed just after being compiled (including on brand new app deploys where /public/packs started out empty).

On heroku, and sometimes locally, these groups of files can easily span more than the default of 2. This task now removes all files except from the last 2 seconds.

Here is a example console session from a brand new heroku app where the `assets:clean` task was skipped, but mimicking the code that *would* have run if it were left alone:

```
irb(main):010:0> all_files       = Dir.glob("public/packs/**/*")
irb(main):011:0> manifest_config = Dir.glob("public/packs/manifest.json*")
irb(main):012:0> packs = all_files - manifest_config
irb(main):013:0> pp packs.group_by { |file| File.mtime(file).utc.to_i }.sort.reverse.map(&:first); nil
[1579728071, 1579728056, 1579728055, 1579728054]
```

webpack outputted files over a period of 17 seconds here, and this clean task would have deleted all the files from the first two groups, including the main .js entrypoints and a lot of other useful stuff.

I think this is due to changes in #2389. Possibly could decrease the resolution of that grouping to something more pessimistic, like 1 minute, or add a min_age param so that it ignores basically new files.

In the meantime have added `Rake::Task['assets:clean'].clear` to the app to avoid this.

gauravtiwari:
Could you please review this? https://github.com/rails/webpacker/pull/2443

sbeam:
I left a comment. To be honest I don't think the method of grouping my file mtime is a bulletproof one. Would probably be best to go back to filename matching like sprockets does, which is ugly but is the only reliable way. I can work something up but will take a few days.

justin808:
@gauravtiwari I just ran into this also! It's quite unexpected that the default uses just 2 different "second times" to group. As far as I know:

1. Webpack makes no guarantees to output ALL files within the same second.
2. If you use the multi-compile feature of webpack, you almost will certainly get build files outside of the same second.
3. If you don't hash some output files, like a bundle for SSR, then you can accidentally delete that bundle.
4. Some files created by webpack are not in the manifest, depending on the plugin, such as "loadable-components."

[...]

PR #2443, opened 2020-01-26
Improve webpacker:clear task

gauravtiwari:
Fixes: https://github.com/rails/webpacker/issues/2441

By default it will keep the latest version, 2 backups created within the past hour (same as sprockets). To force only 1 backup to be kept, set `count=1` and `age=0`. To only keep files created within the last 10 minutes, set `count=0` and `age=600`.

```bash
bundle exec rails "webpacker:clean[1, 0]"
bundle exec rails "webpacker:clean[0, 10]"
bundle exec rails webpacker:clean
```

### DIFF
--- a/lib/webpacker/commands.rb
+++ b/lib/webpacker/commands.rb
@@ -5,12 +5,34 @@ class Webpacker::Commands
     @webpacker = webpacker
   end
 
-  def clean(count = 2)
-    if config.public_output_path.exist? && config.public_manifest_path.exist? && versions.count > count
-      versions.drop(count).flat_map(&:last).each do |file|
-        File.delete(file) if File.file?(file)
-        logger.info "Removed #{file}"
-      end
+  # Cleanup old assets in the compile directory. By default it will
+  # keep the latest version, 2 backups created within the past hour.
+  #
+  # Examples
+  #
+  #   To force only 1 backup to be kept, set count=1 and age=0.
+  #
+  #   To only keep files created within the last 10 minutes, set count=0 and
+  #   age=600.
+  #
+  def clean(count = 2, age = 3600)
+    if config.public_output_path.exist? && config.public_manifest_path.exist?
+      versions
+        .sort
+        .reverse
+        .each_with_index
+        .drop_while do |(mtime, _), index|
+          max_age = [0, Time.now - Time.at(mtime)].max
+          max_age < age && index < count
+        end
+        .each do |(_, files), index|
+          files.each do |file|
+            if File.file?(file)
+              File.delete(file)
+              logger.info "Removed #{file}"
+            end
+          end
+        end
     end
 
     true
@@ -37,14 +59,16 @@ class Webpacker::Commands
       manifest_config = Dir.glob("#{config.public_manifest_path}*")
 
       packs = all_files - manifest_config - current_version
-      packs.group_by { |file| File.mtime(file).utc.to_i }.sort.reverse
+      packs.reject { |file| File.directory?(file) }.group_by { |file| File.mtime(file).utc.to_i }
     end
 
     def current_version
-      manifest.refresh.values.map do |value|
+      packs = manifest.refresh.values.map do |value|
         next if value.is_a?(Hash)
 
-        File.join(config.root_path, "public", value)
+        File.join(config.root_path, "public", "#{value}*")
       end.compact
+
+      Dir.glob(packs).uniq
     end
 end
--- a/lib/tasks/webpacker/clean.rake
+++ b/lib/tasks/webpacker/clean.rake
@@ -4,9 +4,9 @@ require "webpacker/configuration"
 
 namespace :webpacker do
   desc "Remove old compiled webpacks"
-  task :clean, [:keep] => ["webpacker:verify_install", :environment] do |_, args|
+  task :clean, [:keep, :age] => ["webpacker:verify_install", :environment] do |_, args|
     Webpacker.ensure_log_goes_to_stdout do
-      Webpacker.clean(Integer(args.keep || 2))
+      Webpacker.clean(Integer(args.keep || 2), Integer(args.age || 3600))
     end
   end
 end


lib/webpacker/commands.rb
```diff
@@ -37,14 +59,16 @@ def versions
       manifest_config = Dir.glob("#{config.public_manifest_path}*")
 
       packs = all_files - manifest_config - current_version
-      packs.group_by { |file| File.mtime(file).utc.to_i }.sort.reverse
+      packs.reject { |file| File.directory?(file) }.group_by { |file| File.mtime(file).utc.to_i }
```
sbeam: This still assumes that all the files for a given compilation run of webpack are going to be outputted within 1 second, and so can be grouped into distinct runs and counted that way. That's not a valid assumption. 

The max_age code helps, but this still could lead to confusing and incorrect behavior. If files are older than the `age` param, they could still get grouped incorrectly and deleted prematurely. 

Also, with default capistrano config or heroku buildpacks or docker deploys, users don't often have the option of (easily) changing the params given to `rake assets:clean`, or passing that param along to the `webpack:clean` task. So it's probably still important to make sure that the `count` method is reliable.

justin808:
@gauravtiwari how does one set the params that clean will use after precompilation is completed?

Also, would it make sense to allow whitelisting some files so that they are never cleaned?

In the case of server-side rendering, there is no need to put a hash on the bundle name. But I just ran into this exact issue a few times where the precompile calls the clean and that deletes the server bundle: https://github.com/shakacode/react_on_rails/issues/1245

I can advise users of React on Rails to always put a hash on the server bundle. 

Alternately, would you be open to a PR that allows some files to be whitelisted from cleaning, maybe in the webpacker.yml file? or an ENV value? That way, a file by the name of `server-bundle.js` would never get deleted.

lib/webpacker/commands.rb
```diff
@@ -5,12 +5,34 @@ def initialize(webpacker)
...
-      end
+  # Cleanup old assets in the compile directory. By default it will
+  # keep the latest version, 2 backups created within the past hour.
+  #
+  # Examples
+  #
+  #   To force only 1 backup to be kept, set count=1 and age=0.
+  #
+  #   To only keep files created within the last 10 minutes, set count=0 and
+  #   age=600.
```
justin808: @gauravtiwari if you really want to clean the directory, saving what Webpack created, why not just skip the whole `clean` and instead clear the destination directory before the build starts?

lib/webpacker/commands.rb
justin808: @sbeam see my comment above. How about cleaning before compilation, rather than after with all this logic.

sbeam: @justin808 that would work for some situations (including mine), but as I understand it, the point of keeping the last 2-3 versions of each compiled asses is in cases of rolling deploys or clients that have partially cached some of them, and there are still references to them out in the wild. So suddenly 404ing all old assets on each deploy isn't good. This is potentially an issue if you have e.g. capistrano deploys to multiple servers behind a naive load balancer that complete at different times.

That's why sprockets has [similar code](https://github.com/rails/sprockets/blob/master/lib/sprockets/manifest.rb#L235-L265) more or less for years.

sbeam: @juliantrueflynn if you do want to clean the dir entirely, I think you could use the [clean webpack plugin](https://github.com/johnagan/clean-webpack-plugin)

justin808: @sbeam I think for rails/webpacker, that can work. However, I generally prefer that the webpack config be solely focused on webpack builds and not things like deleting files from some directory. In other words, I find plain shell commands more descriptive. Another issue is that many React + React + SSR projects build the client and server bundles in the same directory. So some thought must be given the order of the webpack builds, if concurrency is used, and ensuring that the clean stuff is configured right.

lib/webpacker/commands.rb
somebody32: I was investigating this code today and I think those comments are misleading (or I'm mixing booleans in my head).

if we pass `age` as 0, then `max_age` will always be bigger, so the whole `max_age < age && index < count` will be false always and then `everything` will be removed independently of the value of `index`. 
In this case, `age` should be something like `Float::INFINITY`

the same story for the second line, you can't pass `count` as 0, should be infinity too

somebody32:
Maybe it is worth a separate issue but I also find relying on `mtime` to be brittle. Depends on how you structure your packs some of them could not be updated for a really long period of time and then the next deploy touching them could potentially have some period of time with 404s.

Sprockets + capistrano [are backuping the manifest file](https://github.com/capistrano/rails/blob/0e14a506b63b17b7c2faeae4d083f569b0aa4920/lib/capistrano/tasks/assets.rake#L75), maybe webpacker could do the same? And then remove everything that is not a union of the current manifest and the backed-up one.

viniciusgama:
I just run into issue @somebody32 pointed. The clean task did not retained the 2 latest versions of a file. It keep the one currently in use but not the previous one and that generated some 404s for us. That happened because like he said, that file was not modified in the last 2 deployments, but then when it was, it was not part of the recently modified files.

lib/webpacker/commands.rb
```diff
@@ -5,12 +5,34 @@ def initialize(webpacker)
...
   end
 
-  def clean(count = 2)
-    if config.public_output_path.exist? && config.public_manifest_path.exist? && versions.count > count
-      versions.drop(count).flat_map(&:last).each do |file|
-        File.delete(file) if File.file?(file)
-        logger.info "Removed #{file}"
-      end
+  # Cleanup old assets in the compile directory. By default it will
+  # keep the latest version, 2 backups created within the past hour.
```
benlangfeld, line 9: In what way is the age of the retained files important? This just causes asset availability downtime and doesn't include any explanation of why such a thing is desirable.

justin808: @gauravtiwari @sbeam, maybe we can use the versions so long as they are OLDER than the age?

Right now, as soon as we find a few files with timestamps that differ by a second, we delete everything that's not in the current manifest.

dkniffin:
Heads up for anyone else looking into this, there seems to be a fix [here](https://github.com/rails/webpacker/pull/2734), but it's currently not available on 5.X.X, only 6.X.X, which hasn't been released (in non-beta) yet.

Pezmc:
@dkniffin https://github.com/rails/webpacker/pull/2912

PR #2444, opened 2020-01-28
Update terser-webpack-plugin (reduce memory usage)

pustovalov:
https://github.com/webpack-contrib/terser-webpack-plugin/issues/143
https://github.com/webpack-contrib/terser-webpack-plugin/pull/211

https://github.com/webpack/webpack/pull/10305#issuecomment-579269226

>Reduce memory usage for big projects around 80-90%

From contributor of terser-webpack-plugin

Upd:
https://github.com/webpack-contrib/terser-webpack-plugin/blob/master/CHANGELOG.md

2.3.4 (2020-01-30)
Bug Fixes
```
respect stdout and stderr of workers and do not create warnings (5708574)
use webpack hash options rather than hard-code MD4(330c1f6)
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "regenerator-runtime": "^0.13.3",
     "sass-loader": "7.3.1",
     "style-loader": "^1.0.0",
-    "terser-webpack-plugin": "^2.3.3",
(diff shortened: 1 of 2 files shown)


njradford:
Thank you @pustovalov 👍

PR #2448, opened 2020-02-02
Move integrations to a separate doc file

justin808:
A separate file for the integrations will make the README.md much more concise.


gauravtiwari:
thanks both

Issue #2452, opened 2020-02-09
Add post-install messages, redo README.md and move documentation to separate file


ghost:
All in all this experience has been terribly confusing, especially for me who is a beginner to Rails and thought everything would be ready to go after `rails new myapp`.

Instead it turned out I had to downgrade Yarn. I also struggled to install basic packages like jQuery and had to resort to outdated and flat out wrong 3rd party tutorials out there on the web.

Is there any way to make this experience easier for average users? Perhaps add post-install messages like most other Rails gems do?

Also could you guys simplify the README.md by turning it into a getting started guide for beginners, and then move the actual documentation into a separate file and link to it from the README? It's currently way too much info there adding more confusion than clarity.

Thanks for listening!

justin808:
@jakeNiemiec, #2448 will help.

arjundedha:
https://speciallyshoes.com/best-basketball-shoes-for-ankle-support/

SWITCHAccess:
I wish I could tell you how much I understand but I'm very unsure murkmyself. Yes I absolutely agree that there has got to be an easier way. I completely agree. I wish I was not new and could offer my hand to help! 
Thanks for listening

Bramdimalone226:
> All in all this experience has been terribly confusing, especially for me who is a beginner to Rails and thought everything would be ready to go after `rails new myapp`.
> 
> Instead it turned out I had to downgrade Yarn. I also struggled to install basic packages like jQuery and had to resort to outdated and flat out wrong 3rd party tutorials out there on the web.
> 
> Is there any way to make this experience easier for average users? Perhaps add post-install messages like most other Rails gems do?
> 
> Also could you guys simplify the README.md by turning it into a getting started guide for beginners, and then move the actual documentation into a separate file and link to it from the README? It's currently way too much info there adding more confusion than clarity.
> 
> Thanks for listening!

Issue #2446, opened 2020-02-01
Why isn't .yarn/ in .gitignore?


ratahtatah:
Why isn't the `.yarn/` folder in `.gitignore`? Am I missing something obvious?

Thanks!

jakeNiemiec:
What version of yarn are you using? (`yarn --version`)

ratahtatah:
Hello  @jakeNiemiec,
```
% yarn --version
2.0.0-rc.27
```

Here's the content of my `.yarn/` folder: https://gist.github.com/ratahtatah/984afa9857f40251f53821b3155b5d0c
And my `node_modules/` folder: https://gist.github.com/ratahtatah/7201c886173cf926a4bbe483942c9680

Shouldn't these be in the root of my homedir instead of in my app's folder?

jakeNiemiec:
Yarn 2.0 will not work with webpacker yet, you will need yarn 1.x.

ratahtatah:
@jakeNiemiec oh I see. Thanks for the info. Any idea when this support will be implemented?

jakeNiemiec:
@ratahtatah thanks for getting back to us about it. This should produce a warning now.

Issue #2354, opened 2019-11-05
`yarn check` deprecated


albus522:
`yarn check` used here https://github.com/rails/webpacker/blob/master/lib/webpacker/railtie.rb#L33 is deprecated https://github.com/yarnpkg/rfcs/pull/106

We ran into an issue in our project where `yarn install` worked fine but `yarn check` was reporting multiple errors. Searching for a solution lead here https://github.com/yarnpkg/yarn/issues/6427#issuecomment-434499073 which lead to the merged deprecation.

paulodeon:
I'd like to add that this has led to several wasted hours figuring this out - its especially problematic as that check command gets run on every rails command - it has even broken deploys for us

rossta:
I suggest turning it off in your `config/webpacker.yml` if it's causing pain: 
```yaml
check_yarn_integrity: false
```

challeng:
`yarn check` is also called out as buggy(https://classic.yarnpkg.com/en/docs/cli/check/) in addition to being deprecated and removed in Yarn 2.0 which has had its first stable release. I found another issue in the repo too which calls out that this will need to change soon https://github.com/rails/webpacker/issues/2007. 

Is there an alternative for verifying integrity? The best alternative I know about is `yarn install --check-files`, but adding that would add the install to this step. I thought about opening a PR to do that, but I wasn't sure if that particular change would go over well

sfcgeorge:
It's constantly causing issues for us now so I tried `check_yarn_integrity: false` in webpacker.yml (and restarted Spring) but it had no effect. Really slowing down development.

PR #2454, opened 2020-02-10
Fix package.json engine semver ranges (node.js >=8.9 <13, yarn >=1 <2)

jbielick:
8.16 is somewhat arbitrary as far as I can tell, so this allows earlier 8.x releases of node.js as the engine so that 8.9, 8.10, 8.12, etc from Ubuntu 18.04 LTS can install this package.

See #2228 comments for interest.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   ],
   "engines": {
     "node": ">=8.9.0",
-    "yarn": ">=1.0.0"
+    "yarn": ">=1 <2"
   },
   "dependencies": {
     "@babel/core": "^7.7.2",
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "lib/install/config/webpacker.yml"
   ],
   "engines": {
-    "node": ">=8.16.0",
+    "node": ">=8.9.0",
     "yarn": ">=1.0.0"
   },
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "lib/install/config/webpacker.yml"
   ],
   "engines": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -8,7 +8,7 @@
     "lib/install/config/webpacker.yml"
   ],
   "engines": {
-    "node": ">=8.16.0",
+    "node": ">=8.9.0",
```
jakeNiemiec:
```suggestion
    "node": ">=8.9 <13",
```

package.json
```diff
@@ -8,7 +8,7 @@
...
     "yarn": ">=1.0.0"
```
jakeNiemiec:
```suggestion
    "yarn": ">=1 <2"
```

jbielick:
@jakeNiemiec Great idea—updated. Thanks for the additional references and suggestions.

jakeNiemiec:
Great, this will also fix things like: 

- https://github.com/rails/webpacker/issues/2452#issue-562149280
- https://github.com/rails/rails/issues/38414#issuecomment-583774707
- https://github.com/rails/webpacker/issues/2354#issuecomment-583517707
- https://github.com/rails/webpacker/issues/1537#issuecomment-539119971
- https://github.com/rails/webpacker/issues/2112#issuecomment-583679919
- https://github.com/rails/webpacker/issues/2446#issuecomment-583822119

jakeNiemiec:
> Unless the user has set the engine-strict config flag, this field is **advisory only and will only produce warnings** when your package is installed as a dependency.

Something to keep in mind, but this sounds exactly like what we need.

PR #2455, opened 2020-02-10
Allow other Webpacker rake tasks to be conditionally skipped

mctaylorpants:
Passing the `WEBPACKER_PRECOMPILE` env variable allows Webpacker-specific tasks to be skipped when running `rake assets:precompile`. I'd like to propose the same option be supported for the other tasks that Webpacker enhances (`clean` and `clobber`), for consistency.

Given that `precompile` can be skipped, I think it makes sense to allow users to skip the other asset management tasks, given that they're often used in concert. For example, it's common to run `rake assets:clean assets:precompile` during a build step.

### DIFF
--- a/lib/tasks/webpacker/clean.rake
+++ b/lib/tasks/webpacker/clean.rake
@@ -11,11 +11,15 @@ namespace :webpacker do
   end
 end
 
-# Run clean if the assets:clean is run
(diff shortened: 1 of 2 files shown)


gauravtiwari:
thanks

PR #2464, opened 2020-02-15
Use headers for file names, not code comments

justin808:
* Updated so it's more clear when different files are used for an example
* Clarify that the customization file is based on the NODE_ENV

### DIFF
--- a/docs/webpack.md
+++ b/docs/webpack.md
@@ -6,7 +6,9 @@
 Webpacker gives you a default set of configuration files for test, development and
 production environments in `config/webpack/*.js`. You can configure each individual
 environment in their respective files or configure them all in the base
-`config/webpack/environment.js` file.
+`config/webpack/environment.js` file. 
+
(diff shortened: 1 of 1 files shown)


jakeNiemiec:
> Updated so it's more clear when different files are used for an example

I personally like the file comments since they are commonly used like that on GH. (but I don't want to nitpick)

docs/webpack.md
```diff
@@ -155,9 +159,8 @@ fileLoader.exclude = /\.(svg)$/i
 
 ### Url Loader
 
+#### config/webpack/loaders/url.js
 ```js
-// config/webpack/loaders/url.js
```
jakeNiemiec: This might be a mistake, but I may have just never used webpacker like this.

justin808: I don't understand this comment. The PR adds headers.

justin808:
@gauravtiwari any comments?

justin808:
Closing in favor of #2748.

PR #2486, opened 2020-03-11
Add possibility to add version to filenames

ChrisBr:
With sprockets / asset pipeline there was a version / prefix config which can be helpful when the asset content hasn't changed, but you want to change the response headers served with the asset, but those headers are cached somewhere like a CDN or user's browser.

This is a very naive approach to introduce a version into the config as it always adds the version even if it does not exist. Please let me know if this feature would be accepted and I'm happy to adapt.

### DIFF
--- a/lib/install/config/webpacker.yml
+++ b/lib/install/config/webpacker.yml
@@ -8,6 +8,7 @@ default: &default
   cache_path: tmp/cache/webpacker
   check_yarn_integrity: false
   webpack_compile_output: true
+  version: 1.0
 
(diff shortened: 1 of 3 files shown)


guillaumebriday:
Why do you want to invalidate the cache if the content hasn't changed? 🤔

ChrisBr:
> Why do you want to invalidate the cache if the content hasn't changed? 🤔

For instance to adapt caching TTL

guillaumebriday:
Thanks for the response !

PR #2415, opened 2019-12-25
EOL for Rails < 5.2 and Ruby < 2.3.0

pustovalov:
### Ruby
https://endoflife.software/programming-languages/server-side-scripting/ruby

Ruby 2.2 
Release date
December 25, 2014

End of life
March 31, 2018

Ruby 2.3
December 25, 2015 - March 31, 2019

Ruby 2.4
December 25, 2016 - March 31, 2020

### Rails
Rails 6 
Release date
Aug 16, 2019
https://github.com/rails/rails/releases/tag/v6.0.0

https://guides.rubyonrails.org/maintenance_policy.html

```
For severe security issues all releases in the current major series, and also the last release in the previous major series will receive patches and new versions. The classification of the security issue is judged by the core team.
Currently included series: 6.0.Z, 5.2.Z.
```

PR #2494, opened 2020-03-16
Travis CI test against ruby 2.7

kimyu92:
- run travis CI against ruby 2.7 and the latest minor version for old ruby releases


kimyu92:
similar to #2415

PR #2499, opened 2020-03-22
Fix wrong indent in README

tricknotes:
These pages are not for "Assets".


gauravtiwari:
thanks @tricknotes looks like this shouldn't be even here but at the top TOC under React.

tricknotes:
Thanks for review, @gauravtiwari 
It seems these pages are not only for React but also other libraries like Vue or Elm .
- https://github.com/rails/webpacker/blob/v5.0.1/docs/typescript.md
- https://github.com/rails/webpacker/blob/master/docs/props.md

Should we do so?

Issue #1951, opened 2019-02-19
Nested dependancies not loaded


iamdriz:
When loading a dependancy that's external you prefix it with a tilde e.g. `@import "~external-package/external-file";`

However if that external package also has external dependancies you will get Module build failed errors when compiling the pack.

To fix this you have to manually go into the package and prefix all additional dependancies with the tilde so that they can be found and imported... which is obviously not a solution.

For example I might have: `@import "~@material/typography/mdc-typography";`

And then I've had to go into `node_modules/@material/typography/_mixins.scss` and do this:
```
@import "~@material/feature-targeting/functions";
@import "~@material/feature-targeting/mixins";
```

Apparently this was an issue with the sass-loader and scoped modules/dependancies: https://github.com/material-components/material-components-web/issues/981 and https://github.com/webpack-contrib/sass-loader/issues/466

They mention in the ticket to add includePaths to sass-loader to load the modules:
~~~
{
                    loader: "sass-loader", // compiles Sass to CSS
                    options: {
                      includePaths: [
                        join(dirname(module.filename), 'node_modules')
                      ]
                    }
                }
~~~

 which in Webpacker I'm assuming is the same as adding `resolved_paths: ['node_modules']` to `webpacker.yml` so that Webpacker knows where to look...

 But this does not fix the issue without doing the manual hack job above to prefix everything...

Does Webpacker have a solution for this?

Quintasan:
I just ran into this today. Is there any solution apart from monkeypatching the packages?

jakeNiemiec:
Just import the bits that your SCSS file use, eg:
![image](https://user-images.githubusercontent.com/16782063/55496261-42acf580-5604-11e9-8f54-b73a96436677.png)

The solution from @iamdriz will slow down your builds.

SCSS just can't tree-shake like JS can (yet?) because the styles can't know which classes will be present in the DOM.

Quintasan:
@jakeNiemiec How is this related to tree-shaking? I think that @iamdriz correctly identified the issue of webpacker not being configured correctly to allow loading nested modules.

jakeNiemiec:
@Quintasan The structure of `@material/feature-targeting` is very different from `bootstrap`. **It seems like the consensus is to use`includePaths` as @iamdriz said.** There seems to be a bunch of users having this problem even within the `material-components/material-components-web` repo. [ref](https://github.com/material-components/material-components-web/issues/3985) [ref](https://github.com/material-components/material-components-web/issues/351)

[...]

PR #2500, opened 2020-03-23
Fix sass option and nix additional yarn install task

gauravtiwari:
Please see: https://webpack.js.org/loaders/sass-loader/#sassoptions

and since we are supporting Rails > 5.2 we can safely remove webpacker yarn install enhance task.


package/rules/module.sass.js
```diff
@@ -1,8 +1,14 @@
...
 
 module.exports = getStyleRule(/\.(scss|sass)$/i, true, [
   {
     loader: 'sass-loader',
-    options: { sourceMap: true }
+    options: {
+      sourceMap: true,
+      sassOptions: {
+        includePaths
+      }
```
aried3r: I believe this means that we should go for sass-loader ^8.0.0, WDYT?

https://github.com/webpack-contrib/sass-loader/blob/master/CHANGELOG.md#-breaking-changes

Previously:
https://github.com/webpack-contrib/sass-loader/tree/v7.3.1#options
Current:
https://github.com/webpack-contrib/sass-loader/tree/v8.0.2#object

aried3r:
Hey! Thanks for the PR!

Does this solve https://github.com/rails/webpacker/issues/1951 by any chance? As in, does `includePaths` include `.node_modules`?

package/rules/module.sass.js
gauravtiwari: Ahh I see. Yeah, lets do that. 

I am thinking to try and upgrade master to webpack 5 beta too and see if everything works as intended but in another pr.

gauravtiwari:
Hey, I am not sure really, will work on this pr bit later today and update 👍

gauravtiwari:
closing in favour of #2508 

Will open another pr with remaining changes.

PR #2508, opened 2020-03-24
Update dependencies

aried3r:
Until issues like https://github.com/rails/webpacker/issues/2191#issuecomment-539140915 are resolved (users of webpacker cannot update dependencies of webpacker itself) we need to regularly update the dependencies so users have the newest versions.

Changelogs can be found in the individual commit messages.

I left out sass-loader for now as there is #2501 and #2500.


aried3r:
I've cherry-picked the changes by @schorsch from #2501 since it contained exactly the parts I needed to update sass-loader 8. #2500 by @gauravtiwari contains one commit but also contains other changes.

Friendly ping @jakeNiemiec.

gauravtiwari:
Thanks @aried3r looks good 👍

Issue #1382, opened 2018-03-27
On Rails 4.2 webpacker erb extension raises an error


mdesantis:
Hello,

I tried with a new Rails 4.2.10 webapp (Ruby 2.3.6, webpacker 3.4.1). This is the error (running `bin/webpack`):

```
> bin/webpack
Running via Spring preloader in process 2848
(erb):1:in `<main>': undefined method `javascript_pack_tag' for main:Object (NoMethodError)
	from /home/user/.asdf/installs/ruby/2.3.6/lib/ruby/2.3.0/erb.rb:864:in `eval'
	from /home/user/.asdf/installs/ruby/2.3.6/lib/ruby/2.3.0/erb.rb:864:in `result'
	from /home/user/Sviluppo/rails-4.2.7.1/rails_demo/node_modules/rails-erb-loader/erb_transformer.rb:19:in `<top (required)>'
	from /home/user/.asdf/installs/ruby/2.3.6/lib/ruby/gems/2.3.0/gems/railties-4.2.10/lib/rails/commands/runner.rb:60:in `load'
	from /home/user/.asdf/installs/ruby/2.3.6/lib/ruby/gems/2.3.0/gems/railties-4.2.10/lib/rails/commands/runner.rb:60:in `<top (required)>'
	from /home/user/.asdf/installs/ruby/2.3.6/lib/ruby/gems/2.3.0/gems/activesupport-4.2.10/lib/active_support/dependencies.rb:274:in `require'
	from /home/user/.asdf/installs/ruby/2.3.6/lib/ruby/gems/2.3.0/gems/activesupport-4.2.10/lib/active_support/dependencies.rb:274:in `block in require'
	from /home/user/.asdf/installs/ruby/2.3.6/lib/ruby/gems/2.3.0/gems/activesupport-4.2.10/lib/active_support/dependencies.rb:240:in `load_dependency'
	from /home/user/.asdf/installs/ruby/2.3.6/lib/ruby/gems/2.3.0/gems/activesupport-4.2.10/lib/active_support/dependencies.rb:274:in `require'
	from /home/user/.asdf/installs/ruby/2.3.6/lib/ruby/gems/2.3.0/gems/railties-4.2.10/lib/rails/commands/commands_tasks.rb:123:in `require_command!'
	from /home/user/.asdf/installs/ruby/2.3.6/lib/ruby/gems/2.3.0/gems/railties-4.2.10/lib/rails/commands/commands_tasks.rb:90:in `runner'
[...]

PR #2509, opened 2020-03-24
fixing typo

VRRmax:
comments in .erb is with <%# ... %>


jakeNiemiec:
See: https://github.com/rails/webpacker/pull/2263

VRRmax:
I don't understand why you use **<%%=** instead of **<%#=** which is the standard way for ERB commenting. That's the point of my PR, which is different than the one you pointed to me.
**<%%=** looks like a typo, and you will have PR again and again, people thinking you meant '<%=' by '<%%='.

Please tell me if I have a misconception somewhere.

VRRmax:
https://github.com/rails/webpacker/issues/1382#issuecomment-604475039

olleolleolle:
The `%%` is meant to be there, so that the output can be visible to the user reading it as `%`.

The output text is a JS comment to be read by the developer learning from the example.

Hope this helps explain why the `%%` is not a typo.

PR #2511, opened 2020-03-25
avoid asset_path route helper collisions

granth:
If a Rails application has an `assets` route resource, than the generated `asset_path` route helper will override the asset helper of the same name.

Rails provides the aliases `path_to_asset` and `url_to_asset` to avoid this, and uses them internally. We can do the same here.

### DIFF
--- a/lib/webpacker/helper.rb
+++ b/lib/webpacker/helper.rb
@@ -7,8 +7,8 @@ module Webpacker::Helper
   end
 
   # Computes the relative path for a given Webpacker asset.
-  # Returns the relative path using manifest.json and passes it to asset_path helper.
-  # This will use asset_path internally, so most of their behaviors will be the same.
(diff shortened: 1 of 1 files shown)


lib/webpacker/helper.rb
```diff
@@ -7,8 +7,8 @@ def current_webpacker_instance
   end
 
   # Computes the relative path for a given Webpacker asset.
-  # Returns the relative path using manifest.json and passes it to asset_path helper.
-  # This will use asset_path internally, so most of their behaviors will be the same.
+  # Returns the relative path using manifest.json and passes it to path_to_asset helper.
+  # This will use path_to_asset internally, so most of their behaviors will be the same.
```
granth: I dithered on whether to rename the method in the comments, too, since `path_to_asset` is probably not as well known as `asset_path`. I thought it was better to be consistent.

gauravtiwari:
thanks @granth

PR #2513, opened 2020-03-26
DRAFT: Document which yaml config used by Ruby vs. JS

justin808:
Some of the yaml settings matter only for the Ruby parts of
rails/webpacker. Others are needed for the JS parts of building the
bundles via rails/webpacker.

If one's app is using a custom webpack build, then the non-ruby settings
can be ignored. On the other hand, the ruby settings are critical to all
Rails apps.

### DIFF
--- a/lib/install/config/webpacker.yml
+++ b/lib/install/config/webpacker.yml
@@ -1,7 +1,7 @@
 # Note: You must restart bin/webpack-dev-server for changes to take effect
 
 default: &default
-  # Values used by Ruby part of rails/webpacker
+  # Values used by Ruby part of rails/webpacker as well as the webpack configuration
   source_path: app/javascript
   public_root_path: public
   public_output_path: packs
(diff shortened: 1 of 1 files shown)


lib/install/config/webpacker.yml
```diff
@@ -1,21 +1,35 @@
 # Note: You must restart bin/webpack-dev-server for changes to take effect
 
 default: &default
+  # Values used by Ruby part of rails/webpacker
```
rossta, line 4: I was confused by this comment; as I understand it, `source_path`, `public_root_path`, `public_output_path`, and `cache_path` (listed below)) are also consumed by the `@rails/webpacker` package as part of the default webpack config.

justin808: Good point. I clarified that. I wanted to point out the subset of values that are used by the Ruby on Rails code.

justin808:
I can revisit this at a later date.

PR #2515, opened 2020-03-28
Fix incorrect host forwarding

olzv:
Hi!

`HTTP_X_FORWARDED_HOST` should not contain port because otherwise `Rack::Request` will add a default port causing an error like on the screenshot:
Screenshot from 2020-03-28 01-52-06

So `HTTP_X_FORWARDED_PORT` should be also sent in order for `Rack::Request` to properly see it..
And casted to String explicitly to satisfy `Net::HTTPHeader`.

Useful when running webpack dev server as a separate service in Docker Compose like here for example:
https://github.com/olzv/rails-vue-template/pull/3/files

### DIFF
--- a/lib/webpacker/dev_server_proxy.rb
+++ b/lib/webpacker/dev_server_proxy.rb
@@ -11,7 +11,9 @@ class Webpacker::DevServerProxy < Rack::Proxy
 
   def perform_request(env)
     if env["PATH_INFO"].start_with?("/#{public_output_uri_path}") && dev_server.running?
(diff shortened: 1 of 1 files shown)


vovimayhem:
I've actually noticed this an hour ago... this should be an issue to be fixed!

vovimayhem:
Looks like this is caused by a change introduced in rack v2.2.2:
https://github.com/rack/rack/commit/290523f67cc43c5847b2be2d12964d1232061fe1

In my case, the app updated to this version when I updated rails to version 6.0.2.2

PR #2529, opened 2020-04-11
Remove nodejs buildpack from Heroku deploy doc

dleavitt:
I don't think you've had to explicitly add the `heroku/nodejs` buildpack for a while now?
https://github.com/heroku/heroku-buildpack-ruby/pull/608/files
Or are there use-cases where it's still need?

Also not sure if it's still necessary to add the `heroku/ruby` buildpack since that will be inferred.


gauravtiwari:
@dleavitt Thank you but I think it depends, if you need more control like set node version than separate node buildpack is needed, which I think would be very common use case. Please see: https://devcenter.heroku.com/articles/ruby-support#installed-binaries

PR #2538, opened 2020-04-19
Update dotenv example in env.md

rossta:
Removes use of `EnvironmentPlugin` from the dotenv example in the docs.

The base webpack config already adds the `EnvironmentPlugin` with JSON-ified `process.env`: https://github.com/rails/webpacker/blob/a84a4bb6b385ae17dd775a6034a0b159b88c6ea9/package/environments/base.js#L31-L34

### DIFF
--- a/docs/env.md
+++ b/docs/env.md
@@ -49,8 +49,6 @@ dotenvFiles.forEach((dotenvFile) => {
   dotenv.config({ path: dotenvFile, silent: true })
 })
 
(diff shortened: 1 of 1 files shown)


np6176a:
I'm not sure why this was merged. This feature doesn't work with this line removed.

rromanchuk:
@np6176a  are you on 6 or 5?  i can't access `process.env` except for in config. No idea what i'm doing lol

PR #2548, opened 2020-04-21
Update README for optional TS type checking

### LINKED ISSUE #2547: How to do TypeScript type checking in 5.1+?
Hello,

I have a TypeScript project and recently [followed the upgrade instructions](https://github.com/rails/webpacker/blob/master/docs/typescript.md) since TS compilation is now [babel instead of ts-loader](https://github.com/rails/webpacker/pull/2449). 

However, I was also using `ts-loader` for type checking during the webpack compilation. In the previous setup, the compiler wouldn't succeed if there was a TypeScript error.

I am wondering how can I get that functionality with the new setup?

Thank you

hueter:
closes #2547

### DIFF
--- a/docs/typescript.md
+++ b/docs/typescript.md
@@ -10,6 +10,33 @@ bundle exec rails webpacker:install:typescript
 
 After that, a new file called `hello_typescript.ts` will be present in your `packs` directory (or rather the `source_entry_path` of your `webpacker.yml` configuration). You're now ready to write TypeScript. 
 
+## (Optional) Adding Compile-Time Type Checking
+
+The default installation only transpiles your TypeScript code using Babel. If you would like to enable type checking as part of the Webpack compilation process (i.e. fail the build if there are TS errors), you can do the following:
+
+1. Install the Fork TS Checker Webpack Plugin
+
+```sh
+yarn install -D fork-ts-checker-webpack-plugin
+```
+
+2. Then add it to your development environment config in `config/webpack/development.js`
+
+```js
+const ForkTsCheckerWebpackPlugin = require("fork-ts-checker-webpack-plugin");
+const path = require("path");
+
+environment.plugins.append(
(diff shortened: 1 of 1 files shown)


aried3r:
Hey @hueter, thanks for the guide! I was wondering, I switched to the lastest webpacker and followed your guide, however I was facing a lot of type errors.

I found https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/issues/128 and added `    "skipLibCheck": true` to `tsconfig.js`. Did you also have to do that in your setup? It seems that this is [`true` in `ts-loader`](https://github.com/TypeStrong/ts-loader/blob/bbc6d8161b758893f6825ccc6bb088272ba210d2/src/compilerSetup.ts#L60). 

Maybe we should add it to the guide as well, WDYT?

hueter:
hi @aried3r no problem, glad it was useful!

Full context (forgive me if you know all this already, feel free to skip this part 😉 ): the `skipLibCheck` is actually just a TypeScript compiler setting that varies per project -- I don't have it enabled on mine. It checks that the npm packages from `node_modules` that you're importing don't have type errors, which can usually be avoided by installing the [DefinitelyTyped (@types/packagename)](https://microsoft.github.io/TypeSearch/); in some cases by manually adding type declarations; or in rare cases by downgrading/upgrading packages if they have type errors in certain releases (currently I get type errors from GraphQL 15 so I haven't upgraded that from 14.6 for instance).  

Interesting that `skipLibCheck` was enabled in `ts-loader` settings previously; I'm not sure why they did that. But in regards to your question, I think it's a personal preference like the other TS configs like `strict: true`, etc. My PR was just concerned with adding the ability to do type checking _at all_ with the new setup, not necessarily to prescribe what should be type checked. That being said, you do have a point that it was the rule before, so it might be helpful context to note in the "migrating" section of the docs if you want to add it! 🙂

Issue #2530, opened 2020-04-12
Webpacker cache_path


pustovalov:
Hey,

By default webpacker use this path to save cache:
`tmp/cache/webpacker`
https://github.com/rails/webpacker/blob/master/lib/install/config/webpacker.yml#L8

the cache from the `babel-loader` gets  into this folder:
https://github.com/rails/webpacker/blob/master/package/rules/node_modules.js#L17
https://github.com/rails/webpacker/blob/master/package/rules/babel.js#L15

But most js libraries use `node_modules/.cache`:

`babel-loader`:
https://github.com/babel/babel-loader#options
```
the loader will use the default cache directory in node_modules/.cache/babel-loader 
or fallback to the default OS temporary file directory if no node_modules 
folder could be found in any root directory.
```

`terser-webpack-plugin`
https://github.com/webpack-contrib/terser-webpack-plugin#cache
```
Enable file caching. Default path to cache directory: 
node_modules/.cache/terser-webpack-plugin.
```

`compression-webpack-plugin`
https://github.com/webpack-contrib/compression-webpack-plugin#cache

```
Enable file caching. The default path to cache directory: 
node_modules/.cache/compression-webpack-plugin.
```

`Webpack 5`
https://webpack.js.org/configuration/other-options/#cachecachedirectory
```
Base directory for the cache. Defaults to node_modules/.cache/webpack.
```

I saw the comment from @jakeNiemiec 
https://github.com/rails/webpacker/issues/2178#issuecomment-511898515

`cache_path` can be changed to `node_modules/.cache`

Is there any reason why `cache_path` has `tmp/cache/webpacker` value by default?

jhirn:
This speed up my build times in Circle from 5 minutes down to 2 minutes because I wasn't caching the tmp directory. It's a quick win.

PR #2556, opened 2020-04-25
Use different path for babel-loader cache by default

pustovalov:
Hey,

relative to this issue: https://github.com/rails/webpacker/issues/2530

### DIFF
--- a/lib/webpacker/configuration.rb
+++ b/lib/webpacker/configuration.rb
@@ -59,6 +59,10 @@ class Webpacker::Configuration
     root_path.join(fetch(:cache_path))
   end
 
+  def node_modules_cache_path
+    root_path.join("node_modules/.cache/webpacker")
+  end
+
   def extensions
     fetch(:extensions)
   end
--- a/lib/webpacker/configuration.rb
+++ b/lib/webpacker/configuration.rb
@@ -59,10 +59,6 @@ class Webpacker::Configuration
     root_path.join(fetch(:cache_path))
   end
 
-  def node_modules_cache_path
-    root_path.join("node_modules/.cache/webpacker")
-  end
-
   def extensions
     fetch(:extensions)
   end
--- a/package/config.js
+++ b/package/config.js
@@ -36,6 +36,5 @@ const getPublicPath = () => {
 
 config.publicPath = getPublicPath()
 config.publicPathWithoutCDN = `/${config.public_output_path}/`
-config.cachePath = 'node_modules/.cache/webpacker'
 
 module.exports = config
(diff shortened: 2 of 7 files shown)


gauravtiwari:
@pustovalov Thanks for this but I am wondering if we remove all cache from tmp and move it to `node_modules/.cache` by default except compiler hash that can live in tmp. We can hardcode these paths since they are pretty standard and remove config option from webpacker.yml

pustovalov:
@gauravtiwari updated, WDYT?

gauravtiwari:
Looks great, thank you for updating. One minor thing, could you please put path in one place and use `config.cachePath` so we are not hardcoding at several places but instead using variable like before?

pustovalov:
Friendly ping @gauravtiwari

package/config.js
```diff
@@ -36,5 +36,6 @@ const getPublicPath = () => {
 
 config.publicPath = getPublicPath()
 config.publicPathWithoutCDN = `/${config.public_output_path}/`
+config.cachePath = 'node_modules/.cache/webpacker'
```
gauravtiwari, line 39: Thanks for the changes but I think this is handled internally if we set: `{cacheDirectory: true}` so no need to set this for loaders

https://webpack.js.org/loaders/babel-loader/#options

lib/webpacker/configuration.rb
```diff
@@ -59,6 +59,10 @@ def cache_path
     root_path.join(fetch(:cache_path))
   end
 
+  def node_modules_cache_path
```
gauravtiwari, line 62: We can remove this one?

gauravtiwari:
Sorry @pustovalov I didn't notice this before but I think we can simplify this even further by just setting cacheDirectory option tot true and let the loader do the job. I hope that's okay?

pustovalov:
@gauravtiwari updated, WDYT?

gauravtiwari:
Great, thank you 🙏 . Will merge once passes

pustovalov:
it will failed like in master:
https://travis-ci.org/github/rails/webpacker/jobs/692947115

master specs:
https://travis-ci.org/github/rails/webpacker/jobs/692943219

```
ruby-head - #gemset created /home/travis/.rvm/gems/ruby-head@global
ruby-head - #importing gemset /home/travis/.rvm/gemsets/global.gems.............................
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received
The build has been terminated
```

PR #2560, opened 2020-04-27
Merge resolved_paths and watched_paths into additional_paths

CGA1123:
Merges the configuration of resolved_paths and watched_paths together
into additional_paths, while keeping backwards compatibility for both.

watched_paths and resolved_paths are deprecated in favour of
additional_paths.

### DIFF
--- a/lib/webpacker/configuration.rb
+++ b/lib/webpacker/configuration.rb
@@ -27,12 +27,12 @@ class Webpacker::Configuration
     globbed_path_with_extensions(source_path.relative_path_from(root_path))
   end
 
-  def resolved_paths
-    fetch(:resolved_paths)
+  def additional_paths
+    fetch(:additional_paths) + resolved_paths
   end
 
-  def resolved_paths_globbed
-    resolved_paths.map { |p| globbed_path_with_extensions(p) }
+  def additional_paths_globbed
+    additional_paths.map { |p| globbed_path_with_extensions(p) }
   end
 
   def source_entry_path
@@ -76,6 +76,14 @@ class Webpacker::Configuration
   end
 
   private
+    def resolved_paths
+      paths = data.fetch(:resolved_paths, [])
+
+      warn "The resolved_paths option has been deprecated. Use additional_paths instead." unless paths.empty?
+
+      paths
+    end
+
     def fetch(key)
       data.fetch(key, defaults[key])
     end
--- a/lib/webpacker/configuration.rb
+++ b/lib/webpacker/configuration.rb
@@ -27,14 +27,6 @@ class Webpacker::Configuration
     globbed_path_with_extensions(source_path.relative_path_from(root_path))
   end
 
-  def watched_paths
-    fetch(:watched_paths)
-  end
-
-  def watched_paths_globbed
-    watched_paths.map { |p| globbed_path_with_extensions(p) }
-  end
-
   def resolved_paths
     fetch(:resolved_paths)
   end
--- a/lib/webpacker/configuration.rb
+++ b/lib/webpacker/configuration.rb
@@ -27,6 +27,14 @@ class Webpacker::Configuration
     globbed_path_with_extensions(source_path.relative_path_from(root_path))
   end
 
+  def watched_paths
+    fetch(:watched_paths)
+  end
+
+  def watched_paths_globbed
+    watched_paths.map { |p| globbed_path_with_extensions(p) }
+  end
+
   def resolved_paths
     fetch(:resolved_paths)
   end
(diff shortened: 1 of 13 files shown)


gauravtiwari:
@CGA1123 Thanks for working on this. 

I have been thinking to actually consolidate `watched` and `resolved` paths since they are pretty much same and server similar purpose. Perhaps, we can add `additional_paths` that covers both and remove both `watched` and `resolved` option from the settings?

CGA1123:
@jakeNiemiec `bower_components` is the example used in the documentation https://github.com/rails/webpacker/blob/740b686dd9ba5e97dbc679aa94e4091c0ca8415c/lib/webpacker/compiler.rb#L6

@gauravtiwari That sounds good 😄Should this be done in a backwards-compatible way? Or is it okay to fully remove without deprecation of those options?

gauravtiwari:
Thanks @CGA1123 Lets do it backwards-compatible way in this PR, please.

CGA1123:
@gauravtiwari One difference between `resolved_paths` and `watched_paths` is that `resolved_paths` is actually passed to `webpack` in order to resolve module names, while `watched_paths` is used only by `webpacker` to determine whether there is a need to recompile.

Is it okay to merge those _slightly_ different concerns? Is there any issue with passing on those extra 'watched' paths to webpack?

gauravtiwari:
Yes that's correct. 

I guess whatever modules are resolved or watched should be available to all. I think that's what you are trying to solve in this pr, being resolved paths to compiler? 

So, if we just use one setting to add additional paths that can be used by - webpack,  on-demand compiler and sass loader then it would remove complexity and confusion.

gauravtiwari:
Am I missing something?

CGA1123:
@gauravtiwari No, that makes sense.

gauravtiwari:
Great, thank you 👍 

Can we also add deprecation warnings to old options?

CGA1123:
I've added warnings to `resolved_paths` and `watched_paths` which will emit when they are non-empty.

lib/webpacker/configuration.rb
```diff
@@ -76,6 +76,14 @@ def extract_css?
   end
 
   private
+    def resolved_paths
+      resolved = data.fetch(:resolved_paths, [])
+
+      warn "The resolved_paths option has been deprecated. Use additional_paths instead." unless resolved.empty?
+
+      resoved
```
gauravtiwari, line 84: typo here ^^ 

Maybe lets call it paths instead of `resolved`?

gauravtiwari:
Thanks @CGA1123, looking great. I will review this later today and get it merged.

CGA1123:
Sound good @gauravtiwari 😄Thanks for your time!

cmitz:
👋🏽  Can someone please put this _somewhere_ into a changelog as breaking change?

CGA1123:
Hi @cmitz 

When initially implemented this was done in a backwards-compatible manner.

It seems more recently in e81793ba3742a1e33176e7796f2b8586c73712b3 / #2802 backwards compatibility was removed in preparation with the release of the new major version 6.

I don't see any reason a PR to add this to the `CHANGELOG` or `6_0_upgrade` would not be accepted before moving the 6.0.0 beta release into a full release 😄

I am assuming that you are running the `6.x` beta release, as the `5.x` release line should maintain backwards compatibility.

cmitz:
I encountered this when upgrading from 5.1.1 to 5.2.1. We did not have a `additional_paths` entry in `webpacker.yml` yet, resulting in the log:

```
NoMethodError: undefined method `+' for nil:NilClass
/srv/.../shared/bundle/ruby/2.7.0/gems/webpacker-5.2.1/lib/webpacker/configuration.rb:31:in `additional_paths'
...
```

I'm happy to create a PR to add it to the Changelog for 5.2.0, but if you say "was done in a backwards-compatible manner" I guess a new issue is more appropriate?

CGA1123:
@cmitz Yes, in this case, that is a bug and not intended behaviour (my bad). I think a new issue would be worth it 😄 

That is an interesting `nil` issue 🤔 
The `additional_paths` should be falling back to defaults set in the YAML afaiu:
https://github.com/rails/webpacker/blob/3d73dbad1eeaf3ef1b7aeaba4ba9262d507f999e/lib/webpacker/configuration.rb#L30-L32
https://github.com/rails/webpacker/blob/3d73dbad1eeaf3ef1b7aeaba4ba9262d507f999e/lib/webpacker/configuration.rb#L87-L89
https://github.com/rails/webpacker/blob/3d73dbad1eeaf3ef1b7aeaba4ba9262d507f999e/lib/webpacker/configuration.rb#L109-L112
https://github.com/rails/webpacker/blob/3d73dbad1eeaf3ef1b7aeaba4ba9262d507f999e/lib/install/config/webpacker.yml#L11-L13

cmitz:
Thanks for clearing that up 😄 , I've made an issue for it: https://github.com/rails/webpacker/issues/2866

PR #2561, opened 2020-04-28
Allow configuration of the WEBPACKER_CONFIG via env

justin808:
If you're using the rails/webpacker node package from an installation
where the node_modules is a subdirectory of the main rails directory,
then the package has an error reading the main config/webpacker.yml.

This is because config.js and env.js assume (maybe incorrectly) that the
node_modules directory is at the top level of the rails project.

Instead, an installation may like the node_modules for the frontend or
client to be in a directory named as such.

This change sets the ENV value in the 3 places that webpack is called:

1. bin/webpack
2. bin/webpack-dev-server
3. compiler called from within the Rails code when the bundles are
   stale.

### DIFF
--- a/lib/webpacker/compiler.rb
+++ b/lib/webpacker/compiler.rb
@@ -105,6 +105,7 @@ class Webpacker::Compiler
       return env unless defined?(ActionController::Base)
 
       env.merge("WEBPACKER_ASSET_HOST"        => ENV.fetch("WEBPACKER_ASSET_HOST", ActionController::Base.helpers.compute_asset_host),
-                "WEBPACKER_RELATIVE_URL_ROOT" => ENV.fetch("WEBPACKER_RELATIVE_URL_ROOT", ActionController::Base.relative_url_root))
+                "WEBPACKER_RELATIVE_URL_ROOT" => ENV.fetch("WEBPACKER_RELATIVE_URL_ROOT", ActionController::Base.relative_url_root),
+                "WEBPACKER_CONFIG" => webpacker.config_path)
     end
 end
--- a/lib/webpacker/dev_server_runner.rb
+++ b/lib/webpacker/dev_server_runner.rb
@@ -45,6 +45,7 @@ module Webpacker
 
       def execute_cmd
         env = Webpacker::Compiler.env
+        env.merge("WEBPACKER_CONFIG" => @webpacker_config)
 
         cmd = if node_modules_bin_exist?
           ["#{@node_modules_bin_path}/webpack-dev-server"]
--- a/lib/webpacker/dev_server_runner.rb
+++ b/lib/webpacker/dev_server_runner.rb
@@ -45,7 +45,7 @@ module Webpacker
 
       def execute_cmd
         env = Webpacker::Compiler.env
-        env["WEBPACKER_CONFIG"] = @webpacker_config
+        env.merge("WEBPACKER_CONFIG" => @webpacker_config)
 
         cmd = if node_modules_bin_exist?
           ["#{@node_modules_bin_path}/webpack-dev-server"]
--- a/lib/webpacker/dev_server_runner.rb
+++ b/lib/webpacker/dev_server_runner.rb
@@ -45,7 +45,7 @@ module Webpacker
 
       def execute_cmd
         env = Webpacker::Compiler.env
-        env.merge("WEBPACKER_CONFIG" => @webpacker_config)
+        env["WEBPACKER_CONFIG"] = @webpacker_config
 
         cmd = if node_modules_bin_exist?
           ["#{@node_modules_bin_path}/webpack-dev-server"]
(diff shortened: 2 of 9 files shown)


justin808:
@gauravtiwari @jakeNiemiec I really need this update badly.

Any chance that you can do a quick review of this one?

lib/webpacker/dev_server_runner.rb
```diff
@@ -45,6 +45,7 @@ def detect_port!
 
       def execute_cmd
         env = Webpacker::Compiler.env
+        env.merge("WEBPACKER_CONFIG" => @webpacker_config)
```
gauravtiwari, line 48: @justin808 don't think this would mutate the env? Maybe simply assign instead of merge.

```rb
env["WEBPACKER_CONFIG"]  =  @webpacker_config)
```

lib/webpacker/webpack_runner.rb
```diff
@@ -5,7 +5,7 @@ module Webpacker
   class WebpackRunner < Webpacker::Runner
     def run
       env = Webpacker::Compiler.env
-
+      env.merge("WEBPACKER_CONFIG" => @webpacker_config)
```
gauravtiwari, line 8: Same here ^^

justin808: @gauravtiwari fixed!

lib/webpacker/dev_server_runner.rb
justin808: @gauravtiwari fixed!

gauravtiwari:
Great thanks, will merge once tests passes.

gauravtiwari:
Could you please get the tests passing?

justin808:
@gauravtiwari I reverted to using merge as that works better with the existing test suite.

env.merge is used in places like this: compiler.rb

```ruby
    def webpack_env
      return env unless defined?(ActionController::Base)

      env.merge("WEBPACKER_ASSET_HOST"        => ENV.fetch("WEBPACKER_ASSET_HOST", ActionController::Base.helpers.compute_asset_host),
                "WEBPACKER_RELATIVE_URL_ROOT" => ENV.fetch("WEBPACKER_RELATIVE_URL_ROOT", ActionController::Base.relative_url_root),
                "WEBPACKER_CONFIG" => webpacker.config_path)
    end
```

gauravtiwari:
Thanks @justin808 

Yes, the method you have mentioned returns the result of `merge` where as the other changes we have are simply calling `env.merge` which means the returned value would have changes but not the actual `env` variable since it's not mutated.


```rb
env = {}
new_env = env.merge(foo: 'bar')
puts env.inspect 
  => {}
puts new_env.inspect
  => {:foo=>"bar"}
```

justin808:
Looks good @gauravtiwari!

justin808:
@gauravtiwari a good question is why we're not just modifying the env. Is it because we don't want to modify what the rest of the Rails app sees?

lib/webpacker/compiler.rb
```diff
@@ -105,6 +105,7 @@ def webpack_env
       return env unless defined?(ActionController::Base)
 
       env.merge("WEBPACKER_ASSET_HOST"        => ENV.fetch("WEBPACKER_ASSET_HOST", ActionController::Base.helpers.compute_asset_host),
-                "WEBPACKER_RELATIVE_URL_ROOT" => ENV.fetch("WEBPACKER_RELATIVE_URL_ROOT", ActionController::Base.relative_url_root))
+                "WEBPACKER_RELATIVE_URL_ROOT" => ENV.fetch("WEBPACKER_RELATIVE_URL_ROOT", ActionController::Base.relative_url_root),
+                "WEBPACKER_CONFIG" => webpacker.config_path)
```
haberbyte, line 109: JUST FYI...This causes a problem for me in testing: ```ActionView::Template::Error: no implicit conversion of Pathname into String
    /Users/jan/.rbenv/versions/2.7.1/lib/ruby/2.7.0/open3.rb:213:in `spawn'
    /Users/jan/.rbenv/versions/2.7.1/lib/ruby/2.7.0/open3.rb:213:in `popen_run'
    /Users/jan/.rbenv/versions/2.7.1/lib/ruby/2.7.0/open3.rb:101:in `popen3'
    /Users/jan/.rbenv/versions/2.7.1/lib/ruby/2.7.0/open3.rb:281:in `capture3'
    /Users/jan/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/bundler/gems/webpacker-ce0133d1948d/lib/webpacker/compiler.rb:70:in `run_webpack'
    /Users/jan/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/bundler/gems/webpacker-ce0133d1948d/lib/webpacker/compiler.rb:23:in `compile'``` ... adding `webpacker.config_path.to_s` fixes it.

gauravtiwari: Could you please make a pr? thank you

haberbyte: Added this PR https://github.com/rails/webpacker/pull/2573

PR #2573, opened 2020-05-07
Fix implicit conversion of Pathname into String on compile

haberbyte:
Fixes an issue with the recent master when running integration tests in a rails app.
```
ActionView::Template::Error: no implicit conversion of Pathname into String
    /Users/jan/.rbenv/versions/2.7.1/lib/ruby/2.7.0/open3.rb:213:in `spawn'
    /Users/jan/.rbenv/versions/2.7.1/lib/ruby/2.7.0/open3.rb:213:in `popen_run'
    /Users/jan/.rbenv/versions/2.7.1/lib/ruby/2.7.0/open3.rb:101:in `popen3'
    /Users/jan/.rbenv/versions/2.7.1/lib/ruby/2.7.0/open3.rb:281:in `capture3'
    /Users/jan/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/bundler/gems/webpacker-ce0133d1948d/lib/webpacker/compiler.rb:70:in `run_webpack'
    /Users/jan/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/bundler/gems/webpacker-ce0133d1948d/lib/webpacker/compiler.rb:23:in `compile'
```

### DIFF
--- a/test/compiler_test.rb
+++ b/test/compiler_test.rb
@@ -40,6 +40,10 @@ class CompilerTest < Minitest::Test
     assert !Webpacker.compiler.fresh?
   end
 
+  def test_compile
+    assert !Webpacker.compiler.compile
+  end
+
   def test_freshness_on_compile_success
     status = OpenStruct.new(success?: true)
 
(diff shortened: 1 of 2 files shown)


gauravtiwari:
thanks @haberbyte

test/compiler_test.rb
```diff
@@ -40,6 +40,10 @@ def test_freshness
     assert !Webpacker.compiler.fresh?
   end
 
+  def test_compile
+    assert !Webpacker.compiler.compile
```
justin808, line 44: @haberbyte how does this test work to assert that the `compiler.compile` command returns false?

haberbyte: Hm, good question, this is actually just the minimal failing test i came up with that would break if you miss the `.to_s` fix. I also think it is weird that it returns false, but did not think much about it.

After some more investigation, it looks like it returns `false` because the command to run in the `Open3.capture3` block can not find "webpack". Stepping into this with a breakpoint i get `Compilation failed: error Command "webpack" not found.`

If this is really intended, i guess hiding that in my test in not good. Not sure how to go about this, but maybe stubbing the webpack command in the test suite would work.

PR #2589, opened 2020-05-15
Change WEBPACKER_DEV_SERVER_HOST to 0.0.0.0

ParamagicDev:
Setting `WEBPACKER_DEV_SERVER_HOST=webpacker` does not allow Docker to see changes from Webpacker.

When changing `WEBPACKER_DEV_SERVER_HOST=0.0.0.0` everything runs as intended.

### DIFF
--- a/docs/docker.md
+++ b/docs/docker.md
@@ -58,7 +58,7 @@ then add the webpacker host name environment variable to the web/app service:
     ports:
       - "3000:3000"
     environment:
-      - WEBPACKER_DEV_SERVER_HOST=webpacker
+      - WEBPACKER_DEV_SERVER_HOST=0.0.0.0
 ```
 
 Lastly, rebuild your container:


rossta:
Hmmm, I've got a working demo with host set to `webpacker`. As long as the host matches the docker-compose container label, I would expect it to work like this example: https://github.com/rossta/rails6-webpacker-demo/compare/rossta:6d18b2f...rossta:3fa8403#diff-4e5e90c6228fd48698d074241c2ba760R26

Mind sharing your config docker-compose config?

ParamagicDev:
Sure not a problem!

[https://github.com/ParamagicDev/getting-started-with-rails-6/blob/master/docker-compose.yml](https://github.com/ParamagicDev/getting-started-with-rails-6/blob/master/docker-compose.yml)

When I set the `web` services `WEBPACKER_DEV_SERVER_HOST` to `webpacker` it doesn't work but when I changed it to `0.0.0.0` it works.

ParamagicDev:
I should clarify, the app still builds, but it doesn't load `app/javascript/` files. In particular, it would not reflect changes from this file [https://github.com/ParamagicDev/getting-started-with-rails-6/blob/master/app/javascript/stylesheets/global.scss](https://github.com/ParamagicDev/getting-started-with-rails-6/blob/master/app/javascript/stylesheets/global.scss) 

The app would still build and run, but the changes stylesheet wouldn't get applied even with loaded the `stylesheet_pack_tag`

rossta:
The reason the documented docker-compose.yml works with `WEBPACKER_DEV_SERVER_HOST=webpacker` is that it runs the webpacker-dev-server in a separate container called `webpacker`:

```yaml
services:
  webpacker:
    build: .
    environment:
      - NODE_ENV=development
      - RAILS_ENV=development
      - WEBPACKER_DEV_SERVER_HOST=0.0.0.0
    command: ./bin/webpack-dev-server
    volumes:
      - .:/webpacker-example-app
    ports:
      - '3035:3035'
```

Your docker-compose.yml runs both the rails server and webpack-dev-server in the same container, i.e., `./bin/webpack-dev-server & bundle exec rails server -p 3000 -b '0.0.0.0'"` which is why `WEBPACKER_DEV_SERVER_HOST=webpacker` doesn't work for you. Perhaps the suggested doc change is unnecessary then.

ParamagicDev:
And this is why you should never link a master branch. 

That was a later change I made. Regardless, I'll close this as resolved, I think this was the commit that had the issue but can't be 100%.

[Docker-compose.yml](https://github.com/ParamagicDev/getting-started-with-rails-6/blob/3b6958bb56d7d95c6dbf2947f78baed8db0beb69/docker-compose.yml)

Oh well, you live and you learn.

PR #2595, opened 2020-05-17
Forward ARGV to webpack-dev-server command

### LINKED ISSUE #2000: ./bin/webpack-dev-server command line arguments are ignored

rossta:
The `webpack` command runner [forwards additional ARGV options to the underlying executable](https://github.com/rails/webpacker/blob/22ab02b7c84e917f985ecc76f5916d144f43bfbf/lib/webpacker/webpack_runner.rb#L21).

This PR brings parity to usage of the `webpack-dev-server` command runner. Example:

Before, options like the `--quiet` flag would be dropped; now they will be respected. 

`./bin/webpacker-dev-server --quiet`

Fixes #2000

### DIFF
--- a/lib/webpacker/dev_server_runner.rb
+++ b/lib/webpacker/dev_server_runner.rb
@@ -59,6 +59,7 @@ module Webpacker
 
         cmd += ["--config", @webpack_config]
         cmd += ["--progress", "--color"] if @pretty
+        cmd += @argv
 
         Dir.chdir(@app_path) do
           Kernel.exec env, *cmd
--- a/lib/webpacker/dev_server_runner.rb
+++ b/lib/webpacker/dev_server_runner.rb
@@ -7,12 +7,24 @@ require "webpacker/runner"
 module Webpacker
   class DevServerRunner < Webpacker::Runner
     def run
+      detect_unsupported_switches!
       load_config
       detect_port!
       execute_cmd
     end
 
     private
+
+      UNSUPPORTED_SWITCHES = %w[--host --port --https]
+      private_constant :UNSUPPORTED_SWITCHES
+      def detect_unsupported_switches!
+        unsupported_switches = UNSUPPORTED_SWITCHES & @argv
+        if unsupported_switches.any?
+          $stdout.puts "The following CLI switches are not supported by Webpacker: #{unsupported_switches.join(' ')}. Please edit your command and try again."
+          exit!
+        end
+      end
+
       def load_config
         app_root = Pathname.new(@app_path)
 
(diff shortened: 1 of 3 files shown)


gauravtiwari:
Thanks @rossta for this.

I hesitated to merge similar changes in the past because Webpacker relies internally on some of the core options like - `host or port`, which are set via either `webpacker.yml` or an environment variable. I am just concerned people will start passing these as arguments and it may cause runtime issues. 

Perhaps, we can ignore core variables and log a message if we do so?

rossta:
Thanks @gauravtiwari. Good point. I'll take another pass at this to address your comments.

gauravtiwari:
Hi @rossta Would you be able to look at this at some point?

rossta:
@gauravtiwari I updated the PR to bail out when it detects unsupported switches. 

I considered parsing the ARGV and attempting to filter out the unsupported switches and let the command continue but it quickly got unwieldy. This seems like the simplest approach for now.

Let me know if there are additional switches that should be listed as unsupported.

jipiboily:
Hello folks!

Any chance we can get this back ported to 5.2.x?

PR #2618, opened 2020-06-03
Use ruby/setup-ruby@v1 for Github Actions

pustovalov:
https://github.com/ruby/setup-ruby - supports `ruby-head`

Example:
https://github.com/pustovalov/webpacker/actions/runs/124153726

### DIFF
--- a/.github/workflows/jest.yml
+++ b/.github/workflows/jest.yml
@@ -14,9 +14,6 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-        with:
-          fetch-depth: 0
-
       - name: Get yarn cache directory path
         id: yarn-cache-dir-path
         run: echo "::set-output name=dir::$(yarn cache dir)"
--- a/.github/workflows/js-lint.yml
+++ b/.github/workflows/js-lint.yml
@@ -15,9 +15,6 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-        with:
-          fetch-depth: 0
-
       - name: Get yarn cache directory path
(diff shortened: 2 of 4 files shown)


gauravtiwari:
Can we get the failing test passing or is that intended?

pustovalov:
I'll see what can be done, but this problem was before

pustovalov:
My guess is that something is wrong with `test_app` config

If remove:
```
Rails.application.initialize!
```
in `test/test_app/config/environment.rb`

this error is gone

https://github.com/pustovalov/webpacker/runs/788584397?check_suite_focus=true

but what is wrong is not yet clear

error:
```
/home/runner/bundle/ruby/2.8.0/gems/actionpack-5.2.4.3/lib/action_dispatch/middleware/static.rb:111:in `initialize': wrong number of arguments (given 3, expected 2) (ArgumentError)
```

https://github.com/rails/rails/blob/5-2-stable/actionpack/lib/action_dispatch/middleware/static.rb#L111

any suggestion? (the difficulties are added by the fact that I can’t install `nokogiri` gem with `ruby-head` on local mac)

justin808:
@gauravtiwari @pustovalov is ruby-head is not supported by Rails? I don't see why we should bother with ruby-head given the nature and purpose of this gem.

pustovalov:
@justin808 

Rails runs its tests for ruby-head

https://buildkite.com/rails/rails/builds/70815
https://buildkite.com/rails/rails/builds/70815#b9aaad41-7a14-4a36-b9d1-0be471da1cd1

PR #2629, opened 2020-06-14
Update bin/webpack-dev-server with ENV value

justin808:
Setting an ENV value allows one to easily setup up config/webpack/development.js so that it can build a separate client bundle separately from a server bundle. Only the client bundle should get sent to the webpack-dev-server.

Here is an example configuration that would use this env value.

## `config/webpack/development.js`
```js
process.env.NODE_ENV = process.env.NODE_ENV || "development";

const environment = require("./environment");
const merge = require("webpack-merge");

// Replace
// module.exports = environment.toWebpackConfig()

const clientConfig = environment.toWebpackConfig();

const serverConfig = merge(environment.toWebpackConfig(), {
  mode: "development",
  target: "web",
  entry: "./app/javascript/bundles/hello-world-bundle-server.js",
  output: {
    filename: "hello-world-bundle-server.js",
    path: environment.config.output.path,
    globalObject: "this",
  },
  optimization: {
    minimize: false,
  },
});

serverConfig.plugins = serverConfig.plugins.filter(
  (plugin) => plugin.constructor.name !== "WebpackAssetsManifest"
);

// For HMR, we need to separate the the client and server webpack configurations
if (process.env.WEBPACK_DEV_SERVER) {
  module.exports = clientConfig;
} else if (process.env.SERVER_BUNDLE_ONLY) {
  module.exports = serverConfig;
} else {
  module.exports = [clientConfig, serverConfig];
}
```

### DIFF
--- a/lib/install/bin/webpack-dev-server
+++ b/lib/install/bin/webpack-dev-server
@@ -2,6 +2,7 @@
 
 ENV["RAILS_ENV"] ||= ENV["RACK_ENV"] || "development"
 ENV["NODE_ENV"]  ||= "development"
+ENV["WEBPACK_DEV_SERVER"] ||= "TRUE"
 
 require "pathname"
 ENV["BUNDLE_GEMFILE"] ||= File.expand_path("../../Gemfile",


justin808:
@jakeNiemiec and @gauravtiwari Let me know if you think this would be good to include in the gem. While this could be added manually, it seems that we need some way to allow conditional logic in development.js to know if the code is run on the bin/webpack-dev-server.

gauravtiwari:
Hi @justin808 Seems like it's already supported? https://github.com/webpack/webpack-dev-server/pull/1929

justin808:
@gauravtiwari brilliant! I never thought to check! How did you know to check that?

gauravtiwari:
No problem, I just googled if there is another way to check that the `webpack-dev-server` is running - got a few hits ^^

PR #2631, opened 2020-06-17
Update doc for TS to support fork-ts-checker-webpack-plugin 5.0

yuki24:
[`fork-ts-checker-webpack-plugin` 5.0.0](https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/releases/tag/v5.0.0) has been released 3 days ago and it has a few breaking changes. Basically I had to remove the `tsconfig` option from the existing setup:

```js
const ForkTsCheckerWebpackPlugin = require("fork-ts-checker-webpack-plugin");

environment.plugins.append(
  "ForkTsCheckerWebpackPlugin",
  new ForkTsCheckerWebpackPlugin({ async: false }) // non-async so type checking will block compilation
);
```

cc @hueter

### DIFF
--- a/docs/typescript.md
+++ b/docs/typescript.md
@@ -16,26 +16,37 @@ The default installation only transpiles your TypeScript code using Babel. If yo
 
 1. Install the Fork TS Checker Webpack Plugin
 
-```sh
-yarn add --dev fork-ts-checker-webpack-plugin
-```
+    ```sh
+    yarn add --dev fork-ts-checker-webpack-plugin
+    ```
 
 2. Then add it to your development environment config in `config/webpack/development.js`
 
-```js
-const ForkTsCheckerWebpackPlugin = require("fork-ts-checker-webpack-plugin");
-const path = require("path");
+    ```js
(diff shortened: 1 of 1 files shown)


hueter:
Oh yep, I also applied this update to my codebase. In fact the `tsconfig` option is probably not necessary for most people, but FWIW, it got moved under the [typescript](https://github.com/TypeStrong/fork-ts-checker-webpack-plugin#typescript-options) key of the [new options config object](https://github.com/TypeStrong/fork-ts-checker-webpack-plugin#options):


```js
environment.plugins.append(
  "ForkTsCheckerWebpackPlugin",
  new ForkTsCheckerWebpackPlugin({
    typescript: {
      tsconfig: path.resolve(__dirname, "../../tsconfig.json"),
    },
    async: false,
  })
);
```

yuki24:
Oh nice, I just updated the doc. Also, I managed to break the build with a documentation update somehow, but I'm sure they are unrelated.

gauravtiwari:
thanks @yuki24

PR #2633, opened 2020-06-19
move browserslist and babel config into package.json

### LINKED ISSUE #2532: Create babel preset instead of copying entire config into project
Currently, webpacker copies the pre made babel.config.js into your project when the install script is run, however this leads to some issues:
- users who would never touch their babel.config.js now have a complicated file they have to maintain
- updates to babel.config.js in webpacker repo have to be manually grabbed by users
- the babel.config.js in the project uses `require` on webpacker dependencies, which is not ideal and is a blocker to adopting Yarn 2 #2112 

To solve these issues, I think webpacker (or an accompanying package) should provide a babel preset that projects can then use:
```
// babel.config.js
module.exports = {
  presets: ["@rails/babel-preset-webpacker"]
}
```
or
```
// babel.config.js
module.exports = {
  presets: ["./node_modules/@rails/webpacker/.../babel.config.js"]
}
```

This would greatly simplify the configuration for basic cases, while still allowing the flexibility to add additional configuration or ignoring the preset completely for advanced cases.

I am willing to help out with this and can create a PR if needed.

SkipKayhil:
Running webpacker:install will now add browserslist and babel configuration into package.json instead of copying files. browserslist [recommends](https://github.com/browserslist/browserslist#queries) using package.json key over rc file.

Use babel.config.js in npm package as a preset, providing updates and less complexity to apps. Fixes #2532 (at least partially)

Additionally, updated rubocop.yml to target ruby 2.4 since it is now the minimum supported version

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,6 +1,6 @@
 require: rubocop-performance
 AllCops:
-  TargetRubyVersion: 2.2
+  TargetRubyVersion: 2.4
   # RuboCop has a bunch of cops enabled by default. This setting tells RuboCop
   # to ignore them, so only the ones explicitly set in this file are enabled.
   DisabledByDefault: true
--- a/lib/install/config/.browserslistrc
+++ b/lib/install/config/.browserslistrc
@@ -1,1 +1,0 @@
-defaults
--- a/lib/install/react.rb
+++ b/lib/install/react.rb
@@ -1,10 +1,12 @@
 require "webpacker/configuration"
-require "fileutils"
+require "json"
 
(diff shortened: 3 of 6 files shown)


SkipKayhil:
@gauravtiwari any chance I could get some feedback on this? I really think it would be a beneficial change

gauravtiwari:
Thanks @SkipKayhil looks great. I am sorry it took me a while to review. There is one config for react as well: https://github.com/rails/webpacker/blob/master/lib/install/examples/react/babel.config.js that we need to migrate to use package.json.

Issue #1879, opened 2019-01-09
Watch rails views and reload page on changes


mbajur:
I'm currently using webpacker with page reloading on css/js changes and it all works just great. Thanks for this amazing piece of software. I was wondering - would that be possible to do the same for rails views? It would be great if updating erb or slim view could trigger a webpack page refresh (ideally without rebuilding entire js/css bundle).

I'm fully aware that's not the point of using webpack(er) and know there are solutions like guard and livereload for that but i would love to avoid using yet another deamon running in the background.

Thanks in advance for any clues!

scottatron:
Not quite as nice as using webpack itself, but I've got this working by using [watchexec](https://github.com/watchexec/watchexec).

I run up webpack-dev-server, then run:
```
watchexec -w app/views touch app/javascript/packs/application.js
```

mtomov:
There's an even easier way, just set the `devServer` `contentBase` property like this in `config/webpack/development.js`

```js
environment.config.devServer.watchContentBase = true
environment.config.devServer.contentBase = [
  path.join(__dirname, '../../app/views')
]
```

swrobel:
Here's an example that won't cause webpack to do a full recompile. This will simply trigger a browser reload. Your specific watches may vary, but they're pretty easy to configure using an array of glob syntax w/ chokidar (which is already included by webpack) - just note that it doesn't accept regexes:

development.js:
```js
const chokidar = require('chokidar')
environment.config.devServer.before = (app, server) => {
  chokidar.watch([
    'config/locales/*.yml',
    'app/views/**/*.slim'
  ]).on('change', () => server.sockWrite(server.sockets, 'content-changed'))
}
```

jesster2k10:
> Here's an example that won't cause webpack to do a full recompile. This will simply trigger a browser reload. Your specific watches may vary, but they're pretty easy to configure using an array of glob syntax w/ chokidar (which is already included by webpack) - just note that it doesn't accept regexes:
> 
> development.js:
> 
> ```js
> const chokidar = require('chokidar')
> environment.config.devServer.before = (app, server) => {
>   chokidar.watch([
>     'config/locales/*.yml',
>     'app/views/**/*.slim'
>   ]).on('change', () => server.sockWrite(server.sockets, 'content-changed'))
> }
> ```

Hey, I can't seem to get this working. The "chokidar" isn't watching any changes in *.html.erb files

swrobel:
[...]

PR #2644, opened 2020-06-25
Only include dev-server parts if for dev server

justin808:
Without this change, webpack still includes things like window
in the build even if not using the webpack-dev-server.

When that happens, the bundle cannot be used for server-side rendering.

### DIFF
--- a/package/__tests__/development.js
+++ b/package/__tests__/development.js
@@ -11,9 +11,10 @@ describe('Development environment', () => {
   describe('toWebpackConfig', () => {
     beforeEach(() => jest.resetModules())
 
-    test('should use development config and environment', () => {
+    test('should use development config and environment including devServer if WEBPACK_DEV_SERVER', () => {
       process.env.RAILS_ENV = 'development'
       process.env.NODE_ENV = 'development'
+      process.env.WEBPACK_DEV_SERVER = 'YES'
       const { environment } = require('../index')
 
       const config = environment.toWebpackConfig()
@@ -25,6 +26,18 @@ describe('Development environment', () => {
           port: 3035
         }
       })
+    })
+
+    test('should use development config and environment if WEBPACK_DEV_SERVER', () => {
+      process.env.RAILS_ENV = 'development'
+      process.env.NODE_ENV = 'development'
+      process.env.WEBPACK_DEV_SERVER = undefined
+      const { environment } = require('../index')
+
+      const config = environment.toWebpackConfig()
+      expect(config.output.path).toEqual(resolve('public', 'packs'))
+      expect(config.output.publicPath).toEqual('/packs/')
+      expect(config.devServer).toEqual(undefined)
     })
   })
 })
--- a/package/environments/development.js
+++ b/package/environments/development.js
@@ -12,7 +12,8 @@ module.exports = class extends Base {
       devtool: 'cheap-module-source-map'
     })
 
-    if (process.env.WEBPACK_DEV_SERVER) {
+    if (process.env.WEBPACK_DEV_SERVER
+        && process.env.WEBPACK_DEV_SERVER !== 'undefined') {
       if (devServer.hmr) {
         this.plugins.append('HotModuleReplacement', new webpack.HotModuleReplacementPlugin())
         this.config.output.filename = '[name]-[hash].js'
--- a/package/environments/development.js
+++ b/package/environments/development.js
@@ -7,41 +7,46 @@ module.exports = class extends Base {
   constructor() {
     super()
 
-    if (devServer.hmr) {
-      this.plugins.append('HotModuleReplacement', new webpack.HotModuleReplacementPlugin())
-      this.config.output.filename = '[name]-[hash].js'
-    }
-
     this.config.merge({
       mode: 'development',
-      devtool: 'cheap-module-source-map',
-      devServer: {
-        clientLogLevel: 'none',
-        compress: devServer.compress,
-        quiet: devServer.quiet,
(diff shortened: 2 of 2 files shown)


package/environments/development.js
```diff
@@ -7,41 +7,46 @@ module.exports = class extends Base {
...
-          errorDetails: true,
-          modules: false,
-          moduleTrace: false
-        },
-        watchOptions: devServer.watch_options
-      }
+      devtool: 'cheap-module-source-map'
     })
+
+    if (process.env.WEBPACK_DEV_SERVER) {
```
justin808, line 15: See https://github.com/rails/webpacker/pull/2629 for why this variable is there.

and

https://github.com/webpack/webpack-dev-server/pull/1929

justin808:
@gauravtiwari This allows a simple setup to work for SSR. Otherwise, SSR can work if HMR and inline are set to false for the devServer. That's how I figured out this was needed.

justin808:
@gauravtiwari @rossta @JakeLaCombe any chance of a quick review on this? It blocks me from making a super cool demo of React on Rails v12 where server rendering just works, so long as the webpack dev server is not used. It's a little more complicated with that one!

justin808:
How do I fix the ruby specs?

package/environments/development.js
```diff
@@ -7,41 +7,47 @@ module.exports = class extends Base {
...
-          modules: false,
-          moduleTrace: false
-        },
-        watchOptions: devServer.watch_options
-      }
+      devtool: 'cheap-module-source-map'
     })
+
+    if (process.env.WEBPACK_DEV_SERVER
+        && process.env.WEBPACK_DEV_SERVER !== 'undefined') {
```
justin808, line 16: When I ran the tests, for some reason, the undefined comes as a string for comparison.

package/__tests__/development.js
```diff
@@ -26,5 +27,17 @@ describe('Development environment', () => {
         }
       })
     })
+
+    test('should use development config and environment if WEBPACK_DEV_SERVER', () => {
+      process.env.RAILS_ENV = 'development'
+      process.env.NODE_ENV = 'development'
+      process.env.WEBPACK_DEV_SERVER = undefined
```
justin808, line 34: This is per the definition here: https://github.com/webpack/webpack-dev-server/pull/1929/files#diff-15fb51940da53816af13330d8ce69b4eR66

```
if (!process.env.WEBPACK_DEV_SERVER) {
  process.env.WEBPACK_DEV_SERVER = true;
}
```

justin808:
@gauravtiwari @jakeNiemiec Any feedback on why this is not getting merged?

gauravtiwari:
Thanks for this @justin808 and apologies for delay. 

The code looks good that only reason I have been hesitating is the dependence on env variable. I was thinking perhaps, we should have a completely different config class for dev server which extends development.js and then in the dev server config we instantiate that class explicitly. 

https://github.com/rails/webpacker/blob/master/lib/webpacker/runner.rb#L14

```js
      @webpack_config        = File.join(@app_path, "config/webpack/#{ENV["NODE_ENV"]}.js")
```

This line here should be dev server runner and set to dev server config class regardless of NODE_ENV.

justin808:
@gauravtiwari I think this change is small enough that we could merge this PR and do a bigger change as you suggest next. AFAIK, I don't think the webpack-dev-server would remove that ENV value as many other apps might depend on it.

I could try to work on your suggestion as a followup PR.

For right now, it's **very** confusing that the HMR settings break the non-HMR non-webpack-dev-server usage of bin/webpack.

justin808:
Thanks @gauravtiwari!

aried3r:
This change broke our test suite, since `WEBPACK_DEV_SERVER` isn't set during tests anymore after this.

What exactly is the migration path from webpacker 5.1.1 to 5.2.1 regarding this and where to best set this variable in a "webpacker-way"?

The code causing our issues is the following in `config/webpack/development.js`:

```JavaScript
const chokidar = require("chokidar")
environment.config.devServer.before = (app, server) => {
  chokidar
    .watch([
      "config/locales/**/*.yml",
      "app/views/**/*.html.erb",
      "app/assets/**/*.scss",
    ])
    .on("change", () => server.sockWrite(server.sockets, "content-changed"))
}
```

Error:
```

Taken from https://github.com/rails/webpacker/issues/1879#issuecomment-558397652

aried3r:
Okay, in classic fashion of rubber duck debugging, I found the solution shortly after writing it down.

https://github.com/rails/webpacker/blob/9ec5daabc5daea1c96f538d7af1f807ae9b077ec/lib/install/config/webpack/test.js#L1

is causing it and we'll write a guard clause similar to what's in this PR for our use case.

justin808:
@aried3r how does setting the `before` on the [dev_server](https://github.com/rails/webpacker/blob/master/package/dev_server.js) object work?

justin808:
@aried3r the NODE_ENV of test picks the test.js file by default. Resetting the NODE_ENV to 'development' in the test.js file seems very odd.

aried3r:
Hey @justin808, thank you for your comment. I believe https://github.com/rails/webpacker/pull/2721 and https://github.com/rails/webpacker/issues/2654 answer my question much better than I did in my follow-up comment.

> how does setting the `before` on the [dev_server](https://github.com/rails/webpacker/blob/master/package/dev_server.js?rgh-link-date=2020-09-05T21%3A16%3A52Z) object work?

I don't quite understand, could you elaborate?

justin808:
@aried3r, what code uses the function set in property `environment.config.devServer.before`?

PR #2645, opened 2020-06-25
Update CHANGELOG.md

rossta:
Add bullet for upcoming 6.0.0 release about changes to the install task defaulting `extract_css` to true and generating the separate application.css file.


justin808:
@rossta @gauravtiwari I'm also getting the same Ruby specs CI failure. So that's not relevant to stopping a merge.

PR #2647, opened 2020-06-27
Remove unsupported Ruby version head

justin808:
GH Actions does not support "head"
https://github.com/actions/setup-ruby


pustovalov:
https://github.com/rails/webpacker/pull/2618

justin808:
@JakeLaCombe @gauravtiwari can we do a quick fix and turn off ruby-head? Do we need it?

Issue #2640, opened 2020-06-23
Allow a custom compilation command


justin808:
Context: Many applications have custom webpack build commands that are not easy to get retrofitted into the stale checking and assets:precompile of integrations of rails/webpacker.

Could we allow a setting in the `config/webpacker.yml` file to allow a custom compilation command?

`custom_webpack_compile`: "yarn run my_default_custom_webpack_command"
`test.custom_webpack_compile: "yarn run my_custom_webpack_command_to_build_test_bundles`
`production.custom_webpack_compile: "yarn run my_custom_webpack_command_to_build_prodution_bundles`

This command would be run with the RAILS_ENV and the NODE_ENV.

The use case for this is that many React on Rails installations have custom webpack setups. For React on Rails, two configuration values set the compiler command for test and production envs. Development webpack compilation is typically handled by the Procfile.dev.

Suppose that React on Rails install with a custom webpack config wants to leverage the compiler's stale checking. The command to get the webpack config currently is the `config/webpack/production.js` Potentially, the code in those files could be adjusted to run the current webpack configurations. Allowing for a custom command would be much more direct.

@gauravtiwari @jakeNiemiec If you're agreeable to this, I'd create a PR with this enhancement.

justin808:
The compiler calls the bin/webpack script.

So maybe one just needs to replace that script?

https://github.com/rails/webpacker/blob/master/lib/webpacker/compiler.rb#L70-L74

```ruby
      stdout, stderr, status = Open3.capture3(
        webpack_env,
        "#{RbConfig.ruby} ./bin/webpack",
        chdir: File.expand_path(config.root_path)
      )
```

Let me know your thoughts on that.

justin808:
The problem with replacing the installed binstub is that only a Ruby script works.

PR #2650 addresses that limitation.

justin808:
@gauravtiwari @jakeNiemiec Another thought is that the `compile` option makes the invocation of the `bin/webpack` use the `RbConfig.ruby` but many times people run `bin/webpack --watch` and that won't use the special ruby version. This seems sort of problematic. Maybe we don't ever need to configure the ruby version in the `compiler.rb` file at all?

justin808:
OK -- I see why this was done. Windows which I cannot test

https://github.com/rails/webpacker/pull/1286#issuecomment-367951995

```
gauravtiwari commented on Feb 22, 2018
[...]

PR #2650, opened 2020-06-27
Only prepend the ruby runner if a ruby script

justin808:
Check for the ruby shebang on the first line.

This allows a standard bash script instead.


justin808:
This relates to #2640.

justin808:
@gauravtiwari @jakeNiemiec What do you think of this PR? If we can get this merged, then I can remove some code from React on Rails for custom webpack configurations.

justin808:
@gauravtiwari @jakeNiemiec Any thoughts on this one? This small PR would enable a custom webpack build script tied into rails/webpacker.

PR #2653, opened 2020-06-28
Update README

kazuhitonakayama:
※If you can't execute the command ``yarn add https://github.com/rails/webpacker.git`` , your node version is maybe out of date. Expected version "^6 || ^7 || ^8 || ^9 || ^10 || ^11 || ^12 || >=13.7".


pustovalov:
Webpacker supports:

`"node": ">=10.17.0"`
https://github.com/rails/webpacker/blob/master/package.json#L11

jakeNiemiec:
To clarify further, Webpacker supports all `Active/Maintenance LTS` releases of node.js. You can view these here: https://nodejs.org/en/about/releases/

PR #2659, opened 2020-07-08
Update CSS docs for importing from NPM package

### LINKED ISSUE #2552: How to use css from an npm package - there is not piece of documentation that touches on this
1. Migrating to rails 6 
2. Using jquery.fancytree - https://github.com/mar10/fancytree
3. Successfully migrate to using the JS
4. Now has to use the css - 

```
import 'jquery.fancytree/dist/skin-lion/ui.fancytree.less';  // CSS or LESS
```

There is little in the documentation that shares how to use a css from a npm package in a rails view or to refer it in another css document. And this is the second thing we should do when migrating. We should use the js from the npm package and the css from somewhere else which is not the right solution.  

1. New rails 6 apps are not helpful since they use sprockets. 
2. https://github.com/rails/webpacker#usage is about writing a new css.
3. https://github.com/rails/webpacker/blob/master/docs/css.md is trying to touch on the subject of css with "Link styles in your Rails views" and "Add bootstrap", but it is detached  - I guess the "Add bootstrap" should also include more information about he configuratoin for the app/javascript/app.sass

rossta:
Provides examples to address how importing CSS provided by an NPM package would work.

Fixes #2552

### DIFF
--- a/docs/css.md
+++ b/docs/css.md
@@ -50,6 +50,17 @@ app/
 @import './comments';
 ```
 
+### Importing CSS provided by an NPM package from SCSS/CSS
+
+Given your application installs an NPM package that provides CSS, such as `flatpickr`, you can import the CSS file(s) by path from the package directory within `node_modules/`:
+
+```js
+/* app/javascript/stylesheets/application.scss */
+
+@import "flatpickr/dist/flatpickr.css"
+```
+
+
 ### Importing CSS from JS
 
 ```sass
@@ -74,6 +85,16 @@ const Hello = props => (
     <p>Hello {props.name}!</p>
   </div>
 )
+```
+
+### Importing CSS provided by an NPM package from JS
+
+Given your application installs an NPM package that provides CSS, such as `flatpickr`, you can import the CSS file(s) by path from the package directory within `node_modules/`. This is an alternative to importing from within a CSS file, as above:
+
+```js
+// app/javascript/packs/application.js
+
+import "flatpickr/dist/flatpickr.css"
 ```
 
 ## Import scoped styles into your JS app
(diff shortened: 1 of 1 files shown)


docs/css.md
```diff
@@ -50,6 +50,17 @@ app/
...
 ```
 
+### Importing CSS provided by an NPM package from SCSS/CSS
+
+Given your application installs an NPM package that provides CSS, such as `flatpickr`, you can import the CSS file(s) by path from the package directory within `node_modules/`:
+
+```js
+/* app/javascript/stylesheets/application.scss */
+
+@import "flatpickr/dist/flatpickr.css"
```
nguyenj, line 60: This needs to have a prepending tilde, so it would be like this `@import "~flatpickr/dist/flatpickr.css"` 

Reference from issue, https://github.com/rails/webpacker/issues/2552#issuecomment-655660887.

rossta: @nguyenj

Thanks for your comments. 

> TBH, this is actually explained already

I think this info is better explained in the CSS docs as I believe assets.md is geared more towards static assets like images; perhaps the next step is to remove CSS info from assets.md or move examples to css.md.

> This needs to have a prepending tilde

My understanding is that the prepending tilde is now incorrect since the introduction of `postcss-loader` with [`postcss-import` in the default `postcss.config.js`](https://github.com/rails/webpacker/blob/bf278f9787704ed0f78038ad7d36c008abc2edfd/lib/install/config/postcss.config.js#L3) as of [Webpacker v4.0.0.rc1](https://github.com/rails/webpacker/commit/8d496a0b871085648d6ee21c0e1e469d11cfe593). Existing documentation that refers to tilde should either be removed or notated to reflect it's relevance only to pre v4.0.0 unless I'm missing something.

I have a working demo for importing flatpickr css without the tilde prefix from the NPM package within an SCSS file running on Webpacker v5. https://github.com/rossta/rails6-webpacker-demo/compare/example/flatpickr

nguyenj: @rossta 

> My understanding is that the prepending tilde is now either unnecessary or, for many projects, incorrect since the introduction of postcss-loader with postcss-import in the default postcss.config.js as of Webpacker v4.0.0.rc1. Existing documentation that refers to tilde should either be removed or notated to reflect it's relevance only to pre v4.0.0 unless I'm missing something.

This is great! Thanks for taking the time to explain. And yes, I think that assets documentation should be updated too.

Issue #2613, opened 2020-06-01
esModule file-loader compat


eric-hemasystems:
I recently upgraded an app from webpacker 4 -> 5. When I did that my "fontawesome" icons went away. I found the font was not loading despite being in the `static_assets_extensions`.  When I looked at the compiled CSS it looked like:

```css
@font-face {
  font-family: 'FontAwesome';
  src: url([object Module]);
  src: url("[object Module]?#iefix&v=4.7.0") format("embedded-opentype"), url([object Module]) format("woff2"), url([object Module]) format("woff"), url([object Module]) format("truetype"), url("[object Module]#fontawesomeregular") 
```

I did a search and found a related issue on stack overflow that indicates adding `esModule: false` to the fileloader options:

https://stackoverflow.com/questions/57671995/webpack-4-gives-background-urlobject-module-as-bg-image

I looked at the file-loader package and it appears in v5 they made a breaking change documented as:

> switch to ES modules by default (the option esModule is true by default)

Since the file-loader config is controlled by webpacker making that config change would not be easy to do. Instead I downgraded to version 4.3 of `file-loader` and things are now working fine.

Seems either webpacker needs to be updated to use this `esModule: false` option or some other change needs to be made so it is compat with the new way `file-loader` is doing things.

jakeNiemiec:
Are you using the `sprite-loader` from that SO question? How are you `require`'ing fontawesome? (CSS or JS)

eric-hemasystems:
Not that I know of and I don't see it in my yarn.lock file.

* The `dependencies` line on my `package.json` is `"font-awesome": "^4.7.0"`
* `app/javascripts/packs/application.js` has `import 'font-awesome/scss/font-awesome.scss'`.
* `config/webpacker.yml` has `extract_css: true`
* `app/layouts/application.html.erb` has `<%= stylesheet_pack_tag 'application', media: 'all', 'data-turbolinks-track': 'reload' %>`

wokenqingtian:
I met a similar issue with `file-loader` and found a solution [here](https://github.com/vuejs/vue-loader/issues/1612#issuecomment-614818907). It may be useful for you.

```js
environment.loaders.get('file').use.find(item => item.loader === 'file-loader').options.esModule = false

```

PascalPixel:
Same here, no special loaders, but `scss` files are outputting `src: url([object Module]);` since upgrading webpacker. The above comment provided the solution by disabling esModule in `file-loader`.

[...]

PR #2668, opened 2020-07-21
Fix file-loader compat issue in webpacker 5

amoshydra:
Change file-loader `esModule` option to `false`

Resolve https://github.com/rails/webpacker/issues/2613

### DIFF
--- a/package/rules/file.js
+++ b/package/rules/file.js
@@ -13,6 +13,7 @@ module.exports = {
           }
           return 'media/[folder]/[name]-[hash:8].[ext]'
         },
+        esModule: false,
         context: join(sourcePath)
       }
     }


aried3r:
@gauravtiwari, what do you think about https://github.com/rails/webpacker/issues/2613#issuecomment-668092394? It seems going forward, many loaders will have `esModule: true`, or the option added, and the problem here seemed to be that we treated font files etc as ES modules.

https://github.com/webpack-contrib/css-loader/blob/master/CHANGELOG.md#400-2020-07-25
https://github.com/webpack-contrib/mini-css-extract-plugin/pull/475
https://github.com/webpack-contrib/style-loader/pull/441
https://github.com/webpack-contrib/url-loader/blob/master/CHANGELOG.md#300-2019-11-26
https://github.com/webpack-contrib/file-loader/blob/master/CHANGELOG.md#500-2019-11-22

Friendly ping to @jakeNiemiec.

gauravtiwari:
Yes, I did, thank you. 

I think at the moment it's pretty much static files: https://github.com/rails/webpacker/blob/master/lib/install/config/webpacker.yml#L21 so for general usage lets make it default to false. 

I believe a lot of people would run this problem but happy to look into a simple configurable solution later.

renchap:
This is a breaking change and should have been in a major version, or at least documented in the changelog.

If you used for example `require("./file.svg").default` to get the path of a SVG file, this changes your path to be `undefined`, and images to be empty if you used it like ``

PR #2694, opened 2020-08-13
Export additional utility classes

justin808:
Exporting these allows for more flexibility to customize Webpack.

### DIFF
--- a/package/environments/base.js
+++ b/package/environments/base.js
@@ -157,6 +157,7 @@ module.exports = class Base {
     return this.config.merge(deepMerge(defaultConfig, appConfig))
   }
 
+  // This should be a NEW object and separate from multiple calls to this
   toWebpackConfig() {
     return this.config.merge({
       entry: this.entry.toObject(),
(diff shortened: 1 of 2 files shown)


justin808:
@gauravtiwari @jakeNiemiec In order to create a server-bundle based on the rails/webpacker client configuration, take a look at what I had to do here:

https://github.com/shakacode/react_on_rails_tutorial_with_ssr_and_hmr_fast_refresh/blob/master/config/webpack/serverWebpackConfiguration.js#L9-L13 

The code in this file could be simplified if I could use new instances of the classes to modify.

I think the use of a global singleton variable of `Environment` violates the principle of least surprise.

Instead, the object should be created when used and passed around. While this is a simple change, it would require everybody upgrading to change from the current default of:


## environment.js
```js
```

to:

```js
const { createEnvironment } = require('@rails/webpacker')
const environment = createEnvironment()
```

To do this bigger change would require changing the:

1. default install files
2. documentation
3. remove the `environment` export from `@rails/webpacker` so anybody upgrading gets a very obvious error

justin808:
@gauravtiwari If you're working on a new API, please consider the case of building a derivation of the default, webpacker client bundle that can be used for server-side rendering.

Issue #2693, opened 2020-08-12
Latest release (5.1.1) is using dependency, which has older version of dependency with known issue


hasanen:
Latest release (5.1.1) is using 3.x branch of `compression-webpack-plugin`, which is using 2.x branch of `serialize-javascript`.  And now `yarn audit` it gives a notice from it. 

In `master` branch `compression-webpack-plugin` is updated to 4.x, which in other hand is using 3.x of `serialize-javascript` and thus has non-vulnerable version. 

Could you make a new release? Or is there a way where I/we could get that 2.x of `serialize-javascript` updated?

jipiboily:
The master branch seems to be for 6.0.0, which I assume is not ready.

Wondering if we can easily backport the dependency update from https://github.com/rails/webpacker/pull/2609 to 5.1.X (or 5.2.0).

akoskm:
4.2.2 has the same problem.

gauravtiwari:
Thanks for pointing out. I will check if I can create a 5.0 stable branch and selectively merge in changes. Reg: 4.2 you can make a PR against this branch to update deps: https://github.com/rails/webpacker/tree/4-x-stable

artob:
The security advisory: https://nvd.nist.gov/vuln/detail/CVE-2020-7660

The relevant upstream issues and commits:

- https://github.com/webpack-contrib/compression-webpack-plugin/issues/176

- https://github.com/yahoo/serialize-javascript/pull/79

- https://github.com/yahoo/serialize-javascript/commit/f21a6fb3ace2353413761e79717b2d210ba6ccbd

gauravtiwari:
Just realised 5.2.0

hasanen:
Thank you @gauravtiwari for the quick action <3

vitobotta:
Hi, I am using 5.2.1 and I still get the warning about serialize-javascript. I followed the instructions as per https://github.com/rails/webpacker to upgrade. What else can I do? Thanks!

hasanen:
@vitobotta have you checked that there are no other dependencies that are requiring the old version?

jaredbeck:
> @vitobotta have you checked that there are no other dependencies that are requiring the old version?

@vitobotta Try `yarn why` or `npm ls`

npearson72:
@gauravtiwari 

> Just realised 5.2.0. Please see 5-x-stable branch

I don't think this issue is resolved.

[...]

PR #2695, opened 2020-08-13
bump compression-webpack-plugin to ^4.0.0

akoskm:
https://github.com/rails/webpacker/issues/2693

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,13 @@
 language: ruby
 dist: xenial
 before_install:
-  - gem install rubygems-update && update_rubygems
   # Rails 4.2 doesn't support bundler 2.0, so we need to lock bundler to
   # v1.17.3. This is just for Ruby 2.5 which ships with bundler 2.x on Travis
   # CI while Ruby 2.6 does not.
   # https://github.com/travis-ci/travis-rubies/issues/57#issuecomment-458981237
-  - yes | rvm @global do gem install bundler -v 1.17.3 || true
+  - rvm gemset create travis
+  - rvm gemset use travis
+  - gem install bundler -v 1.17.3
 rvm:
   - 2.3.8
   - 2.4.6
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "babel-plugin-dynamic-import-node": "^2.3.0",
     "babel-plugin-macros": "^2.6.1",
     "case-sensitive-paths-webpack-plugin": "^2.2.0",
-    "compression-webpack-plugin": "^3.0.0",
+    "compression-webpack-plugin": "^4.0.0",
     "core-js": "^3.4.0",
     "css-loader": "^3.2.0",
     "file-loader": "^4.2.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -848,6 +848,13 @@
     "@types/istanbul-reports" "^1.1.1"
     "@types/yargs" "^13.0.0"
 
+"@npmcli/move-file@^1.0.1":
+  version "1.0.1"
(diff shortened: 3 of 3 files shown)


akoskm:
@gauravtiwari I made this PR on top of https://github.com/rails/webpacker/tree/4-x-stable as you suggested. The build failures are the same for each job https://travis-ci.org/github/rails/webpacker/jobs/717516894 did I mess up something?

DeeDeeG:
The build failure also happens on the stable branch this PR is made against.

https://travis-ci.org/github/rails/webpacker/builds/665541517

I'm confused what happens to cause this. CI updates bundler to latest (v2) and rubygems to latest, then uses that to install a copy of bundler v1.17.3, but Bundler 2 can't find bundler v1.3.0+?

https://travis-ci.org/github/rails/webpacker/jobs/665541518#L1044-L1055

Maybe trying to install Rails ~4.2.0 with bundlr v2 will always fail? In which case I suppose CI should ensure it does not install Bundler v2 on this branch. Not 100% sure, but maybe that's it.

DeeDeeG:
CI configs have a comment that is referring to this problem and explaining how it attempts to fix it.

https://github.com/rails/webpacker/blob/8db818bb214497a54791b9ebd2e604594ee0eca9/.travis.yml#L3-L9

The problem is, bundler v2 is used anyhow.

Thinking out loud:

I think `gem install rubygems-update && update_rubygems` is installing bundler v2. We don't want that, apparently.

According to [this StackOverflow answer](https://stackoverflow.com/a/37149817), it's probably better to do `gem update --system` if we need updated rubygems, then install bundler v1.17.3.

**_NOT_** `gem install rubygems-update && update_rubygems`.

(We may be able to skip updating Rubygems alotgether, but I'm not sure unless somebody tries it and it works).

DeeDeeG:
I have a fix that appears to be working.

commit: https://github.com/DeeDeeG/webpacker/commit/b21cb2536fe1f22a96d6ceb12d7c3b4be1232267

CI: https://travis-ci.com/github/DeeDeeG/webpacker/builds/179666602

DeeDeeG:
Unrelated to this PR (`compression-webpack-plugin` bump), there is a genuine install failure for the combination of Rails 5.1.x and Ruby 4.2.

There is an unconstrained `sprockets` dependency, which resolves to the latest `sprockets` (`sprockets` 4), which does require Ruby 2.5 or newer.

https://rubygems.org/gems/sprockets/versions/4.0.0

I'm suprised bundler isn't clever enough to just use an older `sprockets`? These are unsupported older versions of Rails _and_ Ruby, so I'm not sure being able to install Rails 5.1 under Ruby 4.2 is this repo's responsibility.

Still, I wonder if this would be fixed with latest `rubygems`. I might see if I can get that last failure to go away, but it's not related to this PR, or even this repo, not really.

akoskm:
@DeeDeeG thanks, it looks like a flake to me. There is a timeout error in the log:
```
Retrying download gem from https://rubygems.org/ due to error (2/4): Gem::RemoteFetcher::UnknownHostError timed out (https://rubygems.org/gems/unicode-display_width-1.5.0.gem)
```
probably Travis couldn't fetch a package during the build. Unfortunately, I cannot rerun the build.

gauravtiwari:
released: 4.3.0

ansonhoyt:
Why would this go into `4-x-stable` when it isn't until 5.0.0 that webpacker drops Node 8 support? This is a breaking change (though I admit Node 8 is now EOL). See [compression-webpack-plugin 4.0.0 changelog](https://github.com/webpack-contrib/compression-webpack-plugin/blob/master/CHANGELOG.md).

We have a few things running on older servers, so this change means we'll need to pin webpacker to `~4.2.0` (or mess with "resolutions" in package.json) until we get around to updating them.

PR #2696, opened 2020-08-13
Update babel.config.js

justin808:
Prettier would add a space after function with no function name before the parens.


gauravtiwari:
Fixed by #2633

Issue #2706, opened 2020-08-22
Deployment fails with webpacker 5.2.1 when `additional_paths` is `nil`


brandonzylstra:
The following commit causes deployment to fail

```
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -559,7 +559,7 @@ GEM
       nokogiri (~> 1.6)
       rubyzip (>= 1.3.0)
       selenium-webdriver (>= 3.0, < 4.0)
-    webpacker (5.1.1)
+    webpacker (5.2.1)
       activesupport (>= 5.2)
       rack-proxy (>= 0.6.1)
       railties (>= 5.2)
```

Here's the failure:
```
rake aborted!
NoMethodError: undefined method `+' for nil:NilClass
/var/www/rsp/shared/bundle/ruby/2.6.0/gems/webpacker-5.2.1/lib/webpacker/configuration.rb:31:in `additional_paths'
/var/www/rsp/shared/bundle/ruby/2.6.0/gems/webpacker-5.2.1/lib/webpacker/configuration.rb:35:in `additional_paths_globbed'
/var/www/rsp/shared/bundle/ruby/2.6.0/gems/webpacker-5.2.1/lib/webpacker/compiler.rb:93:in `default_watched_paths'
/var/www/rsp/shared/bundle/ruby/2.6.0/gems/webpacker-5.2.1/lib/webpacker/compiler.rb:57:in `watched_files_digest'
/var/www/rsp/shared/bundle/ruby/2.6.0/gems/webpacker-5.2.1/lib/webpacker/compiler.rb:38:in `fresh?'
/var/www/rsp/shared/bundle/ruby/2.6.0/gems/webpacker-5.2.1/lib/webpacker/compiler.rb:43:in `stale?'
/var/www/rsp/shared/bundle/ruby/2.6.0/gems/webpacker-5.2.1/lib/webpacker/compiler.rb:22:in `compile'
/var/www/rsp/shared/bundle/ruby/2.6.0/gems/webpacker-5.2.1/lib/webpacker/commands.rb:51:in `compile'
/var/www/rsp/shared/bundle/ruby/2.6.0/gems/webpacker-5.2.1/lib/webpacker.rb:35:in `compile'
/var/www/rsp/shared/bundle/ruby/2.6.0/gems/webpacker-5.2.1/lib/tasks/webpacker/compile.rake:23:in `block (4 levels) in <main>'
[...]

Issue #2699, opened 2020-08-17
Webpacker 5.2.0 error


kapso:
After upgrade to `5.2.0`, getting the following error when deploying my app.

```
NoMethodError: undefined method `+' for nil:NilClass
remote:        /tmp/build_9e71071d/vendor/bundle/ruby/2.7.0/gems/webpacker-5.2.0/lib/webpacker/configuration.rb:31:in `additional_paths'
remote:        /tmp/build_9e71071d/vendor/bundle/ruby/2.7.0/gems/webpacker-5.2.0/lib/webpacker/configuration.rb:35:in `additional_paths_globbed'
remote:        /tmp/build_9e71071d/vendor/bundle/ruby/2.7.0/gems/webpacker-5.2.0/lib/webpacker/compiler.rb:93:in `default_watched_paths'
remote:        /tmp/build_9e71071d/vendor/bundle/ruby/2.7.0/gems/webpacker-5.2.0/lib/webpacker/compiler.rb:57:in `watched_files_digest'
remote:        /tmp/build_9e71071d/vendor/bundle/ruby/2.7.0/gems/webpacker-5.2.0/lib/webpacker/compiler.rb:38:in `fresh?'
remote:        /tmp/build_9e71071d/vendor/bundle/ruby/2.7.0/gems/webpacker-5.2.0/lib/webpacker/compiler.rb:43:in `stale?'
remote:        /tmp/build_9e71071d/vendor/bundle/ruby/2.7.0/gems/webpacker-5.2.0/lib/webpacker/compiler.rb:22:in `compile'
```

gauravtiwari:
@kapso that's very strange since that change is a non-breaking one. I will take a look 👍

legendado:
@gauravtiwari Rename `resolved_paths => additional_paths` in `webpacker.yml` fix the problem, but is this correct?

gauravtiwari:
Yep, that's correct but the change was backwards-compatible since we have this defined as default in `webpacker.yml`: https://github.com/rails/webpacker/blob/master/lib/install/config/webpacker.yml#L13

legendado:
Seems like we have old configuration file and `additional_paths` was missing. Thanks for help :+1:

gauravtiwari:
Just checked, the upgrade was pretty smooth for me from 5.1.1. 

Makes sense, I was getting on the same conclusion. Perhaps, @kapso you might have some legacy settings in your config.

dzemlianoi-double:
[...]

PR #2707, opened 2020-08-22
handle nils in additional_paths method (#2699)

brandondrew:
There is a failing test in master—it's not failing based on anything I did in this branch:

```
F

Failure:
RakeTasksTest#test_rake_tasks [/Users/brandon/Code/Ruby/webpacker/test/rake_tasks_test.rb:6]:
Expected "" to include "webpacker".
```

This is meant to resolve the problem described in https://github.com/rails/webpacker/issues/2706.

This is the first time I've used Minitest, so if there is anything you'd like me to do differently, please let me know.  I'm happy to adapt style or techniques.

### DIFF
--- a/lib/webpacker/configuration.rb
+++ b/lib/webpacker/configuration.rb
@@ -28,7 +28,7 @@ class Webpacker::Configuration
   end
 
   def additional_paths
-    fetch(:additional_paths) + resolved_paths
+    (fetch(:additional_paths) || []) + (resolved_paths || [])
   end
 
   def additional_paths_globbed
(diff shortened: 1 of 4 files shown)


brandondrew:
Apparently this issue was already reported: https://github.com/rails/webpacker/issues/2699.

I'll change my title from #2706  to #2699.

gauravtiwari:
@brandondrew Thanks for the pr. 

I am actually trying to understand the exact problem because from the code it feels like, it should handle this behind the scenes by using entry from default webpacker.yml, if one is not present:

https://github.com/rails/webpacker/blob/master/lib/install/config/webpacker.yml#L13

and 

https://github.com/rails/webpacker/blob/bdb8673f8ae42af8b2d5c388c9938c3f3826a14f/lib/webpacker/configuration.rb#L87

I might be missing something.

PR #2709, opened 2020-08-24
Change @babel/preset-env modules to 'auto'

rossta:
@babel/preset-env already defaults to `'auto'` and encourages its use when using babel with a bundler such as webpack.

According to babel docs (source: https://babeljs.io/docs/en/babel-preset-env#modules)

> Setting this to false will preserve ES modules. Use this only if you intend to ship native ES Modules to browsers. If you are using a bundler with Babel, the default modules: "auto" is always preferred.

I believe this would also be [consistent with create-react-app](https://github.com/facebook/create-react-app/blob/edfc30abf91c1081bee3e50358f981cf3c6e1b05/packages/babel-preset-react-app/create.js#L78-L89) which appears to use the default.

Given this info, I believe Webpacker should follow suit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ environment.loaders.append('nodeModules', nodeModules)
 - If you have added `environment.loaders.delete('nodeModules')` to your `environment.js`, this must be removed or you will receive an error (`Item nodeModules not found`).
 - The install task will now set the `extract_css` default to `true` in all environments and generate a separate `application.css` file for the default `application` pack, as supported by multiple files per entry introduced in 5.0.0.  [#2608](https://github.com/rails/webpacker/pull/2608)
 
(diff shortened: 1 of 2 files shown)


justin808:
@gauravtiwari @rossta If the default is 'auto', why specify it?

create-react-app specifies less:

https://github.com/facebook/create-react-app/blob/42dcf79b43171449a116ddcb51dce7b043cdb1d2/packages/babel-preset-react-app/create.js#L80-L88

```js
        require('@babel/preset-env').default,
        {
          // Allow importing core-js in entrypoint and use browserlist to select polyfills
          useBuiltIns: 'entry',
          // Set the corejs version we are using to avoid warnings in console
          corejs: 3,
          // Exclude transforms that make all code slower
          exclude: ['transform-typeof-symbol'],
        },
```

PR #2716, opened 2020-08-29
Use Dart Sass by default

patsplat:
`@rails/webpacker` should switch to Dart Sass (aka the `sass` package).

First off, note that per [sass](https://sass-lang.com/dart-sass) "Dart Sass is the primary implementation of Sass".  The most recent versions of [sass-loader](https://github.com/rails/webpacker/pull/2689) also remark "prefer sass (dart-sass) by default, it is strongly recommended to migrate on sass (dart-sass)"

Next, `node-sass` has many open security issues with no remediations.  Per [snyk](https://snyk.io/vuln/npm:node-sass) as of today there are 4 high priority and 10 medium priority issues affecting all node-sass versions.  Even though one can install and configure `sass` directly, the presence of `node-sass` triggers security audits.

Finally if a user really wants to use `node-sass` it can still be [installed and configured](https://webpack.js.org/loaders/sass-loader/#implementation) in applications.

For these reasons, the `@rails/webpacker` package should choose the primary and more secure implementation of `sass` as the default dependency as recommended.


SampsonCrowley:
this seems to have been reverted.

The "current" package.json lists v5.1.1, and tags v5.1.1 - v5.2.1 all have node-sass

SampsonCrowley:
@nevermind, I was just confused by the fact that the "main" version is lower than the current tag version. this was added _after_ v5.2.1

@gauravtiwari is there a timeline on actually _releasing_ this update so people don't have to install python2 and use an old bug filled package when the fix has already been implemented?

ariccio:
bump? This still seems to be causing annoying downstream issues with node-gyp and such.

PR #2719, opened 2020-09-01
Add missing_entry_behavior configuration

stokarenko:
Introduces `missing_entry_behavior` setting to be set within `webpacker.yml`.
It behaves like [ActiveSupport::Deprecation.behavior](https://api.rubyonrails.org/classes/ActiveSupport/Deprecation/Behavior.html) and provides the options:

* 'raise' - default value keeping the existing behavior, raises the exception on a try to render the missing bundle;
* 'log' - warns the logger instead of raising the exception, `lookup!` method will return the name of bundle;
* 'silence' - does nothing more than returning the name of bundle as a result of `lookup!`.

Why do we need this:

1) `log` option can be useful for production environment;
2) `silence` option looks suitable for some test environments.

For instance, `silence` option  in combination with `compile: false` saves for us up to 2 minutes per every single CI build. We still keep the compilation tested by dedicated CI step for sure )

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ in which case you may not even need the asset pipeline. This is mostly relevant 
 - [Paths](#paths)
   - [Resolved](#resolved)
   - [Watched](#watched)
+- [Missing entry behavior](#missing-entry-behavior)
 - [Deployment](#deployment)
 - [Docs](#docs)
 - [Contributing](#contributing)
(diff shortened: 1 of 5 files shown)


stokarenko:
Hi all )

If this PR appears to be valuable - is possible to push the same against `v4-stable` branch? Ready to do that any moment )

PR #2729, opened 2020-09-16
Additional Paths turned into Resolved Paths.

markskayff:
For webpacker.yml I'm not seeing additional_paths option working on the latest versions.
I had to rename it to resolved_paths to work.

Here updated the README.md to reflect this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -370,12 +370,12 @@ If you have `hmr` turned to true, then the `stylesheet_pack_tag` generates no ou
 
 If you are adding Webpacker to an existing app that has most of the assets inside
 `app/assets` or inside an engine, and you want to share that
-with webpack modules, you can use the `additional_paths`
+with webpack modules, you can use the `resolved_paths`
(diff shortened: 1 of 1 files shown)


aried3r:
Hey @markskayff, I believe `additional_paths` is correct. It was changed from webpacker 5.1 to 5.2 I believe. See https://github.com/rails/webpacker/pull/2560.

v5.2.1 
https://github.com/rails/webpacker/blob/9ec5daabc5daea1c96f538d7af1f807ae9b077ec/lib/install/config/webpacker.yml#L11-L13

v5.1.1
https://github.com/rails/webpacker/blob/40a171021f6a89117aed1317957199cf2ca72b98/lib/install/config/webpacker.yml#L11-L13

markskayff:
Hi @aried3r. Thanks for that. I just noticed we are using an old one. We are on 4.3
For some reason I thought we were on the latest.

That makes sense. Thanks for the clarification.

PR #2733, opened 2020-09-18
add image_pack_path and image_pack_url

### LINKED ISSUE #2562: Consider adding an image_pack_path helper
This is very useful, for instance to write in inline CSS or to handcraft your own img tag.

Here's the (probably too simple) one I use:

```ruby
def image_pack_path(name)
  asset_pack_path "media/images/#{name}"
end
```

SampsonCrowley:
add expected path and url helpers for image pack assets

resolves #2562


docs/assets.md
```diff
@@ -108,10 +109,25 @@ require.context('../images', true)
 <img src="<%= asset_pack_path 'media/images/calendar.png' %>" />
 <% # => <img src="/packs/media/images/calendar-k344a6d59eef8632c9d1.png" /> %>
 
+<img src="<%= asset_pack_url 'media/images/calendar.png' %>" />
+<% # => <img src="https://example.com/packs/media/images/calendar-k344a6d59eef8632c9d1.png" /> %>
+
+<img src="<%= image_pack_path 'media/images/calendar.png' %>" />
```
jamiew, line 115: I'm not too familiar with how `resolve_path_to_image` works internally; is the `media/images` prefix required here? I found that slightly confusing since there is no `media` directory and you need to understand Webpack internals to know about that

SampsonCrowley: 1. `media` is the folder under public that webpack outputs compiled assets

2. The docs literally explain directly below that the `media/images` is automatically added if needed, just like it *already* does with `image_pack_tag`

3. I did not change a single thing with how assets are resolved, only added methods that use the exact same rules as `image_pack_tag` for lookup

I know this is coming off a bit standoffish, and am sorry about that, don't mean to be intentionally rude, but I'm low on time and I don't want this PR denied over comments on things that already exist

I recommend you open a PR to update the docs so you don't miss that fact if you didn't need or read the  [Upgrade Guide](https://github.com/rails/webpacker/blob/54c3ca9245e9ee330f8ca63b447c202290f7b624/docs/v4-upgrade.md#namespaces)

jamiew: No problem on terseness, I should've led that I am a big fan of the patch :) 

I saw the later examples without `media/*` which is the only reason I asked

SampsonCrowley: `image_pack_path` calls the same function as `image_pack_tag` to resolve image assets

That (already existing) resolution function first prepends `media/images` *if* that isn't in the current path, then if it doesn't exist there, it runs again looking only for the given path

gauravtiwari:
thanks both 🙏

SampsonCrowley:
@gauravtiwari any idea of the timeline on a new patch/minor release to go with it?

PR #2734, opened 2020-09-18
Keep backups, even when they're old.

braddunbar:
Recently, we encountered an issue with webpacker removing our previous assets during the webpacker:clean task, and not keeping the specified number of backups. We expected the [behavior from sprockets][sprockets], in which assets are kept if under age or within the count limit. However, webpacker only keeps assets if under age *AND* within the count limit.

I think this should work like sprockets and has perhaps been overlooked.

[sprockets]: https://github.com/rails/sprockets/blob/358f83ff09a77f69ac17543a9b1d127737060f00/lib/sprockets/manifest.rb#L258-L259

### DIFF
--- a/lib/webpacker/commands.rb
+++ b/lib/webpacker/commands.rb
@@ -23,7 +23,7 @@ class Webpacker::Commands
         .each_with_index
         .drop_while do |(mtime, _), index|
           max_age = [0, Time.now - Time.at(mtime)].max
-          max_age < age && index < count
+          max_age < age || index < count
         end
         .each do |(_, files), index|
           files.each do |file|


gauravtiwari:
thank you 🍰

Pezmc:
@gauravtiwari Can you point me to any docs on releasing webpacker updates? I'd love to help backport this bugfix to 5.X.X as we're currently using a custom fork to get this fix, and cut a release!

dkniffin:
@gauravtiwari @Pezmc +1. We just encountered this issue as well and it caused lots of missing assets on our prod env. I'd really like to see it back-ported to 5.X.X as well.

oboxodo:
+1 to backporting this to 5.x! we were bitten by this issue in production today!

Pezmc:
@oboxodo https://github.com/rails/webpacker/pull/2900#issuecomment-773190194

dkniffin:
It turns out our issue was a bit different than this. This issue certainly didn't help us, but wasn't the root cause of our issues. I've created a GH issue over at #2913 describing what we ran into.

PR #2737, opened 2020-09-22
Fix duplicated test names for image pack helpers

SampsonCrowley:
@gauravtiwari **emergency commit**

in my last pull request to add missing helper functions I completely missed that I accidentally left duplicated test names in the `helper_test.rb` when I copied the structure of the other asset path tests for the image pack versions

Also added the use of github relative links in docs so that instructions aren't pointing at incorrect assets because the link in the `assets` docs to an old version of the helper test file is how I came across the duplication when switching to the master version

If the docs updates needs to be a separate PR I can split them out, but the test file needs to be updated ASAP

### DIFF
--- a/docs/assets.md
+++ b/docs/assets.md
@@ -132,4 +132,4 @@ require.context('../images', true)
 <% # => <img src="/packs/media/images/calendar-k344a6d59eef8632c9d1.png" /> %>
 ```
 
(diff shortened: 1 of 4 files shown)


gauravtiwari (approved):

gauravtiwari:
Thank you

PR #2743, opened 2020-09-27
Only set writeToDisk for the webpack-dev-server

justin808:
Otherwise the manifest.json is written twice for every compilation.

See https://github.com/webdeveric/webpack-assets-manifest/issues/89
for more details.


justin808:
@jakeNiemiec @gauravtiwari any reason not to merge this?

Issue #2745, opened 2020-09-28
postcss-loader 4 support?


saiqulhaq:
when I updated my postcss-loader deps to 4 version
I got this error
![image](https://user-images.githubusercontent.com/1275215/94412273-ccab2f80-01a3-11eb-83cd-9732fff0314d.png)

could be because this one https://github.com/rails/webpacker/blob/d05068209544acb8cf32f10d6b722b72beed3314/package/utils/get_style_rule.js#L24 ?

PR #2749, opened 2020-10-02
Update postcss dependencies, drop EOL nodes

pustovalov:
closes: https://github.com/rails/webpacker/issues/2745

https://github.com/webpack-contrib/postcss-loader/blob/master/CHANGELOG.md
>the config option doesn't support Object type anymore, config.path and config.ctx options were removed

https://github.com/postcss/postcss-safe-parser/blob/master/CHANGELOG.md
https://github.com/postcss/postcss/blob/master/CHANGELOG.md

Node: 11, 13, no longer supported and not receiving security update

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ in which case you may not even need the asset pipeline. This is mostly relevant 
 
 - Ruby 2.4+
 - Rails 5.2+
-- Node.js 10.17.0+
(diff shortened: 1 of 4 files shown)


gauravtiwari:
Thanks @pustovalov Could you please rebase and I will merge, thank you.

pustovalov:
@gauravtiwari done

Issue #2241, opened 2019-08-19
Using json-loader is not a good example


awendt:
I followed the instructions in [Webpack: Loaders](https://github.com/rails/webpacker/blob/master/docs/webpack.md#loaders) and found the following steps are not working with `json-loader` anymore:

1. `yarn add json-loader`
2. call `environment.loaders.append` 

because Webpack supports JSON out-of-the-box since v3 (see https://github.com/webpack/webpack/issues/3363 and [release notes](https://webpack.js.org/migrate/3/#json-loader-is-not-required-anymore)). In fact, following all the steps results in breaking things.

Nevertheless, I learned something about Webpack, and I think that just removing those steps will confuse others who will try to add different loaders.

I think the docs should be changed to use another loader but I don't feel confident enough to do that myself.

smondal:
We can use any other loader  from the loader list in webpack except json-loader. Removed json-loader related entries from documents.
We can use coffee-loader to replacement with json-loader.

Here is the full loader list from webpack
https://webpack.js.org/loaders

PR #2752, opened 2020-10-05
Use file-loader instead of json-loader in loaders docs

basherru:
closes: #2241

updated version: https://github.com/rails/webpacker/blob/f6f99179683f4968094217d790eba812d86e4c73/docs/webpack.md#loaders


guilleiguaran:
Thanks for your contribution!!!

PR #2762, opened 2020-10-14
Use css-modules suggested localIdentName default

justin808:
From docs: https://github.com/webpack-contrib/css-loader#localidentname

localIdentName
Type: String Default: '[hash:base64]'

Recommendations:

use '[path][name]__[local]' for development
use '[hash:base64]' for production

Also, refactored check for productionEnv.

### DIFF
--- a/package/env.js
+++ b/package/env.js
@@ -6,7 +6,9 @@ const DEFAULT = 'production'
 const configPath = require('./configPath')
 
 const railsEnv = process.env.RAILS_ENV
-const nodeEnv = process.env.NODE_ENV
+const rawNodeEnv = process.env.NODE_ENV
+const nodeEnv = rawNodeEnv && NODE_ENVIRONMENTS.includes(rawNodeEnv) ? rawNodeEnv : DEFAULT
+const isProduction = nodeEnv === 'production'
 
 const config = safeLoad(readFileSync(configPath), 'utf8')
 const availableEnvironments = Object.keys(config).join('|')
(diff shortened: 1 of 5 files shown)


package/env.js
```diff
@@ -6,13 +6,17 @@ const DEFAULT = 'production'
 const configPath = require('./configPath')
 
 const railsEnv = process.env.RAILS_ENV
-const nodeEnv = process.env.NODE_ENV
+const rawNodeEnv = process.env.NODE_ENV
+const nodeEnv = rawNodeEnv && NODE_ENVIRONMENTS.includes(rawNodeEnv) ? rawNodeEnv : DEFAULT
+const isProduction = nodeEnv === 'production'
```
justin808, line 11: DRY up the code a bit, removing duplication for checking for the production NODE ENV.

package/utils/get_style_rule.js
```diff
@@ -14,7 +15,7 @@ const getStyleRule = (test, modules = false, preprocessors = []) => {
         sourceMap: true,
         importLoaders: 2,
         modules: modules ? {
-          localIdentName: '[name]__[local]___[hash:base64:5]'
+          localIdentName: isProduction ? '[hash:base64]' : '[path][name]__[local]'
```
justin808, line 18: The path of the resource is more useful than a hash in development.

https://github.com/webpack/loader-utils#interpolatename

```
[path] the path of the resource relative to the context query parameter or option.
```

package/utils/get_style_rule.js
```diff
@@ -28,7 +29,7 @@ const getStyleRule = (test, modules = false, preprocessors = []) => {
     ...preprocessors
   ]
 
-  const options = modules ? { include: /\.module\.[a-z]+$/ } : { exclude: /\.module\.[a-z]+$/ }
+  const options = modules ? { include: /\.module\.[a-z]+$/i } : { exclude: /\.module\.[a-z]+$/i }
```
justin808, line 32: The default for css-modules is to use a case-insensitive regexp.

gauravtiwari (approved):

PR #2771, opened 2020-10-30
Allow passing files to additional_files

palkan:
Currently, everything passed to `additional_files` array is converted into a directory glob pattern, so, it's not possible to add a single file from the project's root to the list of watched files.

For example, if you're using Tailwind, you're likely to have `<root>/tailwind.config.js` which contains important configuration. Currently, changes to this file are ignored when calculated the digest which could result in "Everything up-to-date..." while we must recompile the assets.


gauravtiwari:
thanks @palkan

PR #2772, opened 2020-11-08
Specify Dart Sass for sass-loader

pustovalov:
>Beware the situation when node-sass and sass were installed! By default the sass-loader prefers sass. In order to avoid this situation you can use the implementation option.

https://github.com/webpack-contrib/sass-loader#implementation


gauravtiwari:
thanks @pustovalov

PR #2785, opened 2020-11-21
bump gem to 5.2.1

masterkain:
switched from gem to github source and was surprised about the version number

also: should this be 6.0.0 instead of 5.2.1?

Screenshot 2020-11-21 at 08 00 59

### DIFF
--- a/lib/webpacker/version.rb
+++ b/lib/webpacker/version.rb
@@ -1,4 +1,4 @@
 module Webpacker
   # Change the version in package.json too, please!
-  VERSION = "5.1.1".freeze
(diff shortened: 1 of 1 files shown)


lib/webpacker/version.rb
```diff
@@ -1,4 +1,4 @@
 module Webpacker
   # Change the version in package.json too, please!
-  VERSION = "5.1.1".freeze
+  VERSION = "5.2.1".freeze
 end
```
majdi1401, line 4: yes

majdi1401 (approved):

majdi1401 (approved):

PR #2788, opened 2020-11-24
Fix for Webpacker not detecting up changes in Rails engines

### LINKED ISSUE #2487: Webpacker not detecting up changes in engines
I followed the instructions outlined in [webpacker/engines](https://github.com/rails/webpacker/blob/master/docs/engines.md).  rails asset precompile picks up and compiles files initially, but it does not detect and/or precompile changes unless I clobber the assets first and then precompile again. Webpacker is not detecting javascript changes in an engine.

Steps to reproduce:
1. run `bundle exec rake RAILS_ENV=production assets:precompile`
2. Make a change to a javascript file
3. run `bundle exec rake RAILS_ENV=production assets:precompile` 

After step 1, all javascript files are compiled
After step 3, nothing is compiled until after I clobber the assets.  It seems all files are cached.

Current workaround: clobber precompiled assets/packs and the recompile.

How can I configure webpacker on the engine to detect precompile changed files, while still caching the unchanged files.

krists:
Fixes #2487

### DIFF
--- a/lib/webpacker/compiler.rb
+++ b/lib/webpacker/compiler.rb
@@ -53,10 +53,11 @@ class Webpacker::Compiler
 
     def watched_files_digest
       warn "Webpacker::Compiler.watched_paths has been deprecated. Set additional_paths in webpacker.yml instead." unless watched_paths.empty?
-
-      files = Dir[*default_watched_paths, *watched_paths].reject { |f| File.directory?(f) }
-      file_ids = files.sort.map { |f| "#{File.basename(f)}/#{Digest::SHA1.file(f).hexdigest}" }
-      Digest::SHA1.hexdigest(file_ids.join("/"))
+      Dir.chdir File.expand_path(config.root_path) do
+        files = Dir[*default_watched_paths, *watched_paths].reject { |f| File.directory?(f) }
+        file_ids = files.sort.map { |f| "#{File.basename(f)}/#{Digest::SHA1.file(f).hexdigest}" }
+        Digest::SHA1.hexdigest(file_ids.join("/"))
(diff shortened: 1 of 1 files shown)


n-rodriguez:
I think it should be reverted since it's worse after than before (Ruby 3 compat) : https://github.com/rails/webpacker/issues/2801

krists:
Hi @n-rodriguez!

There are no warnings in webpacker build.

From: https://ruby-doc.org/core-3.0.0/Dir.html#method-c-5B-5D
```
chdir blocks can be nested, but in a multi-threaded program an error will be raised if a thread attempts to open a chdir block while another thread has one open or a call to chdir without a block occurs inside a block passed to chdir (even in the same thread).
```

Could it be that your application have multiple threads trying to compile assets at the same time?

n-rodriguez:
> Could it be that your application have multiple threads trying to compile assets at the same time?

Yeah. Actually it's due to an Ajax call that triggers a call to `asset_pack_path` on the Ruby side. Since I run Puma, it spawns a new thread to answer the Ajax request and triggers this warning. I solved it by removing the call to `asset_pack_path`.

Kukunin:
I believe the same solution can be implemented without using Dir.chdir which is not thread-safe. Something like:

```
root_path = File.expand_path(config.root_path)
files = Dir[*default_watched_paths, *watched_paths]
  .map { |path| root_path.join(path) }
  .reject { |f| File.directory?(f) }
```

jessewmc:
I have a working implementation of the above here in this PR: https://github.com/rails/webpacker/pull/3218

Issue #2751, opened 2020-10-05
Vue.js 3


tchukuchuk:
Hi,

Do you know where I can find a working example with vuejs 3 and typescript ?

Vannsl:
Hi @tchukuchuk,

just a couple of days ago, I added Vue 3 to a Rails app. I want to create a PR to install vue3 with webpacker and write an article/tutorial about it. Please give me 2-3 days to get that done.

Best
Vannsl

simonhutchings:
How did you add it without webpacker? That would work for me for the time being!

Vannsl:
https://dev.to/vannsl/vue3-on-rails-l9d

I just published this, I haven't checked for typos yet ;)
Typescript follows!

simonhutchings:
Great work!

guillaumebriday:
v3 installation shouldn't be different from v2

We could close this issue I think 👍

Vannsl:
@guillaumebriday I would suggest to leave it open. I wanted to create a PR to add the installer on webpacker. The installation is different from vue2. The packages are different and the command to create an app.

Vannsl:
(If you just want to close the issues for some reason, it's fine for me. I would still create the PR)

guillaumebriday:
I mean the configuration related to webpacker is the same. Only the [install](https://github.com/rails/webpacker/blob/master/lib/install/vue.rb) has changed.

Since the whole Vue ecosysteme is not ready for 3 yet, I suggest to add a `vue3` install option and keep the actual `vue` option for a bit.

Vannsl:
Hi @guillaumebriday, thanks for the quick reply. I actually propose the same. The vue3 ecosystem is not ready yet, therefore vue3 is behind vue@next - so a vue3-install-option is the way to go

guillaumebriday:
So cool! Thanks 👍

tchukuchuk:
Hi @Vannsl ,

Thanks a lot for your article !

Do you know when you'll publish the next one about Typescript support ?

Vannsl:
Hi @tchukuchuk, I plan it to do upcoming weekend. 

But as far as I understood it:

- Add a `tsconfig.json` in the root directory with settings like:

```json
{
  "compilerOptions": {
    "target": "esnext",
    "module": "esnext",
    "strict": true,
    "jsx": "preserve",
    "importHelpers": true,
    "moduleResolution": "node",
    "skipLibCheck": true,
    "esModuleInterop": true,
[...]

PR #2792, opened 2020-11-25
Add Vue3 install task

t0yohei:
ref #2751

Add task to use Vue 3 Installer.


t0yohei:
I close this because of the reason below.
https://github.com/rails/webpacker/issues/2751#issuecomment-734255808

Issue #2757, opened 2020-10-10
Webpack v5 is out!


pedrofurtado:
A great news, everyone! 🎉 

Webpack v5 is ready: https://github.com/webpack/webpack/releases/tag/v5.0.0

Happy to see upgrade plans to this version 🥇 

I can help with testing (not to much with code, in this particular case 😄)

addamove:
Waiting...

Ruff9:
It would help fix a lot of security issues introduced by webpack 4.44.2

gauravtiwari:
Thanks @pedrofurtado

I am working towards a 6.0 release that will include Webpack 5 support but we are still a couple of weeks away.

guillaumebriday:
@gauravtiwari If you need help, feel free to ping me 👍 

I tried to upgrade a webpack-boilerplate to Webpack 5, and most of plugins and loaders are not compatible yet :/

I guess we must wait a bit.

brentdodell:
@gauravtiwari, do you have a date that you'd like to have Webpack 5 support out? Are there a list of action items that need to happen to get it there?

gauravtiwari:
Thanks, @guillaumebriday will definitely do

@brentdodell 

The main goal with 6.0 is to make Webpacker leaner:
- Only JS support out of the box  (inline with https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#using-webpacker)
- Optional CSS/SASS support if dependencies are installed (`css-loader` or `sass-loader` or `less-loader`)
- Optional static assets support if `file-loader` is installed

*optional features are basically like you would add `haml-rails` gem to use haml templating engine*

which means we will move away from rake installers - erb, react etc. In my opinionm, installers don't do much anyway and anyone can simply follow the docs to add bunch of dependencies to get started (much better for learning). 

We will document all installer steps in the docs. And, you can also use NPM, if needed. 

---

I can't promise the exact date but Rails 6.1 RC is due soon (end of October/first week november) so hopefully around that time.

gauravtiwari:
@pedrofurtado wondering if we can remove the title of the issue to reflect Webpacker 6.0 instead of Webpack 5?

pustovalov:
one of the blockers:
 
`webpack-assets-manifest` - currently doesn't support webpack 5, https://github.com/webdeveric/webpack-assets-manifest/pull/59#issuecomment-707821385

guillaumebriday:
[...]

PR #2797, opened 2020-12-01
Removing installers

guillaumebriday:
Removing installers as suggested in https://github.com/rails/webpacker/issues/2757


guillaumebriday:
Did I forget something @gauravtiwari ? Or did something wrong? Maybe we should add docs to explain how to install these tools?

SleeplessByte:
Shouldn't the docs be updated to reflect _how_ to integrate with X now that there's no longer an installer? I think that was the gist of https://github.com/rails/webpacker/issues/2757#issuecomment-710784864.

gauravtiwari:
Closing in favour of #2802

PR #2798, opened 2020-12-01
Adding webp and avif support by default

guillaumebriday:


gauravtiwari:
Already included this in  #2802

PR #2799, opened 2020-12-02
Add support for Yarn v2

doits:
This should make Webpacker support both Yarn 1 and 2. It comes down to the two changes:

- Use correct `yarn install` flags
- Add `webpack` and `webpack-cli` as direct per project dependencies, otherwise Yarn 2 complains about not finding the binary `webpack`

In addition `nodeLinker: node-modules` must be added to `.yarnrc.yml` for maximum compatibility, but this is left as an exercise for the user who wants to experiment with Yarn 2 for now.

---

I think supporting both Yarn versions – if those are really the only two changes – should be no problem. Later one could force to only supprt Yarn 2, but Rails must gain support for Yarn 2 before. See https://github.com/rails/rails/pull/40722 for Yarn 2 support in Rails.


n-rodriguez:
@doits I've tested your branch and it works for me :+1:

n-rodriguez:
I'm waiting for https://github.com/rails/rails/pull/40722 to be merged but I'm not sure it will be in time for Rails 6.1 :/

doits:
https://github.com/rails/rails/pull/40722 is merged 👏  even into https://github.com/rails/rails/tree/6-1-stable – maybe this can be merged, too?

n-rodriguez:
Hi there! Any news?

stouset:
Yarn 3 is now out. :(

pedrofurtado:
@stouset A PR is already open for this purpose 🎉 🍻 : https://github.com/rails/webpacker/pull/3098

stouset:
Great news!

PR #2815, opened 2020-12-20
Fix references to webpack and webpacker versions

tagliala:
Webpacker 6.x supports webpack 5.x

[ci skip]


gauravtiwari:
thank you

PR #2820, opened 2020-12-22
More descriptive glob and no swallowed exceptions

justin808:
1. globSync is a clearer variable name than sync
2. Never swallow errors across multiple lines
3. Refactored duplicated code blocks regarding the check for
   non-installed packages
4. Removed reference to non-existent html-loader

The code for checking the css-loader exists comes is based on the API of
https://nodejs.org/api/modules.html#modules_require_resolve_request_options

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,7 +158,7 @@ can use the `asset_pack_path` helper:
 <img src="<%= asset_pack_path 'images/logo.svg' %>" />
 ```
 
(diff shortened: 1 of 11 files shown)


justin808:
@gauravtiwari @rossta I hate swallowed exceptions. 😄

gauravtiwari:
thanks @justin808

Issue #2775, opened 2020-11-08
How to customize babel config with new webpacker configuration?


pustovalov:
Hey,

injection of babel config recently was changed:

https://github.com/rails/webpacker/pull/2633/files#diff-5b9c334d542cd6b11f4035ee36a8630bc46c395443623be9cf4e7acc9b0751bcR45

>Running webpacker:install will now add browserslist and babel configuration into package.json instead of copying files.

How the user can now extend default babel configuration? For example to add `React Refresh Webpack Plugin
`
https://github.com/rails/webpacker/blob/master/docs/react.md#react-refresh-webpack-plugin

>Update babel.config.js adding
```
plugins: [
  process.env.WEBPACK_DEV_SERVER && 'react-refresh/babel',
  // other plugin
```

but new users will not have `babel.config.js`, `process.env` not available inside `package.json`

How the user can add some babel plugin for a specific environment?

after these changes default setup has been simplified, but it became more difficult to add something custom

tomprats:
This was definitely easier than I thought. I was worried I'd have to copy and paste the default setup, but you can just do something like this in your babel.config.js:

```
module.exports = function config(api) {
  plugins: [
    api.env('development') && 'react-refresh/babel',
  ].filter(Boolean),
  presets: [
    "./node_modules/@rails/webpacker/package/babel/preset.js"
  ]
}
```

Or if you don't need the environment, you can just use an object with whatever plugins

```
module.exports = {
  plugins: [
    "@babel/plugin-proposal-export-default-from"
  ],
  presets: [
    "./node_modules/@rails/webpacker/package/babel/preset.js"
  ]
}
```

justin808:
I think this is really a documentation issue.

I added a PR to address that: https://github.com/rails/webpacker/pull/3247.

PR #2826, opened 2020-12-23
Rewrite docs (part 1)

gauravtiwari:
This PR deletes old docs and adds a starting point for writing some fresh docs for version 6.x

- [x] Simple documentation
- [x] Don't duplicate webpack documentation since most of webpacker is webpack instead link
- [ ] Link to integrations (React on Rails etc.)


gauravtiwari:
cc// @rossta @justin808 @jakeNiemiec @guillaumebriday  Would be nice to get some feedback on this ^^

gauravtiwari:
@justin808 Please could you link to react on rails from Readme and other react-rails integrations.

README.md
```diff
@@ -248,7 +208,201 @@ WEBPACKER_DEV_SERVER_HOST=0.0.0.0 ./bin/webpack-dev-server
...
+#### Postcss
+
+```
+yarn add postcss-loader
+```
+
+#### Sass
+
+```
+yarn add sass-loader
```
pustovalov, line 336: @gauravtiwari should it be `yarn add sass-loader --dev`?

https://github.com/webpack-contrib/sass-loader#getting-started

and for other loaders

package/babel/preset.js
```diff
@@ -27,18 +28,28 @@ module.exports = function config(api) {
...
       'babel-plugin-macros',
-      [
-        '@babel/plugin-proposal-class-properties',
-        { loose: true }
-      ],
-      [
-        '@babel/plugin-transform-runtime',
-        { helpers: false }
-      ]
+      ['@babel/plugin-proposal-class-properties', { loose: true }],
```
pustovalov, line 46: it would also be nice to document how developer can customize this babel preset

https://github.com/rails/webpacker/issues/2775

README.md
gauravtiwari: Yes, this has been a historical confusion. I think we can say that if you are building locally then use `--dev` otherwise it should be added with main dependencies: https://github.com/rails/webpacker/issues/117

package/babel/preset.js
gauravtiwari: Should be easy, I think one can import and extend the base config just like webpack or just copy over and customise as needed.

docs/react.md
```diff
@@ -1,183 +0,0 @@
...
-           'ReactRefreshWebpackPlugin',
-           new ReactRefreshWebpackPlugin({
-               overlay: {
-                   sockPort: 3035
-               }
-           })
-       );
-   }
-   ```
-
```
pustovalov, line 114: this part was also helpful

README.md
```diff
@@ -158,31 +143,6 @@ can use the `asset_pack_path` helper:
...
-helper otherwise you will get duplicated chunks on the page.
-
-```erb
-<%# DO %>
-<%= javascript_packs_with_chunks_tag 'calendar', 'map' %>
-
-<%# DON'T %>
-<%= javascript_packs_with_chunks_tag 'calendar' %>
-<%= javascript_packs_with_chunks_tag 'map' %>
-```
```
pustovalov, line 184: can it now be used like this?

```
<%= javascript_packs_with_chunks_tag 'calendar' %>
<%= javascript_packs_with_chunks_tag 'map' %>
```

rossta: This is a good question. If it doesn't work currently, maybe we can figure out a way to make it work as expected. 

Also, if splitChunks is the default, perhaps we rename this method `javascript_pack_tag` and remove the old implementation?

README.md
jakeNiemiec: For this reason, the language here might be useful for the new docs: https://github.com/rails/webpacker/blob/fb9aa9622a7760cb6c21d304250e1c9b92b1a6ac/docs/yarn.md#add-an-npm-module-to-devdependencies

> Be careful not to add any build or app related JS modules in this fashion. Adding JS modules to `devDependencies` [will block them from being installed in **any** production environment](https://yarnpkg.com/lang/en/docs/cli/install/#toc-yarn-install-production-true-false).
>
> Docs from JS modules may instruct you to use `--dev` or `devDependencies`, but this is generally under the assumption that you are using a `node.js` workflow.

Issue #2832, opened 2020-12-24
Webpacker 6.x - Dev server not reloads automatically in browser (refreshing the webpage without press f5 key)


pedrofurtado:
Hey @gauravtiwari ! 🤝 

Once I updated to webpacker 6.x beta 2, the dev server not automatically reloads the browser. But if I revert to 5.x this works well. Any configuration about it was changed (new defaults, required external libs, and so on)?

pedrofurtado:
@gauravtiwari  I think this bug can in `webpacker` be related with theses facts:

1 - The `webpack-dev-server` is compatible with `webpack-cli v4` only in your 4.x version (today, the last release of `webpack-dev-server` in NPM is 4.0.0-beta.0).
2 - This PR ( https://github.com/rails/webpacker/pull/2783 ) changes the default behavior of dev server in `webpacker`, to use the `@webpack-cli/serve` for it (that behaves like a proxy to `webpack-dev-server`).
3 - But the current codebase of `webpacker` gem is only compatible with `webpack-dev-server` 3.x. In order to be compatible with `webpack-dev-server` 4.x, there are some breaking changes in configuration options. That's why, I supposed, the webpacker 6.x not works the live reload on browser.

🔍

pedrofurtado:
I am investigating what keys we will need to change here: https://github.com/rails/webpacker/blob/master/package/environments/development.js#L25-L52

But any help @gauravtiwari  will be awesome hehehe I am new in codebase of webpacker 🤝

pedrofurtado:
@gauravtiwari  Well, researching here, we need make webpacker compatible with webpack-dev-server v4 (even in beta), because with the compatibility of gem with webpack v5, the old webpack-dev-server v3 not works anymore (because it is not compatible with webpack v5) 🙈 

Ps: Merry Christmas for all 🎄 🎅 ☃️

apuntovanini:
Up, HMR seems broken with version 6

pedrofurtado:
@gauravtiwari 🤝

pedrofurtado:
@gauravtiwari 🤝

gauravtiwari:
Seems to broken for me too 😄  will take a look at this. Thanks for reporting @pedrofurtado

pedrofurtado:
@gauravtiwari 🤝 👍

justin808:
@gauravtiwari any progress on this? Need any help?

scottrobertson:
```
[...]

PR #2849, opened 2020-12-30
Update 6_0_upgrade.md: webpack-dev-server

rossta:
Depending on local installations, outdated versions of webpack-dev-server might not work with webpack 5. 

Uninstalling the current version is one way to help developers avoid running into issues when trying to run the dev server after upgrading since the install task will upgrade in this case.


pedrofurtado:
@rossta Is this PR related to this issue? Or not? -> https://github.com/rails/webpacker/issues/2832

🤝

PR #2839, opened 2020-12-27
Updated docs for 6.x

justin808:

Issue #202, opened 2017-03-29
Inline presets take precedence over .babelrc and it's not clear from the get-go.


sergiotapia:
Using Webpacker 1.1 and a brand new Rails 5.0.2.

Here's my simple React component for testing purposes.

```
import React, { Component, PropTypes } from 'react'
import DayPicker, { DateUtils } from 'react-day-picker'
import 'react-day-picker/lib/style.css'

export default class Example extends Component {
  constructor(props) {
    super(props)
  }

  sayHello = () => {
    console.log("Hello")
  }

  render() {
    return (
      <p onClick={this.sayHello}>Working</p>
    )
  }
}
`
```

The error I get is:

```
21:21:01 hot.1  | ERROR in ./app/javascript/packs/components/example/example.js
21:21:01 hot.1  | Module build failed: SyntaxError: Missing class properties transform.
21:21:01 hot.1  |
21:21:01 hot.1  |   11 |   }
21:21:01 hot.1  |   12 |
21:21:01 hot.1  | > 13 |   sayHello = () => {
21:21:01 hot.1  |      |   ^
```

My .babelrc file looks like:

```
{
  "presets": ["es2015", "stage-2", "env", "react"]
}
```

My package.json

```
{
  "dependencies": {
    *snip*
  },
  "devDependencies": {
    "babel-preset-es2015": "^6.24.0",
    "babel-preset-stage-2": "^6.22.0",
    "webpack-dev-server": "^2.4.2"
  }
}

```

Any ideas on how to make the class arrow functions working with Webpacker? It's like the stage-2 preset isn't being applied proper and I'm out of ideas.

Thanks for the help!

sergiotapia:
To fix, I had to add the presets to the individual loader files, specifically.

```
// config/webpack/loaders/babel.js
module.exports = {
  test: /\.js(\.erb)?$/,
  exclude: /node_modules/,
  loader: 'babel-loader',
  options: {
    presets: [
      'es2015',
      'stage-2'
    ]
  }
}

// config/webpack/loaders/react.js
module.exports = {
  test: /\.(js|jsx)?(\.erb)?$/,
  exclude: /node_modules/,
  loader: 'babel-loader',
  options: {
    presets: [
      'es2015',
      'react',
      'stage-2'
    ]
  }
}
```

Now I can use the nice shorthand arrow class functions.

gauravtiwari:
[...]

PR #2850, opened 2020-12-30
modify WEBPACKER PARALLEL environment variable.

masato-hi:
Will be possible to specify the WEBPACKER PARALLEL environment variable from the command line.

We also added a document about WEBPACKER_PARALLEL.

The master branch is currently working on 6.x, so we need to merge after merging this pull request.
https://github.com/rails/webpacker/pull/2839

I'm not good at English, so I'm looking for opinions on better documentation.

ref: https://github.com/rails/webpacker/pull/2093

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -111,6 +111,19 @@ mv $HOME/your_rails_app/node_modules/.bin/elm-make $HOME/your_rails_app/node_mod
 printf "#\041/bin/bash\n\necho \"Running elm-make with sysconfcpus -n 2\"\n\n$HOME/dependencies/sysconfcpus/bin/sysconfcpus -n 2 $HOME/your_rails_app/node_modules/.bin/elm-make-old \"\$@\"" > $HOME/your_rails_app/node_modules/.bin/elm-make
 chmod +x $HOME/your_rails_app/node_modules/.bin/elm-make
 ```
+## If an error occurs or times out during a production build on Continuous Integration (CI) services such as CircleCI, CodeShip, Travis CI
+
+The parallel option of Tersor-Plugin is enabled in the production build, which may cause memory shortage.
+To solve this problem, you can specify the WEBPACKER_PARALLEL environment variable.
+Possible values ​​are true, false, or a number.
+
+```
+RAILS_ENV=production WEBPACKER_PARALLEL=false bundle exec rake webpacker:compile
+```
+
+Check the link for details.
+
+https://webpack.js.org/plugins/terser-webpack-plugin/#parallel
 
 ## Rake assets:precompile fails. ExecJS::RuntimeError
 This error occurs because you are trying to minify by terser a pack that's already been minified by Webpacker. To avoid this conflict and prevent appearing of ExecJS::RuntimeError error, you will need to disable uglifier from Rails config:
--- a/package/utils/helpers.js
+++ b/package/utils/helpers.js
@@ -1,5 +1,6 @@
 const isArray = (value) => Array.isArray(value)
 const isBoolean = (str) => /^true/.test(str) || /^false/.test(str)
+const isNumber = (str) => /^-?\d+(\.\d+)?$/.test(str)
 const chdirTestApp = () => {
   try {
     return process.chdir('test/test_app')
@@ -39,11 +40,24 @@ const canProcess = (rule, fn) => {
   return null
 }
 
+const asBoolean = (str) => {
+  if (/^true/.test(str)) return true
+  return false
+}
+
+const asBooleanOrNumber = (value) => {
+  if (isBoolean(value)) return asBoolean(value)
+  if (isNumber(value)) return Number(value)
+  throw new Error(`${value} is not boolean or numeric strings`)
+}
+
 module.exports = {
+  asBooleanOrNumber,
   chdirTestApp,
   chdirCwd,
   isArray,
   isBoolean,
+  isNumber,
   ensureTrailingSlash,
   canProcess,
   moduleExists,
--- a/lib/webpacker/compiler.rb
+++ b/lib/webpacker/compiler.rb
@@ -113,6 +113,7 @@ class Webpacker::Compiler
 
       env.merge("WEBPACKER_ASSET_HOST"        => ENV.fetch("WEBPACKER_ASSET_HOST", ActionController::Base.helpers.compute_asset_host),
                 "WEBPACKER_RELATIVE_URL_ROOT" => ENV.fetch("WEBPACKER_RELATIVE_URL_ROOT", ActionController::Base.relative_url_root),
-                "WEBPACKER_CONFIG" => webpacker.config_path.to_s)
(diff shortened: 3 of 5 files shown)


package/utils/helpers.js
```diff
@@ -37,11 +38,24 @@ const canProcess = (rule, fn) => {
   return null
 }
 
+const asBoolean = (str) => {
```
gauravtiwari, line 41: thanks @masato-hi Could you please use the above function `isBoolean` (line 2)?

masato-hi: Thanks for review.
Cannot pass `"false"` in the `isBoolean` function.

pedrofurtado:
Hey, @masato-hi ! 👋 

Thanks for your contribution!

For this purpose (modify WEBPACKER_PARALLEL environment variable for a specific environment, for example), this works for your scenario?

- Edit your `config/webpack/[ENVIRONMENT].js` and put at the top the desired value of `WEBPACKER_PARALLEL`, like this snippet code below:
```javascript
process.env.WEBPACKER_PARALLEL = 'desired-value'
// ... existing code of file ...
```

Or this not works for you, and the PR in this case solves some problem related to it?

Let me know what you think about 🤝 🍻

masato-hi:
@pedrofurtado Thanks for confirming.

It looks like it's working fine in my environment. :thinking:  (The `sed` command installed on MacOSX may not execute the following code successfully.)

```
rails new myapp --webpack
cd myapp
sed -i -e "s/gem 'webpacker', '~> 5.0'/gem 'webpacker', github: 'masato-hi\/webpacker', branch: 'modify-webpacker-parallel-env'/" Gemfile
bundle install
yes | bundle exec rake webpacker:install
sed -i -e 's/"@rails\/webpacker": ".*"/"@rails\/webpacker": "masato-hi\/webpacker#modify-webpacker-parallel-env"/' package.json
yarn
echo "
process.env.NODE_ENV = process.env.NODE_ENV || 'production'
process.env.WEBPACKER_PARALLEL = '2'

const webpackConfig = require('./base')

console.log(webpackConfig.optimization.minimizer[0])
module.exports = webpackConfig
" > config/webpack/production.js
rm .browserslistrc # cheap tricks for "Error [BrowserslistError]: contains both .browserslistrc and package.json with browsers"
RAILS_ENV=production bundle exec rake webpacker:clean webpacker:compile
```

Output:
```
Compiling...
Compiled all packs in /tmp/myapp/public/packs
TerserPlugin {
  options: {
    test: /\.[cm]?js(\?.*)?$/i,
    extractComments: true,
    parallel: 2,
    include: undefined,
    exclude: undefined,
    minify: undefined,
    terserOptions: {
      parse: [Object],
      compress: [Object],
      mangle: [Object],
      output: [Object]
    }
  }
}
asset manifest.json 23 bytes [emitted]
webpack 5.21.2 compiled successfully in 26 ms
```
The value of "parallel" is 2.

The configuration only affects production builds.
You also need to set process.env.WEBPACKER_PARALLEL before `require('@rails/webpacker')` / `require('./ base')` is executed.

masato-hi:
@gauravtiwari 
The changes in docs/troubleshooting.md have been merged into master and are now reflected.
What do you think?

docs/troubleshooting.md
```diff
@@ -111,6 +111,19 @@ mv $HOME/your_rails_app/node_modules/.bin/elm-make $HOME/your_rails_app/node_mod
...
+To solve this problem, you can specify the WEBPACKER_PARALLEL environment variable.
+Possible values ​​are true, false, or a number.
+
+```
+RAILS_ENV=production WEBPACKER_PARALLEL=false bundle exec rake webpacker:compile
+```
+
+Check the link for details.
+
+https://webpack.js.org/plugins/terser-webpack-plugin/#parallel
```
olleolleolle, line 126:
```suggestion
[`terser-webpack-plugin` documentation](https://webpack.js.org/plugins/terser-webpack-plugin/) has more details.
```

docs/troubleshooting.md
```diff
@@ -111,6 +111,19 @@ mv $HOME/your_rails_app/node_modules/.bin/elm-make $HOME/your_rails_app/node_mod
 printf "#\041/bin/bash\n\necho \"Running elm-make with sysconfcpus -n 2\"\n\n$HOME/dependencies/sysconfcpus/bin/sysconfcpus -n 2 $HOME/your_rails_app/node_modules/.bin/elm-make-old \"\$@\"" > $HOME/your_rails_app/node_modules/.bin/elm-make
 chmod +x $HOME/your_rails_app/node_modules/.bin/elm-make
 ```
+## If an error occurs or times out during a production build on Continuous Integration (CI) services such as CircleCI, CodeShip, Travis CI
+
+The parallel option of Tersor-Plugin is enabled in the production build, which may cause memory shortage.
```
olleolleolle, line 116:
```suggestion
The parallel option of TerserWebpackPlugin is enabled in the production build, which may cause memory shortage.
```

docs/troubleshooting.md
```diff
@@ -111,6 +111,19 @@ mv $HOME/your_rails_app/node_modules/.bin/elm-make $HOME/your_rails_app/node_mod
 printf "#\041/bin/bash\n\necho \"Running elm-make with sysconfcpus -n 2\"\n\n$HOME/dependencies/sysconfcpus/bin/sysconfcpus -n 2 $HOME/your_rails_app/node_modules/.bin/elm-make-old \"\$@\"" > $HOME/your_rails_app/node_modules/.bin/elm-make
 chmod +x $HOME/your_rails_app/node_modules/.bin/elm-make
 ```
+## If an error occurs or times out during a production build on Continuous Integration (CI) services such as CircleCI, CodeShip, Travis CI
```
olleolleolle, line 114:
```suggestion
## If an error occurs or times out during a production build
```

olleolleolle (commented):
I made some inline suggestions, hope this helps.

justin808:
@guillaumebriday, For sure, we should do something with PR. Running out of memory causes a very confusing build just seems to hang when it's swapping memory.

I don't like the current default: 
>        parallel: Number.parseInt(process.env.WEBPACKER_PARALLEL, 10) || true,

Defaults should be defaults, which is `true` for this option.

How about this:

1. default to true
2. all env override of process.env.WEBPACKER_PARALLEL to either be `false` or some lower number to set on CI machines?

https://webpack.js.org/plugins/terser-webpack-plugin/#parallel

> parallel
> Type: Boolean|Number Default: true
> 
> Use multi-process parallel running to improve the build speed. Default number of concurrent runs: os.cpus().length - 1.
> 
> ℹ️ Parallelization can speedup your build significantly and is therefore highly recommended.
> 
> ⚠️ If you use Circle CI or any other environment that doesn't provide real available count of CPUs then you need to setup explicitly number of CPUs to avoid Error: Call retries were exceeded (see #143, #202).

justin808:
Feel free to reopen this PR here: 
https://github.com/shakacode/shakapacker/pulls

PR #2863, opened 2021-01-05
Better group assets for clean

zunda:
For cleaning, group assets by paths with removing the hash part and
sort them with mtime within each group to determine their versions.


zunda:
This bit me again when deploying an app to Heroku where writing assets to the file system may take more than 2 seconds which resulted with some of latest assets being removed with `rake assets:clean`.

PR #2802, opened 2020-12-06
Compatibility with Webpack 5

gauravtiwari:
### 1. Automatic rules

See the approach in: `package/rules/index.js`

The idea is that a rule is only loaded if you have declared the relevant dependency in package.json file. This aligns webpacker with the intention of its creation i.e. to only use it for JS compilation. No need to keep extra deps around. 

There is still some work left to do to ensure errors are more clear and concise. 
Please feel free to add comments.  

### 2. Simpler webpack config 

Change from class-based config files to regular object-based config 
as seen in webpack documentation to make learning and customisation easier 
for everyone.

### 3. And, webpack 5

Obviously, ensuring the latest release is compatible with Webpack 5.0

Hopefully, I will get this ready and merged this week and then we can push out 6.0 release

Supersedes #2797 
See #2757 

## Out of the box compatibility checklist: 

- [x] JS
- [x] CSS
- [x] SASS
- [x] Tailwindcss
- [x] React
- [x] Assets
- [x] SVG

PR #2877, opened 2021-01-13
Constrain supported node versions

### LINKED ISSUE #2874: Webpacker 5.2.1 compilation issues with Node 15?
This follows up on https://github.com/rails/rails/issues/41060.

On a macbook with Node 15.5.1 and Rails 6.1.0 `rails new` will create an app that fails to start.

According to the Prerequisites in Webpacker's README, supported Node versions are `Node.js 10.22.1+ || 12+ || 14+`. Does this include or exclude Node 15?

If Node 15 is supported, it should be added to the CI?

If Node 15 is not supported, webpacker should raise an error upon startup saying it's not compatible with Node 15?

rossta:
Fixes #2874

Until Webpacker uses node-sass 5.0+, node version 15 isn't supported.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "lib/install/config/webpacker.yml"
   ],
   "engines": {
-    "node": ">=10.22.1 || ^12 || >=14",
+    "node": ">=10.22.1 || ^12 || >=14 || <15",
     "yarn": ">=1 <3"
   },
   "dependencies": {


gmcgibbon:
I think there's been a misunderstanding here.

Webpacker is compatible with Node 15, and the issue this was originally linked to was related to node-sass's use of node-gyp (which requires python to build native extensions for node-sass). Node-sass isn't on master anymore (see https://github.com/rails/webpacker/pull/2772). Sass support was also made optional in https://github.com/rails/webpacker/pull/2802, so any dart-related issues should be sidestepped by default too. 

The changeset I had in mind was more like this: https://github.com/gmcgibbon/webpacker/commit/310d3c63f39ef9d3fcea0f52560105824b8dd9c9

I think we can add Node 15 to CI (simply because it is good to test webpacker under the latest Node) and revert this patch.

rossta:
@gmcgibbon Thanks for the clarification and the quick fix.

PR #2890, opened 2021-01-15
Dev server detection workaround

crawler:
Not sure how it works for everyone else. But as I can see, the webpack-cli@4.2.0 requires the webpack-dev-server package after our development.js file is loaded, so process.env.WEBPACK_DEV_SERVER variable is not yet available. This resulting devServer config section skipped when bin/webpack-dev-server is executed without manually setting env variables.

Looks like this problem is solved in webpack-cli@4.3.0 with a WEBPACK_SERVE environment variable (webpack/webpack-cli/pull/2027). But because the master branch is linked to 4.2.0 I came up with this solution. And I think that this is maybe a good idea to not rely on the environment variable that is set somewhere in the internals of the other project when it can be coupled within one repo.

### DIFF
--- a/lib/install/config/webpacker.yml
+++ b/lib/install/config/webpacker.yml
@@ -25,6 +25,8 @@ development:
     host: localhost
     port: 3035
     public: localhost:3035
+    # Inject browserside javascript that required by both HMR and Live(full) reload
+    inject_client: true
     # Hot Module Replacement updates modules while the application is running without a full reload
     hmr: false
     # Inline should be set to true if using HMR; it inserts a script to take care of live reloading
--- a/lib/webpacker/dev_server_runner.rb
+++ b/lib/webpacker/dev_server_runner.rb
@@ -64,6 +64,7 @@ module Webpacker
       def execute_cmd
         env = Webpacker::Compiler.env
         env["WEBPACKER_CONFIG"] = @webpacker_config
+        env["WEBPACK_DEV_SERVER"] = 'true'
 
         cmd = if node_modules_bin_exist?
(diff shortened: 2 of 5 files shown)


pedrofurtado:
@crawler A doubt: is this PR related to this issue https://github.com/rails/webpacker/issues/2832 ? Or not? 🤝

crawler:
@pedrofurtado 
Partially, both issues are caused by migration to the newer webpack-dev-server / webpack cli. My pull request does not solve the live reload issue, it only ensures that our dev server configuration is not skipped when you run bin/webpack-dev-server command. Exactly in my case, even port was 8080

pedrofurtado:
@crawler Thanks for clarification! Anyway, thanks for your contribution in this PR!🤝 👍

pedrofurtado:
@gauravtiwari Your eyes/reviews on it will be awesome 🎉

gauravtiwari:
Thanks both, I looked into it last week and this issue is related to: https://github.com/webpack/webpack-cli/issues/1973 but wasn't sure about the exact solution the problem since webpack dev server 4 is still unreleased. Also, this PR adds a new env variable: https://github.com/webpack/webpack-cli/pull/2027

I need to dig deeper to understand what's going on and how this is changed in 4.0

gauravtiwari:
@crawler Yes, that makes sense and I was thinking along the same lines to add our own env variable inside webpack dev server binstub.

crawler:
Hi @gauravtiwari, thank you for looking into this. Sorry for messing up with this PR. I created a new one #2898

Issue #2338, opened 2019-10-30
WebPackerClean#clean does not remove files that are not in manifest


spaghetticode:
https://github.com/rails/webpacker/blob/23d5fdd66b0184ac08fb0fce855c813b4a3ee36d/lib/webpacker/commands.rb#L8

I think there's a minor issue with this method. 

The code tracks files that should be removed by using the assets manifest, and removes old versions of still-existing assets. But if a source file was removed and it is not referenced anymore in the assets manifest, its compiled files will be never deleted. Same goes for arbitrary files that may end up in the folder. 

Changing specularly the way that the code works, ie keep only files that are listed in the manifest and their more recent versions won't be optimal either,  as this would also remove compiled files whose sources have been recently removed and were supposed to be kept.

spaghetticode:
@justin808 I've seen this is closed now, can you confirm this is not an issue anymore?

justin808:
@spaghetticode Definitely, I have not confirmed. However, the code changed significantly since you last posted this issue.

Can you take a look and confirm?

PR #2892, opened 2021-01-18
webpack-dev-server live reload issue fix

### LINKED ISSUE #2832: Webpacker 6.x - Dev server not reloads automatically in browser (refreshing the webpage without press f5 key)

crawler:
In combination with #2890 fixes #2832.
Since webpack 5, the default compiler target is browserslist (when available), and it is not being recognized by the webpack-dev-server as a target for client injection by default. Because it searches through:

`
[
          'web',
          'webworker',
          'electron-renderer',
          'node-webkit',
          undefined,
          null
        ]
`

So, even this can be a webpack-dev-server issue, we need to set this setting explicitly to have Live Reload or HMR working.

### DIFF
--- a/lib/install/config/webpacker.yml
+++ b/lib/install/config/webpacker.yml
@@ -25,6 +25,8 @@ development:
     host: localhost
     port: 3035
     public: localhost:3035
+    # Inject browserside javascript that required by both HMR and Live(full) reload
+    inject_client: true
     # Hot Module Replacement updates modules while the application is running without a full reload
     hmr: false
     # Inline should be set to true if using HMR; it inserts a script to take care of live reloading
--- a/package/__tests__/development.js
+++ b/package/__tests__/development.js
@@ -22,7 +22,8 @@ describe('Development environment', () => {
       expect(webpackConfig).toMatchObject({
         devServer: {
           host: 'localhost',
-          port: 3035
+          port: 3035,
+          injectClient: true
         }
       })
     })
--- a/package/environments/development.js
+++ b/package/environments/development.js
@@ -34,6 +34,7 @@ if (
       hot: devServer.hmr,
       contentBase,
       inline: devServer.inline,
+      injectClient: devServer.inject_client,
       useLocalIp: devServer.use_local_ip,
       public: devServer.public,
       publicPath,


pedrofurtado:
@gauravtiwari Happy to this reviewed and merged asap 🎉

crawler:
@gauravtiwari 

For the webpack-dev-server 3 the automatic client injection is made here:
https://github.com/webpack/webpack-dev-server/blob/7e70eeea457a6a154709bd6069c05855f12f7842/lib/utils/addEntries.js#L130

This works well with webpack 4 as the config.target is 'web' and the webTarget var becomes true.
In the webpack-dev-server 4, this code is moved to a different place, but the behavior is exactly the same:
https://github.com/webpack/webpack-dev-server/blob/master/lib/utils/DevServerPlugin.js#L142

As far as I can see, this "target" check was introduced to solve problems building "isomorphic" applications, when the server and the client-side build is going in the same env.
https://github.com/webpack/webpack-dev-server/pull/1775

So if I understand this correctly, this is completely not related to the apps with the ruby backend. And the injectClient setting can be forced without any side-effects.

pedrofurtado:
@gauravtiwari Happy to this reviewed and merged asap 🎉

pedrofurtado:
@crawler This PR targets compatibility with webpack-dev-server v3, v4 or both? 🤝

crawler:
@pedrofurtado Both

pedrofurtado:
@crawler Strange 🤔  When i set in Gemfile and package.json the webpacker getting from `https://github.com/crawler/webpacker -> branch fix-live-reloading`, and executing the webpack-dev-server, the port is always set as 8080, instead of 3035 as defined in webpacker.yml

crawler:
Plase try the master branch

pedrofurtado:
@crawler Now works like a charm! 🎉 🎉  @gauravtiwari I can confirm this PR works well with latest stable version of webpack-dev-server and this gem itself.

Thanks for your contribution @crawler 🤝 !

crawler:
@pedrofurtado thank you for checking this out 🤝

gauravtiwari:
thanks, both 🙏

justin808:
@gauravtiwari @pedrofurtado this does not work for HMR of CSS modules. More details in #2338.

PR #2897, opened 2021-01-19
Updating the default source_path

guillaumebriday:
From `app/javascript` to `app/packs`

PR #2901, opened 2021-01-21
Fix installed file does not match configuration source path in yml file

Zhong-z:
The new default source_path has been updated from `app/javascript` to `app/packs` via #2897.

However the installation template was still referencing the old path.

This PR should fix it.


lib/install/template.rb
```diff
@@ -8,7 +8,7 @@
   say "The JavaScript app source directory already exists"
```
guillaumebriday, line 8:
```diff
-   say "The JavaScript app source directory already exists"
+   say "The packs app source directory already exists"
```

guillaumebriday:
Nice catch, my bad!

lib/install/template.rb
Zhong-z: Thanks. updated via e91a315, also updated the `Javascript` to `packs` in line 10

Issue #2835, opened 2020-12-25
Webpacker 6.0.0.beta.2 - vue-loader doesn't work


holidayworking:
I'm trying Webpacker 6.0.0.beta.2 and I'm getting the following error:

```bash
$ ./bin/webpack
warning package.json: No license field
[webpack-cli] Error: [VueLoaderPlugin Error] No matching use for vue-loader is found.
Make sure the rule matching .vue files include vue-loader in its use.
    at VueLoaderPlugin.apply (/Users/hidekazu/workspace/webpacker-6.0.0-vue-bug-report-app/node_modules/vue-loader/lib/plugin-webpack5.js:93:13)
    at createCompiler (/Users/hidekazu/workspace/webpacker-6.0.0-vue-bug-report-app/node_modules/webpack/lib/webpack.js:69:12)
    at create (/Users/hidekazu/workspace/webpacker-6.0.0-vue-bug-report-app/node_modules/webpack/lib/webpack.js:113:15)
    at webpack (/Users/hidekazu/workspace/webpacker-6.0.0-vue-bug-report-app/node_modules/webpack/lib/webpack.js:121:46)
    at f (/Users/hidekazu/workspace/webpacker-6.0.0-vue-bug-report-app/node_modules/webpack/lib/index.js:37:15)
    at WebpackCLI.createCompiler (/Users/hidekazu/workspace/webpacker-6.0.0-vue-bug-report-app/node_modules/webpack-cli/lib/webpack-cli.js:1093:24)
    at async WebpackCLI.bundleCommand (/Users/hidekazu/workspace/webpacker-6.0.0-vue-bug-report-app/node_modules/webpack-cli/lib/webpack-cli.js:1217:20)
    at async Command.<anonymous> (/Users/hidekazu/workspace/webpacker-6.0.0-vue-bug-report-app/node_modules/webpack-cli/lib/webpack-cli.js:271:21)
    at async Promise.all (index 1)
    at async Command.<anonymous> (/Users/hidekazu/workspace/webpacker-6.0.0-vue-bug-report-app/node_modules/webpack-cli/lib/webpack-cli.js:652:13)
```

I created a [sample app](https://github.com/holidayworking/webpacker-6.0.0-vue-bug-report-app) to report this issue.
The files related to webpacker are as follows.
- package.json
    ```javascript
    {
      "dependencies": {
        "@rails/webpacker": "^6.0.0-beta.2",
        "vue-loader": "^15.9.6",
[...]

PR #2915, opened 2021-02-05
Update docs: run vue-loader before asset/source

Juice10:
Stopgap until https://github.com/rails/webpacker/issues/2835 is fixed.
Credits https://github.com/rails/webpacker/issues/2835#issuecomment-759772592


Juice10:
This stopgap triggers the following issue: https://github.com/rails/webpacker/issues/2835#issuecomment-759790323

PR #2919, opened 2021-02-08
Properly look up entry in manifest when name is given with extension

leonid-shevtsov:
There was a fairly obvious (once you encounter it) bug with separating extension from entry name: `File.basename` requires not just the extension, but the full suffix, dot included:

```
> File.basename('file.rb', 'rb')
=> "file."
> File.basename('file.rb', '.rb')
=> "file"
```

(I've encountered the entry name with extension [in ActiveAdmin](https://github.com/activeadmin/activeadmin/blob/a8e9d2f78960968c2c9409b2b7e3c5cf1b5c38ba/lib/active_admin/application.rb#L167))

### DIFF
--- a/lib/webpacker/manifest.rb
+++ b/lib/webpacker/manifest.rb
@@ -92,7 +92,7 @@ class Webpacker::Manifest
     # When the user provides a name with a file extension, we want to try to strip it off.
     def manifest_name(name, pack_type)
       return name if File.extname(name.to_s).empty?
-      File.basename(name, pack_type)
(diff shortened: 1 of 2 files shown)


pedrofurtado:
Hey, @leonid-shevtsov ! 👋 

Thanks for contribution 👍 

That is really a tricky bug. The PR is ready to merge ✅ 

I will clone your PR to provide this fix to our stable version (5.x), this bug affects both versions (5.x and 6.x-betas).

PR #2921, opened 2021-02-09
[5.x] Properly look up entry in manifest when name is given with extension

pedrofurtado:
Clone of https://github.com/rails/webpacker/pull/2919, to provide the bugfix also in stable 5.x versions.

Thanks @leonid-shevtsov 🤝 🍻


pedrofurtado:
Oh, I created with wrong branch 😄

PR #2923, opened 2021-02-10
Allow "stylus" as additional file ending

gr8bit:
I found that the file extension and the language code (in Vue files) "stylus" (instead of "styl") becomes more common (e.g. RubyMine suggests it in autocompletion), so this PR simply adds support for using "stylus" in addition to "styl".

Bildschirmfoto 2021-02-11 um 13 47 13
(Screenshot of RubyMine's autocompletion suggestions)


guillaumebriday:
LGTM

PR #2931, opened 2021-02-18
Improve error messaging when running webpacker:install

mctaylorpants:
Hello! I noticed that running `rails webpacker:install` will always report success, even if an underlying yarn command fails:

Screen Shot 2021-02-18 at 08 58 03

This can happen for many reasons, including network timeouts, or using a private registry and encountering authentication failure, etc.

We already output enough info from the yarn command itself to debug, but the "success" message is a bit misleading. I thought it would be nicer to tell the user that something went wrong instead:

Screen Shot 2021-02-18 at 08 57 02

### DIFF
--- a/lib/install/template.rb
+++ b/lib/install/template.rb
@@ -72,4 +72,5 @@ if results.all?
   say "Webpacker successfully installed 🎉 🍰", :green
 else
   say "Webpacker installation failed 😭 See above for details.", :red
+  exit 1
 end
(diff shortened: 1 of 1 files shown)


pedrofurtado:
Hey @mctaylorpants ! 👋 

Thank you for your contribution, this fix a really tricky scenario 🍻 

It looks good to be merged ☑️ @gauravtiwari

PR #2933, opened 2021-02-19
Wrap the path in quotes

raulriera:
## Description

While trying to upgrade to Rails 6 one of my projects, I noticed that `bin/rails webpacker:install` was failing with

```shell
Thor::Error: Could not find "/Users/raulriera/Library/Mobile" in any of your source paths. Your current source paths are: 
/Users/raulriera/Library/Mobile Documents/com~apple~CloudDocs/Documents/Work/odonto.me/odontome/vendor/cache/ruby/2.7.0/gems/railties-6.1.3/lib/rails/generators/rails/app/templates
/Users/raulriera/Library/Mobile Documents/com~apple~CloudDocs/Documents/Work/odonto.me/odontome/vendor/cache/ruby/2.7.0/gems/thor-1.1.0/lib/thor/actions.rb:156:in `find_in_source_paths'
/Users/raulriera/Library/Mobile Documents/com~apple~CloudDocs/Documents/Work/odonto.me/odontome/vendor/cache/ruby/2.7.0/gems/thor-1.1.0/lib/thor/actions.rb:215:in `apply'
/Users/raulriera/Library/Mobile Documents/com~apple~CloudDocs/Documents/Work/odonto.me/odontome/vendor/cache/ruby/2.7.0/gems/railties-6.1.3/lib/rails/tasks/framework.rake:15:in `block (2 levels) in <top (required)>'
```

Turns out the `LOCATION` environment variable when set without quotes will get trimmed.  I was unable to find a test for the install action 🤔 

This should probably be merged into the 5.x branch as well (if accepted)

### DIFF
--- a/lib/tasks/webpacker/binstubs.rake
+++ b/lib/tasks/webpacker/binstubs.rake
@@ -7,9 +7,9 @@ namespace :webpacker do
     prefix = task.name.split(/#|webpacker:binstubs/).first
 
     if Rails::VERSION::MAJOR >= 5
-      exec "#{RbConfig.ruby} #{bin_path}/rails #{prefix}app:template LOCATION=#{binstubs_template_path}"
+      exec "#{RbConfig.ruby} #{bin_path}/rails #{prefix}app:template LOCATION='#{binstubs_template_path}'"
     else
-      exec "#{RbConfig.ruby} #{bin_path}/rake #{prefix}rails:template LOCATION=#{binstubs_template_path}"
+      exec "#{RbConfig.ruby} #{bin_path}/rake #{prefix}rails:template LOCATION='#{binstubs_template_path}'"
     end
   end
 end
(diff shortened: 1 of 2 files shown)


pedrofurtado:
Hey, @raulriera ! 👋 

Thanks a lot for your contribution.

Just for understand the motivation behind PR: the quotes is required because the path defined in LOCATION environment (in your machine, in this case) contains whitespaces, right? Then, with quotes, the whitespace not causes tricky bugs and so on?

Anyway, thanks a lot for your PR! Let me know these details, because (depending of your response) we can provide this also in stable version (5.x at this moment) 🤝

raulriera:
Hey Pedro,

Thanks for the kinds words. You are right, that is the exact issue. Without the quote, my path which has spaces would not work

pedrofurtado:
@raulriera Thanks for feedback!

I think we are ready to merge @gauravtiwari  👍

henrik:
Will this break if there’s a `’` in the path? Not common but presumably possible. Perhaps  https://ruby-doc.org/stdlib-2.5.1/libdoc/shellwords/rdoc/Shellwords.html#method-c-shellescape?

PR #2783, opened 2020-11-19
Use webpack serve for dev server

excid3:
webpack-dev-server has been deprecated in favor of `webpack serve` according to this issue: https://github.com/webpack/webpack-dev-server/issues/2759 

This PR updates the `DevServerRunner` to be compatible with webpack-cli v4. Fixes #2776

PR #2937, opened 2021-02-24
Enable devServer.stats.colors by default

### LINKED ISSUE #2936: Webpacker 6: webpack dev server output not colored
Hi,

I don't know if this change is intended. I've tried the search functionality but I wasn't able to find a duplicate or a commit documenting the change

### How to replicate

```
$ rails new test-webpacker6-webpack-dev-server --skip-active-record --skip-javascript
$ cd test-webpacker6-webpack-dev-server/
$ echo "gem 'webpacker', git: 'https://github.com/rails/webpacker.git'" >> Gemfile
$ bundle
$ rails webpacker:install
$ bin/webpack-dev-server
```

### Actual Result

image

### Expected Result

image

rossta:
## Problem

The webpack-dev-server changed the default option for devServer.stats.colors from true to false in [a recent changeset](https://github.com/webpack/webpack-dev-server/pull/2660/files#r447299960).

Hence, in previous versions of Webpacker, the webpack-dev-server output was colorized, but is no longer in Webpacker 6.

## Solution

This PR sets the default Webpacker 6 devServer.stats.colors to true to preserve colorized behavior across the upgrade. This is consistent with the behavior of `bin/webpack`, which _is_ colorized by default and, generally, with Rails.logger behavior in development.

### Before

Webpacker 6 `$ bin/webpack-dev-server`
Screen Shot 2021-02-24 at 3 47 03 PM

### After

Webpacker 6 `$ bin/webpack-dev-server`
Screen Shot 2021-02-24 at 3 49 16 PM

See also current behavior of Webpacker 6 `$ bin/webpack`
Screen Shot 2021-02-24 at 3 47 12 PM

Fixes #2936

### DIFF
--- a/package/environments/development.js
+++ b/package/environments/development.js
@@ -42,6 +42,7 @@ if (
       headers: devServer.headers,
       overlay: devServer.overlay,
       stats: {
+        colors: true,
         entrypoints: false,
         errorDetails: true,
         modules: false,


package/environments/development.js
```diff
@@ -42,6 +42,7 @@ if (
       headers: devServer.headers,
       overlay: devServer.overlay,
       stats: {
+        colors: true,
```
tagliala, line 45: This is different from the value that may be generated by `webpack-dev-server`, which is using `supports-color` to determine the default value for this option

on maxOS, using webpacker 5, the colors option looks like: 
```
    colors: { level: 2, hasBasic: true, has256: true, has16m: false }
```

rossta: Thanks for the tip, that does look like a more flexible option.

tagliala: Hi @rossta,

Apologies, I'm not a JavaScript developer.

This will change anyway in webpack-dev-server 4: they have removed [supports-color](https://www.npmjs.com/package/supports-color) in the PR you mentioned at https://github.com/rails/webpacker/issues/2936#issuecomment-785359018, but I did not get the reason

I can confirm that launching `webpack-dev-server` or `webpack serve` with `webpack-dev-server@next` sets the `colors` option to `true`:

```
yarn webpack --version
yarn run v1.22.10
webpack 5.24.1
webpack-cli 4.5.0
webpack-dev-server 4.0.0-beta.0
```

```js
    stats: { preset: 'normal', colors: true },
```

rossta: Good to hear. Looks from the [webpack-dev-server 4.0.0-beta.0 release notes](https://github.com/webpack/webpack-dev-server/releases/tag/v4.0.0-beta.0) that the `stats` option is removed, so these `devServer` options won't be needed. 

> the stats option was removed, please use the stats option from webpack.config.js

I'll await feedback from the maintainers who could either merge this PR as a temporary improvement or close it in favor of the future webpack-dev-server v4 integration (in which case, the workaround noted in #2936 will work in the mean time).

guillaumebriday:
LGTM

We moved from webpack-dev-server to webpack's cli serve [here](https://github.com/rails/webpacker/pull/2783).

But this option should be enabled by default IMO: https://webpack.js.org/configuration/stats/#statscolors

cc @dhh @gauravtiwari

PR #2942, opened 2021-03-02
Preserve paths in entrypoint lookup

bpo:
Looking up `print/application.css` in the manifest should yield the same entry as a lookup for `print/application`. Previously it would strip parent paths and yield the entry for `application`.

Related to #2919, which fixed entrypoints with extensions that lack paths.

I'm new to this codebase, all feedback very welcome!


bpo:
(update just rebased onto main)

Issue #2926, opened 2021-02-12
webpacker:check_yarn doesn't support versioning of yarn nightly builds


pustomytnyk:
The script relies on `SemanticRange.satisfies?` which doesn't seem to support versions like `1.23.0-20210103.1434` (current yarn Classic Nightly version). That leads to `Webpacker requires Yarn ">=1 <3" and you are using 1.23.0-20210103.1434` error.

pedrofurtado:
Hey, @pustomytnyk ! 👋 

Thanks for report. 

After read your description, maybe this can related to `semantic_range` gem, that is a webpacker dependency. I suggest open a issue directly in semantic_range repository 🤝

pustomytnyk:
I see, `satisfies?` in node semver has optional `includePrerelease` flag. I'll try to add it to SemanticRange.

pedrofurtado:
@pustomytnyk Good news!

Please, could you link the issues/PRs that you opened in semantic_range repo, to be visible in this issue?

After the merge and release, we can update the version of semantic_range used in webpacker, to provide this new feature 🚀

Let us know the progress of this 🤝

pedrofurtado:
@pustomytnyk Let us know when you have any progress with this change in semantic_range gem 🤝 🍻  Then, we will be able to include this in webpacker 👍

pustomytnyk:
(librariesio/semantic_range#74 prepares SemanticRange for the new option, then we'll add `include_prerelease` option in another PR, and then we'll finally be able to properly fix `check_yarn.rake`)

jdelStrother:
FWIW, Yarn's latest installation instructions suggest `yarn set version 3.x`, which results in `Webpacker requires Yarn ">=1 <4" and you are using 3.1.0-rc.5`

For now I've stubbed out webpacker's version checks by adding the following to my rakefile:

```ruby
Rake::Task["webpacker:check_yarn"].clear_actions
```

PR #2947, opened 2021-03-05
semantic_range dep: Update; use twiddle-wakka

henrik:
3.0 was recently released, and it was a bit scary to find the version constraint in Webpacker was optimistic – it could have been incompatible.

Turned out it only changed APIs that Webpacker doesn't use. But I think we can trust "semantic_range" to use semantic versioning itself, so let's be careful in future :)

Did not change any other version constraints to keep this focused, but that may be a good idea too.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -5,7 +5,7 @@ PATH
       activesupport (>= 5.2)
       rack-proxy (>= 0.6.1)
       railties (>= 5.2)
-      semantic_range (>= 2.3.0)
+      semantic_range (~> 3.0)
 
 GEM
   remote: https://rubygems.org/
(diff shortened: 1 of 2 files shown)


pustomytnyk:
It was a preparation release to fix #2926. So eventually you'll need to update it again.

pedrofurtado:
Hi @pustomytnyk @henrik ! 🤝 

Thanks for contribution!

With this update, @pustomytnyk , it was possible to use yarn nightly builds? 👍  If so, then we will be ready for merge ✅

pustomytnyk:
Hi, not yet.

dhh:
@henrik I've found that I'd generally rather deal with incompatibilities that might come up, than be stuck with incompatible demands for specific versions. Not a big fan of upper-bounded dependencies for this reason.

henrik:
Fair enough! Thanks for the explanation.

PR #2948, opened 2021-03-08
Add support to run webpack-dev-server via webpack-dashboard

texpert:
Add support for this convenient [Webpack-Dashboard](https://github.com/FormidableLabs/webpack-dashboard)

### DIFF
--- a/README.md
+++ b/README.md
@@ -474,6 +474,35 @@ const vueConfig = require('./rules/vue')
 module.exports = merge(webpackConfig, vueConfig)
 ```
 
+#### Webpack-Dashboard
+
+When running webpack-dev-server in a separate process, a CLI dashboard could be used - [Webpack-Dashboard](https://github.com/FormidableLabs/webpack-dashboard)
+
+Add it to the gem's package.json:
+
+```bash
+yarn add --dev webpack-dashboard
+```
+
+Configure it as a webpack plugin:
+
+```js
+// config/webpack/development.js
+const webpackConfig = require('./base')
+const { merge } = require('webpack-merge')
+const DashboardPlugin = require("webpack-dashboard/plugin")
+
+module.exports = merge(webpackConfig, {
+  plugins: [new DashboardPlugin()],
+})
+```
+
+Add it as an option when starting the dev server:
+
+```bash
+bin/webpack-dev-server --webpack-dashboard
+```
+
 ### Custom Rails environments
 
 Out of the box Webpacker ships with - development, test and production environments in `config/webpacker.yml` however, in most production apps extra environments are needed as part of deployment workflow. Webpacker supports this out of the box from version 3.4.0+ onwards.
--- a/lib/webpacker/dev_server_runner.rb
+++ b/lib/webpacker/dev_server_runner.rb
@@ -77,6 +77,17 @@ module Webpacker
           @argv.delete "--debug-webpacker"
         end
 
+        if @argv.include?("--webpack-dashboard")
+          cmd = if node_modules_bin_exist?
+            %W[#{@node_modules_bin_path}/webpack-dashboard --] + cmd
+          else
+            idx_after_yarn = cmd.index("yarn") + 1
+            cmd.insert(idx_after_yarn, "webpack-dashboard", "--")
+          end
+
+          @argv.delete "--webpack-dashboard"
+        end
+
         cmd += ["--config", @webpack_config]
         cmd += ["--progress", "--color"] if @pretty
         cmd += @argv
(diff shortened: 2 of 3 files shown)


mcmire:
I came across this PR and had some thoughts. Does it make sense to have `bin/webpack-dev-server` run `webpack-dashboard` instead of `webpack-dev-server`, despite the `--webpack-dashboard` option? It seems like in that case it would be more consistent to rename `bin/webpack-dev-server` to something like `bin/webpack-watcher`. In other words, instead of treating `webpack-dashboard` as an override, reframe this so that the executable now has two strategies for how the watcher will be run: by default, it uses the `webpack-dev-server` strategy, but you can swap that out with the `webpack-dashboard` strategy if you like. As an example of this reframing, perhaps the user would run `bin/webpack-watcher --strategy webpack-dashboard` to make that swap-out.

Alternatively, is there a way to use `webpack-dashboard` today without having to modify the inner workings of `bin/webpack-dev-server`?

texpert:
> I came across this PR and had some thoughts. Does it make sense to have `bin/webpack-dev-server` run `webpack-dashboard` instead of `webpack-dev-server`, despite the `--webpack-dashboard` option? It seems like in that case it would be more consistent to rename `bin/webpack-dev-server` to something like `bin/webpack-watcher`. In other words, instead of treating `webpack-dashboard` as an override, reframe this so that the executable now has two strategies for how the watcher will be run: by default, it uses the `webpack-dev-server` strategy, but you can swap that out with the `webpack-dashboard` strategy if you like. As an example of this reframing, perhaps the user would run `bin/webpack-watcher --strategy webpack-dashboard` to make that swap-out.
> 

bin/webpack-dev-server` is running `webpack-dashboard` not instead of `webpack-dev-server`, it is running `webpack-dev-server` via `webpack-dashboard`.  As for your proposal of strategy, don't quite get it - what's the purpose and benefit?

> Alternatively, is there a way to use `webpack-dashboard` today without having to modify the inner workings of `bin/webpack-dev-server`?

Alas, I've just spent some hours trying to run the `webpack-dashboard` via the `package.json` scripts, but with not much success. I've made it start, but then only the log windows is working, other windows remaining blank. Here's the try on one of my demo apps - https://github.com/texpert/rails_6_rss_reader/pull/93/files

mcmire:
> `bin/webpack-dev-server` is running `webpack-dashboard` not instead of `webpack-dev-server`, it is running `webpack-dev-server` via `webpack-dashboard`. As for your proposal of strategy, don't quite get it - what's the purpose and benefit?

Sure, here is more of what was in my head:

1. It seems that `webpack-dashboard` works much differently than `webpack-dev-server` and cannot be used in the same way. For instance, when using Webpacker I have found it to be a speedier experience to run WDS side-by-side with the Rails server rather than having Webpacker kick in on each request, using something like `heroku local` or Overmind. I don't know how many other people do this, but I imagine it is not just me. However, if I understand how `webpack-dashboard` works, I don't think that would be possible — it's not designed to work that way. So, given these differences, it places the accuracy of the name of `bin/webpack-dev-server` in question. I know the intent is the same — you would still be running WDS — but the behavior is different, so it could be confusing for anyone maintaining this in the future to see code around a different executable in the same file. 
2. I was also thinking, what if other approaches to running `webpack-dev-server` emerge in the future and someone else wants to support that in Webpacker. So perhaps we would want `bin/webpack-dev-server` to have an API that reflects a strategy pattern (`--strategy webpack-dashboard` or `--via webpack-dashboard`) so that other ways to run WDS could be slotted in later (and the API wouldn't allow you to combine them in a weird way, for instance, `bin/webpack-dev-server --webpack-dashboard --some-other-mechanism`).

> Alas, I've just spent some hours trying to run the webpack-dashboard via the package.json scripts, but with not much success.

It seems that in your changes you are merely prepending `bin/webpack-dev-server` with `webpack-dashboard`. Also it seems like the command that comes after `webpack-dashboard` could theoretically be anything. Could you make your own bin script which essentially does this?

``` bash
#!/bin/sh

# replace with npm as needed 
yarn webpack-dashboard -- bin/webpack-dev-server
```

texpert:
> It seems that in your changes you are merely prepending `bin/webpack-dev-server` with `webpack-dashboard`. Also it seems like the command that comes after `webpack-dashboard` could theoretically be anything. Could you make your own bin script which essentially does this?
> 
> ```shell
> #!/bin/sh
> 
> # replace with npm as needed 
> yarn webpack-dashboard -- bin/webpack-dev-server
> ```

Yeah, thanks a lot for the hint - didn't think yarn is able to run Ruby scripts. Worked like a charm!

Also, added it the same way to the `scripts`, so `yarn dev` is now running them.

Closing then this PR as redundant.

PR #2949, opened 2021-03-08
Enable all default rubocop performance cops

tlatsas:
This enables all cops which are enabled by default in the rubocop-performance gem.
Currently, we require the performance extension but no cops run due to the `DisabledByDefault: true` setting.


pedrofurtado:
Hi @tlatsas ! 🍻 

Thanks for contribution! It is a great change 👍 

I think we are ready to merge it, @gauravtiwari ✔️

tlatsas:
Thank you! @pedrofurtado @gauravtiwari 🎉

PR #2954, opened 2021-03-12
Always use content hash for MiniCssExtractPlugin

### LINKED ISSUE #2928: Question: Why does CSS only have contentHash in production mode? (webpacker 6)
Sorry if this has been answered before. (Context: webpacker 6 beta 5)

Looking at the impllmentation of the MiniCssExtractPlugin...

https://github.com/rails/webpacker/blob/master/package/environments/base.js#L66

```js
  if (moduleExists('css-loader') && moduleExists('mini-css-extract-plugin')) {
    const MiniCssExtractPlugin = require('mini-css-extract-plugin')
    plugins.push(
      new MiniCssExtractPlugin({
        filename: isDevelopment
          ? 'css/[name].css'
          : 'css/[name]-[contenthash:8].css',
        chunkFilename: isDevelopment
          ? 'css/[id].css'
          : 'css/[id]-[contenthash:8].css'
      })
    )
  }
```

Why is it that the files development mode do not include `contentHash`? If I make a change to a CSS file, the view will get the same filename as before and will get a cached copy of the old styles. This is the opposite of the JS files that, on change, gives me a new file based on the hash.

I'm able to work around it be removing the existing plugin and merging in my own instance where the contentHash is always part of the file...

```js
// base.js

const { webpackConfig, merge } = require('@rails/webpacker')
const MiniCssExtractPlugin = require('mini-css-extract-plugin');
webpackConfig.plugins = webpackConfig.plugins.filter(p => p.constructor.name != 'MiniCssExtractPlugin')

const customConfig = {
  plugins: [
    new MiniCssExtractPlugin({
      filename: 'css/[name]-[contenthash:8].css',
      chunkFilename: 'css/[id]-[contenthash:8].css'
    })
  ],
  resolve: {
    extensions: ['.css']
  }
}

module.exports = merge(webpackConfig, customConfig)
```

I'm more curious as to why it's like this?

t27duck:
Closes #2928 

With Javascript packs in development mode, files are outputted with
hashes in the file names so browser cache is automatically busted when
code is changed and new files are made.

However, the mini-css-extract-plugin plugin is only configured to output
files with hashes in production mode. This means that any changes to CSS
files locally will appear to not have any effect on the page you are
working on because the browser is assuming the resulting CSS file with
the same name has not changed and reloads the cached version, forcing us
to manually bust browser cache to see our changes.

This code change enables the hash-based file names for the plugin
regardless of environment.

### DIFF
--- a/package/environments/base.js
+++ b/package/environments/base.js
@@ -67,12 +67,8 @@ const getPlugins = () => {
     const MiniCssExtractPlugin = require('mini-css-extract-plugin')
     plugins.push(
       new MiniCssExtractPlugin({
-        filename: isDevelopment
-          ? 'css/[name].css'
-          : 'css/[name]-[contenthash:8].css',
-        chunkFilename: isDevelopment
-          ? 'css/[id].css'
-          : 'css/[id]-[contenthash:8].css'
+        filename: 'css/[name]-[contenthash:8].css',
+        chunkFilename: 'css/[id]-[contenthash:8].css'
       })
     )
   }
--- a/package/environments/base.js
+++ b/package/environments/base.js
@@ -10,7 +10,6 @@ const WebpackAssetsManifest = require('webpack-assets-manifest')
 const webpack = require('webpack')
 const rules = require('../rules')
 const config = require('../config')
-const { isDevelopment } = require('../env')
 const { moduleExists } = require('../utils/helpers')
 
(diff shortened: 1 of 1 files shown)


excid3:
Hit this several times already, would be nice to have this consistent. 👍

PR #2957, opened 2021-03-17
Make versions explicit in 6.0 upgrade doc

### LINKED ISSUE #2955: Javascript not executing?
I really think this must be something really dumb and am hoping it's an easy fix.

I was having issues with mini-css-extract-plugin and undefined 'tap' method, but everything else was working fine. After some research, I saw the 6_0_upgrade.md doc so figured I'd try that. It did fix my CSS issue but now my Javascript won't execute in the browser, even though it compiles just fine.

I have stripped my application.js down so it is JUST a console.log statement.

```
/* eslint no-console:0 */
// Uncomment to copy all static images under ../images to the output folder and reference
// them with the image_pack_tag helper in views (e.g <%= image_pack_tag 'rails.png' %>)
// or the `imagePath` JavaScript helper below.
//
// const images = require.context('../images', true)
// const imagePath = (name) => images(name, true)

console.log('Hello World, from Webpacker!')
```

I've included this via `<%= javascript_pack_tag "application" %>`.

When I browse to a page, I can see in dev tools sources tab that it has loaded.

However, it just isn't console.logging anything. What could I be missing? This worked before going from webpacker 5.x to 6.0.

rossta:
## Problem

The 6.0 upgrade instructions as written do not guarantee that developers will end up with identical versions for the webpacker gem and `@rails/webpacker` NPM package. Certain version combinations are incompatible causing developers to run into issues with the upgrade. 

Examples 

- #2955 
- https://discuss.rubyonrails.org/t/upgrading-to-webpacker-6-0-0/77392/4

There is unexpected behavior when following the upgrade instructions:

1. Given the instruction `gem 'webpacker', '~> 6.0.0.pre.2'`, bundler does not install the latest prerelease with the pessimistic version constraint. I would expect bundler to install the latest prerelease version, which, as of this note, is `6.0.0.beta.6`, but instead, `6.0.0.pre.2` is installed.
2. Given the instruction `yarn add @rails/webpacker@next`, yarn does not necessarily install the same version as my webpacker gem nor the latest prerelease. As of this note, it installs `6.0.0.beta.5` when I want `6.0.0.beta.6`.

## Solution

To avoid ambiguity and possibly incompatible installations, I've updated the upgrade instructions to be explicit about version numbers. I've added a note to encourage developers to check the releases page for the latest release, should the version number in the doc get out-of-date. I've also added a note to encourage developers to ensure that the gem and NPM version numbers match to avoid compatibility problems.

Fixes #2955

### DIFF
--- a/docs/v6_upgrade.md
+++ b/docs/v6_upgrade.md
@@ -35,24 +35,27 @@ Now you have a straight access to the Webpack configuration and you can change i
 3. Rename `config/webpack` to `config/webpack_old`
 4. Rename `config/webpacker.yml` to `config/webpacker_old.yml`
 5. Uninstall the current version of `webpack-dev-server`: `yarn remove webpack-dev-server`
-6. Upgrade Webpacker
+6. Upgrade webpacker to the latest beta
 
-  ```ruby
-  # Gemfile
-  gem 'webpacker', '~> 6.0.0.pre.2'
-  ```
+- [Check the releases page to verify the latest version](https://github.com/rails/webpacker/releases)
+- Make sure to install identical version numbers of webpacker gem and `@rails/webpacker` npm package
 
-  ```bash
-  bundle install
-  ```
+```ruby
+# Gemfile
+gem 'webpacker', '6.0.0.beta.7'
+```
 
-  ```bash
-  yarn add @rails/webpacker@next
-  ```
+```bash
+bundle install
+```
 
-  ```bash
-  bundle exec rails webpacker:install
-  ```
+```bash
+yarn add @rails/webpacker@6.0.0-beta.7
+```
+
+```bash
+bundle exec rails webpacker:install
+```
 
 - Change `javascript_packs_with_chunks_tag` and `stylesheet_packs_with_chunks_tag` to `javascript_pack_tag` and
   `stylesheet_pack_tag`.
@@ -75,12 +78,13 @@ Now you have a straight access to the Webpack configuration and you can change i
 9. Copy over custom browserlist config from `.browserslistrc` if it exists into the `"browserslist"` key in `package.json` and remove `.browserslistrc`.
 
 10. `extensions` was removed from the webpacker.yml file. Move custom extensions to
-  your configuration by merging an object like this. For more details, see docs for
-  [Webpack Configuration](https://github.com/rails/webpacker/blob/master/README.md#webpack-configuration)
+    your configuration by merging an object like this. For more details, see docs for
+    [Webpack Configuration](https://github.com/rails/webpacker/blob/master/README.md#webpack-configuration)
+
 ```js
 {
   resolve: {
-      extensions: ['.ts', '.tsx']
+    extensions: ['.ts', '.tsx']
   }
 }
 ```


6_0_upgrade.md
```diff
@@ -11,19 +11,22 @@ straightforward.
...
+  gem 'webpacker', '6.0.0.beta.6'
   ```
 
   ```bash
   bundle install
   ```
 
   ```bash
-  yarn add @rails/webpacker@next
+  yarn add @rails/webpacker@6.0.0.beta.6
```
nickbayley, line 29:
```suggestion
  yarn add @rails/webpacker@6.0.0-beta.6
```

The yarn package has a `-` instead of a `.` between `6.0.0` and `beta.6`.

rossta: Good catch!

pedrofurtado:
@rossta  @gauravtiwari I think we are ready to merge ✅

anitagraham:
Note that having installed gem `6.0.0.beta.6`, running 

`bundle exec rails webpacker:install`  adds `"@rails/webpacker": "^6.0.0-beta.5",` to `package.json`.

```
 >bundle info webpacker
  * webpacker (6.0.0.beta.6)

 >bundle exec rails webpacker:install
    conflict  config/webpacker.yml
Overwrite /private/var/www/rummy/config/webpacker.yml? (enter "h" for help) [Ynaqdhm] **a**
       force  config/webpacker.yml
Copying webpack core config
       exist  config/webpack
.....
       apply  /Users/anita/.rbenv/versions/2.6.6/lib/ruby/gems/2.6.0/gems/webpacker-6.0.0.beta.6/lib/install/binstubs.rb
  Copying binstubs
.....  
Installing all JavaScript dependencies [from prerelease rails/webpacker]
         run  yarn add @rails/webpacker@next from "."
yarn add v1.22.10
[2/4] 🚚  Fetching packages...
[3/4] 🔗  Linking dependencies...
[4/4] 🔨  Building fresh packages...
success Saved 1 new dependency.
info Direct dependencies
└─ @rails/webpacker@6.0.0-beta.5
info All dependencies
└─ @rails/webpacker@6.0.0-beta.5
✨  Done in 2.47s.
Installing webpack and webpack-cli as direct dependencies
         run  yarn add webpack@^5.11.0 webpack-cli@^4.2.0 from "."
yarn add v1.22.10
.....
✨  Done in 2.74s.
      insert  package.json
Webpacker successfully installed 🎉 🍰

```

rossta:
Updated to incorporate changes in new upgrade doc

justin808:
I've thought a lot about this problem with React on Rails. I require a precise version match between the gem and the package.  While this is a bit of a pain, it's worse to debug inconsistent versions.

Thoughts? I could submit a PR. 

The major bump to v6 could be a good time to add this. Otherwise, we're sort of breaking sem versioning.

docs/v6_upgrade.md
```diff
@@ -66,21 +69,22 @@ Now you have a straight access to the Webpack configuration and you can change i
 
   ```js
   // config/webpack/base.js
-  const { webpackConfig, merge } = require('@rails/webpacker')
-  const customConfig = require('./custom')
+  const { webpackConfig, merge } = require('@rails/webpacker');
```
mcmire, line 72: It seems like the generated versions of these files have never used semicolons. Should these be removed so as not to confuse people?

docs/v6_upgrade.md
```diff
@@ -66,21 +69,22 @@ Now you have a straight access to the Webpack configuration and you can change i
...
-  your configuration by by merging an object like this. For more details, see docs for
-  [Webpack Configuration](https://github.com/rails/webpacker/blob/master/README.md#webpack-configuration)
+    your configuration by by merging an object like this. For more details, see docs for
+    [Webpack Configuration](https://github.com/rails/webpacker/blob/master/README.md#webpack-configuration)
+
 ```js
 {
   resolve: {
-      extensions: ['.ts', '.tsx']
+    extensions: ['.ts', '.tsx'];
```
mcmire, line 87: Does this semicolon belong here?

mcmire:
Friendly ping on this PR. I feel like there are a few issues that have been raised already that have gone something like "I tried using this gem and ran into this issue! P.S. I'm using pre.2" This would help in triaging issues by cutting down on some of the noise I think.

docs/v6_upgrade.md
rossta: 👍

docs/v6_upgrade.md
rossta: 👍

mcmire:
Another case where someone tried to use an old version: https://github.com/rails/webpacker/issues/3032

PR #2971, opened 2021-03-31
Correct docs re: assets:precompile task

jaredbeck:
Based on my experience only, it seems *to me* that the documentation is incorrect or misleading here.

After uninstalling sprockets, I no longer see the `assets:precompile` task in my `rake -T` list. According to the docs, I should, but I don't.

```plain
bundle info sprockets
Could not find gem 'sprockets'.
bundle info webpacker
  * webpacker (5.2.1)
NODE_ENV=production bin/rake -T | egrep -i compile
rake webpacker:clean[keep,age]                                      # Remove old compiled webpacks
rake webpacker:clobber                                              # Remove the webpack compiled output directory
rake webpacker:compile                                              # Compile JavaScript packs using webpack for production with digests
```

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -2,7 +2,17 @@
 
 
 Webpacker hooks up a new `webpacker:compile` task to `assets:precompile`, which gets run whenever you run `assets:precompile`.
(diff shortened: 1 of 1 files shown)


jaredbeck:
My bad, `assets:precompile` has no description, so I wasn't seeing it under `rake -T` (Instead, must use `rake -A -T`)

PR #2983, opened 2021-04-11
Updating the documentation for v6

guillaumebriday:


rromanchuk:
this is really great, thanks. The changes in v6, even just naming defaults, has really made a difference helping a non-js person like myself hit the ground running.

guillaumebriday:
thanks @rromanchuk 🙏

rromanchuk:
Just found minor config issue when moving to non local environment.  I moved to `app/packs` right away since `app/javascripts` is a mental mouse trap for rails people. I just forgot about autoloading for Zeitwerk, since i've moved off the `rails new` defaults, my packs folder was being picked up. Just gave it my new webpacker path to ignore. 

`Rails.autoloaders.main.ignore(Rails.root.join('app/packs'))`

Issue #2814, opened 2020-12-18
When not using dev server, compile is called for every single asset tag and dev and test page rendering is very slow


davetron5000:
# problem

When not using `webpack-dev-server`, it seems that Webpacker calls compile for every single asset tag in dev and test, making dev and test very slow.

Based on commit history, this seems intentional or at least this behavior has existed for a while.  Is there a way to disable it in test at least?

## background

We have a run-of-the-mill ERB-based app.  We are on Webpacker 5.2.1, Rails 6.0.2.

Our header prefetches several fonts and our pages load many images.  Webpacker and Webpack 100% work to compile things and make the site work as intended.  Production is not affected.

The problem is when a page is loaded in dev or test, each time a webpacker-provided helper is called, compile is called:

Like so:

First, `lookup!` is called on the manifest:

https://github.com/rails/webpacker/blob/b6c21802bfd4968c6e22d6faf69a45ee1ff3eb19/lib/webpacker/helper.rb#L20-L24

`lookup!` calls `lookup`, which calls `compile if compiling?`

https://github.com/rails/webpacker/blob/b6c21802bfd4968c6e22d6faf69a45ee1ff3eb19/lib/webpacker/manifest.rb#L41-L50

`compiling?` is true if we are set to compile (true dev and test) and the dev server is not running:

https://github.com/rails/webpacker/blob/b6c21802bfd4968c6e22d6faf69a45ee1ff3eb19/lib/webpacker/manifest.rb#L53-L55

The result is that our image-heavy pages show `[Webpacker] Everything's up-to-date. Nothing to do` *many* times and a page that renders in < 1s in production takes 10s locally.

## What I need help with

* Although *not* using the dev server may be unusual, this issue must happen for *everyone* running system tests. In our app, there is one test of our home page that takes 56seconds.  If I set `compile: false` in `webpacker.yml` it takes 1.5 seconds.
* Is Webpacker aware of the request/response cycle or is it only aware of a particular asset helper being called?  This is question for someone who knows the code to point me in the right direction.

I think a solution might be to only call compile once per request, so if someone knows the codebase well, I can provide a patch.

## See Also

* #2392

rossta:
> this issue must happen for everyone running system tests

Yes, you are correct. This issue would affect anyone not using the dev server for development as well.

> Is Webpacker aware of the request/response cycle or is it only aware of a particular asset helper being called? 

[...]

Issue #2392, opened 2019-12-11
Stop logging "Nothing to do"


reggieb:
At the moment, a ["Nothing to do" log entry](https://github.com/rails/webpacker/blob/d836e34dc85b97f8c29a02286d8c93ecef980e58/lib/webpacker/compiler.rb#L30) is being generated every time `Webpacker::Compiler#compile` is called on none stale compiled packs.

```
def compile
    if stale?
      run_webpack.tap do |success|
        # We used to only record the digest on success
        # However, the output file is still written on error, (at least with ts-loader), meaning that the
        # digest should still be updated. If it's not, you can end up in a situation where a recompile doesn't
        # take place when it should.
        # See https://github.com/rails/webpacker/issues/2113
        record_compilation_digest
      end
    else
      logger.info "Everything's up-to-date. Nothing to do"
      true
    end
  end
```

This log entry is unnecessary and I think it should be removed or at the very least set at the `:debug` level.

burmashave:
> This log entry is unnecessary and I think it should be removed or at the very least set at the `:debug` level.

Agreed.

jakeNiemiec:
If you are seeing this in your log files, it would indicate that webpack is being run when it is not needed. In what context does this cause issues?

burmashave:
It's not an 'issue' so much as it's unnecessary. We're being informed of something it _didn't_ do with every request. 

This was [added in July](https://github.com/rails/webpacker/commit/2ee77e616fa7fb9fb519ad6f3fa6090570680a9e#diff-d372bcc3fac8ad71a96a2cc49d03bfac) to "keep the user informed, vs. have them wonder if anything happened". It's just as informative to not see a message when it does nothing as it is to send a message telling us it did nothing.

jakeNiemiec:
Understand that this is hard to balance with people who say things aren't verbose enough https://github.com/rails/webpacker/issues/955#issuecomment-503830871, https://github.com/rails/webpacker/issues/1824, https://github.com/rails/webpacker/issues/2312

> We're being informed of something it didn't do with every request.

Can you think of any middle ground? I agree that it seems unnecessary for every single request.

reggieb:
I'd suggest the middle ground is to log it `:debug` level.

burmashave:
> Understand that this is hard to balance with people who say things aren't verbose enough [#955 (comment)](https://github.com/rails/webpacker/issues/955#issuecomment-503830871), #1824, #2312

[...]

PR #3015, opened 2021-05-10
Better static assets serving

crawler:
Hi, and ❤️   to Rails people.

This change makes it easier to serve all project frontend assets with a Webpacker. The advantages of such an approach include the ability to switch from sassc to the last Dart Sass version shipped as node package, without losing the ability to use of same assets within the view and the CSS. Moreover, the same static assets will be available from JS if needed, the project can be configured to use assets from the node packages like the icon bundles within a view helpers.  This also provides some type of unification in my opinion.

## Current state
With some configuration tuning and the webpack-dev-server, everything works almost fine. The only issue is while you don't want to enable the cache_manifest option for the development env, every call of `resolve_path_to_image` causes `JSON.parse` to be called.

But when the dev server is enabled but is not running, the thing became unusable. Every asset call causes recalculation of the sources directory hash, and with the use of responsive images which require every picture to contain a set of different sources, the single page load takes forever to complete.

### Webpacker::Listen
Optional Webpacker::Listen class, code of which extracted from the Spring. It relies on the "listen" gem that is already installed for the dev environment in almost every Rails project (because of the spring-watcher-listen gem). It listens to the source changes and tells the Webpacker::Compiler and the Webpacker::Manifest objects that is time to refresh.

### Webpacker::DevServer change
**See next comment 👇**
~~A break between "is webpack dev server running" checks, which can be adjusted with `Webpacker.dev_server.retry_after = 0~~

~~Helps to save time when the project is configured with the dev_server option but the webpack dev server is not started.~~

### rules/file.js change
Modified file.js rule that adds relative directories to the resulting asset name so it can be accessed via view helpers as well as from scss. 

**For example:**
source: `assets/images/layout/burger.svg`
compiled: `/packs/media/images/layout/burger-2bacd1ce7bda5fa5c1f8.svg `
in the manifest: `"media/images/layout/burger.svg": "http://spitfire.lan/packs/media/images/layout/burger-2bacd1ce7bda5fa5c1f8.svg"`
can be accessed via: `image_pack_tag('layout/burger.svg')`

**Current:**
source: `assets/images/layout/burger.svg`
compiled `/packs/media/images/2bacd1ce7bda5fa5c1f8.svg"`
in the manifest: `"media/images/burger.svg": "http://spitfire.lan/packs/media/images/2bacd1ce7bda5fa5c1f8.svg"`
accessible via `image_pack_tag('burger.svg')`
_The subdirectories are ignored_

### webpacker.yml options
- static_assets_extensions - ablity to specify own list of static asset extensions to test;
- static_assets_output_prefix - option to change media/images to something else;
- use_listen - enable listen to monitor source changes;

Also I added install/packs/entrypoints/images_manifst.js - example of how to force webpack to add all images to the build.

All current tests are passing. I wrote no tests for the new features because not sure if my proposition will be accepted.

This change will close "slow Everything's up-to-date. Nothing to do" issues like this:
#2814
#2392 

In terms of speed, not the noise. I can provide demo app and the benchmarks if needed.

@gauravtiwari


crawler:
Hi! 
While I did not receive any feedback on my proposition, I ran this on two production projects found some issues, and made some improvements:

I moved dev server connection monitoring into a separate class: DevServerMonitor. It starts a Thread that checks the dev server connection every second (configurable). And if needed, this feature may be converted to opt-in with only one "if". This makes the experience away more smooth when starting, restarting, developing... as is usually the case. 

Also, I made the Manifest class listen for the manifest.json changes, so he can be aware of changes no matter the dev server generated the manifest, bin/webpack was executed or the assets were compiled from the lookup call.

As this feature is in the proposal stage, I added some verbose logging to be able to see what is going on. And muted the "Everything's up-to-date. Nothing to do" with a little trick ;)

crawler:
Video demonstration of the difference [https://www.youtube.com/watch?v=bBYBVtFV4tE](https://www.youtube.com/watch?v=bBYBVtFV4tE)

crawler:
Hi, @gauravtiwari can you please pay some attention to this?

gauravtiwari:
Thanks Anton, Apologies for the delay. It's in my radar just need to find sometime - hopefully this week :)

Thanks for your patience

crawler:
Thank you for letting me know!

crawler:
@gauravtiwari since this pull request was created, we have been using a fork in production for two projects without any issues. The HTML coder girl pretty happy with the speed improvement)

crawler:
@dhh ping. Can you please take a look at this?

imajes:
@crawler a couple notes that are worth considering here.

1. you're doing a lot in this PR. could you make the listen changes in one, the rules in another, and so forth?
2. there are no tests here. it's going to be hard to imagine moving forward without any- you are adding new behavior, so it's worth considering.
3. `static_assets_extensions` seems to be a regression of sorts - it looks like the webpacker team are trying to get that out of the config. I'd be cautious to imagine it might come back?

anyhow. feels like it should be a good fix, but my guess is that for the webpacker team to roll with it, there are a couple steps that might make their approval easier :)

crawler:
@dhh @gauravtiwari 
Hi, Here some real-world app benchmarks with the running dev server.

### Given
Hardware: 2.6 GHz 6-Core Intel Core i7 MBP
Almost static page with 2 CSS files, 1 JS file, and a lot of responsive images, SVG sprites, etc.
All assets are served with webpacker.

Summary of the calls to the webpacker helpers (excluding helpers that calling other helpers)
`{:path_to_asset=>148, :stylesheet_pack_tag=>1, :javascript_pack_tag=>1}`
Benchmarking was made with rack-mini-profiler


### Cold master branch
Cold means the app just started and no prior requests were made.

master

### Cold this PR

better

### Second request master branch

master

### Second request this PR

better2

Not sure why rack-mini-profiler skipping some methods from the report. Maybe because of too low execution time.

I Will try to set up CPU throttling or run in the VM

crawler:
@imajes, sure your right. I was already thinking about splitting this out to separate pull requests. But I wanted to hear some feedback from the webpacker team decision-makers. If anything from what I proposing makes sense for the project.

justin808:
@crawler good work!
I second @imajes on his comment https://github.com/rails/webpacker/pull/3015#issuecomment-862668435

Please:
1. break to smaller PRs. Sounds like this could be several based on your description. 
2. explain how your solution compares to using  `webpack --watch` (non-hmr)
3. provide a sample repo that demonstrates the fix
4. explain how to verify the fix.

And why can the webpack-dev-server configured appropriately solve the problem for the triggering of the current check (which I agree is HORRIBLE).

crawler:
@justin808 sorry missed out on your message. I am happy to do what you both asking, but not sure that maintainers are interested in this. @guillaumebriday what do you think?

PR #3031, opened 2021-05-26
CSS Modules HMR working again

justin808:
Demo: https://youtu.be/lIkhnLXIH6I

# Changes
1. Enable CSS modules HMR if webpack-dev-server running and HMR is on. Enabled by using the style-loader rather than extract CSS
2. Add docs for developing packages
3. export `inliningCss` from JS package
4. Remove inject_client webpacker.yml setting as this can be set by the HMR value
5. Added setting of injectClient and injectHot per the HMR setting
6. Simplifications and fixes for HMR 
   1. HMR default to true for webpacker.yml
   1. inline, injectClient, and injectHot default to the value of hmr
   1.  hmr turned on for the webpack dev server if the hmr config is true
   1.  using command line option for --hot as the docs recommend that over
     the plugin.

# To Test
Tested on https://github.com/shakacode/react_on_rails_tutorial_with_ssr_and_hmr_fast_refresh/pull/12

1. Checkout branch `justin808/change-css-loader`
2. bundle && yarn
3. Run `foreman start -f Procfile.dev-hmr`
4. Open up http://localhost:3000/hello_world and keep it visible to see the hot reloading
5. Edit the string on [line 17 of HelloWorld.tsx](https://github.com/shakacode/react_on_rails_tutorial_with_ssr_and_hmr_fast_refresh/blob/master/app/javascript/bundles/HelloWorld/components/HelloWorld.tsx#L17) and see the string change in the browser change upon saving the file.
6. Edit the text color on [line 2 of HelloWorld.module.css](https://github.com/shakacode/react_on_rails_tutorial_with_ssr_and_hmr_fast_refresh/blob/master/app/javascript/bundles/HelloWorld/components/HelloWorld.module.css#L2) and see the browser change upon saving the file.

### DIFF
--- a/docs/developing_webpacker.md
+++ b/docs/developing_webpacker.md
@@ -1,0 +1,29 @@
+# Developing Webpacker
+
+It's a little trickier for Rails developers to work on the JS code of a project like rails/webpacker. So here are some tips!
+
+## Use some test app
+For example, for React on Rails Changes, I'm using [shakacode/react_on_rails_tutorial_with_ssr_and_hmr_fast_refresh](https://github.com/shakacode/react_on_rails_tutorial_with_ssr_and_hmr_fast_refresh).
+This directory is the `TEST_APP_DIR`.
+
+## Fork rails/webpacker
+Let's call the rails/webpacker directory `WEBPACKER_DIR` which has rails/webpacker's `package.json`.
+
+## Changing the Package
+### Setup with Yalc
+Use [`yalc`](https://github.com/wclr/yalc) unless you like yak shaving weird errors. 
+1. In `WEBPACKER_DIR`, run `yalc publish` 
+2. In `TEST_APP_DIR`, run `yarn link @rails/webpacker`
+   
+## Update the Package Code
+1. Make some JS change in WEBPACKER_DIR
+2. Run `yalc push` and your changes will to your `TEST_APP_DIR`'s node_modules.
+3. You may need to run `yarn` in `TEST_APP_DIR` if you added or removed dependencies of rails/webpacker.
+
+## Updating the Ruby Code
+
+For the Ruby part, just change the gem reference `TEST_APP_DIR`, like:
+
+```ruby
+gem "webpacker", path: "../../forks/webpacker"
+```
--- a/docs/developing_webpacker.md
+++ b/docs/developing_webpacker.md
@@ -17,7 +17,7 @@ Use [`yalc`](https://github.com/wclr/yalc) unless you like yak shaving weird err
    
 ## Update the Package Code
 1. Make some JS change in WEBPACKER_DIR
-2. Run `yalc push` and your changes will to your `TEST_APP_DIR`'s node_modules.
+2. Run `yalc push` and your changes will be pushed to your `TEST_APP_DIR`'s node_modules.
 3. You may need to run `yarn` in `TEST_APP_DIR` if you added or removed dependencies of rails/webpacker.
 
 ## Updating the Ruby Code
--- a/lib/install/config/webpacker.yml
+++ b/lib/install/config/webpacker.yml
@@ -25,8 +25,6 @@ development:
     host: localhost
     port: 3035
     public: localhost:3035
-    # Inject browserside javascript that is required by both HMR and Live (full) reload
-    inject_client: true
     # Hot Module Replacement updates modules while the application is running without a full reload
     hmr: false
     # Inline should be set to true if using HMR; it inserts a script to take care of live reloading
--- a/lib/install/config/webpacker.yml
+++ b/lib/install/config/webpacker.yml
@@ -26,7 +26,7 @@ development:
     port: 3035
     public: localhost:3035
     # Hot Module Replacement updates modules while the application is running without a full reload
-    hmr: true
+    hmr: false
     # Should we show a full-screen overlay in the browser when there are compiler errors or warnings?
     overlay: true
     # Should we use gzip compression?
--- a/lib/install/config/webpacker.yml
+++ b/lib/install/config/webpacker.yml
@@ -26,9 +26,7 @@ development:
     port: 3035
     public: localhost:3035
     # Hot Module Replacement updates modules while the application is running without a full reload
-    hmr: false
-    # Inline should be set to true if using HMR; it inserts a script to take care of live reloading
-    inline: true
+    hmr: true
     # Should we show a full-screen overlay in the browser when there are compiler errors or warnings?
     overlay: true
     # Should we use gzip compression?
@@ -40,7 +38,7 @@ development:
     # When enabled, nothing except the initial startup information will be written to the console.
     # This also means that errors or warnings from webpack are not visible.
     quiet: false
-    pretty: false
+    pretty: true
     headers:
       'Access-Control-Allow-Origin': '*'
     watch_options:
--- a/lib/webpacker/dev_server.rb
+++ b/lib/webpacker/dev_server.rb
@@ -61,6 +61,8 @@ class Webpacker::DevServer
 
   private
     def fetch(key)
+      return nil unless config.dev_server.present?
+
       ENV["#{env_prefix}_#{key.upcase}"] || config.dev_server.fetch(key, defaults[key])
     end
 
--- a/lib/webpacker/dev_server.rb
+++ b/lib/webpacker/dev_server.rb
@@ -11,7 +11,7 @@ class Webpacker::DevServer
     @config = config
   end
 
-  def configured_and_running?
+  def running?
     if config.dev_server.present?
       Socket.tcp(host, port, connect_timeout: connect_timeout).close
       true
--- a/lib/webpacker/dev_server.rb
+++ b/lib/webpacker/dev_server.rb
@@ -11,7 +11,7 @@ class Webpacker::DevServer
     @config = config
   end
 
-  def running?
+  def configured_and_running?
     if config.dev_server.present?
       Socket.tcp(host, port, connect_timeout: connect_timeout).close
       true
--- a/lib/webpacker/dev_server.rb
+++ b/lib/webpacker/dev_server.rb
@@ -51,6 +51,10 @@ class Webpacker::DevServer
     fetch(:pretty)
   end
 
+  def hmr?
+    fetch(:hmr)
+  end
+
   def env_prefix
     config.dev_server.fetch(:env_prefix, DEFAULT_ENV_PREFIX)
   end
--- a/lib/webpacker/dev_server_runner.rb
+++ b/lib/webpacker/dev_server_runner.rb
@@ -74,12 +74,13 @@ module Webpacker
         end
 
         if @argv.include?("--debug-webpacker")
-          cmd = [ "node", "--inspect-brk"] + cmd
+          cmd = [ "node", "--inspect-brk", "--trace-warnings" ] + cmd
           @argv.delete "--debug-webpacker"
         end
 
         cmd += ["--config", @webpack_config]
         cmd += ["--progress", "--color"] if @pretty
+
         cmd += ["--hot"] if @hot
         cmd += @argv
 
--- a/lib/webpacker/dev_server_runner.rb
+++ b/lib/webpacker/dev_server_runner.rb
@@ -30,6 +30,7 @@ module Webpacker
         @port              = dev_server.port
         @pretty            = dev_server.pretty?
         @https             = dev_server.https?
+        @hot               = dev_server.hmr?
 
       rescue Errno::ENOENT, NoMethodError
         $stdout.puts "webpack dev_server configuration not found in #{@config.config_path}[#{ENV["RAILS_ENV"]}]."
@@ -79,6 +80,7 @@ module Webpacker
 
         cmd += ["--config", @webpack_config]
         cmd += ["--progress", "--color"] if @pretty
+        cmd += ["--hot"] if @hot
         cmd += @argv
 
         Dir.chdir(@app_path) do
--- a/lib/webpacker/helper.rb
+++ b/lib/webpacker/helper.rb
@@ -137,7 +137,9 @@ module Webpacker::Helper
   #   <%= stylesheet_pack_tag 'calendar' %>
   #   <%= stylesheet_pack_tag 'map' %>
   def stylesheet_pack_tag(*names, **options)
-    stylesheet_link_tag(*sources_from_manifest_entrypoints(names, type: :stylesheet), **options)
+    unless current_webpacker_instance.dev_server.running?
+      stylesheet_link_tag(*sources_from_manifest_entrypoints(names, type: :stylesheet), **options)
+    end
   end
 
   private
--- a/lib/webpacker/helper.rb
+++ b/lib/webpacker/helper.rb
@@ -141,10 +141,10 @@ module Webpacker::Helper
   #   <%= stylesheet_pack_tag 'calendar' %>
   #   <%= stylesheet_pack_tag 'map' %>
   def stylesheet_pack_tag(*names, **options)
-    css_extracted = !Webpacker.dev_server.configured_and_running?
-    if css_extracted
-      stylesheet_link_tag(*sources_from_manifest_entrypoints(names, type: :stylesheet), **options)
-    end
+    css_inline = Webpacker.dev_server.hmr? && Webpacker.dev_server.configured_and_running?
+    return "" if css_inline
+
+    stylesheet_link_tag(*sources_from_manifest_entrypoints(names, type: :stylesheet), **options)
   end
 
   private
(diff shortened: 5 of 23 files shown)


justin808:
@gauravtiwari can you give me a review, please?

lib/webpacker/helper.rb
```diff
@@ -137,7 +137,9 @@ def preload_pack_asset(name, **options)
   #   <%= stylesheet_pack_tag 'calendar' %>
   #   <%= stylesheet_pack_tag 'map' %>
   def stylesheet_pack_tag(*names, **options)
-    stylesheet_link_tag(*sources_from_manifest_entrypoints(names, type: :stylesheet), **options)
+    unless current_webpacker_instance.dev_server.running?
```
dhh, line 140: This seems risky. Must avoid ever having a path that leads to a port check happening in production. Seems like something that should happen outside the pack call itself.

lib/install/config/webpacker.yml
```diff
@@ -25,12 +25,8 @@ development:
     host: localhost
     port: 3035
     public: localhost:3035
-    # Inject browserside javascript that is required by both HMR and Live (full) reload
-    inject_client: true
     # Hot Module Replacement updates modules while the application is running without a full reload
-    hmr: false
-    # Inline should be set to true if using HMR; it inserts a script to take care of live reloading
-    inline: true
+    hmr: true
```
justin808, line 29: I see no reason to have the extra settings in the YML file since they are not used by the Ruby code and they can be customized in a custom Webpack config.

lib/webpacker/dev_server.rb
```diff
@@ -11,7 +11,7 @@ def initialize(config)
     @config = config
   end
 
-  def running?
+  def configured_and_running?
```
justin808, line 14: By renaming this method, it's clear that there's no cost for checking if the devServer is "configured" on a production environment.

lib/webpacker/dev_server_runner.rb
```diff
@@ -30,6 +30,7 @@ def load_config
         @port              = dev_server.port
         @pretty            = dev_server.pretty?
         @https             = dev_server.https?
+        @hot               = dev_server.hmr?
```
justin808, line 33: Recent docs on using HMR with the webpack-dev-server all suggest using the command line option rather than the plugin.

lib/webpacker/railtie.rb
```diff
@@ -8,7 +8,7 @@ class Webpacker::Engine < ::Rails::Engine
   config.webpacker = ActiveSupport::OrderedOptions.new
 
   initializer "webpacker.proxy" do |app|
-    insert_middleware = Webpacker.config.dev_server.present? rescue nil
```
justin808, line 11: There is no need to rescue and swallow errors here.

package/environments/development.js
```diff
@@ -33,8 +29,9 @@ if (
       https: devServer.https,
       hot: devServer.hmr,
       contentBase,
-      inline: devServer.inline,
-      injectClient: devServer.inject_client,
+      inline: devServer.inline || devServer.hmr,
+      injectClient: devServer.hmr,
+      injectHot: devServer.hmr,
```
justin808, line 34: Let's base the default "it just works" settings based on the one `hmr` setting.

package/utils/helpers.js
```diff
@@ -39,6 +39,9 @@ const canProcess = (rule, fn) => {
   return null
 }
 
+const runningWebpackDevServer = process.env.WEBPACK_DEV_SERVER &&
+  process.env.WEBPACK_DEV_SERVER !== 'undefined'
```
justin808, line 43: When running the webpack-dev-server, this ENV value is set.

justin808 (commented):
@dhh Please let me know if I addressed your concerns. I also improved the HMR setup.

lib/webpacker/helper.rb
justin808: @dhh, does the refactoring clarify that there cannot be a port check during production?

rossta:
Thanks @justin808 for working on getting CSS module HMR working. I understand it's an important feature to projects like react_on_rails and I appreciate your efforts here.

I do feel that changing the default to orient toward HMR should be reconsidered. My strong personal preference is that extracting css in development and test should be the default, even if the dev server is running. That means I also prefer that HMR should not be enabled by default.

As we saw in last year's May-of-WTFs, Webpacker has not provided a great out-of-the-box experience for many Rails developers. Anything we can do improve the OOB experience would be a win in terms of adoption and getting more folks helping improve the project. Extracting CSS and disabling HMR, even if the dev server is running, is closer to what I expect newcomers would expect and decrease the likelihood of those WTF moments that turn developers off. Enabling HMR makes sense for advanced users who know how to leverage its benefits but, I believe, creates confusion for folks encountering tools like webpack and Webpacker for the first time. 

To summarize, I'd prefer the view helper logic look something like

```ruby
extract_css = !Webpacker.dev_server.running_and_hmr_enabled? # not necessarily the suggested method name
```

And in `config/webpacker.yml` for development.

```
hmr: false
```

justin808:
> I do feel that changing the default to orient toward HMR should be
reconsidered. My strong personal preference is that extracting css in
development and test should be the default, even if the dev server is running.
That means I also prefer that HMR should not be enabled by default.

The big question is:

If you want simplicity and lack of WTFs, why not use `bin/webpack --watch` and see that webpack generated the static (actual) files upon saving changes?

Is there another benefit to using the webpack-dev-server without HMR? Is it any faster than watch mode?

Maybe we can make the default for development mode to be `bin/webpack --watch` and leave HMR as the default if somebody wants to run bin/webpack-dev-server?

For tests, static files are the way to go. 

> As we saw in last year's May-of-WTFs, Webpacker has not provided a great
out-of-the-box experience for many Rails developers. Anything we can do improve
the OOB experience would be a win in terms of adoption and getting more folks
helping improve the project. Extracting CSS and disabling HMR, even if the dev
server is running, is closer to what I expect newcomers would expect and
decrease the likelihood of those WTF moments that turn developers off. When
extract_css used to be false by default, this lead to a lot of reported issues
that boiled down to lack of understanding. Enabling HMR makes sense for
advanced users who know how to leverage its benefits but, I believe, creates
confusion for folks encountering tools like webpack and Webpacker for the first
time.

I agree with "Anything we can do improve the OOB experience." That's why we should remove extra devServer options in the `config/webpacker.yml` that can be easily set in the JS files for the webpack configuration. To this point, having one `development.devServer.hmr = true` option that controls everything to make HMR function is a big win.

I agree that having a default of CSS extraction is a big win, as that is the typical scenario of both `bin/webpack --watch` and production.

However, JS devs would expect that HMR is on by default, as that's the default for creating-react-app, gatsby, and next.js.

So not having HMR work by default would be a WTF for others. Having finicky things like HMR "it just works" is a big reason to use a library like rails/webpacker.

So can we have: 
* `bin/webpack --watch` default for development and testing
* `bin/webpack-dev-server` for advanced JS devs with features like HMR.

I'm also OK with @rossta's idea of adding the check that HMR is on before disabling the CSS extraction and style tag. 

Any opinions?

justin808:
Per my previous comment, there might be a slight performance advantage in using the webpack-dev-server per [these docs](https://webpack.js.org/guides/build-performance/#compile-in-memory).

So I'm OK with having HMR be off for the default, and if HMR is off, then CSS is extracted.

aried3r:
> However, JS devs would expect that HMR is on by default, as that's the default for creating-react-app, gatsby, and next.js.

I agree with this statement. Having worked mostly with Rails and some Next.js recently, there is a definitive gap between the default webpacker experience and the default experience of the mentioned frameworks.

But I also understand the concern that onboarding people coming from the assets pipeline is important.

justin808:
@rossta @dhh I changed the PR to address @rossta comments.

lib/webpacker/helper.rb
```diff
@@ -137,6 +141,9 @@ def preload_pack_asset(name, **options)
   #   <%= stylesheet_pack_tag 'calendar' %>
   #   <%= stylesheet_pack_tag 'map' %>
   def stylesheet_pack_tag(*names, **options)
+    css_inline = Webpacker.dev_server.hmr? && Webpacker.dev_server.configured_and_running?
+    return "" if css_inline
+
```
justin808, line 146: @dhh and @rossta, does the above make it clear when to not have the style tag if we want CSS inlined?

package/environments/development.js
```diff
@@ -11,14 +11,10 @@ let devConfig = {
...
 
-if (
-  process.env.WEBPACK_DEV_SERVER &&
-  process.env.WEBPACK_DEV_SERVER !== 'undefined'
-) {
+if (runningWebpackDevServer) {
   if (devServer.hmr) {
     devConfig = merge(devConfig, {
-      output: { filename: '[name]-[hash].js' },
-      plugins: [new webpack.HotModuleReplacementPlugin()]
```
justin808, line 21: Previously, we always included the HotModuleReplacementPlugin even if HMR was off. So it's no change if the command line option `--hot` is always added.

lib/webpacker/helper.rb
guillaumebriday: LGTM

package/utils/helpers.js
guillaumebriday, line 43: Is `undefined` a string? And not the `undefined` keywork? 🤔

```node
> process.env.WEBPACK_DEV_SERVER
undefined
> process.env.WEBPACK_DEV_SERVER !== 'undefined'
true
> process.env.WEBPACK_DEV_SERVER !== undefined
false
>
```

```suggestion
// When running the webpack-dev-server, this ENV value is set.
const runningWebpackDevServer = process.env.WEBPACK_DEV_SERVER !== undefined
```

lib/webpacker/dev_server.rb
dhh: I know I flagged this, but seeing it in context, it actually looks worse. I think we were better off with just #running?. Which is kinda ironic because I vaguely remember flipflopping on this exact question when we originally made web packer!

lib/webpacker/helper.rb
dhh: I would pull this up into Webpacker, so it's `Webpacker.inlining_css?`.

lib/webpacker/railtie.rb
dhh: Don't think the local variable explains anything. Should just inline into the conditional.

justin808: Agree.

package/utils/helpers.js
justin808: Updated to check against 'true'.

justin808:
@guillaumebriday Almost ready. Just got to fix the jest tests.

lib/webpacker/helper.rb
justin808: Done. How's this look?

package/env.js
```diff
@@ -16,9 +16,15 @@ const config = safeLoad(readFileSync(configPath), 'utf8')
...
+const runningWebpackDevServer = process.env.WEBPACK_DEV_SERVER === 'true' ||
+  process.env.WEBPACK_DEV_SERVE === 'true'
+
 module.exports = {
   railsEnv: railsEnv && railsEnv.match(regex) ? railsEnv : DEFAULT,
   nodeEnv,
   isProduction,
-  isDevelopment
+  isDevelopment,
+  runningWebpackDevServer
```
justin808, line 29: runningWebpackDevServer feels like an env value.

justin808:
@guillaumebriday @dhh this PR now has all the requested changes.

docs/developing_webpacker.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Changing the Package
+### Setup with Yalc
+Use [`yalc`](https://github.com/wclr/yalc) unless you like yak shaving weird errors. 
+1. In `WEBPACKER_DIR`, run `yalc publish` 
+2. In `TEST_APP_DIR`, run `yarn link @rails/webpacker`
+   
+## Update the Package Code
+1. Make some JS change in WEBPACKER_DIR
+2. Run `yalc push` and your changes will to your `TEST_APP_DIR`'s node_modules.
```
iguanus, line 20: I think we missed a verb in this phrase.

... and your changes will be **pushed** to your...

iguanus:
Looks solid to me. Kudos on the helper variables naming!

tomgilligan:
responding to https://github.com/rails/webpacker/issues/3055#issuecomment-867423734

Followed the instructions to test:
- [x] Checkout branch justin808/change-css-loader
- [x] bundle && yarn
   (had to `bundle update webpacker` first because `fatal: Could not parse object '1a680abb22154b97b6e86ced38b15cfa23a441da'.` )
- [x] Run foreman start -f Procfile.dev-hmr
  [Foreman output](https://gist.github.com/tomgilligan/4e9e104cda267da043eff86a188e7f25)
- [ ] Open up http://localhost:3000/hello_world and keep it visible to see the hot reloading
   at this point i think i should see something that is green but i don't

#### aborted testing here

- [ ] Edit the string on line 17 of HelloWorld.tsx and see the string change in the browser change upon saving the file.
- [ ] Edit the text color on line 2 of HelloWorld.module.css and see the browser change upon saving the file.


i tried with
node v14.15.4
and again with a fresh install of node v16.4.0

justin808:
Hi @tomgilligan, please pul the latest code from  https://github.com/shakacode/react_on_rails_tutorial_with_ssr_and_hmr_fast_refresh/pull/12

I had updated rails/webpacker, but I forgot to push my test code!

@guillaumebriday Can you give it a try?

tomgilligan:
Thank you for the update @justin808 .  I now get this for both node v14.15.4 and v16.4.0 .

```
22:15:40 wp-client.1 | <s> [webpack.Progress] 9% setup compilation DefinePlugin
22:11:25 wp-client.1 | [webpack-cli] TypeError: Cannot read property 'get' of undefined
22:11:25 wp-client.1 |     at exports.provide (/Users/tomgilligan/react_on_rails_tutorial_with_ssr_and_hmr_fast_refresh/node_modules/@rails/webpacker/node_modules/webpack/lib/util/MapHelpers.js:17:20)
22:11:25 wp-client.1 |     at /Users/tomgilligan/react_on_rails_tutorial_with_ssr_and_hmr_fast_refresh/node_modules/@rails/webpacker/node_modules/webpack/lib/DefinePlugin.js:290:6
22:11:25 wp-client.1 |     at Hook.eval [as call] (eval at create (/Users/tomgilligan/react_on_rails_tutorial_with_ssr_and_hmr_fast_refresh/node_modules/tapable/lib/HookCodeFactory.js:19:10), <anonymous>:300:1)
22:11:25 wp-client.1 |     at Hook.CALL_DELEGATE [as _call] (/Users/tomgilligan/react_on_rails_tutorial_with_ssr_and_hmr_fast_refresh/node_modules/tapable/lib/Hook.js:14:14)
22:11:25 wp-client.1 |     at Compiler.newCompilation (/Users/tomgilligan/react_on_rails_tutorial_with_ssr_and_hmr_fast_refresh/node_modules/webpack/lib/Compiler.js:988:26)
22:11:25 wp-client.1 |     at /Users/tomgilligan/react_on_rails_tutorial_with_ssr_and_hmr_fast_refresh/node_modules/webpack/lib/Compiler.js:1029:29
22:11:25 wp-client.1 |     at Hook.eval [as callAsync] (eval at create (/Users/tomgilligan/react_on_rails_tutorial_with_ssr_and_hmr_fast_refresh/node_modules/tapable/lib/HookCodeFactory.js:33:10), <anonymous>:62:1)
22:11:25 wp-client.1 |     at Hook.CALL_ASYNC_DELEGATE [as _callAsync] (/Users/tomgilligan/react_on_rails_tutorial_with_ssr_and_hmr_fast_refresh/node_modules/tapable/lib/Hook.js:18:14)
22:11:25 wp-client.1 |     at Compiler.compile (/Users/tomgilligan/react_on_rails_tutorial_with_ssr_and_hmr_fast_refresh/node_modules/webpack/lib/Compiler.js:1024:28)
22:11:25 wp-client.1 |     at /Users/tomgilligan/react_on_rails_tutorial_with_ssr_and_hmr_fast_refresh/node_modules/webpack/lib/Watching.js:112:19
22:11:25 wp-client.1 | exited with code 2
```

justin808:
@tomgilligan Please pull the latest from my branch. Updating dependencies fixed the issue.

justin808:
I made a video showing this working: https://youtu.be/lIkhnLXIH6I

vtamara:
I was able to run the proposed test on the following platform:

* OS: OpenBSD/adJ 6.9
* Ruby: 3.0.1
* Node: 12.16.1

I had to do the following changes:
1. In the `Gemfile` remove `ruby 2.7.3` and replace `mini_racer` with `execjs`
2. In `config/webpacker.yml` adding `poll: true` under `watch_options:`
3. In `package.json` adding `"style-loader": "^3.0.0"`


Thank you for the good job.

Issue #3035, opened 2021-06-01
v6 Docs Blocker -- big omission, function javascript_packs_tag ==> javascript_packs_with_chunks_tag not documented 


justin808:
We renamed 

> Change javascript_packs_with_chunks_tag and stylesheet_packs_with_chunks_tag to javascript_pack_tag and stylesheet_pack_tag.

https://github.com/rails/webpacker/blob/5-x-stable/docs/webpack.md#add-splitchunks-webpack-v4
But we don't document this anywhere...

justin808:
This was completed in #3036.

PR #3036, opened 2021-06-02
Update README.md for javascript_pack_tag

justin808:
Addresses #3035

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,16 +150,19 @@ The result looks like this:
 <script src="/packs/map-16838bab065ae1e314.js" data-turbolinks-track="reload"></script>
 ```
 
-**Important:** Pass all your pack names when using this helper otherwise you will
-get duplicated chunks on the page.
+**Important:** Pass all your pack names as multiple arguments, not multiple calls, when using **`javascript_pack_tag`** and the **`stylesheet_pack_tag`**. Otherwise, you will
+get duplicated chunks on the page. Be especially careful if you might be calling these view helpers from your view, partials, and the layout for a page. You will need some logic to ensure you call the helpers only once with multiple arguments.
 
 ```erb
 <%# DO %>
-<%= javascript_packs_with_chunks_tag 'calendar', 'map' %>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -135,6 +135,33 @@ You can then link the JavaScript pack in Rails views using the `javascript_pack_
...
+<script src="/packs/map~runtime-16838bab065ae1e314.js" data-turbolinks-track="reload"></script>
+<script src="/packs/map-16838bab065ae1e314.js" data-turbolinks-track="reload"></script>
+```
+
+**Important:** Pass all your pack names when using this helper otherwise you will
+get duplicated chunks on the page.
+
+```erb
+<%# DO %>
+<%= javascript_packs_with_chunks_tag 'calendar', 'map' %>
```
wasifhossain, line 158: guess it would be `javascript_pack_tag` rather than `javascript_packs_with_chunks_tag`

README.md
```diff
@@ -135,6 +135,33 @@ You can then link the JavaScript pack in Rails views using the `javascript_pack_
...
+
+<%# DON'T %>
+<%= javascript_packs_with_chunks_tag 'calendar' %>
+<%= javascript_packs_with_chunks_tag 'map' %>
```
wasifhossain, line 162: same here as above

justin808:
@wasifhossain Thanks for the review. I updated this. @gauravtiwari @jakeNiemiec Can we merge this ASAP?

justin808:
On a related note, we changed the `javascript_packs_with_chunks_tag` to be `javascript_pack_tag`, but maybe we should consider renaming it to `javascript_packs_tag` (packs, not pack) to be clear that this call takes multiple arguments and will generate multiple HTML tags. 

The advantage of this for migration purposes is that many apps will update to v6 and probably miss the need to have all the packs included in one call to the helper which is going to result in horrible performance as the browser will parse the JS multiple times.

dhh:
I think it's better to have poor performance than to break compatibility. But let's stress it loud and clear in the update docs!

PR #3041, opened 2021-06-09
Add config option to disable healthcheck

scarybot:
It appears that the rails proxy middleware is opening a dummy TCP connection to the webpack-dev-server upon every request, in order to test that it is still running. I think for many purposes this will be unneccessary, and we can introduce a configuration option to disable this behaviour. Removing this extra connection may reduce load and response times when using the webpack-dev-server, especially when working with large projects.

### DIFF
--- a/lib/install/config/webpacker.yml
+++ b/lib/install/config/webpacker.yml
@@ -37,6 +37,9 @@ development:
     compress: true
     # Note that apps that do not check the host are vulnerable to DNS rebinding attacks
     disable_host_check: true
+    # When this option is true, the rails proxy will assume that the dev server is always running.
+    # Otherwise, it will check connectivity by opening a test connection before every request.
+    disable_health_check: false
     # This option lets the browser open with your local IP
     use_local_ip: false
     # When enabled, nothing except the initial startup information will be written to the console.
@@ -44,9 +47,9 @@ development:
     quiet: false
     pretty: false
     headers:
-      'Access-Control-Allow-Origin': '*'
+      "Access-Control-Allow-Origin": "*"
     watch_options:
-      ignored: '**/node_modules/**'
+      ignored: "**/node_modules/**"
 
 test:
   <<: *default
--- a/lib/install/config/webpacker.yml
+++ b/lib/install/config/webpacker.yml
@@ -47,9 +47,9 @@ development:
     quiet: false
     pretty: false
     headers:
-      "Access-Control-Allow-Origin": "*"
+      'Access-Control-Allow-Origin': '*'
     watch_options:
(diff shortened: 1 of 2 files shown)


dhh:
Hey James. Do you have any stats or measurements on what performance boost you're seeing when you turn this off? I would think that TCP check would be near instant.

crawler:
@dhh i did measurements. And can provide them for you. But i came up with the different solution, a DevServerMonitor that does checks in the separate thread.
 
https://github.com/rails/webpacker/pull/3015/files#diff-c21b6aa9416a45bc8b0bf0398e0758ed3c4c3ec97c86c3f26ea932527e8a4798

crawler:
`puts Benchmark.measure { 10000.times { Webpacker.dev_server.running? } }`
`0.945449   3.024040   3.969489 (  7.010174)`

Looks like it takes almost no time: 0.0007010174s

IMHO the problem is that you can't avoid the dev server start or seamlessly handle it when it goes down without changing the configuration because then the timeout became the thing. But this is offtopic for this pr.

dhh:
I think the bar for introducing a configuration flag needs to be higher than shaving 0.0007s off per request, particularly in development mode! But if we uncover some other issue with this, happy to reconsider ✌️

dhh:
Ah, interesting. Do you have some stats on that setup? Just want to make sure we are curing something more than an aesthetic concept of whether these checks should be made or not.

dhh:
The other idea would be to simply memoize this for a set period of time. Even just 1s. So if you have 100 requests, it's still just 1 check per request.

crawler:
@dhh I had the iteration with memoization and it was working ok, but I ended up with the dev server listener, which can tell if the dev server is up or down even before the page is requested. But I don't know why this option is not taken in the count by you.

dhh:
You're talking about https://github.com/rails/webpacker/pull/3015/files#diff-c21b6aa9416a45bc8b0bf0398e0758ed3c4c3ec97c86c3f26ea932527e8a4798R1? That's just a massive amount of code to take on for this, which we're still trying to quantify how much of an issue it actually it is.

crawler:
@dhh. Yes, about that. I understand, please let me know when you will decide something.

crawler:
BTW i added some benchmarks https://github.com/rails/webpacker/pull/3015#issuecomment-863099263

dhh:
Right, need to set time aside to review that. Much larger PR than this.

crawler:
If required I can split it into independent parts. There are 3 of them.

crawler:
But that PR gives a whole view of performance improvement

PR #3058, opened 2021-06-18
Cleaning troubleshooting and adding global variables example

guillaumebriday:


docs/troubleshooting.md
```diff
@@ -56,25 +59,23 @@ completed the compilation successfully before loading a view.
...
-* This could happen if  `webpacker:install` step is skipped. Please run `bundle exec rails webpacker:install` to fix the issue.
+* This could happen if `webpacker:install` step is skipped. Please run `bundle exec rails webpacker:install` to fix the issue.
 
 * If you encounter the above error on heroku after upgrading from Rails 4.x to 5.1.x, then the problem might be related to missing `yarn` binstub. Please run following commands to update/add binstubs:
 
 ```bash
-bundle config --delete bin
-./bin/rails app:update:bin # or rails app:update:bin
+$ bundle config --delete bin
+$ ./bin/rails app:update:bin # or rails app:update:bin
```
jaredbeck, line 76: There is a slight downside to including `$` as a prompt in bash examples. It is more difficult to copy/paste the example. Not a big deal, just something I've noticed recently.

docs/troubleshooting.md
```diff
@@ -158,3 +157,34 @@ default: &default
   public_output_path: complaints_packs
   webpack_compile_output: true
 ```
+
+### Global Variables
+
+If you want to access any dependency without importing it everywhere or use it directly in your dev tools, please check: [https://webpack.js.org/plugins/provide-plugin/](https://webpack.js.org/plugins/provide-plugin/)
+
+**You don't need to assign dependencies on `window`.**
```
jaredbeck, line 165: Thanks for clarifying this!

docs/troubleshooting.md
justin808: I agree with @jaredbeck. What's the upside of the `$`? Many people use fancy prompts without the `$`.

docs/troubleshooting.md
```diff
@@ -101,53 +101,52 @@ Basic fix involves:
...
+$ make && make install
 
 # use sysconfcpus with elm-make
-mv $HOME/your_rails_app/node_modules/.bin/elm-make $HOME/your_rails_app/node_modules/.bin/elm-make-old
-printf "#\041/bin/bash\n\necho \"Running elm-make with sysconfcpus -n 2\"\n\n$HOME/dependencies/sysconfcpus/bin/sysconfcpus -n 2 $HOME/your_rails_app/node_modules/.bin/elm-make-old \"\$@\"" > $HOME/your_rails_app/node_modules/.bin/elm-make
-chmod +x $HOME/your_rails_app/node_modules/.bin/elm-make
+$ mv $HOME/your_rails_app/node_modules/.bin/elm-make $HOME/your_rails_app/node_modules/.bin/elm-make-old
+$ printf "#\041/bin/bash\n\necho \"Running elm-make with sysconfcpus -n 2\"\n\n$HOME/dependencies/sysconfcpus/bin/sysconfcpus -n 2 $HOME/your_rails_app/node_modules/.bin/elm-make-old \"\$@\"" > $HOME/your_rails_app/node_modules/.bin/elm-make
+$ chmod +x $HOME/your_rails_app/node_modules/.bin/elm-make
 ```
```
justin808, line 114: Same comment as above on adding the `$`.

docs/troubleshooting.md
```diff
@@ -158,3 +157,34 @@ default: &default
   public_output_path: complaints_packs
   webpack_compile_output: true
 ```
+
+### Global Variables
```
justin808, line 161: How about:

```
### Using Global Variables for Dependencies, like `$`
```

docs/troubleshooting.md
guillaumebriday: I find it easier to see the beginning of the command. I can remove it if needed

jaredbeck: > I can remove it if needed

It's a lot easier to copy/paste without it, thanks :)

PR #3059, opened 2021-06-18
Updating NODE ENV in test

### LINKED ISSUE #2602: Question: Shouldn't the default value in test.js be test?
Hi and thanks for the webpacker project. 👏 

Quick question. Shouldn't [line 1](https://github.com/rails/webpacker/blob/master/lib/install/config/webpack/test.js#L1) of test.js be `process.env.NODE_ENV = process.env.NODE_ENV || 'test'` ?

guillaumebriday:
Fix #2602

### DIFF
--- a/lib/install/config/webpack/test.js
+++ b/lib/install/config/webpack/test.js
@@ -1,4 +1,4 @@
-process.env.NODE_ENV = process.env.NODE_ENV || 'development'
+process.env.NODE_ENV = process.env.NODE_ENV || 'test'
 
 const webpackConfig = require('./base')
 


crawler:
Hi @gavinanthony,

Maybe I wrong but as I understand process.env.NODE_ENV == 'test' is used to run JS tests like a jest, but the test.js in the webpacker is for the rails application tests so JS should not be in the test mode. When I was needing to know if it the test environment in my JS for stubbing Stripe token or something, I set process.env.RAILS_ENV end exported it with webpack.EnvironmentPlugin. Looks like now it's done out of the box.

guillaumebriday:
You're right, my bad for the misunderstanding.

PR #3060, opened 2021-06-21
Skip contenthash when not production

justin808:
Per the webpack docs on better performance, the contenthash should be
skipped when not on production.

https://webpack.js.org/guides/build-performance/#avoid-production-specific-tooling

### DIFF
--- a/package/environments/base.js
+++ b/package/environments/base.js
@@ -77,12 +77,12 @@ const getPlugins = () => {
 
 // Don't use contentHash except for production for performance
 // https://webpack.js.org/guides/build-performance/#avoid-production-specific-tooling
-const filenameHash = isProduction ? '-[contenthash]' : '';
+const hash = isProduction ? '-[contenthash]' : ''
 module.exports = {
   mode: 'production',
   output: {
-    filename: `js/[name]${filenameHash}.js`,
-    chunkFilename: `js/[name]${filenameHash}.chunk.js`,
+    filename: `js/[name]${hash}.js`,
+    chunkFilename: `js/[name]${hash}.chunk.js`,
 
     // https://webpack.js.org/configuration/output/#outputhotupdatechunkfilename
     hotUpdateChunkFilename: 'js/[id].[fullhash].hot-update.js',
--- a/package/environments/base.js
+++ b/package/environments/base.js
@@ -8,6 +8,7 @@ const { sync: globSync } = require('glob')
 const WebpackAssetsManifest = require('webpack-assets-manifest')
 const webpack = require('webpack')
 const rules = require('../rules')
+const { isProduction } = require('../env')
 const config = require('../config')
 const { moduleExists } = require('../utils/helpers')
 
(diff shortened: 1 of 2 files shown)


thomasklemm:
Hey @justin808, this PR looks like it reverts the changes made in https://github.com/rails/webpacker/pull/2954, which implemented a fix for stylesheets not being reloaded in certain browsers in development. Would the issue described in that PR return if these changes were merged? Code-wise it seems to revert settings for the MiniCssExtractPlugin to what they were before https://github.com/rails/webpacker/pull/2954 was merged if I see correctly?

package/environments/base.js
```diff
@@ -63,24 +64,30 @@ const getPlugins = () => {
...
       })
     )
   }
 
   return plugins
 }
 
+// Don't use contentHash except for production for performance
+// https://webpack.js.org/guides/build-performance/#avoid-production-specific-tooling
+const filenameHash = isProduction ? '-[contenthash]' : '';
```
guillaumebriday, line 82:
```suggestion
const hash = isProduction ? '-[contenthash]' : '';
```

For consistency

dhh: We are not using semicolons at all. So should strip instead.

justin808:
> Hey @justin808, this PR looks like it reverts the changes made in #2954, which implemented a fix for stylesheets not being reloaded in certain browsers in development. Would the issue described in that PR return if these changes were merged? Code-wise it seems to revert settings for the MiniCssExtractPlugin to what they were before #2954 was merged if I see correctly?

Per
https://github.com/rails/webpacker/pull/3060/files#diff-7af8667a3e36201db57c02b68dd8651883d7bfc00dc9653661be11cd31feeccdR71-R72

CSS gets the `hash`, not `contenthash`.

justin808:
@guillaumebriday Comments addressed. Tomorrow, HST, I'll try to update https://github.com/shakacode/react_on_rails_tutorial_with_ssr_and_hmr_fast_refresh with this change to test this out.

t27duck:
I'm slightly perplexed as to what the end result is going to be now. For those of use using webpacker for straight up JS and CSS use (no react), does this plus HMR allow in dev mode to let us change JS and CSS and have those changes be shown in the browser?

If the filenames rename the same, browsers will treat it as a cached file and not attempt to pull in the new code. That would force developers to run their browsers with caching disabled via the network tab in the inspector.

That was the whole point of adding the hashes in https://github.com/rails/webpacker/pull/2954 for CSS as the JS code was being refreshed on save but not the CSS.

t27duck:
A more specific example:

I have apps thare are simply turbo, stimulus, and tailwind all going through the current webpacker beta. No react. No SSR. Just straight up bundled JS and CSS. I start up `bin/webpacker-dev-server` and start editing my JS and CSS. On save, the page refreshes and my changes show up because the CSS and JS filenames change. With this change, it looks to me like they will continue to alwyas be `application.js` and `application.css` which would cause the browser to not pull the latest files by default.

justin808:
I'm pretty sure that the Rails server lets the browser know to not cache the files during development.

t27duck:
Appolgies if I came off strong earlier. I managed (barely) to get master hooked up to an app and did confirm that these changes (plus #3031) does appear to properly refresh any changed CSS and JS (at least with some limited local testing).

The whole webpack/JS ecosystem is still a bit of a black box to me, but somehow this is pushing changes without having to have file names with different hashes in them. 🤷

PR #3083, opened 2021-07-06
Raise an error when javascript_pack_tag or stylesheet_pack_tag are called more than once

### LINKED ISSUE #3068: Need to have webpacker better support the situation of calling javascript_pack_tag multiple times during rendering
It's too difficult to depend on library users to know that they are calling `javascript_pack_tag` and `stylesheet_pack_tag` only once during the rendering cycle.

This is extremely deadly when using React as you get error like:

```
Original message: Hooks can only be called inside the body of a function component. (https://fb.me/react-invalid-hook-call)
```

Or other vague errors with React indicating that maybe multiple copies of react are included.

We have 2 ways to fix this:

1. Throw an error if this is detected during view rendering. While this seems unfriendly, it's WAY better than obscure errors. `Logger.warn` could also be done.
2.  Allow calling the view helper multiple times, so that "it just works."

# Current docs --> too hard

**Important:** Pass all your pack names as multiple arguments, not multiple calls, when using **`javascript_pack_tag`** and the **`stylesheet_pack_tag`**. Otherwise, you will
get duplicated chunks on the page. Be especially careful if you might be calling these view helpers from your view, partials, and the layout for a page. You will need some logic to ensure you call the helpers only once with multiple arguments.

```erb
<%# DO %>
<%= javascript_pack_tag 'calendar', 'map' %>
<%= stylesheet_pack_tag 'calendar', 'map' %>

<%# DON'T %>
<%= javascript_pack_tag 'calendar' %>
<%= javascript_pack_tag 'map' %>
<%= stylesheet_pack_tag 'calendar' %>
<%= stylesheet_pack_tag 'map' %>

### PR DESCRIPTION
**Pull request by guillaumebriday**
fix #3068

### CODE REVIEW

File: lib/webpacker/helper.rb [hunk 1]
```diff
@@ -141,6 +144,10 @@ def preload_pack_asset(name, **options)
   #   <%= stylesheet_pack_tag 'calendar' %>
   #   <%= stylesheet_pack_tag 'map' %>
   def stylesheet_pack_tag(*names, **options)
+    byebug
```
DylanFM, line 147: `byebug` call left in here

guillaumebriday: Oups my bad, fixed

lib/webpacker/helper.rb
```diff
@@ -96,6 +96,9 @@ def favicon_pack_tag(name, **options)
   #   <%= javascript_pack_tag 'calendar' %>
   #   <%= javascript_pack_tag 'map' %>
   def javascript_pack_tag(*names, **options)
+    raise "To prevent duplicated chunks on the page, you should call javascript_pack_tag only once on the page." if defined?(@javascript_pack_tag_loaded)
```
mcmire, line 99: Instead of raising an error that says "you shouldn't do this", is it possible to link to some kind of documentation that explains why this restriction is in place and a couple of ways the developer can resolve this? For instance, the new section that is getting added to the Usage section in the README. This is a change in behavior from the existing behavior in v5, so my fear is that even people who have been keeping up with the changes in the beta will be frustrated upon receiving this error.

olivierlacan: I agree. This is a fairly strong reaction to something that will not crash an app, but only cause rendering issues. 

For the record we could link to the section added by https://github.com/rails/webpacker/pull/3084.

dhh:
This could use a test.

lib/webpacker/helper.rb
justin808: @guillaumebriday How about if we **raise** by default for Rails.env.test? and Rails.env and warn otherwise?

justin808:
Feel free to reopen this PR here: 
https://github.com/shakacode/shakapacker/pulls

justin808:
Continue here: https://github.com/rails/webpacker/issues/3265

PR #3090, opened 2021-07-14
Update v6_upgrade.md

jaredbeck:
This config fixes the following fatal `webpacker:install` error.

```plain
bundle exec rails webpacker:install

Zeitwerk::NameError: wrong constant name Xyz-dashboard-app inferred by Module from directory

  /Users/jaredbeck/redacted/redacted/app/packs/xyz-dashboard-app

/Users/jaredbeck/redacted/redacted/config/environment.rb:7:in `<main>'
/Users/jaredbeck/redacted/redacted/bin/rails:4:in `<main>'
```


dhh:
Hey @fxn just added you here so we can track this is needed to fix before 6.0.final.

guillaumebriday:
@jaredbeck could you fix the conflict please?

fxn:
In this particular case, I believe it is a better fix to do this:

```ruby
ActiveSupport::Dependencies.autoload_paths.delete(...)
```

That is, that directory should not be in the autoload paths at the Rails level. Then, Rails behind the scenes configures Zeitwerk with what the user configured in the autoload paths, but that is part of the integration, not user-level.

fxn:
Let me say that I believe Rails is lacking interface here, you have `config.autoload_paths` for adding, but opting-out or removing is missing good API.

Maybe when after we stabilize in Rails 7, we can improve these last details to have it all round.

jaredbeck:
> @jaredbeck could you fix the conflict please?

Conflict resolved

guillaumebriday:
@fxn Do you suggest to merge this PR anyway?

We could upgrade that once it gets stabilized in Rails 7 maybe?

fxn:
@guillaumebriday I'd prefer the other one liner.

Rails has a public interface for configuring autoload paths, which consists of `config.autoload_paths` and `ActiveSupport::Dependencies.autoload_paths`. The directory `app/packs` should not belong to the autoload paths, so you have to say so using the existing public interface.

Note that `app/packs` should not belong to the autoload paths _regardless of the autoloader_. It shouldn't be there for classic either, right? It is not an autoload path, so it should not be among the autoload paths.

What I was saying is that the call to `delete` is not elegant for my taste, it is what Rails has had all this time, but we need a better interface at some point in my view. However, that's the one that we have today and can document.

fxn:
Even more, since `app/packs` does not belong to the autoload paths, ideally `webpacker` would do that `delete` itself, users should not need to do anything. Not familiar with the gem, does that make sense?

ahukkanen:
@fxn The fix you've added at 77231548ee9981568ffbcadd7256a3ededcee834 requires the `webpacker.yml` to exist already prior to running any `rails` commands, such as `rails webpacker:install`.

This is because the added initializer initializes `Webpacker::Configuration` which cannot load the configuration file here:
https://github.com/rails/webpacker/blob/1a7402cfd7c7ac4c04b6510fd53c2554069f7f7c/lib/webpacker/configuration.rb#L79

So, after the `webpacker` gem (6.0.0.rc.6) is added to the Gemfile, all `rails` commands are broken after that, unless the `webpacker.yml` file is added manually. This is a headache for app/plugin generators.

How to replicate the issue:

```bash
$ rails -v
Rails 6.1.4.1
$ rails new testapp --skip-webpack-install
[...]
$ cd testapp
$ sed -i "s/gem 'webpacker'.*/gem 'webpacker', '6.0.0.rc.6'/" Gemfile
$ cat Gemfile | grep webpacker
# Transpile app-like JavaScript. Read more: https://github.com/rails/webpacker
gem 'webpacker', '6.0.0.rc.6'
$ bundle
[...]
Installing webpacker 6.0.0.rc.6 (was 5.4.3)
Bundle complete! 17 Gemfile dependencies, 75 gems now installed.
Use `bundle info [gemname]` to see where a bundled gem is installed.
$ bundle exec rails webpacker:install
rails aborted!
Webpacker configuration file not found /.../testapp/config/webpacker.yml. Please run rails webpacker:install Error: No such file or directory @ rb_check_realpath_internal - /.../testapp/config/webpacker.yml
/.../testapp/config/environment.rb:5:in `<main>'
/.../testapp/bin/rails:5:in `<top (required)>'
/.../testapp/bin/spring:10:in `require'
/.../testapp/bin/spring:10:in `block in <top (required)>'
/.../testapp/bin/spring:7:in `tap'
/.../testapp/bin/spring:7:in `<top (required)>'
/.../testapp/bin/rails:2:in `load'
/.../testapp/bin/rails:2:in `<main>'
```

fxn:
@ahukkanen thanks, fixed in 9d55518ca3849a00f847cddfc1c9582ce36f0a26.

ahukkanen:
@fxn Thanks!

@fxn and/or @dhh could you cut a new RC release please, so that we could get app generators working?

fxn:
Whatever David says, he's the one with perspective about this gem.

PR #3091, opened 2021-07-15
Docs: v6 upgrade: fix instructions re: React

jaredbeck:
Without this fix, if a React user followed the instructions, they'd get an error from Babel about JSX being unsupported.


jaredbeck:
> Without this fix, if a React user followed the instructions, they'd get an error from Babel about JSX being unsupported.

This statement is correct for `6.0.0.pre.2`, but incorrect for `6.0.0.beta.7`, because `preset-react.js` no longer exists.

PR #3113, opened 2021-08-17
Create SECURITY.md

zidingz:
A simple instruction for security researchers.


dhh:
We rely on https://rubyonrails.org/security/ for all Rails gems.

PR #3115, opened 2021-08-18
Separate svg loader from file loader and add a custom generator

tonibardina:
Looks like you get rid off svg loader on `https://github.com/rails/webpacker/pull/2830`, I think its a good idea to keep loading it separately since svg files could be manipulated differently than other assets files. Also it would probably be better to add the `svg` files inlined thus we can get rid of some `HTTP` requests.

### DIFF
--- a/package/environments/__tests__/base.js
+++ b/package/environments/__tests__/base.js
@@ -8,6 +8,7 @@ const { chdirTestApp, chdirCwd } = require('../../utils/helpers')
 chdirTestApp()
 
 const { resolve } = require('path')
+
 const rules = require('../../rules')
 const baseConfig = require('../base')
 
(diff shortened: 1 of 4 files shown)


guillaumebriday:
I think it's the kind of thing that should not be configured by default with Webpacker because it will increase confusion for most people. If people need to inline svg, it's up to them to configure it. The [official Webpack documention](https://webpack.js.org/guides/asset-modules/#inlining-assets) is really neat about it

tonibardina:
Hey @guillaumebriday  thanks for the feedback. Why do you think its going to be confusing? Right now Webpacker is loading `svgs` as assets, why is one way more confusing than the other one? Loading `svgs` inline is something that was done on version `6.beta.2` but was wrongly implemented. I think loading `svgs` inline might be a better way since we can apply performance improvements on them

PR #3121, opened 2021-08-19
Specify webpack-dev-server to be v3

adrianbautista:
Otherwise when using Webpacker v5, it will install [webpack-dev-server v4](https://github.com/webpack/webpack-dev-server/blob/master/CHANGELOG.md#400-2021-08-18) which then requires `@webpack-cli/serve` to also be installed to properly run `bin/webpack-dev-server`.

When using Webpacker v5 with Webpack v4 and trying to run `bin/webpack-dev-server`:

```shell
$ bin/webpack-dev-server
The command moved into a separate package: @webpack-cli/serve
Would you like to install serve? (That will run npm install -D @webpack-cli/serve) (yes/NO) :
```

### DIFF
--- a/lib/install/template.rb
+++ b/lib/install/template.rb
@@ -51,7 +51,7 @@ say "Installing webpack and webpack-cli as direct dependencies"
 run "yarn add webpack@#{webpack_version} webpack-cli@#{webpack_cli_version}"
 
 say "Installing dev server for live reloading"
-run "yarn add --dev webpack-dev-server"
+run "yarn add --dev webpack-dev-server@^3"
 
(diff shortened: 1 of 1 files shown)


DavidColby:
Would it be possible to get a new release out with this change?

Creating a new Rails 6.1 application results in a broken `webpack-dev-server` since the latest release of webpacker (5.4.2) does not include the change in this PR. Figuring out what's going on with the error loop can be tricky for folks who are new to Rails or who are coming to Rails 6 for the first time from an older, webpacker-less version of Rails.

Recreate the issue with:
```
rails new some_new_app
bin/webpack-dev-server # Enter the error loop this PR fixes
```

dhh:
Yeah, should do that.

PR #3122, opened 2021-08-20
Update webpack-dev-server config to work with version 4

t27duck:
`webpack-dev-server` 4 dropped very recently. With it came a very large set of breaking changes regarding the configuration.

This updates the dev server configurtation in `development.js` and the default `webpacker.yml` files to work with it.

I can at least confirm that using this against a relatively basic Rails app using version 4 with tailwind + turbo works as "expected."

The upgrade guide for going from webpacker 5 to 6 pretty much does a redo on the installation (including uninstalling and reinstalling the latest `webpack-dev-server`), so in theory those on the upgrade path would need to review and reimplement any custom config changes anyway for their `webpacker.yml` file?

### DIFF
--- a/lib/install/config/webpacker.yml
+++ b/lib/install/config/webpacker.yml
@@ -24,25 +24,21 @@ development:
     https: false
     host: localhost
     port: 3035
-    public: localhost:3035
     # Hot Module Replacement updates modules while the application is running without a full reload
     hmr: false
-    # Should we show a full-screen overlay in the browser when there are compiler errors or warnings?
-    overlay: true
+    client:
+      # Should we show a full-screen overlay in the browser when there are compiler errors or warnings?
+      overlay: true
     # Should we use gzip compression?
     compress: true
     # Note that apps that do not check the host are vulnerable to DNS rebinding attacks
-    disable_host_check: true
-    # This option lets the browser open with your local IP
-    use_local_ip: false
-    # When enabled, nothing except the initial startup information will be written to the console.
-    # This also means that errors or warnings from webpack are not visible.
-    quiet: false
+    allowed_hosts: "all"
     pretty: true
     headers:
       'Access-Control-Allow-Origin': '*'
-    watch_options:
-      ignored: '**/node_modules/**'
+    static:
+      watch:
+        ignored: '**/node_modules/**'
 
 test:
   <<: *default
--- a/lib/install/config/webpacker.yml
+++ b/lib/install/config/webpacker.yml
@@ -29,6 +29,11 @@ development:
     client:
       # Should we show a full-screen overlay in the browser when there are compiler errors or warnings?
       overlay: true
+      # May also be a string
+      # webSocketURL:
+      #  hostname: "0.0.0.0"
+      #  pathname: "/ws"
+      #  port: 8080
     # Should we use gzip compression?
     compress: true
     # Note that apps that do not check the host are vulnerable to DNS rebinding attacks
--- a/package/environments/development.js
+++ b/package/environments/development.js
@@ -18,9 +18,29 @@ if (runningWebpackDevServer) {
     })
   }
 
-  let staticConfig = { publicPath: contentBase }
+  const devServerConfig = {
+    devMiddleware: {
+      publicPath
+    },
+    compress: devServer.compress,
+    allowedHosts: devServer.allowed_hosts,
+    host: devServer.host,
+    port: devServer.port,
+    https: devServer.https,
+    hot: devServer.hmr,
+    historyApiFallback: { disableDotRule: true },
+    headers: devServer.headers,
+    static: {
+      publicPath: contentBase
+    }
+  }
+
   if (devServer.static) {
-    staticConfig = { ...staticConfig, ...devServer.static }
+    devServerConfig.static = { ...devServerConfig.static, ...devServer.static }
+  }
+
+  if (devServer.client) {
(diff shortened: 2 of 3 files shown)


t27duck:
The failing Ruby tests are the same ones failing in master (the `defer` attribute on the script tags aren't found) and unrelated to this branch.

PikachuEXE:
For local testing I did

`Gemfile`:
```ruby
gem(
  "webpacker",
  ">= 6.0.0.rc1",
  git: "https://github.com/t27duck/webpacker.git",
  branch: "webpack-dev-server-update",
)
```

`package.json`
```js
"@rails/webpacker": "https://github.com/t27duck/webpacker.git#webpack-dev-server-update",
```


Test Execution for `webpack-dev-server`
```sh
ruby ./bin/webpack-dev-server
```

But I can't terminate with Ctrl+C immediately like v3 (it still terminates but takes 10+ seconds)
Not sure if issue for `webpacker` or `webpack-dev-server`

t27duck:
@PikachuEXE - Can't seem to reproduce either. Granted, I was on WSL2 for this branch but it does sound like a `webpack-dev-server` issue in general.

AlecRust:
When Ctrl+C from the dev server pressing Enter key seems to confirm the exit (there's a prompt that's not worded very well).

This branch worked well in my testing, though I had to install `core-js` which doesn't look to be part of the migration guide.

Looks like the Ruby tests are fixed on `master` by https://github.com/rails/webpacker/commit/f0a29a589e36421980213350c304f69d631e6561 so this branch could do with a rebase.

t27duck:
Rebased. Now everything is at least green.

ledermann:
I have tested this branch with one of my applications, and it appears that some things are still missing. My application uses `devServer.public` in the `webpacker.yml` file to configure a local domain instead of `localhost:3035`. This works with webpacker 6.0.0.rc.1 and webpack-dev-server v3, but not with your branch and v4. You find my repo here:
https://github.com/ledermann/templatus/pull/71

v4 supports this, but there are changes in the configuration:
https://github.com/webpack/webpack-dev-server/blob/master/migration-v4.md

It seems to me that `devServer.client.webSocketURL` should be set for this purpose.

t27duck:
@ledermann - Thanks for pointing that out. I've made some adjustments to the branch so now pretty much anything in `client:` in the yaml file will be a pass-thru. I left an example for `webSocketURL` in the default `webpacker.yml` file in this branch which should do it for you.

ledermann:
@t27duck Cool, this works. But there is one more thing: When setting `hmr` to `true`, live reloading is still enabled, which does not make any sense, see here:
https://webpack.js.org/configuration/dev-server/#devserverlivereload

I suggest adding this line to `development.js`:

```json
liveReload: !devServer.hmr
```

t27duck:
@ledermann - Ok, `devServer.liveReload` should now be set to the inverse of `devServer.hot`.

(If it's not obvious yet, I tend to run webpacker with defaults so any additional knobs and settings that others need please let me know).

PikachuEXE (approved):
LGTM after testing the latest change (`liveReload` value)

lib/install/config/webpacker.yml
```diff
@@ -24,25 +24,26 @@ development:
...
-    quiet: false
+    allowed_hosts: "all"
     pretty: true
     headers:
       'Access-Control-Allow-Origin': '*'
-    watch_options:
-      ignored: '**/node_modules/**'
+    static:
+      watch:
+        ignored: '**/node_modules/**'
```
justin808, line 46: @t27duck @guillaumebriday Do we need to be on the lookout for changes to the webpacker.yml file as updating this won't be obvious to current installations? Surely not existing beta users. I guess we'll need to double check the upgrade docs one final time before the final release.

guillaumebriday: Sure, I will open a PR for that to update the `upgrade.md` file. 👍

t27duck: @justin808 what I've been telling people in the various discords and slacks I'm in is to treat it like the "upgrade" from 5 to the betas where you back up the original file, take the new default from the gem, and reapply any custom changes.

package/environments/development.js
```diff
@@ -18,35 +18,41 @@ if (runningWebpackDevServer) {
...
+
+  if (devServer.static) {
+    devServerConfig.static = { ...devServerConfig.static, ...devServer.static }
+  }
+
+  if (devServer.client) {
+    devServerConfig.client = devServer.client
+  }
+
+  devConfig = merge(devConfig, {
```
olivierlacan, line 47: Doesn't this override all local (user-side) changes to the config though? We're trying to change this config and our changes keep getting overridden by defaults.

t27duck: What specific changes are you trying to make that are being overridden? I was under the impression that `merge` would more or less take the existing `devConfig` which is from the `yaml` file and stick on the `stats` and `devServer` values.

From what I can see, this is a similar pattern to webpacker 4 (old line 21 of this diff)

PikachuEXE:
Just found out `liveReload` is enabled accidentally due to 
```
liveReload: !devServer.hmr
```
What's supposed to be done when I want both HMR and liveReload disabled?

t27duck:
@PikachuEXE - Let me know if PR https://github.com/rails/webpacker/pull/3154 works for you.

PR #3134, opened 2021-08-24
Bump dependencies for both Node and Ruby

guillaumebriday:


guillaumebriday:
Oups rubocop is not happy, need to fix that

PR #3136, opened 2021-08-25
Update broken link on Readme.md

rapito:


guillaumebriday:
https://github.com/shakacode/react_on_rails/blob/master/docs/guides/rails-webpacker-react-integration-options.md#hmr-and-react-hot-reloading

Here is the correct link pointing to master

rapito:
> https://github.com/shakacode/react_on_rails/blob/master/docs/guides/rails-webpacker-react-integration-options.md#hmr-and-react-hot-reloading
> 
> Here is the correct link pointing to master, could you update your PR with this one?

I sent it to that commit on purpose so the issue wont happen again.
At the very least a tagged version should reference another tagged version.

dhh:
Removed for now.

PR #3141, opened 2021-08-28
Update README.md per behavior of HMR

justin808:
Note, this change only includes the README. 

No testing was done to confirm that current behavior is as described.

Most older installations have the source_entry_path under the source_path.

For example:
source_path: /app/javascript
source_entry_path:  /app/javascript/packs

We need to check that if one moves the source_entry_path outside of the source_path, that HMR behaves correctly? and did anybody check that additional_paths will also work for this purpose? If the webpack-dev-server recompiles when files change in those, then this is working, and then my PR is valid.


justin808:
Just updated to master to get this doc change to pass CI.

Issue #3138, opened 2021-08-25
Docs: `--colors` is not a valid options in webpack 5


tagliala:
https://github.com/rails/webpacker/blame/master/README.md#L214

```
$ ./bin/webpack --watch --colors --progress
[webpack-cli] Error: Unknown option '--colors'
[webpack-cli] Run 'webpack --help' to see available commands and options
```

Also, `colors` seems enabled by default

It may be also worth to note that `./bin/webpack --help` does not work anymore

### Webpacker v6

```
$ ./bin/webpack --help
[webpack-cli] Incorrect use of help
[webpack-cli] Please use: 'webpack help [command] [option]' | 'webpack [command] --help'
[webpack-cli] Run 'webpack --help' to see available commands and options
```

### Webpacker v5

```
$ bin/webpack --help
webpack-cli 3.3.12

Usage: webpack-cli [options]
...
```

guillaumebriday:
Ok I did some tests. Colors are enabled by default in Webpack 5, but we can disable them with `--no-color` and remove `--color` (note that is singular now) from your command.


The help command changed a lot between Webpack 4 and 5. I opened a PR to allow usage of these commands: #3144

PR #3144, opened 2021-08-28
Feature/webpack runner rework

guillaumebriday:
## Topics:

- Cleaning the WebpackRunner class
- Adding `--no-deprecation` flag.
- Allow the usage of Webpack default commands without `--config` by default.

Fix: #3138


tagliala:
Hi @guillaumebriday , shouldn't `--colors` be removed from the Readme?

guillaumebriday:
Nice catch, I will remove it too 👍

PR #3150, opened 2021-08-31
Fix GitHub workflow

crawler:


crawler:
@dhh sorry for the offtopic. But I'm don't know how to contact you in a different way to be heard. This summer, I published a stimulus helper gem that was inspired by the from_for helper. I just wanted to know, what do you think about adding its code to the stimulus-rails gem? 

IMHO the code of my views became a little bit less wordy after using it.

https://github.com/crawler/stimulus_tag_helper

PR #3151, opened 2021-08-31
Remove fingerprint only if  webpack-dev-server

justin808:
Standard for Rails development, unlike webpack guidelines, is to use
hashes for development.

We should be able to skip hashes for the webpack-dev-server.

### DIFF
--- a/package/environments/base.js
+++ b/package/environments/base.js
@@ -8,7 +8,7 @@ const { sync: globSync } = require('glob')
 const WebpackAssetsManifest = require('webpack-assets-manifest')
 const webpack = require('webpack')
 const rules = require('../rules')
-const { isProduction } = require('../env')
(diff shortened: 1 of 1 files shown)


justin808:
@Judahmeek please update the project that had troubles with this branch and check if this fixes the issue when using static compilation.

guillaumebriday:
LGTM 👌

t27duck:
A rebase may make the the checks green here. It did at least on my recent PR.

justin808:
Rebased and pushed.

justin808:
I just tested this again with both static builds and webpack-dev-serve per https://github.com/shakacode/react_on_rails_tutorial_with_ssr_and_hmr_fast_refresh/pull/16 and this PR is not necessary. See https://github.com/rails/webpacker/pull/3169.

Issue #2876, opened 2021-01-11
manifest.json is potentially sensitive


grncdr:
I'm not sure I'd classify this as a "security issue" as it's somewhat context sensitive, but I feel it deserves some consideration.

Webpacker uses the [WebpackAssetsManifest](https://github.com/webdeveric/webpack-assets-manifest) plugin to communicate the generated filenames from webpack back to Rails for use in view helpers (`javascripts_packs_with_chunks_tag` etc). This manifest file will contain entrypoint information about all packs present in `app/javascript/packs`, and will be served on the web at `/packs/manifest.json`.

This _could_ be considered a leak of private information if a user has multiple packs and expects that packs are only visible on "private" pages that include them using a view helper.

While obviously it's good practice to avoid committing anything sensitive into _any_ JS file, there are degrees of sensitivity and a single Rails app commonly serves multiple audiences (e.g. users/admins). It might be fine for an authenticated admin to receive some information in a JS bundle that should not be exposed to all users. Currently a user (or security researcher) can trivially view the `manifest.json` and download all assets for all packs.

As far as I can tell, nothing in the default webpacker setup _relies_ on being able to fetch `manifest.json` from the client, so I would propose that webpacker should change it's defaults to generate/read that file from another directory (e.g. `/config/webpack/manifest.json`).

What are your thoughts?

rossta:
Thanks for your question.

We have [precedent from a related issue regarding sourcemap defaults in Webpacker](https://github.com/rails/webpacker/issues/769#issuecomment-458216151) which can be summarized generally as "assets open by default."

While I agree with your assessment "nothing in the default webpacker setup relies on being able to fetch manifest.json from the client", I have used this feature for debugging on numerous occasions. 

If you have access to settings on your webserver, it should be possible to disable public access though I understand that does require extra work on the developer and may not be feasible in all environments.

Perhaps making the manifest path easier to customize could be a compromise.

grncdr:
A configuration option would be a good first step, but I still think the default should eventually be changed.

I feel that the issue here is qualitatively different from the one about source maps.* This is more like serving a default Apache `index.html` that lists all files in the assets tree. For me personally that is an unexpected behaviour, and does not align with other Rails defaults.

* something I've been doing and advocating for years now.

jdelStrother:
I've got a PR for discussion up here - https://github.com/rails/webpacker/pull/3152 - see what you think.

justin808:
[...]

PR #3152, opened 2021-09-01
Default the manifest path to app/packs/manifest.json

jdelStrother:
This makes the manifest path configurable, as well as defaulting it to app/packs/manifest.json

Previously discussed here: https://github.com/rails/webpacker/issues/2876

Not sure if app/packs is the best place for it, I'm open to other suggestions.

I wondered if we could get away with changing the default to app/packs since I _think_ by default it will continue to work with many people's setups. I think the main issue would be if, eg, someone is running `assets:precompile` on a CI server and tarring up public/packs to ship off to their production servers, they'll no longer have the manifest.json included in that tar file. But for anyone that just runs `assets:precompile` on all their production servers as part of a deploy, they won't notice much difference unless they're actually trying to load manifest.json out of their public path for some reason.

If we were able to get this into the 6.0 release that would be amazing because it would mean we don't have to argue about changing the default behaviour, but I know that 6.0 has already got RCs out so maybe it's a bit late.

One issue is that you can end up overwriting manifest.json with the test environment's version of it - prior to this, they'd be kept separate in public/packs/manifest.json & public/packs-test/manifest.json.  In my own app I tweak the test environment to avoid this:

```yaml
test:
  <<: *default
  compile: true

  # Compile test packs to a separate directory
  public_output_path: packs-test
  manifest_path: app/packs/manifest-test.json
```

Possibly that should be included in the default config.


jdelStrother:
If I fix the merge conflicts is there anything I can do to help get this merged, or is it a wontfix?

justin808:
@jdelStrother I'm not clear on why you would want the packs-test (for rails testing) manifest to be in the main packs (development and production) directory.

Is this a real security concern to be able to see the manifest file?

and if so, then it would make more sense to just move the location of the manifest to from /public/packs to /manifests/packs and /manifests/packs-test.

I'm not convinced that we need this.

@guillaumebriday I recommend that we close and reopen if a real security issue.

jdelStrother:
> I'm not clear on why you would want the packs-test (for rails testing) manifest to be in the main packs (development and production) directory.

I'm not sure I follow..? It's in RAILS_ROOT/app/packs, which is shared between all environments.

> Is this a real security concern to be able to see the manifest file?

TBD, I guess. In my opinion it's leaking information that might not be desirable to be leaked, but see https://github.com/rails/webpacker/issues/2876 for more discussion.

> and if so, then it would make more sense to just move the location of the manifest to from /public/packs to /manifests/packs and /manifests/packs-test.

As in RAILS_ROOT/manifests/? Are you suggesting introducing a new top-level directory, or is there other stuff in the rails ecosystem that writes to RAILS_ROOT/manifests/ ?

lib/install/config/webpacker.yml
```diff
@@ -3,6 +3,7 @@
 default: &default
   source_path: app/javascript
   source_entry_path: /
+  manifest_path: app/javascript/manifest.json
```
justin808, line 6: I think it's awkward to include a build-generated, not in git, file inside of `/app`

lib/install/template.rb
```diff
@@ -18,12 +18,13 @@
...
 if File.exists?(git_ignore_path)
   append_to_file git_ignore_path do
-    "\n"                   +
-    "/public/packs\n"      +
-    "/public/packs-test\n" +
-    "/node_modules\n"      +
-    "/yarn-error.log\n"    +
-    "yarn-debug.log*\n"    +
+    "\n"                              +
+    "/app/javascript/manifest.json\n" +
```
justin808, line 22: Per prior comment.

justin808 (changes requested):
I think having the manifest path be configurable is reasonable.

However, I'd have the default stay as the current location, rather than changing the default.

Why? Many build processes cache the webpack generated files and putting the output off somewhere else adds (unnecessary) complexity.

This PR is also missing a changelog entry and documentation.

PR #3154, opened 2021-09-02
Allow liveReload to be configurable in dev server

t27duck:
By default, liveReload for the `webpack_dev_server` config is the inverse of the hmr setting. There may be a usecase where one may want to control both values manually.

This allows liveReload to be configurable through `webpacker.yml` via the `live_reload` option. If not set, then the default of the inverse of hmr is used.

### DIFF
--- a/package/environments/development.js
+++ b/package/environments/development.js
@@ -12,6 +12,11 @@ let devConfig = {
 }
 
 if (runningWebpackDevServer) {
+  let liveReload = !devServer.hmr
+  if (devServer.live_reload !== undefined) {
+    liveReload = devServer.live_reload
+  }
+
   const devServerConfig = {
     devMiddleware: {
       publicPath
@@ -22,7 +27,7 @@ if (runningWebpackDevServer) {
     port: devServer.port,
     https: devServer.https,
     hot: devServer.hmr,
-    liveReload: !devServer.hmr,
+    liveReload,
     historyApiFallback: { disableDotRule: true },
     headers: devServer.headers,
     static: {
--- a/lib/install/config/webpacker.yml
+++ b/lib/install/config/webpacker.yml
@@ -26,6 +26,8 @@ development:
     port: 3035
     # Hot Module Replacement updates modules while the application is running without a full reload
     hmr: false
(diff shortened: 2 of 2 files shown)


package/environments/development.js
```diff
@@ -14,10 +14,15 @@ let devConfig = {
 if (runningWebpackDevServer) {
   if (devServer.hmr) {
     devConfig = merge(devConfig, {
-      output: { filename: '[name]-[hash].js' }
+      output: { filename: '[name]-[contentHash].js' }
```
PikachuEXE, line 17: I am not sure if contentHash should be used.
As #3060 skips `contentHash` for non prod environments.

PikachuEXE (commented):
Looks good except unsure about the `contentHash` change

package/environments/development.js
t27duck: Right now it's bringing up a deprecation warning. This at least eliminates that while maintaining the existing functionality.

It's also possible that something down the line is overriding this anyway...

PikachuEXE (approved):

PikachuEXE (approved):
Manually tested locally

PikachuEXE:
For those who want to test this locally simply update `package.json`:
```
    "@rails/webpacker": "https://github.com/t27duck/webpacker.git#live_reload_configurable",
```
Update `config/webpacker.yml` if necessary

t27duck:
Rebased with `master`. PR #3169 removed the hash part of the config so the "fix the deprication warning" is also removed from this PR.

PikachuEXE:
Anything else needs to be done to get this merged?

guillaumebriday:
Sounds good!

PikachuEXE:
Yup, that comment right there
I have liveReload disabled for ages in current version of webpacker
It's annoying if any small change (sometimes just adding/updating code comment) triggers a reload and lose the state on my dev website page.

justin808:
@t27duck @PikachuEXE, OK, I get it. If not HMR, then live reload may be on or off.

But what is the point of having liveReload true if HMR is true?

PR #3159, opened 2021-09-09
CI: Simplify ruby.yml workflow

olleolleolle:
This PR simplifies the **CI Ruby Workflow**.

- 🛠️  Quote `3.0` to avoid risking this Float becoming the String "3" when interpolating.
- 🧹  Format YAML lists so the GitHub editing UI likes it.
- 🧹  Remove unused "exclude:" directive for Ruby v2.4 entries.
- 🧹 Remove unused Node.js.
- ✅  Use setup-ruby's `bundler-cache: true` to do the whole of the caching & install.

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -11,7 +11,6 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        node: [14]
         ruby:
           - 2.7
           - "3.0"
@@ -39,14 +38,6 @@ jobs:
       BUNDLE_GEMFILE: ${{ matrix.gemfile }}
     steps:
       - uses: actions/checkout@v2
-
-      - name: Use Node.js ${{ matrix.node }}
-        uses: actions/setup-node@v1
-        with:
-          node-version: ${{ matrix.node }}
-
-      - name: Install yarn maybe
-        run: which yarn || npm install -g yarn
 
       - uses: ruby/setup-ruby@v1
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/ruby.yml
```diff
@@ -7,51 +7,36 @@ jobs:
...
             os: ubuntu-latest
             gemfile: gemfiles/Gemfile-rails-edge
             experimental: true
-          - ruby: 3.0
+          - ruby: "3.0"
             os: ubuntu-latest
             gemfile: gemfiles/Gemfile-rails-edge
             experimental: true
 
     steps:
```
bquorning, line 38: `BUNDLE_GEMFILE` needs to be set:
```suggestion
    env:
      BUNDLE_GEMFILE: ${{ matrix.gemfile }}
    steps:
```

olleolleolle: Perfect, that was it!

olleolleolle: That fixed my cut-and-paste mistake, thanks!

bquorning:
Perhaps too much for this PR, but… do you know why Node and Yarn is installed when we only run a Ruby process (`bundle exec rake`)? Perhaps it’s a copy/pasta from the JS test config files.

olleolleolle:
@bquorning https://github.com/rails/webpacker/blob/master/test/webpacker_test.rb#L18 in there, there's a Development Webpack server, which Ruby communicates with over a socket, perhaps that's a Node.js need?

bquorning (approved):

Issue #3005, opened 2021-05-03
Manifest file includes asset host


willcosgrove:
Hi, we're using this in a Rails application that is deployed on Heroku using their Pipeline feature.  After upgrading to the latest beta, I noticed that the manifest file that is generated includes the asset host configured for the application.

The issue we're having with this is that the asset host is different between application stages (staging uses one CDN, production uses another). So if I take a build of the application meant for staging, and try to directly promote it to master, it will be using the wrong CDN. Before upgrading, we were able to promote without having this issue.

My understanding of why this is needed is so that if you're doing some kind of code splitting where JS files are lazy loaded, the compiled JS files need to know where to grab the other split files from. Our JS files are split now after performing this upgrade to Webpacker 6, but they appear to all be eager loaded in the HTML by the `javascript_pack_tag` helper.

I realize that depending on your setup, to some this is a feature and not a bug. But because we are not lazy loading JS files from JS, this ends up causing more problems than it fixes. Is there a way to disable this functionality if we don't need it?

Thank you!

willcosgrove:
I think I've achieved the desired effect by setting the `WEBPACKER_ASSET_HOST` ENV variable to an empty string.

I tried that before I posted but it didn't work. I realized later that the JS hadn't rebuilt because no JS changes had happened, so it didn't rebuild the manifest file with the new ENV variable present. Once I forced the JS to rebuilt, the manifest was built without the CDN host.

PR #3164, opened 2021-09-17
Add troubleshooting section about CDN URL gotcha

ramhoj:
This PR adds a new section to the troubleshooting document explaining a gotcha where the javascript asset file can be incorrectly linked when `config.asset_host` is configured with different URLs in different environments and assets are not re-compiled between those environments.

As the new section explains this is common on Heroku where the user is encouraged to use their pipeline promote feature that does not re-compile assets between apps.

A common pattern is to use one app as a staging environment to evaluate builds. Upon acceptance, the staging app is “promoted” to the production environment. If `WEBPACKER_ASSET_HOST` is not set to an empty value in the staging application that application’s CDN URL will be compiled into the manifest.json file and reused in the production application.

This issue can be tricky to detect as there might be no immediate error. The file probably still exists as it’s used by staging. When the staging environment later is re-deployed with new finger-printed assets the CDN will eventually remove the asset files that production is still dependent on.

All credit to @willcosgrove who provided this solution in https://github.com/rails/webpacker/issues/3005.

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -188,3 +188,21 @@ module.exports = merge(webpackConfig, {
   ],
 })
 ```
+
+## Wrong CDN src from javascript_pack_tag
+
+If your deployment doesn't rebuild assets between environments (such as when
+using Heroku's Pipeline promote feature). You might find that your production
+application is using your staging `config.asset_host` URL when using
+`javascript_pack_tag`.
+
+This can be fixed by setting the environment variable `WEBPACKER_ASSET_HOST` to
+an empty string where your assets are compiled. On Heroku this is done under
+*Settings* -> *Config Vars*.
+
+This way webpacker won't hard-code the CDN URL into the manifest file used by
+`javascript_pack_tag`, but instead fetch the CDN at runtime, resolving the
+issue.
+
+See [this issue](https://github.com/rails/webpacker/issues/3005) for more
+details.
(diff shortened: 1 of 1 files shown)


guillaumebriday:
I might be wrong but if the `WEBPACKER_ASSET_HOST` is empty the output will look something like that `/packs/foo` right?

Is this something we really want? If the CDN is different between staging and production maybe we should recompile assets no?

ramhoj:
In our application (Rails 6.1.4, webpacker 6.0.0.rc.5) the output includes the `config.asset_host` URL resulting in e.g *https://d29i806g85j6o3.cloudfront.net/packs/js/927-232a4435482e775b09be.js* as output in the browser. The output to the manifest file should be what you describe, as the host not being present in the compiled manifest file is what fixed the issue.

Is it for performance reasons that the full URL is normally compiled into the manifest file? While recompiling the assets per environment would most probably fix the issue it goes against Heroku's default workflow of *promoting* one environment to another. Heroku customers can achieve this by pushing directly to the production environment to trigger a build there. However this is way slower, our slug tages about 10 minutes to compile on Heroku but only a couple of seconds to promote.

justin808:
@guillaumebriday this change sounds reasonable and it should get merged.

It's worth noting that for React on Rails Pro applications that use dynamic code splitting using loadable-components, it's very important to ensure that the JS files don't get any full URLs compiled into them.

PR #3181, opened 2021-10-02
Add name for chunks

pustovalov:
https://webpack.js.org/configuration/optimization/#optimizationchunkids

before:
```
//manifest.json

    "bundle_highcharts": {
      "assets": {
        "js": [
          "/packs-test/js/runtime-961a03111ca01e88c396.js",
          "/packs-test/js/8840-7edb6b003836ece1c4e4.js",
          "/packs-test/js/bundle_highcharts-cfe86310052c07d694aa.js"
        ]
      }
    },

```

after:
```
//manifest.json

    "bundle_highcharts": {
      "assets": {
        "js": [
          "/packs-test/js/runtime-f6a92c002c82d2eef072.js",
          "/packs-test/js/vendors-node_modules_highcharts_highcharts_js-9c63cc616e88fd4499b6.js",
          "/packs-test/js/bundle_highcharts-a3894b57812733f3fe81.js"
        ]
      }
    },

```

### DIFF
--- a/package/environments/base.js
+++ b/package/environments/base.js
@@ -104,6 +104,7 @@ module.exports = {
   },
 
   optimization: {
+    chunkIds: 'named',
     splitChunks: { chunks: 'all' },
 
     runtimeChunk: 'single'


justin808:
@pustovalov how would this change benefit a project? What if many names are involved? In other words, how does this help your debugging?

From the docs https://webpack.js.org/configuration/optimization/#optimizationchunkids
> Also if the environment is development then optimization.chunkIds is set to 'named', while in production it is set to 'deterministic'

So the defaults are what we want, rather than setting the production value to be what's the default for development.

@guillaumebriday This should probably be closed as it conflicts with the defaults for webpack which seem appropriate.

PR #3182, opened 2021-10-02
Add core-js as webpacker dependency, update core-js version for @babe…

pustovalov:
…l/preset-env

In webpacker v6 users are advised to use the webpacker babel preset instead of their config:

https://github.com/rails/webpacker/blob/155a4f8a576fb5224442a79e075d10671d5e6c62/docs/v6_upgrade.md
```
Remove babel.config.js if you never changed it. Be sure to have this config in your package.json

"babel": {
  "presets": [
    "./node_modules/@rails/webpacker/package/babel/preset.js"
  ]
}
```

But webpacker babel config has this one:
https://github.com/rails/webpacker/blob/155a4f8a576fb5224442a79e075d10671d5e6c62/package/babel/preset.js#L24

to use `useBuiltIns: 'entry'` need to import `core-js` to an application:
https://babeljs.io/docs/en/babel-preset-env#usebuiltins-entry

and this one:
https://github.com/rails/webpacker/blob/155a4f8a576fb5224442a79e075d10671d5e6c62/package/babel/preset.js#L25
because of this line `core-js` should be dependency of webpacker in my opinion

```
It is recommended to specify the minor version otherwise "3" will be interpreted as "3.0" which may not include polyfills for the latest features.
```
https://babeljs.io/docs/en/babel-preset-env#corejs


pustovalov:
@guillaumebriday WDYT?

justin808:
@pustovalov can you:

1. rebase on master
2. Provide a sample app to verify your change.

Note, we're considering #3234, which would change this recommendation.

justin808:
Not needed with peer dependencies in https://github.com/shakacode/shakapacker

Issue #3183, opened 2021-10-03
[webpacker 6.0.0.rc.5] stylesheet_pack_tag doesn't work with HMR


pustovalov:
```
// webpacker.yml

hmr: false
```
![image](https://user-images.githubusercontent.com/1568885/135753441-0d2a38fb-5034-4d2f-a1f6-43008108d8fc.png)

```
// webpacker.yml

hmr: true
```

![image](https://user-images.githubusercontent.com/1568885/135753473-6154be93-65a9-44cd-8827-2a5e46ed12c7.png)

Demo repo:
https://github.com/pustovalov/webpacker-hmr-reproduce

guillaumebriday:
Could you take a look @justin808 please?

justin808:
Should we close this one? 

I don't see any documentation of the issue above.

justin808:
Seems like a misunderstanding that HMR page should have the CSS  on the HTML. It just shouldn't.

pustovalov:
>Should we close this one?

Need to describe it in the documentation at least. Webpacker v6 doesn't support HMR and CSS outside of JS together. Because this combination worked good in Webpacker v5.

justin808:
@pustovalov I think this should be closed. We've got support for CSS with HMR.

For React, here's the configuration for HMR.

https://github.com/rails/webpacker/pull/3224

pustovalov:
>We've got support for CSS with HMR.

@justin808 where was this support added? `stylesheet_pack_tag` still returns `nil`: https://github.com/rails/webpacker/blob/2f4df42c9a4b8d6f9838be0b60c599c8f1c5e363/lib/webpacker/helper.rb#L144

justin808:
It's supposed to return blank if that value is true.

Just setup your webpscker config so that value is false.

PR #3184, opened 2021-10-03
support stylesheet_pack_tag with HMR

pustovalov:
Closes https://github.com/rails/webpacker/issues/3183

Webpack with HRM can work with styles that are imported from JS, but can't work with styles outside of JS scope
https://webpack.js.org/guides/hot-module-replacement/#hmr-with-stylesheets

### DIFF
--- a/lib/webpacker/helper.rb
+++ b/lib/webpacker/helper.rb
@@ -141,8 +141,6 @@ module Webpacker::Helper
   #   <%= stylesheet_pack_tag 'calendar' %>
   #   <%= stylesheet_pack_tag 'map' %>
   def stylesheet_pack_tag(*names, **options)
-    return "" if Webpacker.inlining_css?
-
     stylesheet_link_tag(*sources_from_manifest_entrypoints(names, type: :stylesheet), **options)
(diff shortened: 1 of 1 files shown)


guillaumebriday:
@justin808 is that ok for you?

lib/webpacker/helper.rb
```diff
@@ -141,8 +141,6 @@ def preload_pack_asset(name, **options)
   #   <%= stylesheet_pack_tag 'calendar' %>
   #   <%= stylesheet_pack_tag 'map' %>
   def stylesheet_pack_tag(*names, **options)
-    return "" if Webpacker.inlining_css?
```
justin808, line 144: Removing this line definitely breaks HMR with CSS modules.

If somebody doesn't want HMR, the CSS is exported. So this is a doc issue if anything.

pustovalov: Yeah, this PR will not help with that issue. It would be great to restore the behavior from the previous version of webpacker when HMR works with `stylesheet_pack_tag`

Issue #3153, opened 2021-09-01
Cannot use Stimulus 3.0.0-beta.1 with Webpacker 6.0.0.rc.5


dwightwatson:
I've just tried upgrading my Rails app with Webpacker 6.0.0.rc.5 to use Stimulus 3.0.0-beta.1. However, I get an error that I cannot invoke the Controller constructor.

```
application.js:75 Uncaught (in promise) TypeError: Class constructor Controller cannot be invoked without 'new'
    at new _default (application.js:75)
    at new extended (vendors-node_modules_hotwired_stimulus_dist_stimulus_js.js:1327)
    at new Context (vendors-node_modules_hotwired_stimulus_dist_stimulus_js.js:1187)
    at Module.fetchContextForScope (vendors-node_modules_hotwired_stimulus_dist_stimulus_js.js:1389)
    at Module.connectContextForScope (vendors-node_modules_hotwired_stimulus_dist_stimulus_js.js:1375)
    at Router.scopeConnected (vendors-node_modules_hotwired_stimulus_dist_stimulus_js.js:1694)
    at ScopeObserver.elementMatchedValue (vendors-node_modules_hotwired_stimulus_dist_stimulus_js.js:1612)
    at ValueListObserver.tokenMatched (vendors-node_modules_hotwired_stimulus_dist_stimulus_js.js:803)
    at TokenListObserver.tokenMatched (vendors-node_modules_hotwired_stimulus_dist_stimulus_js.js:736)
    at vendors-node_modules_hotwired_stimulus_dist_stimulus_js.js:730
```

It works fine with Stimulus 2. If I downgrade to Webpacker 5 with the Stimulus 3.x beta it also works fine. It is only broken with the Webpacker beta and Stimulus beta.

Here is the initial issue on the hotwired/stimulus#433 repo.

Here is a minimal reproduction with a fresh Rails/Webpacker beta/Stimulus beta setup: https://github.com/dwightwatson/stimulus-3-reproduction

dhh:
I replicated this as well in my testing. Def need to sort it.

dhh:
Choosing to pursue a path that's not contingent on Webpack API changes as explained in https://github.com/hotwired/stimulus/issues/433#issuecomment-917363278

dhh:
I can't replicate this using webpack via jsbundling-rails. Maybe something else is wrongly setup in Webpacker. Anyway, reopening.

adrienpoly:
[...]

PR #3189, opened 2021-10-05
Remove loose: true to support Stimulus 3.x

dwightwatson:
This fixes an issue with the babel preset which prevents Stimulus 3.x from working.

See #3153 for more information about the issue and the proposed fix.

### DIFF
--- a/package/babel/preset.js
+++ b/package/babel/preset.js
@@ -25,7 +25,6 @@ module.exports = function config(api) {
           corejs: '3.8',
           modules: 'auto',
           bugfixes: true,
-          loose: true,
           exclude: ['transform-typeof-symbol']
         }
       ],


guillaumebriday:
From my first tests, this is not the root cause of the issue in many cases, we need to dive deeper to find how why this option could be a problem or not. It could be many other things.

What versions are you using?

dwightwatson:
I'm using Webpacker 6.0.0.rc.5 and Stimulus 3.0.0.

I'm following the commentary from #3153 which suggested that this was fix.

mrhead:
A combination of different things causes the issue with Stimulus 3, and the `loose` option is one of them.

- When `loose` is set to `true` then [webpack produces ES5 code](https://2ality.com/2015/12/babel6-loose-mode.html).
- [Compile target for Stimulus 3 has changed to ES6](https://github.com/hotwired/stimulus/releases/tag/v3.0.0-beta.1).
- Webpacker stopped compiling code in `node_modules`, therefore application code gets compiled to ES5, but code from `node_modules` remains the same (i.e., ES6).

Because of this, ES5 application controller classes don't work with ES6 Stimulus classes. Disabling the `loose` option fixes it because the application code gets compiled to ES6.

PR #3195, opened 2021-10-11
Support node LTS and future "Current" releases

aried3r:
Drops support for Node.js 13 and 15.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "lib/install/config/webpacker.yml"
   ],
   "engines": {
-    "node": ">= 12.13.0 || >=14",
+    "node": "^12.13.0 || ^14 || >=16",
     "yarn": ">=1 <4"
   },
   "dependencies": {


justin808:
@aried3r Why is this required? What benefit?

aried3r:
> Why is this required? What benefit?

It is not required, but EOL node.js versions have been dropped in the past, see https://github.com/rails/webpacker/pull/2749 and https://github.com/rails/webpacker/pull/2993. Since webpacker 6 stable is not yet released, I thought it'd be a good idea to drop EOL node.js versions unless webpacker is going to keep maintaining support for them. WDYT? However they are not part of the test matrix as of https://github.com/rails/webpacker/pull/3185 (that's my own PR, however).

package.json
```diff
@@ -8,7 +8,7 @@
     "lib/install/config/webpacker.yml"
   ],
   "engines": {
-    "node": ">= 12.13.0 || >=14",
+    "node": "^12.13.0 || ^14 || >=16",
```
guillaumebriday, line 11: Why not that?

```suggestion
    "node": "^12.13.0 || ^14 || ^16",
```

guillaumebriday:
And we could change https://github.com/rails/webpacker/blob/master/.github/workflows/js-lint.yml#L12 to `16` right?

package.json
aried3r: That would not allow users to use version 17 of node.js.

PR #3197, opened 2021-10-12
[5-x-stable] Bump optimize-css-assets-webpack-plugin to ^6.0.1

sikachu:
This PR updates `optimize-css-assets-webpack-plugin` package to `^6.0.1`.

We need to update `optimize-css-assets-webpack-plugin` to version 6.0.1 or later as version 5.0.8 eventually resolves to install a vulnerable version of `nth-check` (1.0.2): https://github.com/advisories/GHSA-rp65-9cf3-cjxr

You can see the dependencies from this reconstructed dependency tree:

```
@rails/webpacker@5.4.3
└─ optimize-css-assets-webpack-plugin@5.0.8
   └─ cssnano@4.1.11
      └─ cssnano-preset-default@4.0.8
          └─ postcss-svgo@4.0.3
             ├─ css-select@2.1.0
             │  └─ nth-check@1.0.2
             └─ nth-check@1.0.2
```

Updating `optimize-css-assets-webpack-plugin` to version 6.0.1 will update the dependency tree to look like this:

```
@rails/webpacker
└─ optimize-css-assets-webpack-plugin@6.0.1
   └─ cssnano@5.0.8
      └─ cssnano-preset-default@5.1.4
         └─ postcss-svgo@5.0.2
            └─ svgo@2.7.0
               └─ css-select@4.1.3
                  └─ nth-check@2.0.1
```

I followed the instructions in CONTRIBUTING.md and got a green test suite on both `yarn test` and `rake test`, so I don't think this will cause any issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "glob": "^7.1.7",
     "js-yaml": "^3.14.1",
     "mini-css-extract-plugin": "^0.9.0",
-    "optimize-css-assets-webpack-plugin": "^5.0.8",
+    "optimize-css-assets-webpack-plugin": "^6.0.1",
     "path-complete-extname": "^1.0.0",
     "pnp-webpack-plugin": "^1.7.0",
     "postcss-flexbugs-fixes": "^4.2.1",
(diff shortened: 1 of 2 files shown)


sikachu:
Sure thing. I can look into it.

Alternatively, I think #2802 actually [remove this dependency and made it on-demand](https://github.com/rails/webpacker/pull/2802/files#diff-a930e3d8169bcd5fc623a99a345b8ec03139f79e9dbf6277b4d1b11a59083b45R28) (i.e. it won't be used if it's not installed). Maybe we can also do something similar here as well, but I'm not sure if that would be appropriate for a minor/patch release change.

justin808:
See comment on #3217. This can probably be resolved either by upgrading or using yarn resolutions.

PR #3213, opened 2021-11-03
Remove duplicate browserlist line in v6 upgrade 📖

sunny:
Both steps 5 and 10 of the v6 upgrade guide asked to remove `.browserslistrc`. This PR keeps only the last one.

Also:
- reformatted list a bit (spacing and indented code blocks)
- mention that the `bin/yarn` to create must be executable
- added `assets:precompile` as a final check
- added `assets:clobber` reminder


guillaumebriday:
Thanks @sunny 🙏

PR #3218, opened 2021-11-08
Make Compiler#watched_files_digest thread safe.

jessewmc:
Removes Dir.chdir call which raises error if two threads both try to call it. Instead we manually expand the root path and join relative watch paths to it before calling Dir.

See https://github.com/rails/webpacker/pull/2788#issuecomment-863898053 for motivation.

### DIFF
--- a/lib/webpacker/compiler.rb
+++ b/lib/webpacker/compiler.rb
@@ -53,11 +53,13 @@ class Webpacker::Compiler
 
     def watched_files_digest
       warn "Webpacker::Compiler.watched_paths has been deprecated. Set additional_paths in webpacker.yml instead." unless watched_paths.empty?
-      Dir.chdir File.expand_path(config.root_path) do
-        files = Dir[*default_watched_paths, *watched_paths].reject { |f| File.directory?(f) }
-        file_ids = files.sort.map { |f| "#{File.basename(f)}/#{Digest::SHA1.file(f).hexdigest}" }
-        Digest::SHA1.hexdigest(file_ids.join("/"))
+      root_path = File.expand_path(config.root_path)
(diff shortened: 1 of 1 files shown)


justin808:
@jessewmc can you describe the reproduction steps?

justin808:
@guillaumebriday I recommend that we close this issue as it can be reopened once we have clear repro steps.

lib/webpacker/compiler.rb
```diff
@@ -53,11 +53,13 @@ def last_compilation_digest
 
     def watched_files_digest
       warn "Webpacker::Compiler.watched_paths has been deprecated. Set additional_paths in webpacker.yml instead." unless watched_paths.empty?
-      Dir.chdir File.expand_path(config.root_path) do
-        files = Dir[*default_watched_paths, *watched_paths].reject { |f| File.directory?(f) }
-        file_ids = files.sort.map { |f| "#{File.basename(f)}/#{Digest::SHA1.file(f).hexdigest}" }
-        Digest::SHA1.hexdigest(file_ids.join("/"))
+      root_path = File.expand_path(config.root_path)
+      expanded_paths = [*default_watched_paths, *watched_paths].map do |path|
+        File.join(root_path, path)
```
justin808, line 58: There is a bug here if path and root_path are full paths. 

I fixed this here: 

https://github.com/rails/webpacker/pull/3233/files#diff-e9bc3ece054a6a66b7673121bfb015e820b756e2a0652de74befd08e2cf19386R66-R68

justin808:
@guillaumebriday We can close this one and consider merging my fix: 
https://github.com/rails/webpacker/pull/3233

@jessewmc did tests fail locally for you?

justin808:
See #3233

PR #3220, opened 2021-11-09
Add fast refresh to default babel if packages exist

justin808:
With this change, it's extremely easy to make React work with HMR.

Here's a PR in my example project demonstrating the improvement:

https://github.com/shakacode/react_on_rails_tutorial_with_ssr_and_hmr_fast_refresh/pull/20

### DIFF
--- a/README.md
+++ b/README.md
@@ -323,6 +323,9 @@ Webpacker out of the box supports JS and static assets (fonts, images etc.) comp
 
 ```bash
 yarn add react react-dom @babel/preset-react
+
+## add hmr ones only for development
+yarn add --dev @pmmmwh/react-refresh-webpack-plugin react-refresh
 ```
 
 ...if you are using typescript, update your `tsconfig.json`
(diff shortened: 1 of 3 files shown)


guillaumebriday:
I'm not sure this should be in the code because it seems to be really tight to React and I think Webpacker should not add framework related code in its own codebase.

What do you think?

justin808:
I added because we have other React code:

https://github.com/rails/webpacker/blob/06dcf5ff18123a5a9d193ed84e5c3f9370c26c16/package/babel/preset.js#L36

So we might as well support HMR if we're going to support React.

Or we should remove React.

I think it's fine to support React by checking if the packages are installed.

"It just works"

guillaumebriday:
As discussed, we should close this PR for https://github.com/rails/webpacker/pull/3224 right? 👍

guillaumebriday:
@justin808 Should be close this PR?

justin808:
Sure. Let's close it. I think we should merge #3224.

Issue #940, opened 2017-10-13
Running `bundle exec rake webpacker:install` gives error: `Webpacker configuration file not found /vagrant_host_beeline/config/webpacker.yml. Please run rails webpacker:install Error: No such file or directory @ rb_sysopen - `


Taytay:
The first time I tried to run `webpacker:install`, it told me that my version of yarn was out of date, so I updated it. Then, I re-ran `webpacker:install` and got this message. I assume I'm doing something wrong, as the install script appears to be copying the webpacker.yml that it's later complaining about. I'm a Rails newb, so me making a mistake is pretty likely...

```
$ bundle exec rake webpacker:install
rake aborted!
Webpacker configuration file not found /vagrant_host_beeline/config/webpacker.yml. Please run rails webpacker:install Error: No such file or directory @ rb_sysopen - /vagrant_host_beeline/config/webpacker.yml
/home/vagrant/.rvm/gems/ruby-2.3.1/gems/webpacker-3.0.2/lib/webpacker/configuration.rb:65:in `rescue in load'
/home/vagrant/.rvm/gems/ruby-2.3.1/gems/webpacker-3.0.2/lib/webpacker/configuration.rb:62:in `load'
/home/vagrant/.rvm/gems/ruby-2.3.1/gems/webpacker-3.0.2/lib/webpacker/configuration.rb:9:in `refresh'
/home/vagrant/.rvm/gems/ruby-2.3.1/gems/webpacker-3.0.2/lib/webpacker/commands.rb:14:in `bootstrap'
/home/vagrant/.rvm/gems/ruby-2.3.1/gems/webpacker-3.0.2/lib/webpacker.rb:18:in `bootstrap'
/home/vagrant/.rvm/gems/ruby-2.3.1/gems/webpacker-3.0.2/lib/webpacker/railtie.rb:37:in `block in <class:Engine>'
/home/vagrant/.rvm/gems/ruby-2.3.1/gems/railties-4.2.4/lib/rails/initializable.rb:30:in `instance_exec'
/home/vagrant/.rvm/gems/ruby-2.3.1/gems/railties-4.2.4/lib/rails/initializable.rb:30:in `run'
/home/vagrant/.rvm/gems/ruby-2.3.1/gems/railties-4.2.4/lib/rails/initializable.rb:55:in `block in run_initializers'
/home/vagrant/.rvm/gems/ruby-2.3.1/gems/railties-4.2.4/lib/rails/initializable.rb:54:in `run_initializers'
[...]

PR #3223, opened 2021-11-12
Fix so webpacker:install does not require webpacker.yml pre-existing

justin808:
If you run ./bin/rails webpacker:install on a rails app with no
config/webpacker.yml, we get a crash with this error:

> Webpacker configuration file not found webpacker-v6/config/webpacker.yml.

> Please run rails webpacker:install Error: No such file or directory @
rb_check_realpath_internal - webpacker-v6/config/webpacker.yml

With this fix, allow the installer to run.

Problem solved with a global to indicate installing so that default
configuration is used rather than crashing because the yet to be
installed config/webpacker.yml is not yet installed.

Fixes this old, very confusing issue https://github.com/rails/webpacker/issues/940#issuecomment-757482835.

### DIFF
--- a/lib/tasks/webpacker/install.rake
+++ b/lib/tasks/webpacker/install.rake
@@ -4,7 +4,8 @@ bin_path = ENV["BUNDLE_BIN"] || Rails.root.join("bin")
 namespace :webpacker do
   desc "Install Webpacker in this application"
   task install: [:check_node, :check_yarn] do |task|
-    @@webpacker_installing = true
+    Webpacker::Configuration.installing = true
+
     prefix = task.name.split(/#|webpacker:install/).first
 
     if Rails::VERSION::MAJOR >= 5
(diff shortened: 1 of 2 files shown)


lib/tasks/webpacker/install.rake
```diff
@@ -4,6 +4,7 @@ bin_path = ENV["BUNDLE_BIN"] || Rails.root.join("bin")
 namespace :webpacker do
   desc "Install Webpacker in this application"
   task install: [:check_node, :check_yarn] do |task|
+    @@webpacker_installing = true
```
justin808, line 7: @dhh is this the right way to pipe the flag over that we're doing the installation?

guillaumebriday:
Great! LGTM, waiting for an answer to your question before merging

lib/webpacker/configuration.rb
```diff
@@ -80,10 +84,13 @@ def load
       end
       config[env].deep_symbolize_keys
     rescue Errno::ENOENT => e
-      raise "Webpacker configuration file not found #{config_path}. " \
-            "Please run rails webpacker:install " \
-            "Error: #{e.message}"
-
+      if self.class.installing
```
justin808, line 87: @guillaumebriday @dhh I think this is a better way to plumb the setting of "installing."

Let me know if there is a better way.

justin808:
@guillaumebriday this one should be ready for merging.

guillaumebriday:
Great!

PR #3224, opened 2021-11-15
Move react dependencies to docs.

justin808:
This commit removes React dependencies, replacing them with docs.

Here's what changes in my example project with this change:

https://github.com/shakacode/react_on_rails_tutorial_with_ssr_and_hmr_fast_refresh/pull/21/files

I feel that if we're not going to support HMR, then it makes sense to let React configuration be outside of rails/webpacker.

### DIFF
--- a/README.md
+++ b/README.md
@@ -318,8 +318,7 @@ By default, you will find the Webpacker preset in your `package.json`.
 },
 ```
 
-Optionally, you can change your Babel configuration by removing these lines in your `package.json` and add [a Babel configuration file](https://babeljs.io/docs/en/config-files) in your project.
-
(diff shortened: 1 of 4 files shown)


justin808:
@guillaumebriday @dhh Should we remove all of React from rails/webpacker?

justin808:
@guillaumebriday this is rebased on master.

justin808:
I've got a [draft NPM package](https://www.npmjs.com/package/react-on-webpacker) that will make React with SSR and HMR a one-liner for rails/webpacker users: https://github.com/shakacode/react_on_rails_tutorial_with_ssr_and_hmr_fast_refresh/pull/24

Here's the config that takes the docs change and puts it into the NPM module: https://github.com/shakacode/react-on-webpacker/blob/main/babel/preset.js

PR #3232, opened 2021-11-20
provides two more common errors (fix suggestions) for Webpacker::Manifest::MissingEntryError

jasonfb:
see https://stackoverflow.com/questions/52639256/webpackermanifestmissingentryerror

### DIFF
--- a/lib/webpacker/manifest.rb
+++ b/lib/webpacker/manifest.rb
@@ -105,11 +105,14 @@ class Webpacker::Manifest
     def missing_file_from_manifest_error(bundle_name)
       <<-MSG
 Webpacker can't find #{bundle_name} in #{config.public_manifest_path}. Possible causes:
-1. You want to set webpacker.yml value of compile to true for your environment
-   unless you are using the `webpack -w` or the webpack-dev-server.
-2. webpack has not yet re-run to reflect updates.
-3. You have misconfigured Webpacker's config/webpacker.yml file.
-4. Your webpack configuration is not creating a manifest.
+1. You forgot to install node packages (try either `npm install` or `yarn install`)
+2. Your app has code with a non-standard extension (like a `.jsx` file) but the extension is not in the `extensions` config in `config/webpacker.yml`
+3. You have set compile: false (see `config/webpacker.yml`) for this environment
+   (unless you are using the `webpack -w` or the webpack-dev-server, in which case maybe you aren't running the dev server in the background?)
+4. Webpacker has not yet re-run to reflect updates.
+5. You have misconfigured Webpacker's `config/webpacker.yml` file.
+6. Your webpack configuration is not creating a manifest.
+
 Your manifest contains:
 #{JSON.pretty_generate(@data)}
       MSG
--- a/lib/webpacker/manifest.rb
+++ b/lib/webpacker/manifest.rb
@@ -105,7 +105,7 @@ class Webpacker::Manifest
     def missing_file_from_manifest_error(bundle_name)
       <<-MSG
 Webpacker can't find #{bundle_name} in #{config.public_manifest_path}. Possible causes:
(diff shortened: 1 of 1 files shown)


lib/webpacker/manifest.rb
```diff
@@ -105,11 +105,14 @@ def manifest_type(pack_type)
...
-1. You want to set webpacker.yml value of compile to true for your environment
-   unless you are using the `webpack -w` or the webpack-dev-server.
-2. webpack has not yet re-run to reflect updates.
-3. You have misconfigured Webpacker's config/webpacker.yml file.
-4. Your webpack configuration is not creating a manifest.
+1. You forgot to install node packages (try either `npm install` or `yarn install`)
+2. Your app has code with a non-standard extension (like a `.jsx` file) but the extension is not in the `extensions` config in `config/webpacker.yml`
+3. You have set compile: false (see `config/webpacker.yml`) for this environment
+   (unless you are using the `webpack -w` or the webpack-dev-server, in which case maybe you aren't running the dev server in the background?)
+4. Webpacker has not yet re-run to reflect updates.
```
justin808, line 112: `bin/webpack` not Webpacker?

justin808:
@guillaumebriday We should merge after the comment is addressed.

lib/webpacker/manifest.rb
jasonfb: you're right it was `webpack` before my change on that line was idle and not intentional

jasonfb:
this is mostly about #1 and #2; #3 (was previously #1) I just cleaned up the verbiage to match "Possible causes" as it was written out of style with the others.

lib/webpacker/manifest.rb
```diff
@@ -105,11 +105,14 @@ def manifest_type(pack_type)
     def missing_file_from_manifest_error(bundle_name)
       <<-MSG
 Webpacker can't find #{bundle_name} in #{config.public_manifest_path}. Possible causes:
-1. You want to set webpacker.yml value of compile to true for your environment
-   unless you are using the `webpack -w` or the webpack-dev-server.
-2. webpack has not yet re-run to reflect updates.
-3. You have misconfigured Webpacker's config/webpacker.yml file.
-4. Your webpack configuration is not creating a manifest.
+1. You forgot to install node packages (try either `npm install` or `yarn install`)
```
justin808, line 108: does `npm install` work with webpacker?

justin808 (commented):
one question

LGTM otherwise

lib/webpacker/manifest.rb
guillaumebriday: It should but we never mentioned it anywhere so I suggest to remove it here too

guillaumebriday (changes requested):
Could you remove the `npm install` please?

lib/webpacker/manifest.rb
justin808: @jasonfb this is good, but please remove the "npm" part.

@guillaumebriday I'll fix this if we don't hear from @jasonfb. This is a helpful fix.

jasonfb:
@justin808 @guillaumebriday -- done. Can you tell me just out of curiosity why you are recommending against `npm` in favor of `yarn` (I use yarn myself cause I find it a little smoother). 

FYI... this was researched by (and thanks the folks who fell down this rabbit hole)...
https://stackoverflow.com/questions/54113179/rails-webpackermanifestmissingentryerror-in-homeindex/70326851?noredirect=1#comment124478710_70326851

justin808:
Hi @jasonfb, a number of years ago, yarn worked better. It's hard to support both equally well. If you think these could be supported equally well, please submit a PR with your justifications.

guillaumebriday:
Thanks @jasonfb and @justin808

Issue #3200, opened 2021-10-18
binstubs should NOT honor the RACK_ENV environment variable in addition to the RAILS_ENV environment variable 


saltvedt:
See issue #773 and associated commit: https://github.com/rails/webpacker/commit/f0ce2a5e66f13cd2b562f71ec851f6f18c195121

When this works, it is by accident. The available values of RACK_ENV supported by Rack are `development`, `deployment` and `none`. Of these, only `development` is a common value of RAILS_ENV.

See: https://github.com/rack/rack/blob/master/lib/rack/server.rb#L180

I found this confusing when debugging an unrelated issue.

justin808:
This seems reasonable.

@saltvedt would you like to submit a PR on this?

CC: @guillaumebriday

justin808:
Here's the documentation.

https://github.com/rack/rack/blob/master/lib/rack/server.rb#L184-L190

```
    # * :environment
    #     this selects the middleware that will be wrapped around
    #     your application. Default options available are:
    #       - development: CommonLogger, ShowExceptions, and Lint
    #       - deployment: CommonLogger
    #       - none: no extra middleware
```


```ruby
    def default_options
      environment  = ENV['RACK_ENV'] || 'development'
      default_host = environment == 'development' ? 'localhost' : '0.0.0.0'
```

PR #3239, opened 2021-11-28
Remove setting RAILS_ENV from RACK_ENV, NODE_ENV defaults RAILS_ENV

justin808:
* Fixes https://github.com/rails/webpacker/issues/3200
* NODE_ENV will default to RAILS_ENV which is a reasonable default.

Per https://nodejs.dev/learn/nodejs-the-difference-between-development-and-production

Generally, the only value that matters for NODE_ENV is production. 

~Given the old code, I suspect that many people ran `rake assets:precompile` and had the NODE_ENV incorrectly set to 'development' for production builds.~

Actually, for running the rake task to build, https://github.com/rails/webpacker/blob/master/lib/tasks/webpacker/compile.rake#L23 sets the NODE_ENV to production if not set.

### DIFF
--- a/lib/install/bin/webpack
+++ b/lib/install/bin/webpack
@@ -6,7 +6,7 @@ require "webpacker"
 require "webpacker/webpack_runner"
 
 ENV["RAILS_ENV"] ||= "development"
-ENV["NODE_ENV"]  ||= "development"
+ENV["NODE_ENV"]  ||= ENV["RAILS_ENV"]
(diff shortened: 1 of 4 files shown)


guillaumebriday:
This is great, maybe we should talk about `RACK_ENV` removal in the upgrade docs?

justin808:
> This is great, maybe we should talk about RACK_ENV removal in the upgrade docs?

@guillaumebriday where in the docs?

PR #3243, opened 2021-12-14
Avoid stylesheet_pack_tag use default Webpacker

dukex:


lib/webpacker/helper.rb
```diff
@@ -141,7 +141,7 @@ def preload_pack_asset(name, **options)
   #   <%= stylesheet_pack_tag 'calendar' %>
   #   <%= stylesheet_pack_tag 'map' %>
   def stylesheet_pack_tag(*names, **options)
-    return "" if Webpacker.inlining_css?
```
justin808, line 144: @dukex does this old code crash? what's the problem with it?

justin808:
@dukex we'll reopen this once you respond.

PR #3244, opened 2021-12-21
Reinstate testing docs for Karma

eriknygren:
Found these docs very useful to get going with Karma unit tests in Webpacker 5. So I ported them over from the last version's docs to match what I had to do to get it working in Webpacker 6.

The docs no longer involve Typescript setup, as I don't use it on my webpacker project, but hopefully this will serve as a good starting point for others.


justin808:
I think the doc could be hosted on a blog.

I posted here as an example: 

https://forum.shakacode.com/t/karma-setup-for-webpacker-v5/2227

I don't like this in the repo because: 
1. We have to maintain this.
2. Jest is reportedly preferable over Karma for speed.

We can reopen if others feel differently.

eriknygren:
Ok. Running a unit test suite feels like a use-case common enough to the point it's worth maintaining docs for in my opinion. But I understand if the consensus is that Karma is not the go-to for Webpacker 6.

PR #3245, opened 2021-12-22
Raise a more accurate error when 'which_command' is not installed

katorres02:
I have a custom Centos docker image with a ruby on rails project setup. I was getting the following **error** during the precompile task.

`bundle exec rails assets:precompile`

this is the error output:
```
.
.
.
I, [2021-12-21T17:28:19.287300 #1]  INFO -- : Writing /usr/local/project/public/assets/express/lib/application-98f164ad65010340ab481475a0137ee982cf4de44f0e4c6441263ebb2b676481.js.gz
sh: which: command not found
sh: which: command not found
Node.js not installed. Please download and install Node.js https://nodejs.org/en/download/
Exiting!

The command '/bin/sh -c PRECOMPILE_ASSETS_FLAG=true RAILS_GROUPS=assets bundle exec rake assets:precompile' returned a non-zero code: 1
```

The final message indicates that **Node.js** is not installed which is incorrect. Node is already installed and working, the problem here is my Centos image doesn't have the "**which**" command activated/installed.

According to the GNU error codes:  https://www.gnu.org/software/libc/manual/html_node/Error-Codes.html
The error `Errno::ENOSYS` will be more accurate for this sceneario.

> Macro: int ENOSYS
“Function not implemented.” This indicates that the function called is not implemented at all, either in the C library itself or in the operating system.

### DIFF
--- a/lib/tasks/webpacker/check_node.rake
+++ b/lib/tasks/webpacker/check_node.rake
@@ -7,7 +7,7 @@ namespace :webpacker do
         which_command = 'where'
       else
         which_command = 'which'
-        raise Errno::ENOSYS if `rpm -qa | grep #{which_command}`.strip.empty?
+        raise Errno::ENOSYS if system("type which").nil?
       end
 
       raise Errno::ENOENT if `#{which_command} node || #{which_command} nodejs`.strip.empty?
@@ -36,7 +36,7 @@ namespace :webpacker do
       $stderr.puts "Exiting!"
       exit!
     rescue Errno::ENOSYS
-      $stderr.puts "'#{which_command}' is not installed. Please download and install '#{which_command}' by running the following command 'yum install #{which_command}'"
+      $stderr.puts "'#{which_command}' is not installed. Please download and install '#{which_command}'"
       $stderr.puts "Exiting!"
       exit!
     end
--- a/lib/tasks/webpacker/check_node.rake
+++ b/lib/tasks/webpacker/check_node.rake
@@ -3,7 +3,13 @@ namespace :webpacker do
   desc "Verifies if Node.js is installed"
   task :check_node do
     begin
-      which_command = Gem.win_platform? ? "where" : "which"
+      if Gem.win_platform?
+        which_command = 'where'
+      else
+        which_command = 'which'
+        raise Errno::ENOSYS if `rpm -qa | grep #{which_command}`.strip.empty?
+      end
+
       raise Errno::ENOENT if `#{which_command} node || #{which_command} nodejs`.strip.empty?
 
       node_version = `node -v || nodejs -v`.strip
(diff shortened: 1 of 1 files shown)


lib/tasks/webpacker/check_node.rake
```diff
@@ -3,7 +3,13 @@ namespace :webpacker do
   desc "Verifies if Node.js is installed"
   task :check_node do
     begin
-      which_command = Gem.win_platform? ? "where" : "which"
+      if Gem.win_platform?
+        which_command = 'where'
+      else
+        which_command = 'which'
+        raise Errno::ENOSYS if `rpm -qa | grep #{which_command}`.strip.empty?
```
justin808, line 10: `rpm` is not standard on MacOS

lib/tasks/webpacker/check_node.rake
```diff
@@ -29,6 +35,10 @@ namespace :webpacker do
       $stderr.puts "Node.js not installed. Please download and install Node.js https://nodejs.org/en/download/"
       $stderr.puts "Exiting!"
       exit!
+    rescue Errno::ENOSYS
+      $stderr.puts "'#{which_command}' is not installed. Please download and install '#{which_command}' by running the following command 'yum install #{which_command}'"
```
justin808, line 39: Why should webpacker be responsible for ensuring items like `which`, `node`, and `yarn` are installed?

Could be better just to have the normal error when some required binary is not available?

@dhh @guillaumebriday  any thoughts?

justin808 (changes requested):

justin808:
@katorres02 could you test out using `type` rather than `which` in your PR?

katorres02:
> @katorres02 could you test out using `type` rather than `which` in your PR?

@justin808 thanks for taking the time to review this pr.
in my Centos docker image the `type which` command didn't work using those special quotation marks. (_type is installed_)
image
 
I used `system("type which")` instead. it returns **nil** if the `which_command` binary is not installed


> Why should webpacker be responsible for ensuring items like which, node, and yarn are installed?

if webpacker is not responsable about that, can we consider removing the `check_node` [rake task](https://github.com/rails/webpacker/pull/3245/files/31e9707526da661b05a70e00b05dd2b8ab561309#diff-a06160b0093d6bcdfaaff1c32c0799870e39945252c317f5c4424f46b5eb0202)?

lib/tasks/webpacker/check_node.rake
guillaumebriday: I agree. You could basically raise the standard error when a command does not exist.

There are here anyway https://github.com/rails/webpacker#prerequisites

guillaumebriday:
We probably should remove everything related to the `which_command` here

lib/tasks/webpacker/check_node.rake
katorres02: what is the standard error?

**the Prerequisites list doesn't mention the "which" binary**

not sure I'm following the conversation.

if you don't have "`which`" installed you get the "`Node.js not installed error`" message, that is not good.
This is why I am proposing a different error message in this PR for this scenario. a more accurate message will help to reduce the time to fix this problem. in my case I spent a couple hours trying to understand why nodejs was not installed and turns out that the issue was not related to nodejs at all I didn't have "which" installed

katorres02:
> We probably should remove everything related to the `which_command` here

makes sense

guillaumebriday:
Nice thanks @katorres02

PR #3247, opened 2021-12-22
Add example babel.config.js

justin808:
Shows extending the default babel.config.js


guillaumebriday:
I think this should be in the docs folder not in the readme 👍

justin808:
I'll make the update in the next day or so.

PR #3249, opened 2021-12-28
Upgrading workflows

guillaumebriday:
- Updating the CI with the current LTS of Node
- Cleaning workflows
- Testing against new versions of Rails and Ruby


aried3r:
Cleanup looks good!

But I don't think linting with multiple node and Ruby versions makes sense. RuboCop should yield the same result with any Ruby version. Currently it should check with the Ruby 2.7 parser because of this line:

https://github.com/rails/webpacker/blob/550c53eb7816140a24b3af9691fbbdf23f24a563/.rubocop.yml#L3

Same should go for JS linting. WDYT?

justin808 (approved):
LGTM

PR #3250, opened 2021-12-28
Use `exist?` instead of deprecated `exists?`

koic:
`Dir.exists?` and `File.exists? were removed by https://github.com/ruby/ruby/pull/5352.
It will be the following error in Ruby 3.2.

```console
% ruby -ve 'File.exists?("foo.rb")'
ruby 3.2.0dev (2021-12-28T15:22:18Z master 39b3aa4fb3) [x86_64-darwin19]
-e:1:in `<main>': undefined method `exists?' for File:Class (NoMethodError)

File.exists?("foo.rb")
    ^^^^^^^^
Did you mean?  exist?
```

And this PR enables `Lint/DeprecatedClassMethods` cop to detect these deprecated methods.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -110,6 +110,9 @@ Layout/TrailingWhitespace:
 Style/RedundantPercentQ:
   Enabled: true
 
+Lint/DeprecatedClassMethods:
+  Enabled: true
+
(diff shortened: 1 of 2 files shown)


koic:
The build error is unrelated to this change. #3251 will solve it.

PR #3259, opened 2022-01-13
Binstub checker message is misleading

bopm:
While it's looking for a file with name `webpacker` it still reports a file with name `webpack`.


guillaumebriday:
Thanks

PR #3267, opened 2022-02-02
Update README.md to reflect webpacker/shakapacker options

justin808:


justin808:
Hi @dhh, can you take a look at my slight changes. The goal of this PR is to allow Shakapacker as a consideration for those Rails apps that benefit from the more advanced webpack support provided by Shakapacker.

As the current README stands, it seems "risky" to move to an "unsupported" fork. I've seen this feedback in many different places. And the feedback I've seen also mentions many issues that were resolved in webpacker/shakapacker v6.

dhh:
Whatever risk someone associates with an unsupported (by Rails core) fork is accurate insofar as their fear is based on Rails core not being involved any more. I don't think this really clarifies or further illuminates that. I appreciate all the work you're putting into Shakapacker, but I also don't want to keep fiddling with this on the webpacker side of things.

PR #3294, opened 2022-11-04
update upgrading guide to suggest webpack-dev-server@3

nachiket87:
As per https://github.com/rails/webpacker/pull/3121

updating webpack-dev-server will result in version 4 which causes breakage that is hard to debug. Update readme to only update webpack-dev-server within version 3 to prevent developers from having to debug the issue with webpack-cli/serve


nachiket87:
No idea why travis-ci is failing 😞

PR #3302, opened 2023-01-20
CI against Ruby 3.1 and 3.2

amatsuda:
Just added configurations for newer stable versions of Ruby.


amatsuda:
Actually this patch includes a fix for current master branch CI failure, so let me merge this.

